Issue #17, opened 2014-05-24
Support other element sets


Juanlu001:
See https://github.com/Pybonacci/poliastro/wiki/Different-element-sets for complete explanation. One simple solution that might do the trick:

Maybe a simple fix would be: return to accepting tuples (i.e. revert 64dc66c) and assume p if ecc == 1.0 (or even within an interval). Plus, include an extra parameter anomaly=='true' or 'mean' and assume the corresponding value. I can also use an extra parameter for the shape='sma' or 'p'. All the rest remains equal.

Juanlu001:
See recent ideas in the wiki:

https://github.com/poliastro/poliastro/wiki/Different-element-sets#new-ideas

I think this could be a solution, but definitely it will see the light in 0.4.

Juanlu001:
Phase 1: Split class for vectors and classical.
Phase 2: Support other element set, for example equinoctial.

PR #61, opened 2015-07-01
Refactor `State` class

Juanlu001:
This allows for adding new representations in an easier way (see #17).


Juanlu001:
There are some conversion functions missing but I will leave it as it is now. Lack of numerical examples to provide tests don't motivate me to add them, I will leave an issue open about it. This will probably mean a decrease in coverage.

PR #64, opened 2015-07-07
General propagation

Juanlu001:
Added general two body equations ~~in Gauss form~~ following Cowell's formulation to perform numerical integration and example notebook to showcase this feature.

To be done:

* [ ] Complete docs
* [ ] ~~Ellaborate on Gauss equations~~
* [x] Add unit tests


Juanlu001:
The documentation needs some lifting already, as there is a mix between high level and low level APIs. But that will go to another pull request. Merging this!

Issue #73, opened 2015-11-25
Investigate latest failures (NumPy 1.10?)


Juanlu001:

newlawrence:
This is what I found out so far.

I've modified the scripts to build **Poliastro** using specific versions of **Numpy** and yes, it seems that there's some kind of issue arising with version 10 (here are the **OS X** logs from [Travis](https://travis-ci.org/newlawrence/poliastro/builds/94829872) & **Windows logs** from [Appveyor](https://ci.appveyor.com/project/newlawrence/poliastro/build/0.4.0.dev0-6)).

But it seemed that everything was OK for **Anaconda Cloud**, doesn't it?

Well, the problem is that **Anaconda Cloud** uses the **conda's build** system to build and test the packages and there're big troubles to tell it to build against specific **Numpy** versions.
If you read the logs, a package ***poliastro-0.4.0.dev0-py27_1*** is generated, while if it also was dependent on **Numpy**'s version it should be ***poliastro-0.4.0.dev0-np19py27_1***.

One of the dependencies of **Poliastro** must be incompatible with **Numpy 1.10** as **conda build** always chooses **Numpy 1.9**, so if we apply the following [patch](https://github.com/newlawrence/poliastro/commit/a8a2150fd873b7b34b51d03c0fe57a466b7d066f) to force the specific **Numpy** version, we obtain the same results for **Linux** from [Anaconda Cloud](https://anaconda.org/newlawrence/poliastro/builds/13) that we got from **Travis** and **Appveyor**.

**How the path works:** **conda build** relies on the **CONDA_PY** and **CONDA_NPY** environment variables to build the packages using specific versions of **Python** and **Numpy**. When both of them are set up, the resulting package will be named ***package-version-npYYpyXX_bnum***. If they're not set (neither **--python** nor **--numpy** flags are provided), **conda build** will build the package using the current **Python** version installed and the most recent version of **Numpy** allowed by the dependancies. As **Anaconda Cloud** doesn't set up the **CONDA_NPY** variable, we do it manually. The changes made to the **meta.yaml** ensure the package is built using the specified **Numpy** version.

[...]

PR #75, opened 2015-12-03
Avoid accessing protected class members from outside

proinsias:
Found a few cases of protected class members accessed from outside
their modules. Fixed these by either making the member public or by
using a property to access the member.

### DIFF
--- a/poliastro/twobody/classical.py
+++ b/poliastro/twobody/classical.py
@@ -92,10 +92,10 @@ def coe2mee(p, ecc, inc, raan, argp, nu):
     return p, f, g, h, k, L
 
 
-class _ClassicalState(State):
+class ClassicalState(State):
     def __init__(self, attractor, p, ecc, inc, raan, argp, nu,
                  epoch):
-        super(_ClassicalState, self).__init__(attractor, epoch)
+        super(ClassicalState, self).__init__(attractor, epoch)
         self._p = p
         self._ecc = ecc
         self._inc = inc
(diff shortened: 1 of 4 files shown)


Juanlu001:
Thanks for the pull request @proinsias! Travis CI is failing because of #73, I will fix it this weekend and then review your contribution.

proinsias:
Thanks @Juanlu001! This is only a minor change. This is my first code PR, so I wasn't sure whether I should submit it when the build of the master branch was failing.

Juanlu001:
You've done the right thing! My policy is that `master` is always green, so if it's red it's my fault :)

poliastro/twobody/classical.py
```diff
@@ -134,7 +138,7 @@ def _repr_latex_(self):
 
         """
         elem_names = [r"a", r"e", r"i", r"\Omega", r"\omega", r"\nu"]
-        elem_values = [elem._repr_latex_().strip("$")
+        elem_values = [elem.repr_latex().strip("$")
```
Juanlu001: Not sure about this one, since it's calling `_repr_latex_` of each of the elements, which is an astropy Quantity: https://github.com/astropy/astropy/blob/v1.0.7/astropy/units/quantity.py#L935

poliastro/twobody/classical.py
```diff
@@ -127,6 +127,10 @@ def argp(self):
     def nu(self):
         return self._nu
 
+    @property
```
Juanlu001: See comment below, this wouldn't be necessary. It's an internal method for IPython use: http://ipython.readthedocs.org/en/3.x/config/integrating.html#rich-display

Juanlu001:
I added a couple of comments, unfortunately I don't think the tests cover them so I made them based on my own judgement.

proinsias:
Sounds good! Anything else you need from me? Do you need me to undo the changes you disagreed with above?

Juanlu001:
Yes please! As soon as you push the changes and the builds turn green I will merge the pull request :)

proinsias:
Okay, I reverted those specific changes, and added a comment explaining why we could keep that private method. Let me know if that looks good.

Juanlu001:
Looks good to me, merging!

Issue #74, opened 2015-12-01
Upload packages for Python 3.5


Juanlu001:

newlawrence:
Please, submit a **conda jplephem** package to your channel and we're done 😊.

newlawrence:
**Python 3.5** with **NumPy 1.9** will work now, [see](https://github.com/poliastro/poliastro/pull/76).

Juanlu001:
This comment https://github.com/conda/conda-build/pull/650#issuecomment-160681665 outlines all the possible cases regarding packages depending on NumPy. jplephem is a pure-python package and so is poliastro, so we shall be very careful of artificially pinning the numpy version - it's mostly harmless, but expensive. I'm still studying this topic.

Juanlu001:
Well, it seems matplotlib is broken for Python 3.5 on Anaconda, so poliastro packages will have to wait a bit. ContinuumIO/anaconda-issues#573

Juanlu001:
I have tried reproducing the problem above in a different Linux box and I can't, so it has to be some specific issue in my home machine. I'll try building Python 3.5 packages today.

Juanlu001:
The packages will be uploaded when making the release, as usual. This is no longer a work in progress and will be closed when poliastro 0.4 is released.

Juanlu001:
Python 3.5 packages were updated at the time of the last release, closing this.

PR #77, opened 2015-12-05
Fixes for NumPy 1.10, fix #73

### LINKED ISSUE #73: Investigate latest failures (NumPy 1.10?)

Juanlu001:


Juanlu001:
@newlawrence In the latest commit I reverted what you explained in https://github.com/poliastro/poliastro/issues/73#issuecomment-161945841 according to what I said in https://github.com/poliastro/poliastro/issues/74#issuecomment-162250209 - in principle, we shouldn't pin the NumPy version since poliastro is not a compiled package. Let's see if everything works though.

PR #80, opened 2015-12-06
Fixes for CI tests, fix #79.

### LINKED ISSUE #79: Only NumPy 1.10 used in tests
@Juanlu001 glad to see you managed to unpin the **NumPy** version from the package, but there's still the problem I mentioned you before.

Specifying the **NumPy** version on the tests scripts is irrelevant since it'll be always updated to the latest version as **jplephem** is installed, as you can see in the logs ([Anaconda - see line 448](https://anaconda.org/poliastro/poliastro/builds/7/1), [Travis - see line 209](https://travis-ci.org/poliastro/poliastro/jobs/95123104) and [Appveyor - see line 147](https://ci.appveyor.com/project/poliastro/poliastro/branch/master/job/rlp3syt1ak8ryv3v)).

newlawrence:
Fixes the issue [#79](https://github.com/poliastro/poliastro/issues/79).

Tests under **Anaconda** are now performed the same way as in **Travis**. No more need of building the **conda** package.

### DIFF
--- a/.binstar.yml
+++ b/.binstar.yml
@@ -7,7 +7,7 @@ script:
   - export NUMPY=$(python -c "from __future__ import print_function;import numpy as np;print(np.__version__[:np.__version__.rfind('.')])")
   - conda create -q -n test-environment python=$PYTHON
   - source activate test-environment
-  - conda install numpy=$NUMPY numba>=0.18 astropy>=1.0 matplotlib pytest pip coverage requests pyyaml scipy jplephem
+  - conda install numpy=$NUMPY numba>=0.18 astropy>=1.0 matplotlib pytest pip requests pyyaml scipy jplephem
   - pip install -e .
   - py.test poliastro -vv
 platform: 
--- a/.binstar.yml
+++ b/.binstar.yml
@@ -3,13 +3,13 @@ install:
   - conda config --set always_yes true
   - conda config --add channels poliastro
 script:
-  # Force conda build to use specific versions of Python and Numpy
-  - export CONDA_PY=$(python -c "from __future__ import print_function;from sys import version_info as ver;print(str(ver.major) + str(ver.minor))")
-  - export CONDA_NPY=$(python -c "from __future__ import print_function;import numpy;npver=numpy.__version__;print(npver[:npver.rfind('.')].replace('.',''))")
-  # Build once the env variables are set
-  - conda build buildscripts/condarecipe
-build_targets:
-  - conda
+  - export PYTHON=$(python -c "from __future__ import print_function;from sys import version_info as pyver;print(str(pyver[0]) + '.' + str(pyver[1]))")
+  - export NUMPY=$(python -c "from __future__ import print_function;import numpy as np;print(np.__version__[:np.__version__.rfind('.')])")
+  - conda create -q -n test-environment python=$PYTHON
+  - source activate test-environment
+  - conda install numpy=$NUMPY numba>=0.18 astropy>=1.0 matplotlib pytest pip coverage requests pyyaml scipy jplephem
+  - pip install -e .
+  - py.test poliastro -vv
 platform: 
   - linux-64
 engine:
--- a/.binstar.yml
+++ b/.binstar.yml
@@ -1,12 +1,10 @@
 package: poliastro
 install:
-  - conda config --set always_yes true
+  - conda config --set always_yes yes --set changeps1 no
   - conda config --add channels poliastro
 script:
   - export PYTHON=$(python -c "from __future__ import print_function;from sys import version_info as pyver;print(str(pyver[0]) + '.' + str(pyver[1]))")
   - export NUMPY=$(python -c "from __future__ import print_function;import numpy as np;print(np.__version__[:np.__version__.rfind('.')])")
-  - conda create -q -n test-environment python=$PYTHON
-  - source activate test-environment
   - conda install numpy=$NUMPY numba>=0.18 astropy>=1.0 matplotlib pytest pip requests pyyaml scipy jplephem
   - pip install -e .
   - py.test poliastro -vv
--- a/.binstar.yml
+++ b/.binstar.yml
@@ -1,18 +1,24 @@
 package: poliastro
+platform: linux-64
+engine:
+  - python=2.7
+  - python=3.3
+  - python=3.4
+  - python=2.7
+  - python=3.4
+env:
+  - NUMPY=1.9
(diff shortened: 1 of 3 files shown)


Juanlu001:
Thanks for the changes again. The thing I see here is that in fact there are two different issues here: the easy one is that we forgot to specify the NumPy requirement when installing jplephem, and therefore was upgrading. The second one is the stubborness of Anaconda to use the NumPy versions we specify. The former is nicely solved, and in fact I would still add a `-q` flag to supress some output.

Regarding the latter, I've noticed a change since the last poliastro builds and yours.

* In poliastro builds https://anaconda.org/poliastro/poliastro/builds/7/4 and https://anaconda.org/poliastro/poliastro/builds/7/0, a message is issued about conda-build being out of date, then it updates and overrides the specified NumPy version.
* In your builds https://anaconda.org/newlawrence/poliastro/builds/22/4 and https://anaconda.org/newlawrence/poliastro/builds/22/0 you can see this does not happen.

Apparently, this may vary from machine to machine: https://github.com/Anaconda-Server/anaconda-build/issues/100#issuecomment-147876314. If that's a problem (which it seems it is) we should probably report a proper issue upstream.

On the other hand Anaconda creates an environment by default, but we are repeating this step in the build scrip t (see https://anaconda.org/newlawrence/poliastro/builds/22/4#L176 and https://anaconda.org/newlawrence/poliastro/builds/22/4#L242). I think we should try to simplify that to the bare minimum.

newlawrence:
Oh, I forgot to remove the environment creation line, thanks for the tip! I'll take care of that within a minute.

About the outdated **conda-build** issue you're right, it's not deterministic (or, at least, it doesn't seem to), but I think it isn't a matter. In fact, if the advice doesn't appear in my builds is due to they no longer rely in building the package, so no **conda-build** is used.

I decided to direct test the library (as we do with **Travis**) instead of building it for two reasons:
* The first is that removing the lines that pin the library to the **NumPy** version in the **meta.yaml** file leaves us with no mechanisms to avoid the upgrading to version 1.10 (maybe there's a way, but I didn't found it yet).
* The second is that you were not using **Anaconda** for automated builds currently, so a much simpler approach than building the package could be the way to proceed.

However, if you prefer building the package on **Anaconda** (for future automated builds or whatever), drop here a line and lets see if we can find a workaround to force the tests against a specific **NumPy** version :).

Juanlu001:
You are right about the conda-build update, it's just an error message but it never runs. It took me several iterations to understand the problem but now I think I get it. I'll get back to this issue tonight.

.binstar.yml
```diff
@@ -1,15 +1,13 @@
...
 install:
-  - conda config --set always_yes true
+  - conda config --set always_yes yes --set changeps1 no
   - conda config --add channels poliastro
 script:
-  # Force conda build to use specific versions of Python and Numpy
-  - export CONDA_PY=$(python -c "from __future__ import print_function;from sys import version_info as ver;print(str(ver.major) + str(ver.minor))")
-  - export CONDA_NPY=$(python -c "from __future__ import print_function;import numpy;npver=numpy.__version__;print(npver[:npver.rfind('.')].replace('.',''))")
-  # Build once the env variables are set
-  - conda build buildscripts/condarecipe
```
Juanlu001: Do you think `conda build buildscripts/condarecipe --python $PYTHON --numpy $NUMPY` would work here? http://conda.pydata.org/docs/commands/build/conda-build.html

Juanlu001: Or better: don't compute `$PYTHON` (as it's not needed) and setting `NUMPY` as an environment variable, which in theory plays nicely with the build matrix: http://docs.anaconda.org/build-config.html#BuildMatrix

newlawrence: In answer to your question I can tell you, in my experience dealing with this hell, that this will have no effect; if no strict dependency on **NumPy** is declared it fallbacks to the default behavior and upgrades the library when setting up the building environment. In fact, since the building process doesn't inherit environment variables by default (in order to be completely reproducible) it's truly difficult to establish mechanisms to prevent the upgrade.

About your suggestion, yes it'll result in a cleaner script, I'll take care of it. However, it's a pity that no environments variables based on the engine configuration are set; this would have made all the process more straightforward.

newlawrence:
There you are https://anaconda.org/newlawrence/poliastro/builds/24 :) (they have a bug in the **HTML**, the test excluded is **Python 3.3 NumPy 1.10** instead of **Python 3.4 NumPy 1.10**).

.binstar.yml
```diff
@@ -1,20 +1,24 @@
 package: poliastro
+platform: linux-64
+engine:
+  - python=2.7
+  - python=3.3
+  - python=3.4
+  - python=2.7
```
Juanlu001: Are these extra two still needed?

.binstar.yml
Juanlu001: Yes, calling it a hell is an accurate description... Let's hope we'll have this sorted sometime soon, we'll have to contact Anaconda devs eventually.

.binstar.yml
```diff
@@ -1,20 +1,24 @@
...
-  - python=3.4 numpy=1.10
+  - conda install numpy=$NUMPY numba>=0.18 astropy>=1.0 matplotlib pytest pip requests pyyaml scipy jplephem
+  - pip install -e .
+  - py.test poliastro -vv
+---
+package: poliastro
+platform: linux-64
+engine: python=3.3
+env: NUMPY=1.10
+exclude: true
```
Juanlu001: Smart!

.binstar.yml
newlawrence: Sorry, my mistake 😅. It's not good to be in a hurry.

.binstar.yml
newlawrence: Well, I think the problem is on what we are trying to do. If a package is meant to be used against a library regardless the version, what should we do? Test it against every single one as posible? And then, shall we repeat the process on any posible combination of the whole bunch of dependencies? It makes sense that in case of no version dependency (or in a range of them), to use the most newer as possible.

But, due to the special status of **NumPy**, I believe they should do an exception (what it seems that they're working on). Hope will see a compromise solution to this issue soon :).

.binstar.yml
newlawrence: Fixed! :)

Juanlu001:
Well, this got closer to a sane solution, despite all the problems :) Thanks again and let's merge this in!

PR #81, opened 2015-12-07
Targets Python 3.5

Juanlu001:
This step is required before fixing #74. Also, implemented idea to solve NumPy version madness when testing (cc @newlawrence).


Juanlu001:
Now the builds will take longer unfortunately. I strive for dropping Python 3.3 support by next year as detailed in the wiki https://github.com/poliastro/poliastro/wiki/Python-support-roadmap.

PR #90, opened 2015-12-12
Change plotting logic for non-closed orbits

Juanlu001:


Juanlu001:
I love all those green marks :green_heart: except from @codecov-io :broken_heart:. I will accept a decrease in coverage here though, since who knows how to unit-test plotting code anyway?

PR #91, opened 2015-12-13
WIP: Changed directory structure and improve testing

Juanlu001:
Still missing: coverage reporting. I took @ionelmc articles, specially http://blog.ionelmc.ro/2014/05/25/python-packaging/, as inspiration, but still trying to keep it simpler and match our conda workflow.


Juanlu001:
Notice that https://github.com/Juanlu001/poliastro/commit/347a97565acbc294d1390a8a577cb06be9167b59 triggered AppVeyor and Anaconda Server builds too. This does not normally happen because they expect new commits to be pushed against the original repo, whereas I push new commits to my fork. Travis CI always does the right thing, so despite the inconvenience I will keep developing this way.

ionelmc:
@Juanlu001 fyi, you can have less test environments (than what I suggested in the blogpost) if you do it like this: https://hynek.me/articles/testing-packaging/

You have to fiddle with the coverage configuration a bit more but it's doable. 

PS. I don't know how to get C extension coverage on with that technique.

Juanlu001:
Thanks for the info! Fortunately I don't have C or Fortran extensions so I'm not worried about that.

PR #100, opened 2016-01-16
Drop `util.dot` in favor of `np.dot`

### LINKED ISSUE #89: Drop `util.dot` when next numba is out
See https://github.com/numba/numba/pull/1560.

Juanlu001:
This change requires numba >=0.23 and fixes #89.


Juanlu001:
This new feature does not work on Python 3.3. Time to drop it! :wave:

PR #106, opened 2016-01-23
Restore semimajor axis as shape parameter

### LINKED ISSUE #62: Conversion between coe and rv is not transitive
This is a consequence on my indecision about using the semimajor axis or the semilatus rectum. To avoid singularities and problems (see #37) I should probably move to using the semilatus rectum definitively.

If I make this change, probably the special function introduced in #38 won't be needed anymore.

Juanlu001:
Closes #62. For reference:

> I find myself never dealing with parabolic orbits, yet I introduced this change and I find it extremely annoying for the remaining 100 % of my use cases.

See https://github.com/poliastro/poliastro/wiki/Parabolic-orbits:-final-decision for rationale.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -1,6 +1,18 @@
 What's new
 ==========
 
+New in poliastro 0.5.0
+----------------------
+
+.. warning:: This version is not released yet.
+
+Backward incompatible changes:
+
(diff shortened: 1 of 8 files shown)


Juanlu001:
The build is red because QuantifiedCode is making one recommendation. Merging anyway.

Issue #58, opened 2015-06-27
Seek for better Kepler and Lambert algorithms


Juanlu001:
Battin and others describe better algorithms with better convergence properties and performance. Aside problems:

* Memory profiling with [memory_profiler](https://pypi.python.org/pypi/memory_profiler)
* Online benchmarking with [asv](http://asv.readthedocs.org/)?

Juanlu001:
I can draw some inspiration from the wonderful, BSD-3 licensed https://github.com/jacobwilliams/Fortran-Astrodynamics-Toolkit (including very useful references)

Issue #115, opened 2016-02-05
Consider creating time benchmarks


Juanlu001:
https://github.com/spacetelescope/asv

Juanlu001:
See also https://github.com/ionelmc/pytest-benchmark

Issue #112, opened 2016-01-25
Lambert algorithm fails to converge in some cases


Juanlu001:
[Reproduction steps to be added]

If it's an implementation problem it needs fixing, and if it's an algorithm problem a new algorithm is needed (as per #58).

Juanlu001:
Preliminary tests show the issue is fixed with the new algorithm. Test pending.

Juanlu001:
A solution is found, but it is still not perfect.

![index](https://cloud.githubusercontent.com/assets/316517/13404713/4e707544-df1b-11e5-828e-566a91f722a7.png)

Decreasing the tolerance in the Halley's process didn't lead to a closer flyby. Worth investigating if it's a problem with the algorithm, the problem, the propagation or what. Anyway, although there's no regression test yet this bug can be fixed when #118 is merged.

Juanlu001:
Postponing this to 0.6.

Juanlu001:
I still need to add a test for this. Guess it's the New Horizons Pluto flyby, but will need to check.

Juanlu001:
I cannot find the code where this happened, and also I think it's probably more related to plotting inequalities than convergence issues. I'm closing this.

Juanlu001:
By the way, moral of the story: always include code to reproduce :)

PR #118, opened 2016-02-29
[WIP] Izzo algorithm for initial orbit determination

Juanlu001:
Test pass with the new algorithm, but several tasks are pending:

* [ ] Test Planet 9 notebook with this algorithm
* [x] Complete algorithm for the multi revolution case and add corresponding tests
* [ ] Profile, refine and `numba.njit`
* [x] Complete Householder iteration
* [ ] Create benchmarks (see #115)
* [x] Create package `poliastro.iod` and keep both algorithms
* [ ] Update release notes


Juanlu001:
This PR should fix #58 and, hopefully, #112.

Juanlu001:
After carefully considering some available options (see https://github.com/poliastro/poliastro/wiki/API-Refactor:-Multi-revolution-Lambert-problem) I decided to specify the number of revolutions `N` as part of the solution, instead of computing it. This means that the function will be more predictable, as it won't find all the possible solutions, and allow for keeping moderate backwards compatibility.

The change will involve modifying https://github.com/Juanlu001/poliastro/blob/cf820d6/src/poliastro/iod/izzo.py#L105-L113 and possibly add a exception if no solution is feasible.

Juanlu001:
The convergence properties of the current implementation are suspicious, see https://github.com/poliastro/poliastro/wiki/Optimization-of-Izzo-algorithm.

Juanlu001:
There was a regression because I made the `householder` method too specific and is no longer reusable. Wondering how to fix it.

Juanlu001:
Closing this in favor of #121.

PR #121, opened 2016-03-05
Izzo algorithm for initial orbit determination, 2nd attempt

Juanlu001:
This supersedes #118. Tests pass with the new algorithm, but several tasks are pending:

* [ ] Test Planet 9 notebook with this algorithm
* [x] Complete algorithm for the multi revolution case and add corresponding tests
* [x] Profile, refine and `numba.njit`
* [x] Complete Householder iteration
* [x] Create benchmarks (see #115)
* [x] Create package `poliastro.iod` and keep both algorithms
* [ ] Update release notes


Juanlu001:
With "strange regression" I meant "could not reproduce on my machine", but the key was to update numba to 0.24, see https://github.com/numba/numba/issues/1755.

Juanlu001:
The code is almost complete now, I need to release the necessary dependencies and finish this pull request.

Juanlu001:
After several attempts, it does not seem easy to get CEPHES to compile on Windows and OS X with the given time constraints, so this approach must be abandoned - otherwise the whole point of being cross-platform is lost. Two options:

* Provide our specific `hypf21(3, 1, 5/2, x)` in pure Python and accelerate it with numba.
* Switch to distutils to build CEPHES instead of using the given makefile, mimicking what SciPy already does.

Juanlu001:
I'm leaving the second option as an open issue (https://github.com/poliastro/pycephes/issues/1) and going easy route for now.

PR #135, opened 2016-08-21
Refactor State objects

Juanlu001:
See https://github.com/poliastro/poliastro/wiki/Simplifying-State-objects for rationale.


Juanlu001:
* [ ] Decide on `ss.state` API
* [ ] Restore Jupyter representation

Issue #129, opened 2016-07-14
Second level tests not included in conda package


Juanlu001:
Apparently `setup.py build` is ignoring them:

```
$ python setup.py build
running build
running build_py
creating build/lib
creating build/lib/poliastro
copying src/poliastro/maneuver.py -> build/lib/poliastro
copying src/poliastro/examples.py -> build/lib/poliastro
copying src/poliastro/cli.py -> build/lib/poliastro
copying src/poliastro/util.py -> build/lib/poliastro
copying src/poliastro/plotting.py -> build/lib/poliastro
copying src/poliastro/bodies.py -> build/lib/poliastro
copying src/poliastro/jit.py -> build/lib/poliastro
copying src/poliastro/hyper.py -> build/lib/poliastro
copying src/poliastro/testing.py -> build/lib/poliastro
copying src/poliastro/__init__.py -> build/lib/poliastro
copying src/poliastro/stumpff.py -> build/lib/poliastro
copying src/poliastro/ephem.py -> build/lib/poliastro
creating build/lib/poliastro/iod
copying src/poliastro/iod/vallado.py -> build/lib/poliastro/iod
copying src/poliastro/iod/izzo.py -> build/lib/poliastro/iod
copying src/poliastro/iod/__init__.py -> build/lib/poliastro/iod
creating build/lib/poliastro/twobody
copying src/poliastro/twobody/rv.py -> build/lib/poliastro/twobody
copying src/poliastro/twobody/classical.py -> build/lib/poliastro/twobody
copying src/poliastro/twobody/propagation.py -> build/lib/poliastro/twobody
copying src/poliastro/twobody/equinoctial.py -> build/lib/poliastro/twobody
copying src/poliastro/twobody/angles.py -> build/lib/poliastro/twobody
copying src/poliastro/twobody/__init__.py -> build/lib/poliastro/twobody
copying src/poliastro/twobody/core.py -> build/lib/poliastro/twobody
creating build/lib/poliastro/tests
copying src/poliastro/tests/test_ephem.py -> build/lib/poliastro/tests
copying src/poliastro/tests/test_maneuver.py -> build/lib/poliastro/tests
copying src/poliastro/tests/test_iod.py -> build/lib/poliastro/tests
copying src/poliastro/tests/test_hyper.py -> build/lib/poliastro/tests
copying src/poliastro/tests/test_stumpff.py -> build/lib/poliastro/tests
[...]

PR #139, opened 2016-08-28
#129 Ensure that "package_data" are included

pombredanne:
Based on your question in https://stackoverflow.com/questions/38384677/python-setup-py-build-ignoring-some-files

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,5 +54,7 @@ setup(
       "Topic :: Scientific/Engineering :: Astronomy",
     ],
     long_description=io.open('README', encoding='utf-8').read(),
-    package_data={"poliastro": ['tests/*.py']}
+    package_data={"poliastro": ['tests/*.py']},
+    include_package_data=True,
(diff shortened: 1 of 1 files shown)


Juanlu001:
Thanks a lot for the pull request! Would you mind explaining why did you add the `zip_safe=False` option?

Juanlu001:
Well, now I see that if you don't explicitly state a value then setuptools does some weird analysis to determine it:

http://peak.telecommunity.com/DevCenter/setuptools#setting-the-zip-safe-flag

So I'm merging this PR as is. Thanks again!

pombredanne:
> Would you mind explaining why did you add the `zip_safe=False` option?

Unless you are sure you could accommodate with accessing package data eventually in a zip,  always use `zip_safe=False` to guarantee that the as-installed package will be unzipped (which is kinda of a moot point when using wheels that are always unzipped, but old habit die hard and eggs are still possible)

PR #151, opened 2017-02-26
Correct typo in user_guide.rst

jzmiller1:


Juanlu001:
Good catch :) Thanks!

Issue #120, opened 2016-03-05
Review examples


Juanlu001:
- [ ] ~~Add custom matplotlibrc style, compatible with matplotlib 2.0~~
- [x] Rerun all with matplotlib 2.0
- [ ] ~~Add instructions on index to download SPK files~~
- [ ] Add explanations where missing
- [ ] Rewrite Cowell notebook using new methods, review error checking (more time to see significant changes?) and validation (perhaps take example from https://github.com/Juanlu001/pfc-uc3m/)
- [x] Rerun all notebooks and fix corresponding bugs
- [x] Add interactive widget to Mars notebook as done with the lightning talks
- [x] Add links in index to old notebooks (those checking performance of Python vs Fortran) and remove them from master
- [ ] ~~Move to a separate repository, hopefully preserving history (see https://git-scm.com/docs/git-filter-branch and http://blog.jupyter.org/2015/04/15/the-big-split/ for inspiration)~~

AunSiro:
Tried to run this notebook with library just installed from conda forge:

https://github.com/poliastro/poliastro/blob/master/examples/Solving%20Lambert's%20problem%20in%20Python.ipynb

The comment cell states that it is necessary to activate a different branch, but:
- The link is dead
- There is no branch with the exact name stated
- There is no info or link on how to change the active branch with poliastro installed through conda

AunSiro:
Tried to run the "Going to Mars with Poliastro", had critical failure dowloading ephemeris (404):

https://github.com/poliastro/poliastro/blob/master/examples/Going%20to%20Mars%20with%20Python%20using%20poliastro.ipynb

In code cell 2º had the following warnings:

> No SPICE kernels found under ~/.poliastro.
> Please download them manually or using
> 
>   poliastro download-spk [-d NAME]
> 
> to provide a default kernel, else pass a custom one as
> an argument to `planet_ephem`.
> 
> Downloading de421.bsp from http://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/, please wait...
> Not Found
> Downloading de421.bsp from http://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/a_old_versions/, please wait...
> 

The links appear to be dead as of now. Following this, calling the function:
`ephem.planet_ephem(ephem.EARTH, times_vector)`
will return:

 

> TypeError                                 Traceback (most recent call last)
> <ipython-input-13-7202734be5bc> in <module>()
[...]

Issue #147, opened 2017-02-17
Error with ephemerides


Davide-sd:
Today I've install poliastro and I tryed to follow the example code in the documentation.
I'm having problem with the ephemerides. This is the error i get:

```---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-43-9c36e44cc7f8> in <module>()
      1 print(epoch)
----> 2 r, v = ephem.planet_ephem(ephem.EARTH, epoch)
      3 print(r)
      4 print(v)

/home/davide/anaconda3/lib/python3.6/site-packages/poliastro/ephem.py in planet_ephem(body, epoch, kernel)
    130 
    131     """
--> 132     r, v = kernel[0, body].compute_and_differentiate(epoch.jd1, epoch.jd2)
    133     r = r * u.km
    134     v = v * u.km / u.day

TypeError: 'NoneType' object is not subscriptable
```

Can anybody try the example [in the user guide](http://poliastro.readthedocs.io/en/latest/user_guide.html#where-are-the-planets-computing-ephemerides) and let me know if you replicate this error?

Juanlu001:
Hi @Davide-sd, thanks for the bug report! This is a duplicate of #128, now that you're the second person to report it I'm trying to fix it ASAP. Please subscribe there for upcoming updates, I might release a 0.6.1 with the fix.

Issue #128, opened 2016-07-06
Missing ephemerides if no files on import time


Juanlu001:
Original comment by @Fedor-Lyanguzov on https://github.com/poliastro/poliastro/issues/126#issuecomment-230373112:

> By the way, i've tried "Going to Mars with Python using poliastro" locally, it works, but not on first run. On first run code downloads kernel "de421", but do not make it the default kernel (ephem.default_kernel), because selecting default kernel is done on ephem import.

Juanlu001:
Steps to reproduce:

```
>>> import poliastro.ephem
No SPICE kernels found under ~/.poliastro.
Please download them manually or using

  poliastro download-spk [-d NAME]

to provide a default kernel, else pass a custom one as
an argument to `planet_ephem`.

>>> poliastro.ephem.download_kernel("de421")
Downloading de421.bsp from http://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/, please wait...
Not Found
Downloading de421.bsp from http://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/a_old_versions/, please wait...

>>> from astropy import time
>>> epoch = time.Time("2015-05-09 10:43")
>>> from poliastro import ephem
>>> r, v = ephem.planet_ephem(ephem.EARTH, epoch)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-df0835788b33> in <module>()
      2 epoch = time.Time("2015-05-09 10:43")
      3 from poliastro import ephem
----> 4 r, v = ephem.planet_ephem(ephem.EARTH, epoch)

/home/juanlu/.miniconda35/envs/py36test/lib/python3.6/site-packages/poliastro/ephem.py in planet_ephem(body, epoch, kernel)
    130 
    131     """
--> 132     r, v = kernel[0, body].compute_and_differentiate(epoch.jd1, epoch.jd2)
    133     r = r * u.km
    134     v = v * u.km / u.day

TypeError: 'NoneType' object is not subscriptable
```

PR #152, opened 2017-02-28
"Going to Mars with Poliastro" warning added

AunSiro:
If you had not downloaded the ephemeris and run the notebook
sequentially, you might encounter problems, as I found in #120 

I just added a small warning so you know that you might need to restart the kernel.

### DIFF
--- a/examples/Going to Mars with Python using poliastro.ipynb
+++ b/examples/Going to Mars with Python using poliastro.ipynb
@@ -19,12 +19,46 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "First of all, we import the necessary modules. Apart from poliastro we will make use of astropy to deal with physical units and time definitions and jplephem to compute the positions and velocities of the planets."
+    "We need a binary file from NASA called *SPICE kernel* to compute the position and velocities of the planets. Let's download and read it:"
    ]
   },
   {
    "cell_type": "code",
    "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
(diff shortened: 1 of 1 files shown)


Juanlu001:
Thanks for the pull request @AunSiro! The error that you noticed is actually a duplicate of https://github.com/poliastro/poliastro/issues/147 and https://github.com/poliastro/poliastro/issues/128, and should be fixed since https://github.com/poliastro/poliastro/pull/148. Would you check if this still happens to you with latest `master`?

Juanlu001:
By the way, the failures are related to AWS outage yesterday.

AunSiro:
It happened with the latest Poliastro that conda installs (intalled it yesterday). How can I check if it is up-to-date with master? if it isn't, how can I manually update it?

Juanlu001:
The development version is already ahead from the latest released version, which is 0.6.0. You can check that here:

https://github.com/poliastro/poliastro/compare/0.6.x...master

To install the latest development version, I think this should work:

```
$ pip install git+git://github.com/poliastro/poliastro
```

provided that you have git properly configured on your computer. Otherwise, you could also try this:

```
$ pip install https://github.com/poliastro/poliastro/archive/master.zip
```

AunSiro:
You were right, with  poliastro 0.7. the notebook downloads and uses the ephemeris without any problem. Should I cancel the pull request, or change the warning to add that it applies to anterior versions?

Juanlu001:
We can close this pull request, since the warning will only apply to master, where it is not needed. I guess this problem is affecting many people so I might release a 0.6.1 with a proper fix. Thanks anyway for the effort :)

Issue #150, opened 2017-02-26
Add patched conics computations


Juanlu001:

AunSiro:
Fly-by maneuvers are extremely sensitive to initial conditions. In KSP, I frequently observe that changes of ~0.5 m/s a quarter of an orbit before the encounter can result in dramatic differences, often changing the post-flyby trajectory more than 40º.

Maybe, an interesting approach can be the following:
First, use the desired objective of the maneuver, like the vectorial dV obtained or flyby trayectory, to obtain a hyperbolic flyby orbit, and propagate it backwards in time with precision to the limit of the SOI.

Using the required speed vector, a collision heliocentric orbit is calculated. 

Then calculate the correction maneuvers required to get to the entering SOI point with the required speed vector.

Maybe, the first two steps can be executed in the inverse order, getting a cone of the possible post flyby trajectories given a collision trajectory, and then finding the optimum one from this possibility space.

AunSiro:
I propose having a minimun flyby altitude for every planet equal to 2% of the planet radius (p.ej, for Earth is ~120 km).
Imagine a generic orbit plane. Assuming a hyperbolic orbit, depending on the speed at the periapsis, a speed can be calculated either at both intersections of the orbit with the SOI or at the infinite before and after flyby. Then, a vectorial difference between the entering and leaving velocities can be calculated. Therefore, we can correlate a maximum speed change with every entering speed, that is to say, the maximum delta v obtainable with a flyby without additional propulsion can be calculated depending only on the scalar value of the relative speed between the spacecraft and the planet. 

I think that getting the function that calculate it could be a good starting point!

Juanlu001:
Thanks a lot for these ideas Siro! I am starting a new page in the wiki to draft a proposal of an API:

https://github.com/poliastro/poliastro/wiki/Patched-conics-computations

Let us first try to imagine how one would **use** these functions and then proceed to implement them. We can take other pages of the wiki as inspiration, for example https://github.com/poliastro/poliastro/wiki/Maneuver:-New-ideas or https://github.com/poliastro/poliastro/wiki/API-Refactor:-Maneuver-example.

Working on this again after four years is going to be extremely refreshing! :smile:

AunSiro:
I suggested some ideas in the wiki page, check them and tell me what do you think!

Juanlu001:
This looks promising! Some comments here:

[...]

PR #140, opened 2016-09-11
Propagation improvements

Juanlu001:
See https://github.com/poliastro/poliastro/wiki/Make-propagation-functions-higher-level for rationale.

Issue #131, opened 2016-08-07
Replace ephemerides functions with Astropy equivalents


Juanlu001:
This replacement cannot be done until Astropy supports proper motions and velocities, see this bug:

https://github.com/astropy/astropy/issues/4344

For instance, `get_body_barycentric` does not return the velocity of the planets, and therefore it's useless for our purposes.

Juanlu001:
First steps: barycentric velocities in Astropy https://github.com/astropy/astropy/pull/5231

Juanlu001:
Astropy 1.3 is out :tada:

http://www.astropy.org/announcements/release-1.3.html

In particular:

* Vector arithmetic for coordinates, including taking the norm http://docs.astropy.org/en/stable/coordinates/representations.html#astropy-coordinates-representations-arithmetic
* Barycentric position _and velocity_ of a Solar System body http://docs.astropy.org/en/stable/api/astropy.coordinates.get_body_barycentric_posvel.html

Juanlu001:
Using the class `CartesianRepresentation` is needed for vector arithmetic, but it doesn't work with normal `Quantity` objects for obvious reasons.

Juanlu001:
For some reason I'm not getting the same results with poliastro and the `jpl` ephemeris of Astropy. More debugging required.

```
In [1]: from poliastro.ephem import planet_ephem

In [2]: from astropy.time import Time

In [5]: from poliastro.ephem import EARTH

In [6]: t = Time("2014-09-22 23:22")

In [7]: planet_ephem(EARTH, t)
Out[7]: 
(<Quantity [  1.50492843e+08, -9.68589576e+05, -4.41002632e+05] km>,
 <Quantity [  -26318.45001712, 2353112.69264584, 1020089.67574404] km / d>)

In [8]: from astropy.coordinates import solar_system_ephemeris

In [9]: solar_system_ephemeris.set("jpl")
Downloading http://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/de430.bsp
|=====================================================================================| 119M/119M (100.00%)      1m21s
Out[9]: <ScienceState solar_system_ephemeris: 'jpl'>

In [10]: print(solar_system_ephemeris.bodies)
('sun', 'mercury', 'venus', 'earth-moon-barycenter', 'earth', 'moon', 'mars', 'jupiter', 'saturn', 'uranus', 'neptune', 'pluto')

In [11]: from astropy.coordinates import get_body_barycentric_posvel

[...]

PR #153, opened 2017-03-01
Expanded Bodies Data

AunSiro:
Additional bodies of the Solar System added to bodies.py.
Additional data of existing and new bodies added, including ROI.

k parameter data from: https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/a_old_versions/de421_announcement.pdf

Radius of the Sphere of influence formula from Wikipedia.

Mean radius, semi-mayor axis and eccentricity from Wikipedia, except for Saturn's semi-mayor axis, which is from http://articles.adsabs.harvard.edu/cgi-bin/nph-iarticle_query?1994A%26A...282..663S&amp;data_type=PDF_HIGH&amp;whole_paper=YES&amp;type=PRINTER&amp;filetype=.pdf

### DIFF
--- a/src/poliastro/bodies.py
+++ b/src/poliastro/bodies.py
@@ -9,15 +9,22 @@ Contains some predefined bodies of the Solar System:
 and a way to define new bodies (:py:class:`~Body` class).
 
 """
-from astropy.constants import R_earth
-from astropy import units as u
 
+from astropy import units as u
+import warnings
+from poliastro.twobody import Orbit
+from poliastro.twobody.angles import M_to_nu
+import os.path
+
+au = 149597870.700 * u.km
+body_dict = {}
 
 class Body(object):
     """Class to represent a body of the Solar System.
 
     """
-    def __init__(self, k, name=None, symbol=None, R=0 * u.km):
+    def __init__(self, k, name=None, symbol=None, R=0 * u.km,
+                 parent=None, orbit = None):
         """Constructor.
 
         Parameters
@@ -30,17 +37,24 @@ class Body(object):
             Symbol for the body.
         R : Quantity
             Radius of the body.
+        parent : Body
+            Gravitational center (ej: Sun for the Earth)
+        orbit: Orbit
+            orbit object of the body
 
         """
         self.k = k
         self.name = name
         self.symbol = symbol
         self.R = R
+        self.parent = parent
+        self.orbit = orbit
+        self.soi = None
 
     @classmethod
     @u.quantity_input(k=u.km ** 3 / u.s ** 2, R=u.km)
-    def from_parameters(cls, k, name, symbol, R):
-        return cls(k, name, symbol, R)
+    def from_parameters(cls, k, name, symbol, R, parent, orbit):
+        return cls(k, name, symbol, R, parent, orbit)
 
     def __str__(self):
         return u"{0} ({1})".format(self.name, self.symbol)
@@ -52,81 +66,75 @@ class Body(object):
 
         """
         return self.__str__()
+    def calculate_soi(self):
+        '''Calculate the radius of the sphere of influence.
+        The body must have a parent body and an orbit defined'''
+        if self.parent == None:
+            warnings.warn('Unable to calculate SOI without valid parent body')
+        elif self.orbit == None:
+            warnings.warn('Unable to calculate SOI without valid orbit assigned')
+        else:
+            self.soi= self.orbit.a * (self.k / self.parent.k) ** (2/5)
 
+data_path = os.path.join(os.path.dirname(__file__), "bodies_data.txt")
+with open(data_path, 'r', encoding='utf-8') as raw_data :
+    for line in raw_data:
+        line = line.replace('\ufeff','')
+        line = line.replace('\n','')
+        if line[0] == '#':
+            continue
+        line = line.replace(' ','')
+        body_data = line.split(sep = ',')
+        name, symbol, k, R = body_data[:4]
+        k = float(k) * u.km**3 / u.s**2
+        R = float(R) * u.km
+        
+        if body_data[4] == 'None':
+            parent = None
+        else:
+            parent_name = body_data[4]
+            if parent_name in body_dict:
+                parent = body_dict[parent_name]
+            else:
+                message = 'loading bodies warning: '
+                message += 'object not found during body loading:'
+                message += parent_name +'. Parent reverted to None for ' + name
+                warnings.warn(message)
+                parent = None
+        if body_data[5] == 'None':
+            orbit = None
+        elif parent == None:
+            orbit = None
+            warnings.warn('Unable to create orbit without valid parent body')
+        else:
+            a, ecc, inc, L, long_peri, raan = body_data[5:11]
+            a = float(a) * au
+            ecc = float(ecc) * u.one
+            inc = (float(inc) * u.deg).to(u.rad)
+            L = (float(L) * u.deg).to(u.rad)
+            long_peri = (float(long_peri) * u.deg).to(u.rad)
+            raan = (float(raan) * u.deg).to(u.rad)
+            argp = long_peri - raan
+            M = L - long_peri
+            nu = M_to_nu(M, ecc)
+            orbit = Orbit.from_classical(parent, a, ecc, inc, raan, argp, nu)
+        body_dict[name] = Body.from_parameters(k, name, symbol, R, parent, orbit)
+    
 
-Sun = Body.from_parameters(k=132712440018 * u.km ** 3 / u.s ** 2,
-           name="Sun", symbol=u"\u2609", R=695700 * u.km)
+for body_name in body_dict:
+    body = body_dict[body_name]
+    if body.parent != None and body.orbit != None:
+        body.calculate_soi()
 
-Earth = Body.from_parameters(k=398600.44 * u.km ** 3 / u.s ** 2,
-             name="Earth", symbol=u"\u2641", R=R_earth.to(u.km))
-Earth.a = 149598023 * u.km #Semi-mayor axis
-Earth.e = 0.0167086 * u.dimensionless_unscaled #Eccentricity
-
-Moon = Body.from_parameters(k=4902.8 * u.km ** 3 / u.s ** 2,
-             name="Moon", symbol=u"\u263E", R=1737.1 * u.km)
-Moon.a = 384399 * u.km #Semi-mayor axis
-Moon.e = 0.0549 * u.dimensionless_unscaled #Eccentricity
-
-Mercury = Body.from_parameters(k=22032.09 * u.km ** 3 / u.s ** 2,
-             name="Mercury", symbol=u"\u263F", R=2439.7 * u.km)
-Mercury.a = 57909050 * u.km #Semi-mayor axis
-Mercury.e = 0.205630 * u.dimensionless_unscaled #Eccentricity
-
-Venus = Body.from_parameters(k=324858 * u.km ** 3 / u.s ** 2,
-             name="Venus", symbol=u"\u2640", R=6051.8 * u.km)
-Venus.a = 108208000 * u.km #Semi-mayor axis
-Venus.e = 0.006772 * u.dimensionless_unscaled #Eccentricity
-
-Mars = Body.from_parameters(k=42828.4 * u.km ** 3 / u.s ** 2,
-             name="Mars", symbol=u"\u2642", R=3389.5 * u.km)
-Mars.a = 227.9392 * u.Gm #Semi-mayor axis
-Mars.e = 0.0934 * u.dimensionless_unscaled #Eccentricity
-
-Jupiter = Body.from_parameters(k=126712764 * u.km ** 3 / u.s ** 2,
-             name="Jupiter", symbol=u"\u2643", R=69911 * u.km)
-Jupiter.a = 778.299 * u.Gm #Semi-mayor axis
-Jupiter.e = 0.048498 * u.dimensionless_unscaled #Eccentricity
-
-Saturn = Body.from_parameters(k=37940585 * u.km ** 3 / u.s ** 2,
-             name="Saturn", symbol=u"\u2644", R=58232 * u.km)
-Saturn.a = 1429.395 * u.Gm #Semi-mayor axis
-Saturn.e = 0.05555 * u.dimensionless_unscaled #Eccentricity
-
-Uranus = Body.from_parameters(k=5794548 * u.km ** 3 / u.s ** 2,
-             name="Uranus", symbol=u"\u2645", R=25362 * u.km)
-Uranus.a = 2875.04 * u.Gm #Semi-mayor axis
-Uranus.e = 0.04638 * u.dimensionless_unscaled #Eccentricity
-
-Neptune = Body.from_parameters(k=6836535 * u.km ** 3 / u.s ** 2,
-             name="Neptune", symbol=u"\u2646", R=24622 * u.km)
-Neptune.a = 4504.45 * u.Gm #Semi-mayor axis
-Neptune.e = 0.009456 * u.dimensionless_unscaled #Eccentricity
-
-Pluto = Body.from_parameters(k=977 * u.km ** 3 / u.s ** 2,
-             name="Pluto", symbol=u"\u2647", R=1187 * u.km)
-Pluto.a = 5906.38 * u.Gm #Semi-mayor axis
-Pluto.e = 0.2488 * u.dimensionless_unscaled #Eccentricity
-
-planet_list = [
-        Mercury,
-        Venus,
-        Earth,
-        Mars,
-        Jupiter,
-        Saturn,
-        Uranus,
-        Neptune,
-        Pluto] #Kinda? Just so it doesn't feel alone
-
-for planet in planet_list:
-    planet.r_soi = planet.a * (planet.k / Sun.k) ** (2/5)   
-
-Moon.r_soi = Moon.a * (Moon.k / Earth.k) ** (2/5)     
-
+Sun = body_dict['Sun']
+Earth = body_dict['Earth']
+Jupiter = body_dict['Jupiter']
 
 #Checking the numers:
 if __name__ == '__main__':
-    for planet in planet_list:
-        print(planet.name.rjust(8), round(float(planet.r_soi / planet.R)))
+    for body_name in body_dict:
+        body = body_dict[body_name]
+        if body.soi != None:
+            print(body.name.rjust(8), round(float(body.soi / body.R)))
     #Values should be close to the described at:
     #https://en.wikipedia.org/wiki/Sphere_of_influence_(astrodynamics)
--- a/src/poliastro/bodies.py
+++ b/src/poliastro/bodies.py
@@ -76,65 +76,58 @@ class Body(object):
         else:
             self.soi= self.orbit.a * (self.k / self.parent.k) ** (2/5)
 
-data_path = os.path.join(os.path.dirname(__file__), "bodies_data.txt")
-with open(data_path, 'r', encoding='utf-8') as raw_data :
-    for line in raw_data:
-        line = line.replace('\ufeff','')
-        line = line.replace('\n','')
-        if line[0] == '#':
+_data_path = os.path.join(os.path.dirname(__file__), "bodies_data.txt")
+with open(_data_path, 'r', encoding='utf-8') as _raw_data :
+    for _line in _raw_data:
+        _line = _line.replace('\ufeff','')
+        _line = _line.replace('\n','')
+        if _line[0] == '#':
             continue
-        line = line.replace(' ','')
-        body_data = line.split(sep = ',')
-        name, symbol, k, R = body_data[:4]
-        k = float(k) * u.km**3 / u.s**2
-        R = float(R) * u.km
+        _line = _line.replace(' ','')
+        _body_data = _line.split(sep = ',')
+        _name, _symbol, _k, _R = _body_data[:4]
+        _k = float(_k) * u.km**3 / u.s**2
+        _R = float(_R) * u.km
         
-        if body_data[4] == 'None':
-            parent = None
+        if _body_data[4] == 'None':
+            _parent = None
         else:
-            parent_name = body_data[4]
-            if parent_name in body_dict:
-                parent = body_dict[parent_name]
+            _parent_name = _body_data[4]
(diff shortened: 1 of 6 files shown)


Juanlu001:
Thanks a lot for this pull request Siro! This is probably a good starting point for https://github.com/poliastro/poliastro/issues/150. I have several comments:

* I'm not sure if the semimajor axis and the eccentricity belong here. While properties like the mass and the planet mean radius are clearly property of the planets themselves, the parameters of the orbit are _a consequence of that body being in orbit around a certain attractor_. For instance, you can retrieve the osculating orbit using `ephem.planet_ephem`, as described here http://poliastro.readthedocs.io/en/latest/user_guide.html#where-are-the-planets-computing-ephemerides. One could argue that having the **mean** properties of the orbit attached to the bodies might be practical, but still I'm not convinced from a code point of view. What do you think?
* In any case, if we are adding more properties I think they should be put inside the `__init__` method. Adding them dynamically after instantiation is possible but can confound IDEs and other introspection tools.
* Instead of hardcoding the formula inside `bodies.py` it would be nice to have a separate function to compute the spehere of influence radius. I think it's fine if we leave it inside `bodies`, we probably don't need to create a separate module for that.
* Notice that you changed the value of `k` for the Earth and three tests failed because of small numerical differences. At this point I think we need to choose an authoritative reference to fix these values and document it somewhere, because a) I don't know where did I copy mine from and b) I don't know where did you copy yours :sweat_smile:
* Regarding the last point, _the gravitational parametes depend on the ephemerides_: each one of them uses slightly different values to perform the computations. I already pointed it out here https://github.com/brandon-rhodes/python-jplephem/issues/9 Taking all of these variations would perhaps be too complex but at least we have to keep that in mind, and document it somewhere.
* We need tests for the new things we include, in particular the function that computes the radius of the sphere of influence but also the numerical properties.
* No, we are not including Pluto :)

Juanlu001:
Fix: you linked your source for the gravitational parameters in the first comment. However, notice that DE421 is not our default ephemeris, but DE430:

https://github.com/poliastro/poliastro/blob/v0.6.0/src/poliastro/cli.py#L25

When we switch to Astropy to compute them (#131) the situation will be the same:

http://docs.astropy.org/en/stable/coordinates/solarsystem.html#solar-system-ephemerides

AunSiro:
- I think that we should include the mean orbital parameters because they are very lightweight, good enought for most hobby applications and first approaches and don't force users to download medium-sized aphemeris files . If these files are not included with poliastro because they are too heavy, the users should not be forced to dowload them later in order to use the interesting items. Also, having a back-up will make the package independent from the user ability to connect to the internet and the state of the source files, which cann not be taken as granted (as happened to me a couple of days ago, when the pages 404'd because of the problems with the servers).
While you are right about the orbital parameters being a consecuence of the parent body gravity, I think that storing them inside each planet is the most clean and organiced  way.
-I agree that the way you are describing for storing and generating these new data are cleaner and more consistent, so I will change it and create a commit.
- I am very surprised that the change in the k value made the test fail, I only added a coulpe of decimals (!?)
- I don't really understand anything in the paragraph about how the gravitational parametes depend on the ephemerides. I don't understand either the conversation in the issue you linked T_T
- I will read how the other test are built and try to create a new one for the next commit!
- I think we sould include Pluto for a simple reason: it is very popular and the aditional work and complexity is very low. In order to be fair, I suggest adding the other four main dwarf planets: Ceres, Eris, Haumea and Makemake. I don't mind searching the data for them. If their Ephemeris is not included in the file, I thing the mean Keplerian elements should work fine.


I have a couple of questions:
- Should the body objects have a link to their gravitational parent body object? (may be useful in the future if data about moons and extrasolar sistem are added)
- Should the data be written in the bodys.py like now, or better read from a text file that could display all the data in a table form, also including the origin of the data? (may be more readable or useful if additional data columns are later considered)

jzmiller1:
I've been creating bodies for stars with exoplanets.  While it was useful to have the Sun and Earth predefined as a new user I think that it might be better let everyone create their own bodies as needed.  Perhaps some tools for create bodies from common representations (CSV, some NASA formats) would be helpful.

Data for stars and exoplanets often comes in terms of Solar, Earth or Jupiter masses and radii.  It might be worth thinking about making it easier to define bodies in these terms as well. This is an example of how I'm handling it now...maybe there's a better way I'm overlooking:

```python
from astropy import constants as c
from poliastro.bodies import Body

G = c.G.to((u.km**3)/(u.kg * u.s**2))
TRAPPIST1 = Body.from_parameters(k=0.08*c.M_sun * G,
                                 name='TRAPPIST',
                                 symbol=None,
                                 R=0.114 * c.R_sun.to(u.km))
```

Would a subclass of Body that has a solar mass/radius interface make sense in the poliastro library or would that be something that users should have in their own codebase?

AunSiro:
I think that would be a really nice interface! We are creating the bodies at the moment through the `from_parameters` method, maybe another way to implement your interface would be to have another creation method in the body object, called `from_relative_values` for example. That way, bodies created either way would share the same class.

Either with a different class or as a method, that would work great together with my proposals here, letting the users insert bodies into their projects in three different ways: from the `poliastro.bodies.Body.from_parameters()`,  from the `poliastro.bodies.Body.from_relative_values` or just importing some common predefined bodies, like the ones I propose here, simply through `poliastro.bodies.Mars`

The ability to create them from CSV or NASA formats is also a great idea, you know of any standard form of distributing these data?

jzmiller1:
I didn't really get what the class method was doing.  Are you thinking something more like this?

```python
class Body(object):
    """Class to represent a body of the Solar System.
    """

    def __init__(self, k, name=None, symbol=None, R=0 * u.km, **kwargs):
        """Constructor.
        Parameters
        ----------
        k : Quantity
            Standard gravitational parameter.
        name : str
            Name of the body.
        symbol : str
            Symbol for the body.
        R : Quantity
            Radius of the body.
        """
        self.k = k
        self.name = name
        self.symbol = symbol
        self.R = R
        self.__dict__.update(kwargs)

    @classmethod
    @u.quantity_input(k=u.km ** 3 / u.s ** 2, R=u.km)
    def from_parameters(cls, k, name, symbol, R):
        return cls(k, name, symbol, R)

    @classmethod
    def from_relative(cls, masses, name=None, symbol=None, R=None, body='sun'):
        bodies = {'sun': {'mass': c.M_sun, 'radius': c.R_sun.to(u.km)},
                  'earth': {'mass': c.M_earth, 'radius': c.R_earth.to(u.km)},
                  'jupiter': {'mass': c.M_jup, 'radius': c.R_jup.to(u.km)}
                  }
        k = masses * bodies[body]['mass'] * c.G.to((u.km ** 3) / (u.kg * u.s ** 2))
        R *= bodies[body]['radius']
        kwargs = {'body': body, 'bodies': bodies}
        return cls(k, name, symbol, R, **kwargs)

    def __str__(self):
        return u"{0} ({1})".format(self.name, self.symbol)

    def _repr_latex_(self):
        """Creates a LaTeX representation.
        Used by the IPython notebook.
        """
        return self.__str__()
```

A NASA format might be an ephemeris file from HORIZONS:

http://ssd.jpl.nasa.gov/horizons.cgi#top

http://ssd.jpl.nasa.gov/horizons.cgi#results

I know poliastro installs https://pypi.python.org/pypi/jplephem so this may be something that is already possible.  I'm still figuring things out here.

Juanlu001:
Wow, lots of ideas going on here! I will answer to all tomorrow, but for the time being, let me just state that Astropy uses approximate orbital elements by default which do not need to download the JPL ephemeris

http://docs.astropy.org/en/stable/coordinates/solarsystem.html#solar-system-ephemerides

based on this algorithm

https://github.com/liberfa/erfa/blob/448a4ec/src/plan94.c

which, by the way, is similar in intent to what poliastro 0.1 offered:

https://github.com/poliastro/poliastro/blob/v0.1.2/poliastro/octave/uplanet_2013.m

Juanlu001:
Answering @AunSiro first:

> I think that we should include the mean orbital parameters because they are very lightweight, good enought for most hobby applications and first approaches and don't force users to download medium-sized aphemeris files . If these files are not included with poliastro because they are too heavy, the users should not be forced to dowload them later in order to use the interesting items. Also, having a back-up will make the package independent from the user ability to connect to the internet and the state of the source files, which cann not be taken as granted (as happened to me a couple of days ago, when the pages 404'd because of the problems with the servers).

Does my previous comment address your concerns? When I implement #131, all computations will be feasible offline.

> While you are right about the orbital parameters being a consecuence of the parent body gravity, I think that storing them inside each planet is the most clean and organiced way.

I'm still not sure, can be convenient but it is also redundant. If we are adding the semimajor axis and eccentricity, I don't see why we would not add the other orbital parameters (except the anomaly). And if we are adding those, why not adding the mean equinoctial values too? Different people will suggest having different parameters there, and the `Body` class can end up being a catch-all object.

Instead, I believe retrieving the ephemerides and then turn that into a `State` object where we can get the desired values, although a bit longer, is easier to maintain in the long term.

> I am very surprised that the change in the k value made the test fail, I only added a coulpe of decimals (!?)

The perils of floating point I guess...

> I don't really understand anything in the paragraph about how the gravitational parametes depend on the ephemerides. I don't understand either the conversation in the issue you linked T_T

The JPL ephemerides fit some polynomials to a handful of observed quantities, and they also have to adjust the values of the gravitational parameter of each body. Every edition has different values, see for instance the ones used in DE421 and DE430:

| Ephemeris | Earth GM (km3/s2) |
|-------|-----|
| DE421 | 398600.436233 |
| DE430 | 398600.435436 |

* DE421 https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/planets/a_old_versions/de421_announcement.pdf
* DE430 https://ilrs.cddis.eosdis.nasa.gov/docs/2014/196C.pdf

This is why we have to:

1. Choose which value we want to use, document it and adjust the tests, or
2. Change the values with the ephemerides that are chosen.

Option 2 is much more complicated, so if we are refining the values we use (which I agree it's a good idea) we should probably use those in DE430 and document this fact somewhere.

> I think we sould include Pluto for a simple reason: it is very popular and the aditional work and complexity is very low. In order to be fair, I suggest adding the other four main dwarf planets: Ceres, Eris, Haumea and Makemake. I don't mind searching the data for them. If their Ephemeris is not included in the file, I thing the mean Keplerian elements should work fine.

Fair enough, if we're including other dwarf planets I agree :)

> Should the body objects have a link to their gravitational parent body object? (may be useful in the future if data about moons and extrasolar sistem are added)

Good catch! Maybe a `parent` attribute, empty for the Sun?

> Should the data be written in the bodys.py like now, or better read from a text file that could display all the data in a table form, also including the origin of the data? (may be more readable or useful if additional data columns are later considered)

I'm not sure, maybe yes? We can consider having a CSV with those values, I would not oppose that idea if it does not make the code much more complicated.

> Is there any instructions available on how to use the ephemeris file? Which data does it contain?

I'm not sure what do you mean, can you rephrase?

Juanlu001:
Thanks @jzmiller1 for jumping in! I think having the ability to create objects in terms of others is a good idea. Regarding your question:

> I didn't really get what the class method was doing.

It is only checking that the units are in place. The method was introduced in #140 because I observed that the unit conversion and checking are `extremely` slow, so I wanted to leave an "unsafe" method to create bodies without checking the units.

Maybe we can have something like this?

```
TRAPPIST1 = Body.from_relative(
    parent=Sun,
    k=0.08,  # Relative to the Sun
    name='TRAPPIST',
    symbol=None,
    R=0.114)  # Relative to the Sun

```

Juanlu001:
Edit: I noticed that there can be a confusion in using the `parent` argument when creating a body, because `TRAPPIST1` is defined in terms of solar masses but _does not orbit around the Sun_. Another approach:

```python
TRAPPIST1 = Body.from_relative(
    reference=Sun,
    parent=None,
    k=0.08,  # Relative to the Sun
    name='TRAPPIST',
    symbol=None,
    R=0.114)  # Relative to the Sun

TRAPPIST1D = Body.from_relative(
    reference=Earth,
    parent=TRAPPIST1,  # Orbits TRAPPIST1
    # ...,
    R=0.772)  # Relative to the Earth
```

jzmiller1:
The parent parameter isn't in the existing code and isn't in this pull request.  What is it meant to be used for?

Juanlu001:
It isn't in the existing code, but I suggested its creation after this
discussion :) It's mean to hold the central body (Sun for the Earth, Earth
for the Moon).

AunSiro:
I am near to create a new commit with some changes to this pull request. @jzmiller1 , Should I include your proposal for a function to create a body from relative values of another body, or would you rather create a new pull request with it yourself so it gets linked to you?

jzmiller1:
I can put it in once your PR is merged in unless you prefer to do it.

AunSiro:
I have commited a new version, the only left thing to do is add aditional test to cover the new code.

AunSiro:
I wrote some test that I think that cover most of the new content! Expecting feedback for next iteration or merge at this point!

src/poliastro/bodies.py
```diff
@@ -9,15 +9,22 @@
...
+au = 149597870.700 * u.km
+body_dict = {}
 
 class Body(object):
     """Class to represent a body of the Solar System.
 
     """
-    def __init__(self, k, name=None, symbol=None, R=0 * u.km):
+    def __init__(self, k, name=None, symbol=None, R=0 * u.km,
+                 parent=None, orbit = None):
```
Juanlu001: I still dislike the idea of adding an orbit to the body itself. The mean elements available at a certain epoch are somewhat useful, but having approximate ephemerides through Astropy and ERFA it looks like duplicated data to me. I would still like the `Body` class to be as simple as possible, removing the orbit and the associated imports.

src/poliastro/bodies.py
```diff
@@ -30,17 +37,24 @@ def __init__(self, k, name=None, symbol=None, R=0 * u.km):
...
+        orbit: Orbit
+            orbit object of the body
 
         """
         self.k = k
         self.name = name
         self.symbol = symbol
         self.R = R
+        self.parent = parent
+        self.orbit = orbit
```
Juanlu001: See above.

src/poliastro/bodies.py
```diff
@@ -30,17 +37,24 @@ def __init__(self, k, name=None, symbol=None, R=0 * u.km):
...
+        self.soi = None
```
Juanlu001: Why defer the computation of the SOI? We can do that in `__init__`.

src/poliastro/bodies.py
```diff
@@ -52,9 +66,68 @@ def _repr_latex_(self):
 
         """
         return self.__str__()
+    def calculate_soi(self):
```
Juanlu001: This can be a function outside of the `Body` class:

```
def calculate_soi(body, a, parent=None):
    # Parameter `a` is mandatory now
    # `parent` can be None if `body` already has it
    pass
```

src/poliastro/bodies.py
```diff
@@ -52,9 +66,68 @@ def _repr_latex_(self):
...
+        The body must have a parent body and an orbit defined'''
+        if self.parent == None:
+            warnings.warn('Unable to calculate SOI without valid parent body')
+        elif self.orbit == None:
+            warnings.warn('Unable to calculate SOI without valid orbit assigned')
+        else:
+            self.soi= self.orbit.a * (self.k / self.parent.k) ** (2/5)
+
+_data_path = os.path.join(os.path.dirname(__file__), "bodies_data.txt")
+with open(_data_path, 'r', encoding='utf-8') as _raw_data :
```
Juanlu001: This should be a function.

Also, is there a chance we can use the `csv` module from the stdlib here, or even pandas?

https://docs.python.org/3/library/csv.html

src/poliastro/bodies.py
```diff
@@ -52,9 +66,68 @@ def _repr_latex_(self):
...
+            _orbit = Orbit.from_classical(_parent, _a, _ecc, _inc, _raan, _argp, _nu)
+        body_dict[_name] = Body.from_parameters(_k, _name, _symbol, _R, _parent, _orbit)
+    
+
+for _body_name in body_dict:
+    _body = body_dict[_body_name]
+    if _body.parent != None and _body.orbit != None:
+        _body.calculate_soi()
+
+Sun = body_dict['Sun']
```
Juanlu001: Regarding the r_SOI of the major bodies: it's still questionable which value for the semimajor axis to take, but on the other hand this is just an approximate value. Do you think we can hardcode the important ones in the data file and leave the rest blank? There is a table in Battin (imperial system though).

src/poliastro/tests/test_maneuver.py
```diff
@@ -55,7 +55,7 @@ def test_hohmann_maneuver():
 
     man = Maneuver.hohmann(ss_i, Earth.R + alt_f)
 
-    assert_quantity_allclose(ss_i.apply_maneuver(man).ecc, 0 * u.one, atol=1e-15 * u.one)
+    assert_quantity_allclose(ss_i.apply_maneuver(man).ecc, 0 * u.one, atol=1e-14 * u.one)
```
Juanlu001: This one is failing because of the change in parameters?

src/poliastro/twobody/tests/test_states.py
```diff
@@ -110,7 +110,7 @@ def test_convert_from_coe_to_rv():
     assert_quantity_allclose(inc, expected_inc, rtol=1e-4)
     assert_quantity_allclose(raan, expected_raan, rtol=1e-4)
     assert_quantity_allclose(argp, expected_argp, rtol=1e-4)
-    assert_quantity_allclose(nu, expected_nu, rtol=1e-6)
+    assert_quantity_allclose(nu, expected_nu, rtol=1e-5)
```
Juanlu001: Same question as above

src/poliastro/tests/test_bodies.py
```diff
@@ -33,6 +37,42 @@ def test_body_printing_has_name_and_symbol():
...
+    expected_soi = 1 * u.km
+    assert_quantity_allclose(_soi.to(u.km), expected_soi)
+    
+def test_generalized_Titius_Bode_Law():
+    _bodies = ['Mercury', 'Venus', 'Earth', 'Mars', 'Ceres',
+               'Jupiter', 'Saturn', 'Uranus', 'Neptune']
+    au = 149597870.700 * u.km 
+    expected_approx = au * 0.2075 * 1.7327 ** arange(1,10)
+    obtained = []
+    for _name in _bodies:
```
Juanlu001: I would advise keeping the cyclomatic complexity of the tests as unity and instead parametrize them. See for example:

https://github.com/poliastro/poliastro/blob/v0.6.0/src/poliastro/tests/test_iod.py

src/poliastro/bodies_data.txt
```diff
@@ -0,0 +1,51 @@
...
+#
+#
+#       Stars or parentless Bodies:
+#
+Sun , ☉      , 132712440041, 695700 , None, None
+#
+#
+#       Planets of the Solar System 
+#
+Mercury, ☿      , 22031.78 , 2439.7 , Sun,  0.38709927, 0.20563593,  7.00497902, 252.25032350,  77.45779628,  48.33076593
```
Juanlu001: If following what I said above, we would rather exclude the orbit parameters from here and instead put the radius of the SOI. In the future we might be interested in adding J2 but that is out of scope for this PR.

src/poliastro/bodies.py
AunSiro: I don't like hardcoding the values or r_SOI for two reasons:
- It is not a physical measurable property, but an artificial radius that in the context of patched conics we decide that works well enough. It is a consequence of other values that we already have, not something you could measure in real life.
- If we calculate it trough a function, we can obtain it for any body with mass and semi-major axis, wether or not its value has been documented before. This means that we can use patched conics for mission planning for anything, from Trappist System to Jovian satellites exploration, or even fictional systems.

src/poliastro/tests/test_maneuver.py
AunSiro: Yes, the added precision for K parameter in DE430 made the test fail. Since the documentation of the test quotes a reference for the expected values, I thought that relaxing the tolerance was a better approach than changing the expected values.

src/poliastro/twobody/tests/test_states.py
AunSiro: Same answer...

src/poliastro/bodies.py
AunSiro: Yes, I agree with having it built into a function. I will search how to use those libraries, but it will require to expand the dependencies of Poliastro, are you ok with it?

src/poliastro/tests/test_bodies.py
AunSiro: Ok, I will try to reformulate with a parametric approach...

src/poliastro/bodies.py
Juanlu001: Yeah, it's okay. `csv` is in the `stdlib`, `pandas` would have to be installed but it's not a great deal I think. Try to go the easier route first.

src/poliastro/bodies.py
Juanlu001: Yes, I agree we will need a function to compute them anyway, for the reasons you explained. Maybe hardcoding them in some cases and computing them in others is an unnecessary inconsistency...

Juanlu001:
By the way, the conflicts appeared because I just merged #155. We can ignore them until the last moment.

Juanlu001:
We finally made a decision regarding the use of planetary constants in poliastro.

https://github.com/poliastro/poliastro/pull/197#issuecomment-321748009

So probably this PR will have to be re-purposed to think only what do we do with the radius of the sphere of influence. I suggest we have a conversation about this in Cáceres, @AunSiro :)

Juanlu001:
Also @AunSiro, notice that all this began with an email I sent you July 1st, 2013...

Juanlu001:
Since we [switched to Astropy to compute ephemerides](https://github.com/poliastro/poliastro/pull/155/commits/e353c4e46#diff-80029d89874aeee488c5a7c40d6e1c6f) (which include an approximate model that does not require external downloads) and merged [our latest modifications to the `bodies` module](https://github.com/poliastro/poliastro/blob/97e0a4f/src/poliastro/bodies.py) this PR has conflicts. I am closing it and suggesting an alternative course of action in #150.

Thanks again @AunSiro for your effort, this started many changes in poliastro that will be useful for many :pray: it was, perhaps, too ambitious :)

PR #164, opened 2017-04-18
Updating About me

franktoffel:
I was reading poliastro docs and found a small typo ;)


Juanlu001:
Haha good catch Fran! This could also be labeled as a bug :) Merging!

PR #169, opened 2017-05-03
Added more bodies of the solar system

Hroddric:
Added the rest of the planets of our Solar System to the bodies list.

### DIFF
--- a/src/poliastro/bodies.py
+++ b/src/poliastro/bodies.py
@@ -60,6 +60,30 @@ class Body(object):
 Sun = Body.from_parameters(
     None, k=132712440018 * u.km ** 3 / u.s ** 2,
     name="Sun", symbol=u"\u2609", R=695700 * u.km)
+Mercury = Body.from_parameters(
+    Sun, k=22032 * u.km ** 3 / u.s ** 2,
+    name="Mercury", symbol=u"\u263F", R=2439.7 * u.km)
+Venus = Body.from_parameters(
+    Sun, k=324859 * u.km ** 3 / u.s ** 2,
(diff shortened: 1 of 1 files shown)


src/poliastro/bodies.py
```diff
@@ -60,6 +71,40 @@ def _repr_latex_(self):
...
+Saturn = Body.from_parameters(
+    Sun, k=3793939 * u.km ** 3 / u.s ** 2,
+    name="Saturn", symbol=u"\u2644", R=58232 * u.km)
+Uranus = Body.from_parameters(
+    Sun, k=5793939 * u.km ** 3 / u.s ** 2,
+    name="Uranus", symbol=u"\u26E2", R=25362 * u.km)
+Neptune = Body.from_parameters(
+    Sun, k=6836529 * u.km ** 3 / u.s ** 2,
+    name="Neptune", symbol=u"\u2646", R=24622 * u.km)
+Pluto = Body.from_parameters(
```
Juanlu001: Pluto is not a planet 😜

Juanlu001:
I amend myself: the coverage has increased! But it's kind of misleading because the code is just running, nobody is checking if the values are what they should be.

Juanlu001:
Also, I forgot to say: the important values (those who need greater precision) are in Table 8 of the DE430 report. The radiuses of the bodies are not that important in my view, but still should be read from a reputable source and properly cited.

Juanlu001:
We finally made a decision regarding the use of planetary constants in poliastro.

https://github.com/poliastro/poliastro/pull/197#issuecomment-321748009

So it means we are closing this PR in favor of #197. I appreciate your effort and encourage you to keep on contributing to poliastro!

PR #171, opened 2017-05-20
Fixed two broken links

anhiga:


Juanlu001:
Good catch! Thank you very much 😊

PR #172, opened 2017-05-20
Docs update

anhiga:
- Updated API Reference section (api.rst) structure in order to include all the packages and modules, except for the _base, testing and jit modules, which seem only useful in development.

- Updated user_guide.rst internal references in order to comply with the new structure.

- Minor fixes like adding docstrings in examples.py and testing.py, and changing all "Notes" to "Note".

I'm fairly new on Github (not git), so I don't really know if I should do separate Pull Requests in this case.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -97,12 +97,6 @@ poliastro.hyper module
 .. automodule:: poliastro.hyper
     :members:
 
-poliastro.jit module
----------------------
-
-.. automodule:: poliastro.jit
-    :members:
-
 poliastro.maneuver module
 -------------------------
 
@@ -121,12 +115,6 @@ poliastro.stumpff module
 ------------------------
 
 .. automodule:: poliastro.stumpff
-    :members:
-
-poliastro.testing module
--------------------------
-
-.. automodule:: poliastro.testing
     :members:
 
 poliastro.util module
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -8,43 +8,43 @@ poliastro.twobody package
     :members:
 
 poliastro.twobody.angles module
-********************************
+*******************************
 
 .. automodule:: poliastro.twobody.angles
    :members:
 
 poliastro.twobody.classical module
-***********************************
+**********************************
 
 .. automodule:: poliastro.twobody.classical
    :members:
 
 poliastro.twobody.decorators module
-************************************
+***********************************
 
 .. automodule:: poliastro.twobody.decorators
    :members:
 
 poliastro.twobody.equinoctial module
-*************************************
+************************************
 
 .. automodule:: poliastro.twobody.equinoctial
    :members:
 
 poliastro.twobody.orbit module
-*******************************
+******************************
 
 .. automodule:: poliastro.twobody.orbit
    :members:
 
 poliastro.twobody.propagation module
-*************************************
+************************************
 
 .. automodule:: poliastro.twobody.propagation
    :members:
 
 poliastro.twobody.rv module
-****************************
+***************************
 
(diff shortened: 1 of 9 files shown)


docs/source/api.rst
```diff
@@ -1,32 +1,70 @@
 API Reference
-=============
+#############
```
Juanlu001: Let's better be consistent with the rest and keep the `===` headers for first level sections.

docs/source/api.rst
```diff
@@ -1,32 +1,70 @@
...
 
 poliastro.twobody package
 -------------------------
 
 .. automodule:: poliastro.twobody
+    :members:
+
+poliastro.twobody.angles module
+********************************
```
Juanlu001: Same as above. Also, I see length mismatch, have you tried compiling the docs with these changes? Cannot do it myself now.

docs/source/api.rst
```diff
@@ -35,24 +73,43 @@ poliastro.bodies module
     :members:
     :special-members: __init__
 
-poliastro.plotting module
+poliastro.cli module
+---------------------
```
Juanlu001: Length mismatch again, Sphinx might fail or warn.

src/poliastro/examples.py
```diff
@@ -16,12 +16,14 @@
     [7.37289205, 2.08223573, 4.39999794e-1] * u.km / u.s,
     time.Time("2013-03-18 12:00", scale='utc')
 )
+'''ISS orbit example'''
```
Juanlu001: Do these strings appear in Sphinx?

src/poliastro/twobody/angles.py
```diff
@@ -172,7 +172,7 @@ def fp_angle(nu, ecc):
     ecc : float
         Eccentricity.
 
-    Notes
+    Note
```
Juanlu001: We use "Notes" to be consistent with https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

docs/source/api.rst
anhiga: Completely right, I forgot to change it after doing some experiments. :)

docs/source/api.rst
anhiga: The `***` headers are a sublevel of `---`, but I think I should ask you these kind of changes in the chat before making a Pull Request, noob error. 😄
I have compiled the docs without errors using Sphinx, I think it only warns you when the line is too short. Anyway, I will change it, it's a good practice as you say.

src/poliastro/examples.py
anhiga: Yes, they appear, altough I only wrote those docstrings because the examples.py module appeared without any text in the API Reference.

src/poliastro/twobody/angles.py
anhiga: Same as above, I should ask you these changes in the chat. Only made this change because Sphinx automatically detects "Note" and shows it differently.

anhiga:
Thanks for such detailed comments. 😄  Regarding the order change in the API docs, I put the modules in alphabetical order, I'll try to modify the API.rst document.
I build the docs repeatedly without errors, should I attach some images showing the html as displayed in a browser? As I said, I'm new to Github, so I really appreciate your help. 😄

Juanlu001:
If you already built the docs with Sphinx locally that's fine, now that I'm at home again I can check it by myself :)

docs/source/api.rst
Juanlu001: Oh that's right, you added another level! Fair enough then :)

Juanlu001:
Checked, merging! Thanks @anhiga :rocket:

PR #173, opened 2017-05-21
Added function lagrange_points to threebody/restrictedCircularProblem

danbul:
So I implemented a function that returns the 5 Lagrange Points in the Restricted Circular Problem.
Since I didn't know where to put it, I've also created a new folder for the Three Body Problem. Probably there is a better place for it. And better names for all of these.

I'm looking of a way to prepare a good test that follows the same integration in the package as the other ones (I've already created a file for it), so this is not definitive, although I made some tests myself. 

About the implementation:

- I used some notes and slides from my Orbital Mechanics course, but I'm looking for access to a good reference, so the mathematics would follow something that can be consulted
- I had some problems for defining the plane where the orbits and the L points are. I chose to include a new parameter (`n`), the normal vector to that plane, but I think that this is not the best solution. I'm working on it

Final thoughts: this is in development, but I wanted to show what I did last night so I could hear the first comments on this.

TO DO:

- Prepare the test function
- Look for good references
- Write some information about the physics of the problem in the Wiki
- Look for a better way of defining the plane that contains the two orbits
- Look for how to integrate this correctly in the poliastro package


Juanlu001:
Hi Daniel! Thanks a lot for this pull request :)

I looked through it from my smartphone and have mostly minor style concerns, but overall it looks good! If you don't mind and to let you work on the important bits first, I prefer to hold the review until you add at least one test.

Perhaps there's info in "An Introduction to the Mathematics and Methods of Astrodynamics", by Battin. If I find another suitable reference I will tell you.

danbul:
So, while I was writing the test/example, I wondered about changing the API to something more similar to this:

`lagrange_points(main_body, secondary_body, time)`

instead of the actual API:

`lagrange_points(main_body_mass, main_body_position, sec_body_mass, sec_body_position, normal_to_orbit_plane)`

I think that this would be more consistent with the rest of the package, as it would use the Body class from `poliastro.bodies` and Time from `astropy.time`, and the usage would be easier. What do you think?

Some things would arise, as for the moment I haven't thought about how I would return the position of the Lagrange Points (they are vector and they need to be in a base). This may be a problem in this approach.

Juanlu001:
The first API looks cleaner, but it's limited to bodies of which you have available ephemerides. Two options here:

* I want to consider all sorts of bodies, so better do it in a generic way (version 2)
* The bodies of interest to compute the Lagrange points will be planetary bodies of the Solar System so I can just specify which ones (version 1)

Regarding the position of the Lagrange Points, you can choose to

* return them in the same frame as the input vectors (easy?)
* return them in the correct Coordinate Frame (hard)

The notion of vectors associated to specific coordinate frames requires the use of `SkyCoord` objects from Astropy:

http://docs.astropy.org/en/stable/api/astropy.coordinates.SkyCoord.html?highlight=cartesian

which allow transformations:

http://docs.astropy.org/en/stable/coordinates/transforming.html

But this is unexplored territory in poliastro, mainly because they still don't support velocities:

https://github.com/astropy/astropy/pull/5871

So maybe you might want to start with the easiest approach first and build from there.

Ivan9528:
Hello Daniel!!

Although I am quite new in poliastro, I have come up with some suggestions more related to astrodynamics than to the implementation after looking at the code. Here they go, I hope they are useful:

- Defining a proper reference frame is key. Although it does not matter which one you are using, you must be consistent along the code. Perhaps it is a good idea to let the user choose it as an input of the function. Some reference frames that I could suggest are: J2000, MEE2000, ecliptic, etc.
- As you noticed, it is compulsory to define the orbital plane of the two massive bodies but it could not be easy for users to give the normal vector to that plane directly. I propose to ask for two different positions of the body which is orbiting with respect to the other (Earth in the Sun-Earth system). Then, we will have solved the problem since two relative positions of the orbiting planet will give us the plane by crossing product them. 
- Regarding the test case, the book "An Introduction to the Mathematics and Methods of Astrodynamics" from Richard H. Battin could be a good reference to compare results of well-known 3 body systems.

Juanlu001:
Hi @Ivan9528, thanks for chiming in! Those are good suggestions :) @danbul has been busy and I don't think he will have time to finish this pull request (correct me if I'm wrong!) so everybody is welcome to take over (ask if you need help with git for that).

danbul:
Hi! Sorry, I didn't see these messages at that time.
I've been busy during this past year with several things and I couldn't work on this. This summer I'll have more time, so I'll probably try this again. This year I've learnt some more astrodynamics and got more confident in Python, so...

@Ivan9528 your suggestions are truly appreciated, I'll keep them in mind.
And @Juanlu001 , thank you for being understanding. Hopefully I'll finish this this summer, but as you said, everyone is welcome to participate in this!

Juanlu001:
Glad to read that @danbul! No hurries :)

Juanlu001:
By request of @danbul I gave another look at this pull request. I would like to revisit what I said here:

> * I want to consider all sorts of bodies, so better do it in a generic way (version 2)
> * The bodies of interest to compute the Lagrange points will be planetary bodies of the Solar System so I can just specify which ones (version 1)

We should notice that, at least as a first step, we should limit ourselves to the **circular** restricted three body problem. The general case is more complex, as [Danby "Stability of the Triangular Points in the Elliptic Restricted Problem of Three Bodies" (1963)](http://adsbit.harvard.edu/full/1964AJ.....69..165D) analyzed.

In this sense, the API that we provide should communicate that not all possible orbits are allowed, but only circular ones.

Juanlu001:
Perhaps we can have two "layers" here:

The first layer receives the collinear distance between the two big bodies as both masses, and returns the distances of L1, L2, L3, L4 and L5 to the primary body.

```
l_points = lagrange_points(r_12, m1, m2)
assert len(l_points) == 5  # Does this hold for all mass ratios and distances?
```

screenshot from 2018-08-15 21-26-30

On top of this, we could have another layer that receives the position vector between the attractor and the secondary body, both masses, and the normal vector:

```
moon_orbit = Orbit.from_body_ephem(Moon)
l_points_vec = lagrange_points_vec(moon_orbit.r, Moon.parent.mass, Moon.mass, moon_orbit.h_vec)
assert len(l_points_vec) == 5
```

Notice that:

* We don't have `Earth.mass` or `Moon.mass` yet, we have to add them!
* With `moon_orbit.r`, `Earth.mass` and `Moon.mass` we can compute the position of the barycenter
* With `moon_orbit.r` and `moon_orbit.h_vec` the orbital plane of an hypothetical circular orbit is perfectly defined
* We can return each Lagrange point in the same reference frame as `moon_orbit.r` and just note in the documentation that they are fixed in a specific point in time
* ...or we could compute a rotating reference frame and return the Lagrange points in it, but that might be too complicated

Also, notice that the only needed parameter here is... the `Body`!

```
l_points_earth_moon = lagrange_points_from_body(Moon)

...
def lagrange_points_from_body(body, epoch=None):
    body_orbit = Orbit.from_body_ephem(body, epoch)
    return lagrange_points_vec(body_orbit.r, body.parent.mass, body.mass, body_orbit.h_vec)
```

What do you think @danbul?

danbul:
Well, I think it's pretty well defined!

Regarding the first layer, there's still need for something that completely determines the orbital plane, since L4 and L5 are outside the line main_body - secondary_body.  We'll add to that layer the h vector, so easy fix. Also, this layer is your last comment is `lagrangian_points_vec` if I understand you, so already taken into account.

I think it'd be nice to add to the API something to select the Lagrangian points that are desired. Probably something like:
```python
def lagrangian_points_from_body(body, epoch=None, selec=None):
     ...
```
`selec` would be an argument that would take a list like `["L1", "L4"]` if the user desired L1 and L4 points. The default behavior would be to compute all of them. The reasoning behind this is that now L1, L2 and L3 are computed numerically, and that can be expensive. Also, may be the returned list of L points should be a dictionary so the user is sure of the computed points:
```python
l_points = {"L1": l1, "L4": l4}
```
Probably, for convenience, if only one point is requested, the function should return that point as an array, although that could be confusing.

Something important that you touched is to add some checks for validity. For example, I already know that there is a requirement on the mass ratio. Another one is the eccentricity of the orbit, so I'll look where is the limit in that.

Finally, regarding the frame of reference, I think we should start with the easy option 😅 It's true that a rotating frame of reference would be more interesting though.

Juanlu001:
Couple of comments:

You don't really need any vectors in the first one, because you just need the distances. The structure is known: L1, L2 and L3 are collinear (and the order is well defined) whereas L4 and L5 are triangular (and their positions are agreed). Then you can use the distances in the second layer to return proper vectors.

Regarding the option to return only a few of them, it looks a bit like premature optimization to me. I'd prefer to have the simple function first, then measure how much time does it take, then if we think it's "too slow" we can think how to make it faster.

Juanlu001:
Also, you don't need the eccentricity check: the eccentricity appears nowhere here :)

Juanlu001:
You're right that perhaps we should add some sort of check in `lagrange_points_from_body`, but the other two levels are "safe" in the sense that we never use the full orbit. Perhaps in the paper by Danby, or the papers that have cited it ever since, we can find some sensible upper limit. However, to help getting some of your work merged, I would start by implementing the first two levels, then we can merge, and we can investigate the third part a little bit. Deal? :)

danbul:
Deal :)

src/poliastro/constants.py
```diff
@@ -31,6 +31,9 @@
 
 J2000 = time.Time('J2000', scale='tdb')
 
+G = Constant('G', 'Gravitational Constant', 6.67428e-11, 'm3 / (kg * s2)', 0.00067e-11,
```
Juanlu001: This is already in `astropy.constants`! http://docs.astropy.org/en/stable/constants/#module-astropy.constants

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
...
+        aux *= rho1**2
+        aux /= rho**3 - rho1**3
+        aux /= (rho + rho1)**2
+        aux -= m1 / m2
+        return aux
+
+    l = np.zeros((5,))
+
+    # L1
+    rho2 = root(eq_L1, 0).x
```
Juanlu001: I read in the documentation that the default root finding method for `scipy.optimize.root` is "MINPACK’s hybrd and hybrj routines (modified Powell method)"

https://docs.scipy.org/doc/scipy/reference/optimize.root-hybr.html.

This method works with a wide variety of functions, in particular for non differentiable functions.

https://en.wikipedia.org/wiki/Powell%27s_method

I wonder if we could get better results using `scipy.optimize.newton`, even without providing the derivative.

In fact, according to Curtis "Orbital Mechanics for Engineering Students", we should be able to use an interval method, such as `scipy.optimize.brentq`, without the need of derivatives or initial guesses:

screenshot from 2018-08-17 19-31-38

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
+# coding: utf-8
+
+import numpy as np
+from scipy.optimize import root
+
+
+def lagrange_points(r12, m1, m2):
+    """Computes the Lagrangian points using the formulation
```
Juanlu001: Could we have a short line first, then more explanation in the body? As in https://numpydoc.readthedocs.io/en/latest/format.html#sections

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
+# coding: utf-8
```
Juanlu001: In Python 3 we don't need this anymore, please remove the line :)

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
...
+
+    # L3
+    rho1 = root(eq_L3, -1000).x
+    # rho2 = rho + rho1
+    l[2] = - rho1
+
+    # TODO: add checks to the results given by `root`
+    # TODO: reassure that the initial values given to `root` are good for all cases
+
+    l[3] = l[4] = 0.5 * r12
```
Juanlu001: Don't L4 and L5 distances depend on the mass ratio as well?
screenshot from 2018-08-17 19-37-14

(Again from Curtis)

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
...
+
+    return l
+
+
+def lagrange_points_vec(m1, r1_, m2, r2_, n_):
```
Juanlu001: Can we make this function receive units? `kg `, `km`, ...

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
...
+    n_ : array
+        Normal vector to the plane in which the two orbits of the main
+        and the secondary body are contained
+
+    Returns
+    -------
+    LP: list of all Lagrange points
+    """
+
+    r1 = np.asarray(r1_).reshape((3,))
```
Juanlu001: Why do we need this part?

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
...
+    r2 = np.asarray(r2_).reshape((3,))
+    n = np.asarray(n_).reshape((3,))
+
+    # Define local reference system:
+    # Center: main body
+    # x axis: points to the secondary body
+    ux = r2 - r1
+    r12 = np.linalg.norm(ux)
```
Juanlu001: If we make `ux` a vector with units, we can compute the norm using `poliastro.util.norm`.

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
...
+
+    # y axis: contained in the orbital plane, perpendicular to x axis
+    uy = np.cross(n, ux)
```
Juanlu001: But here beware, because `np.cross` does not retain units...

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
...
+    L1 = r1 + ux * x1
+    L2 = r1 + ux * x2
+    L3 = r1 + ux * x3
+    L4 = r1 + ux * x4 + uy * y45
+    L5 = r1 + ux * x5 - uy * y45
+
+    return [L1, L2, L3, L4, L5]
+
+
+def lagrange_points_from_body(body):
```
Juanlu001: If we are not implementing it for the moment, let's remove it. We won't forget :smiling_imp:

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
...
+    pass
+
+
+if __name__ == "__main__":
+
+    # Earth
+    r1 = np.array([0, 0, 0])
+    m1 = 5.972e24
```
Juanlu001: I think it would be great to have all these constants in `poliastro.constants`. In fact, we should derive them all from GM and G, as astropy does for a couple of planets:

https://github.com/astropy/astropy/blob/b24b87fc847e568c7389350d2d1deb7ca5ed9455/astropy/constants/iau2015.py#L85-L90

src/poliastro/threebody/restricted.py
```diff
@@ -0,0 +1,173 @@
...
+    res = lagrange_points_vec(m1, r1, m2, r2, n)
+    res1 = np.array(res) / d
+
+    [print("{:+10.5e} , {:+10.5e} , {:+10.5e}".format(r[0], r[1], r[2]))
+     for r in res]
+
+    x = np.array([r[0] for r in res])
+    y = np.array([r[1] for r in res])
+
+    # import matplotlib.pyplot as plt
```
Juanlu001: I love this plot :wink: perhaps we can add a function in the `poliastro.plotting` module to plot the Lagrange points of a system like this!

src/poliastro/threebody/restricted.py
danbul: imagen

This is from Battin, so... May be Curtis has more updated information.
Also, I'd like to give it a look before jumping into conclusions, but maybe Curtis' equation uses a different frame of reference? With the origin in the barycenter? I'll give it a look

src/poliastro/threebody/restricted.py
danbul: Makes sense

src/poliastro/threebody/restricted.py
danbul: I had the same in mind. I'm on it!

src/poliastro/threebody/restricted.py
danbul: 😉

src/poliastro/threebody/restricted.py
danbul: Maybe we could add a wrapper to `poliastro.util` that retains the units

src/poliastro/threebody/restricted.py
danbul: I think it was for the `np.cross` function

src/poliastro/threebody/restricted.py
danbul: Yeah, on it

src/poliastro/threebody/restricted.py
danbul: I tried `scipy.optimize.newton` but it didn't work very well, probably because of the initial guess. I'll look into `brentq`

Juanlu001: Notice that `brentq` will need an interval and not an initial guess, so you will need a slightly different approach

src/poliastro/threebody/restricted.py
Juanlu001: Oh, interesting! Whatever the result is, please leave a comment explaining it so nobody makes the mistake of changing it in the future.

src/poliastro/threebody/restricted.py
Juanlu001: I think `util` is better for small functions that are used everywhere. I'd rather have `lagrange_points_vec` here.

danbul: Oh, I meant to add a wrapper of `np.cross` that accepted units to `util`. For the moment, I'll keep it inside the scope of `lagrange_points_vec`

src/poliastro/threebody/restricted.py
danbul: Yeah, Curtis uses a frame of reference (*comoving coordinate system*) with the origin in the barycenter (Fig 2.33), so the formulation seems to be consistent.

src/poliastro/threebody/restricted.py
danbul: So I looked into Curtis formulation to try to reproduce that plot, because he's formulation is adimensional and would simplify the boundary values for `brentq` or the initial estimations for `newton` and `root`. There's no problem in changing to this formulation in the deeper layer, so I might as well play with it.

I published a gist to do this and test the different root finders in `scipy.optimize`, because I'd like to use always the same.

https://gist.github.com/danbul/3e0afc694b8f4d3866a3632b4347dc03

After this quick study, I found that  `brentq` works perfectly for L2 and L3, but performs horribly for L1. For L1, only `root` with a moving initial estimation works well.

So... what do you think @Juanlu001 ?

PS: I couldn't get `newton` to work, so maybe I'm missing the hero of this story.

Juanlu001: Thanks for investigating this! I would also prefer that we stick to `brentq`, let me see if I can help.

src/poliastro/threebody/restricted.py
Juanlu001: Oh, you're right! Didn't think of it, that would be a neat addition. But yes, let's leave it for a future PR.

Juanlu001:
Notice that there are some PEP8 errors :) https://circleci.com/gh/poliastro/poliastro/454 (you can see them in the failed "ci/circleci: quality" check

danbul:
After playing around a bit, I was able to stick to `brentq`, so this is also solved! What I did at the end is change the boundaries depending on the value of pi2 to avoid the two divergences in the function `eq_L123`.

danbul:
After playing around a bit, I was able to stick to `brentq`, so this is also solved! What I did at the end is change the boundaries depending on the value of pi2 to avoid the two divergences in the function `eq_L123`

Juanlu001:
Great! Although I would like to have a proper test for this and move the plotting function elsewhere, I prefer to merge this already and make the edits later myself. What I would do is make a pull request and ask for your review, @danbul. Do you agree? That way the work won't be interrupted by the beginning of the classes :)

danbul:
Fine by me! Also, I don't have a problem doing it myself if you want @Juanlu001 

The __main__ thing was like the test and the plotting, so it should be very easy to put elsewhere.

Issue #176, opened 2017-05-23
Add instructions to build docs to developer guide


Juanlu001:
This is another point for a developer guide, and should also be in the continuous integration. Looking for inspiration.

Juanlu001:
#178 already put CI in place, now we would need more docs. Changing name.

PR #178, opened 2017-05-25
Added doc building CI

anhiga:
Added docs to continuous integration. I don't know if further tests can be required, but I think this is a good starting point.
Related to #176.


src/poliastro/util.py
```diff
@@ -43,7 +43,7 @@ def rotate(vector, angle, axis='z', unit=None):
     vector while the coordinate system remains unchanged. To do the opposite
     operation (*passive* or *alias* transformation) call the function as
     ``rotate(vec, ax, -angle, unit)`` or use the convenience function
-    :py:func:`transform`, see [1].
+    :py:func:`transform`, see [1]_.
```
Juanlu001: Good catch!

Juanlu001:
I love this! Thanks for the pull request, merging :muscle:

PR #179, opened 2017-05-29
Adding Intersphinx extension

anhiga:
Changed conf.py and docstrings in many modules in order to add external references to other documentation pages.


Juanlu001:
I never thought on doing this with the poliastro docs, it's awesome! I think the fixes for conda will be deployed this week, so let's merge this and see in a while how it goes :+1:

PR #181, opened 2017-06-04
Rtd fix

anhiga:
Added necessary dependencies in environment.yml to run nbsphinx.

There is still an issue regarding depth levels in the TOC tree, but it is due to sphinx_rtd_theme being pinned to version 0.1.7 in RTD. Since [v0.1.8](https://github.com/rtfd/sphinx_rtd_theme#v0-1-8) there is support for third and fourth level headers, so it will be fixed when dependencies are unpinned.


Juanlu001:
Manually squashed and merged the pull request :) Thanks!

PR #188, opened 2017-07-27
WIP: added neos.py to be reviewed

anhiga:
- Exceptions, prints, and error handling in general should be unified.

- `get_orbit_from_name(name)` is a virtually empty function.

### DIFF
--- a/src/poliastro/neos.py
+++ b/src/poliastro/neos.py
@@ -13,6 +13,10 @@ from poliastro.twobody.orbit import Orbit
 from poliastro.bodies import Sun
 from poliastro.twobody.angles import M_to_nu
 
+#Base URLs
+NEOWS_URL = 'https://api.nasa.gov/neo/rest/v1/neo/'
+SBDB_URL = 'https://ssd.jpl.nasa.gov/sbdb.cgi'
+
 def get_orbit_from_spk_id(spk_id, epoch=None):
     """Return `~poliastro.twobody.orbit.Orbit` given a SPK-ID.
 
@@ -30,10 +34,9 @@ def get_orbit_from_spk_id(spk_id, epoch=None):
         NEA orbit..
 
     """
-    url = 'https://api.nasa.gov/neo/rest/v1/neo/'
     payload = {'api_key' : 'DEMO_KEY'}
 
-    response = requests.get(url + spk_id, params=payload)
+    response = requests.get(NEOWS_URL + spk_id, params=payload)
     response.raise_for_status()
     if response.status_code == 200:
         orbital_data = response.json()['orbital_data']
@@ -71,12 +74,9 @@ def get_spk_id_from_name(name):
         SPK-ID number.
 
     '''
-    url = 'https://ssd.jpl.nasa.gov/sbdb.cgi'
-    headers = {'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_1) \
-        AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.95 Safari/537.36'}
     payload = {'sstr' : name, 'orb' : '0', 'log' : '0', 'old' : '0', 'cov' : '0', 'cad' : '0'}
-    
-    response = requests.get(url, params=payload, headers=headers)
+
+    response = requests.get(SBDB_URL, params=payload)
     response.raise_for_status()
     soup = BeautifulSoup(response.text, "html.parser")
 
@@ -102,7 +102,7 @@ def get_spk_id_from_name(name):
         print('\nPlease select one of them')
     else:
         print('Object could not be found. You can visit: ',
-                url + "?sstr=" + name, " for more information.")
+                SBDB_URL + "?sstr=" + name, " for more information.")
 
 def get_orbit_from_name(name):
     '''Return `~poliastro.twobody.orbit.Orbit` given a name.
--- a/src/poliastro/neos.py
+++ b/src/poliastro/neos.py
@@ -19,7 +19,7 @@ from poliastro.twobody.angles import M_to_nu
 NEOWS_URL = 'https://api.nasa.gov/neo/rest/v1/neo/'
 SBDB_URL = 'https://ssd.jpl.nasa.gov/sbdb.cgi'
 
-def get_orbit_from_spk_id(spk_id, epoch=None):
+def orbit_from_spk_id(spk_id, epoch=None):
     """Return `~poliastro.twobody.orbit.Orbit` given a SPK-ID.
 
     Retrieve info from NASA NeoWS API, and therefore
@@ -59,7 +59,7 @@ def get_orbit_from_spk_id(spk_id, epoch=None):
         return Orbit.from_classical(attractor, a, ecc, inc,
                                     anl, parg, nu, epoch)
 
-def get_spk_id_from_name(name):
+def spk_id_from_name(name):
     '''Return SPK-ID number given a small-body name.
     
     Retrieve and parse HTML from JPL Small Body Database
@@ -105,7 +105,7 @@ def get_spk_id_from_name(name):
         warnings.warn('Object could not be found. You can visit: '
                       + SBDB_URL + '?sstr=' + name + ' for more information.')
 
-def get_orbit_from_name(name):
+def orbit_from_name(name):
     '''Return `~poliastro.twobody.orbit.Orbit` given a name.
 
     Retrieve info from NASA NeoWS API, and therefore
@@ -122,6 +122,6 @@ def get_orbit_from_name(name):
         NEA orbit.
 
     '''
-    spk_id = get_spk_id_from_name(name)
+    spk_id = spk_id_from_name(name)
     if spk_id is not None:
-        return get_orbit_from_spk_id(spk_id)
+        return orbit_from_spk_id(spk_id)
--- a/src/poliastro/neos.py
+++ b/src/poliastro/neos.py
@@ -47,17 +47,17 @@ def orbit_from_spk_id(spk_id, epoch=None):
     a = float(orbital_data['semi_major_axis']) * u.AU
     ecc = float(orbital_data['eccentricity']) * u.one
     inc = float(orbital_data['inclination']) * u.deg
-    anl = float(orbital_data['ascending_node_longitude']) * u.deg
-    parg = float(orbital_data['perihelion_argument']) * u.deg
+    raan = float(orbital_data['ascending_node_longitude']) * u.deg
+    argp = float(orbital_data['perihelion_argument']) * u.deg
     M = float(orbital_data['mean_anomaly']) * u.deg
     nu = M_to_nu(M.to(u.rad), ecc)
 
     if epoch is None:
         return Orbit.from_classical(attractor, a, ecc, inc,
-                                    anl, parg, nu)
+                                    raan, argp, nu)
     else:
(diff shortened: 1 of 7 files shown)


src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
+'''Code related to NEOs
+
+All new functions coded as part of SOCIS 2017 proposal
+
+'''
+import re
+from bs4 import BeautifulSoup
```
Juanlu001: this dependency is missing from setup.py and the conda package

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+import requests
```
Juanlu001: same as above

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+    Returns
+    -------
+    orbit : ~poliastro.twobody.orbit.Orbit
+        NEA orbit..
+
+    """
+    url = 'https://api.nasa.gov/neo/rest/v1/neo/'
+    payload = {'api_key' : 'DEMO_KEY'}
+    try:
+        response = requests.get(url + spk_id, params=payload)
```
Juanlu001: fails in certain cases:

```
In [3]: get_orbit_from_name("halley")
The following  2  objects match your search string:

2688 Halley (1982 HG1)
1P/Halley

Please select one of them
TypeError                                 Traceback (most recent call last)
<ipython-input-3-792d9803dfbb> in <module>()
----> 1 get_orbit_from_name("halley")

~/Personal/poliastro/src/poliastro/neos.py in get_orbit_from_name(name)
    139 
    140     '''
--> 141     return get_orbit_from_spk_id(get_spk_id_from_name(name))

~/Personal/poliastro/src/poliastro/neos.py in get_orbit_from_spk_id(spk_id, epoch)
     34     payload = {'api_key' : 'DEMO_KEY'}
     35     try:
---> 36         response = requests.get(url + spk_id, params=payload)
     37         if response.status_code == 200:
     38             orbital_data = response.json()['orbital_data']

TypeError: must be str, not NoneType
```

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+            nu = M_to_nu(M.to(u.rad), ecc)
+
+            if epoch is None:
+                return Orbit.from_classical(attractor, a, ecc, inc,
+                                        anl, parg, nu)
+            else:
+                return Orbit.from_classical(attractor, a, ecc, inc,
+                                        anl, parg, nu, epoch)
+        elif response.status_code == 404:
+            print('\nError 404. Object ID does not exists, or it does not match a NEA.')
```
Juanlu001: Should raise an error. What about this? http://docs.python-requests.org/en/latest/api/#requests.Response.raise_for_status

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+        else:
+            print('\nForbidden or unauthorized request: ', response.status_code)
```
Juanlu001: same as above

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+        return None
```
Juanlu001: `return None` is the same as `return` and is the same as doing nothing, because https://stackoverflow.com/a/15300671/554319

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+    except requests.exceptions.RequestException:
```
Juanlu001: The docstring of this error is "There was an ambiguous exception that occurred while handling your request.". Maybe we shouldn't capture it at all.

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+        print('\nSome kind of connection error ocurred, please try later.')
+        return None
+    except KeyError:
+        print('Internal error, please report it to Poliastro team')
```
Juanlu001: If you capture the error like this, there is no traceback and nothing to report to the team :) Better don't catch it and people will come to open an issue.

It's important to keep in mind what the intended audience of poliastro is, at least now: if they could download and install poliastro, especially the development version, they will understand how to open an issue in the repository. I admit this does not work for everybody, but let's create a developer-friendly tool first and later we can add a user-friendly layer above.

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+    name : str
+        Small-body object name. Wildcards "*" and/or "?" can be used.
+
+    Returns
+    -------
+    spk_id : str
+        SPK-ID number.
+
+    '''
+    url = 'https://ssd.jpl.nasa.gov/sbdb.cgi'
```
Juanlu001: I would store this as a module constant

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+    headers = {'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_1) \
```
Juanlu001: Same as above

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+
+            complete_string = ''
+            for string in data[1].stripped_strings:
+                complete_string += string + ' '
+            regex = re.compile('Classification: ([\S\s]+) SPK-ID: (\d+)')
+            return regex.match(complete_string).group(2)
+        # If there is a 'center' sibling, it is a page with a list of possible objects
+        elif page_identifier.find_next_sibling('center') is not None:
+            object_list = page_identifier.find_next_sibling('center').table.find_all('td')
+            print('The following ', len(object_list), ' objects match your search string:\n')
```
Juanlu001: It's OK to print diagnostic messages like this, but the problem is that this will only work interactively. If for instance we add an `input()` call somewhere, it should be outside of this function. Now or later we will have to break this down into smaller pieces, make only functions that don't print or ask for input and then add a script on top of them that calls the necessary steps.

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+            for body in object_list:
+                #body.a['href']
+                print(body.string)
+            print('\nPlease select one of them')
+        else:
+            print('Object could not be found. You can visit: ',
```
Juanlu001: Very nice. I would print a warning here though https://docs.python.org/3/library/warnings.html

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+                 url + "?sstr=" + name, " for more information.")
+        return None
```
Juanlu001: same as above

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+    except requests.exceptions.RequestException:
```
Juanlu001: same as above

src/poliastro/neos.py
```diff
@@ -0,0 +1,141 @@
...
+        print('Some kind of connection error ocurred, please try later.')
+        return None
+    except (AttributeError, KeyError):
+        print('Object could not be found. You can visit: ',
```
Juanlu001: same as above

src/poliastro/neos.py
Juanlu001: Options:

a) Fail now if there are more than one candidate body.
b) Add another parameter to the function so the user can anticipate the selection
c) Retrieve all IDs as a list, or a generator

Juanlu001: Probably others, I don't have a strong preference towards any of them. Up to you :)

anhiga:
I tried to fix all the errors.
Regarding what happen if several objects are found, I have decided to remove the code for the moment:

- I don't think the user can anticipate the selection, and
- retrieving all IDs as a list could be risky, since we should request every object page, and when using wildcards the object list can be really long. (`*a*` yields 16000 results, 720MB of requests 😆 )

Juanlu001:
Yeah, retrieving all IDs definitely is not a good idea :) but what about raising an Exception with the possible body names? That way the flow is immediately interrupted, and the user can see the possibilities. If you want to bring it back you can use `git revert`

anhiga:
I have just noticed some other errors, so I will do that and will make some more commits.

src/poliastro/neos.py
```diff
@@ -30,39 +36,28 @@ def get_orbit_from_spk_id(spk_id, epoch=None):
...
-            parg = float(orbital_data['perihelion_argument']) * u.deg
-            M = float(orbital_data['mean_anomaly']) * u.deg
-            nu = M_to_nu(M.to(u.rad), ecc)
-
-            if epoch is None:
-                return Orbit.from_classical(attractor, a, ecc, inc,
+
+    response = requests.get(NEOWS_URL + spk_id, params=payload)
+    response.raise_for_status()
+    if response.status_code == 200:
```
anhiga: Good point. Thanks!

src/poliastro/tests/test_neos.py
```diff
@@ -0,0 +1,40 @@
...
+            'aphelion_distance': '1.782461395911054',
+            'perihelion_time': '2457873.186399333365',
+            'mean_anomaly': '71.28027812836476',
+            'mean_motion': '.5598795239089109',
+            'equinox': 'J2000'
+        }
+    }
+
+    # mock_response.status_code = 200
+    mock_response.json.return_value = mock_orbital_data
```
Juanlu001: What about `mock_response = mock_get.return_value` and you save the `Mock` on line 10?

Juanlu001: It doesn't work. But this does:

```diff
diff --git a/src/poliastro/tests/test_neos.py b/src/poliastro/tests/test_neos.py
index 2adea4f..3e12261 100644
--- a/src/poliastro/tests/test_neos.py
+++ b/src/poliastro/tests/test_neos.py
@@ -5,9 +5,8 @@ import astropy.units as u
 from poliastro import neos
 
 @mock.patch('poliastro.neos.requests.get')
-def test_orbit_from_spk_id_has_proper_values(mock_get):
-
-    mock_response = mock.Mock(requests.Response())
+@mock.patch('poliastro.neos.requests.Response')
+def test_orbit_from_spk_id_has_proper_values(mock_response, mock_get):
 
     mock_orbital_data = {
         'orbital_data': {
```

src/poliastro/neos.py
```diff
@@ -0,0 +1,127 @@
...
+            complete_string += string + ' '
+        regex = re.compile(r'Classification: ([\S\s]+) SPK-ID: (\d+)')
+        return regex.match(complete_string).group(2)
+    # If there is a 'center' sibling, it is a page with a list of possible objects
+    elif page_identifier.find_next_sibling('center') is not None:
+        object_list = page_identifier.find_next_sibling('center').table.find_all('td')
+        bodies = ''
+        for body in object_list:
+            bodies += body.string + '\n'
+        raise Exception('Several bodies found:\n' + bodies)
```
Juanlu001: In case there are too many bodies, select the first three ones and write `...` or something

src/poliastro/neos.py
```diff
@@ -0,0 +1,127 @@
...
+    else:
+        warnings.warn('Object could not be found. You can visit: '
```
Juanlu001: In retrospect, this was a bad decision because the flow continues. We should raise an error here.

Juanlu001:
> a) Add a script (an entry point in `setup.py`) so this can be used from the command line

On second thought, we can leave that for the future.

PR #197, opened 2017-08-07
New bodies

anhiga:
Added new bodies with data references.

### DIFF
--- a/src/poliastro/bodies.py
+++ b/src/poliastro/bodies.py
@@ -17,19 +17,10 @@ Contains some predefined bodies of the Solar System:
 
 and a way to define new bodies (:py:class:`~Body` class).
 
-Data references are the following:
-
-* Equatorial radius, obtained from `NASA Planetary Fact Sheets`_.
-* Mass paremeter, obtained from `JPL Development Ephemeris 424`_.
-* Symbols, following this `guideline`_.
-
-.. _`NASA Planetary Fact Sheets`: https://nssdc.gsfc.nasa.gov/planetary/planetfact.html
-.. _`JPL Development Ephemeris 424`: ftp://ssd.jpl.nasa.gov/pub/eph/planets/ioms/de424.iom.pdf
-.. _`guideline`: https://solarsystem.nasa.gov/galleries/solar-system-symbols
+Data references can be found in :py:mod:`~poliastro.constants`
 """
-from astropy.constants import R_earth
 from astropy import units as u
-
+from poliastro import constants
 
 class Body(object):
     """Class to represent a body of the Solar System.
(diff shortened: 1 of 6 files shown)


src/poliastro/bodies.py
```diff
@@ -5,9 +5,27 @@
...
 * Earth (♁)
+* Moon (☾)
+* Mercury (☿)
+* Venus (♀)
+* Mars (♂)
+* Jupiter (♃)
+* Saturn (♄)
+* Uranus (⛢)
+* Neptune (♆)
+* Pluto (♇)
```
Juanlu001: Are we adding Ceres and other dwarf planets? :stuck_out_tongue:

anhiga: haha you are right! But, in [DE430] linked in discussions above, there is a table like this:
imagen
so it appears that even for JPL, Pluto has a higher status than Ceres and other dwarf planets 😄

anhiga:
Right! I'll try to improve reference sources.

src/poliastro/bodies.py
Juanlu001: Good catch!! I demand a blank line separating the Sun from the planets and another one separating Pluto from the planets then :smiling_imp:

Juanlu001:
After thinking about this for months ([years?](https://physics.stackexchange.com/q/69792/7641)) we finally decided to use two main sources for the planetary constants:

* "The IAU 2009 system of astronomical constants" (DOI 10.1007/s10569-011-9352-4), which specifies the mass of all planets and some minor bodies with respect to the **measured** mass of the Sun and the Earth.
* The latest "Report of the IAU Working Group on Cartographic Coordinates and Rotational Elements" (DOI 10.1007/s10569-010-9320-4), which specifies the size and shape parameters of the planets and minor bodies ([list of all IAU - WGCCRE reports](https://astrogeology.usgs.gov/groups/IAU-WGCCRE)).

In the process, we have decided against using:

* The constants used in DE430, because that somewhat implies we are tied to a particular set of ephemerides (which we are not).
* The _nominal_ values proposed by IAU 2015 Resolution B3 and adopted by Astropy, which seem unapropriate for this kind of application (and also would lead to the inconsistency of using some nominal values and some measured values).

This settles a long discussion and paves the way to @anhiga to do the final implementation 😄

anhiga:
For some reason, `test_hohmann_maneuver` and `test_bielliptic_maneuver` fail when Earth radius is defined in `km`, but not when defined in `m`. I'm trying to solve it.

Juanlu001:
That looks like a bug! Keep us posted :)

anhiga:
More precisely, fails occur when `k` and `radius` are in different units (km and m, for example). It seems to be due to [astropy not automatically decomposing units in multiplication and division](http://docs.astropy.org/en/stable/units/quantity.html#multiplication-and-division). In `hohmann`:

```Python
In [9]: t_trans = np.pi * np.sqrt((r_i * (1 + R) / 2) ** 3 / k)
In [10]: t_trans
Out[10]: <Quantity 0.5984347217136032 km(3/2) s / m(3/2)>
```
I guess that adding [`decompose()`](http://docs.astropy.org/en/stable/api/astropy.units.quantity.Quantity.html#astropy.units.quantity.Quantity.decompose), will solve the problem.

src/poliastro/constants.py
```diff
@@ -0,0 +1,111 @@
+"""Astronomical and physics constants in SI units.
+
+This module complements constants defined in `astropy.constants`
+
+Unless otherwise specified, gravitational and mass parameters were obtained from:
```
Juanlu001: Isn't it the other way around? Masses and radiuses are flipped I think.

src/poliastro/constants.py
```diff
@@ -0,0 +1,111 @@
...
+
+.. _`DOI: 10.1007/s10569-010-9320-4`: http://dx.doi.org/10.1007/s10569-010-9320-4
+.. _`DOI: 10.1007/s10569-011-9352-4`: http://dx.doi.org/10.1007/s10569-011-9352-4
+
+"""
+
+from astropy.constants import Constant
+
+
+G = Constant('G', 'Constant of gravitation', 6.67428e-11, 'm3 / (kg * s2)', 6.7e-15,
```
Juanlu001: Should we use the one from Astropy, which is more recent? What do you think?

src/poliastro/constants.py
```diff
@@ -0,0 +1,111 @@
...
+             'IAU 2009 system of astronomical constants', system='si')
+
+GM_sun = Constant('GM_sun', 'Heliocentric gravitational constant', 1.32712442099e20, 'm3 / (s2)', 1.0e10,
```
Juanlu001: In Astropy they use the same exponent for the uncertainty. I think it's a good practice that we can keep.

src/poliastro/constants.py
Juanlu001: Also, if some of the masses are of the whole planetary system, we should say so.

Juanlu001:
By the way, the codecov bot is silent because there is no base report:

https://codecov.io/gh/poliastro/poliastro/pulls

src/poliastro/constants.py
anhiga: Ok, I'll remove it then. Or shoud I add it to poliastro as well?

src/poliastro/constants.py
anhiga: You are right!

Juanlu001:
Merging!

PR #199, opened 2017-08-07
Plot attractor according to orbits size

### LINKED ISSUE #130: Fix attractor size when plotting different orbits

anhiga:
Trying to fix #130, I made some changes to `plotting` module.

When a new orbit is plotted, if the new attractor size is smaller than current attractor size, the attractor is replaced, in order to avoid plottings where attractor is bigger than some orbits:

.

### DIFF
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -59,6 +59,7 @@ class OrbitPlotter(object):
         self.num_points = num_points
         self._frame = None
         self._states = []
+        self._attractor_radius = None
 
     def set_frame(self, p_vec, q_vec, w_vec):
         """Sets perifocal frame if not existing.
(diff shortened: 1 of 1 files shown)


src/poliastro/plotting.py
```diff
@@ -83,6 +84,25 @@ def set_frame(self, p_vec, q_vec, w_vec):
         else:
             raise NotImplementedError
 
+    def set_attractor(self, orbit):
+        """Sets plotting attractor.
```
Juanlu001: Right thing here, double quotes for docstrings.

src/poliastro/plotting.py
```diff
@@ -91,6 +111,14 @@ def plot(self, orbit, osculating=True, label=None):
         if not self._frame:
             self.set_frame(*orbit.pqw())
 
+        # if new attractor radius is smaller, plot it
```
Juanlu001: All right!

Juanlu001:
Also, pay attention to the codeclimate warnings. Sometimes it's not worth it to fix them all, but they give useful hints. They are based on the ratings by this library:

https://radon.readthedocs.io/en/latest/commandline.html

You can read on about the general topic of code quality in Spanish here (a bit dated, but still useful)

http://iwt2-javierj.tumblr.com/post/69258809610/analizando-scikit-aero-en-python-2x

anhiga:
Here is the photo: 
figure_1

PR #208, opened 2017-08-16
Neows epoch

anhiga:
Neows returns a `orbit_determination_date` which is used as epoch.
Also I would like to try codecov 😄


Juanlu001:
Awesome! Merging this right away :grin:

PR #210, opened 2017-08-16
Removed conda in Travis CI

anhiga:
Removed conda in Travis. Pandoc is now installed through package manager.
Let's test build time in Travis and see if this is worth.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,15 +2,7 @@
 # is setting up a generic environment
 # This is not a problem because Python will be installed through conda
 
-language: generic
-
-os:
-  - linux
-  - osx
-
-env:
-  - PYTHON=3.5
-  - PYTHON=3.6
+language: python
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


anhiga:
After trying several options (commit log is really a mess, I'll try to make it clearer when PR got merged, if merged) these are reduced:
* Linux: from ~4' 30'' to ~2' 30''
The main problem is that, without conda, OSX only builds with Python 3.6.

Juanlu001:
> Linux: from ~4' 30'' to ~2' 30''

Awesome!

> The main problem is that, without conda, OSX only builds with Python 3.6.

I think we can afford that. In fact, I would also remove Windows x Python 3.5 builds and keep only the latest Python for non-Linux operative systems.

anhiga:
Ok, I'll try to make commit log clearer and remove Windows Python3.5 build.
I'll ping you when is done.

anhiga:
@Juanlu001 It is ready to be reviewed.

.travis.yml
```diff
@@ -20,22 +12,25 @@ branches:
...
+      language: generic
+    - os: linux
+      python: 3.5
+    - os: linux
+      python: 3.6
+
+addons:
+  apt:
+    packages:
+      - pandoc
```
Juanlu001: :heart:

Juanlu001:
For the record: just five days ago a pull request was merged that will enable automatic generation of SciPy wheels for Windows, for all Python versions and architectures.

https://github.com/scipy/scipy/pull/7616

So perhaps as early as this year we will be able to do `pip install scipy` in **all** platforms, without worrying anymore... This is the most exciting infrastructure change I have seen to SciPy since I am following the project, and I think it will be an inflection point.

I am just so happy :blush:

PR #212, opened 2017-08-16
Add PEP8 and MyPy to CI

### LINKED ISSUE #170: Try to compile with MyPy
"Mypy is an experimental optional static type checker for Python that aims to combine the benefits of dynamic (or "duck") typing and static typing". A new version is out:

http://mypy-lang.blogspot.com.es/2017/05/mypy-0510-released.html

The first question is: does this make any sense? The only way to find out is to try it at least once and decide.

Adding stubs might be needed to improve editor type hints and catch bugs:

https://www.caktusgroup.com/blog/2017/02/22/python-type-annotations/

For external libraries, there's `typeshed` but it includes none of our core dependencies:

https://github.com/python/typeshed

Though this might be a good starting point

https://github.com/machinalis/mypy-data

This is a very, very good article on good practices:

http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/

Ideally this would also be integrated in continuous integration, though I don't know how.

### LINKED ISSUE #175: Add PEP8 checks to CI and guide to docs
This would be useful not only in a future developer guide but also as an automated check. There are two routes: reporting failures, or fixing them.

https://gist.github.com/MichaelCurrie/802ce28c993ff2dd632c

Juanlu001:
This should fix #170 and fix #175, but before we can merge this we will have to fix the tons of errors that will appear :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,11 +35,14 @@ before_install:
 install:
   - pip install numpy  # Required
   - pip install .  # Test installation correctness
+  - pip install pep8 mypy
   - pip install sphinx sphinx_rtd_theme nbsphinx ipython  # Install documentation dependencies
 
 script:
+  - pep8 setup.py src/  # Check PEP8 compliance
+  - mypy --ignore-missing-imports --check-untyped-defs src/  # Check MyPy diagnostics
   - python setup.py test -vv  # Test against installed code
-  - sphinx-build -vW -b html docs/source _HTMLTest #Test docs build
+  - sphinx-build -vW -b html docs/source _HTMLTest  # Test docs build
 
 after_success:
   # Uninstall to test coverage against sources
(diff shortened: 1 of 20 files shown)


.travis.yml
```diff
@@ -35,11 +35,14 @@ before_install:
 install:
   - pip install numpy  # Required
   - pip install .  # Test installation correctness
+  - pip install pep8 mypy
   - pip install sphinx sphinx_rtd_theme nbsphinx ipython  # Install documentation dependencies
 
 script:
+  - pep8 setup.py src/  # Check PEP8 compliance
+  - mypy --ignore-missing-imports --check-untyped-defs src/  # Check MyPy diagnostics
   - python setup.py test -vv  # Test against installed code
```
Juanlu001: Let's put this under `before_script`

Issue #205, opened 2017-08-11
Problem with different epochs


anhiga:
If you plot orbits with different epochs, only the last epoch is showed in the title:
```Python
In [1]: from poliastro.neos import dastcom5
In [2]: from poliastro.plotting import OrbitPlotter
In [3]: eros = dastcom5.orbit_from_name('eros')[0]
In [4]: apophis = dastcom5.orbit_from_name('apophis')[0]
In [5]: eros.epoch
Out[5]: <Time object: scale='utc' format='jd' value=2451960.5>
In [6]: apophis.epoch
Out[6]: <Time object: scale='utc' format='jd' value=2454441.5>
In [7]: frame = OrbitPlotter()
In [8]: frame.plot(eros, label='eros')
In [9]: frame.plot(apophis, label='apophis')
```
![epochs](https://user-images.githubusercontent.com/18404378/29220971-3c574f2a-7ebd-11e7-86c2-667abc6faaee.png)

I think you should be able to choose the epoch. Maybe we should use `Orbit.propagate` function.

Juanlu001:
After discussion in the wiki (see https://github.com/poliastro/poliastro/wiki/Plotting-multiple-orbits:-use-cases) I'm labeling this as a bug.

Juanlu001:
I will give you the honor of submitting a plot showcasing the solution and closing this issue @anhiga :)

anhiga:
This is how plots look now.
![figure_1](https://user-images.githubusercontent.com/18404378/30056499-3bd87644-9233-11e7-9b18-ee0cff84825a.png)

Juanlu001:
Using `%matplotlib notebook` and some patience one can arrive to something like this:

![png image 482 x 684 pixels](https://user-images.githubusercontent.com/316517/30056991-4e4d70e8-9235-11e7-983f-2c6dd185168e.png)

While I admit it is not optimal (notice the white space above the plot, apart from the manual tweaking) this fixes the original issue. Interested readers might want to check out #214, #226, #230 and https://github.com/matplotlib/matplotlib/issues/9155.

PR #214, opened 2017-08-23
Epoch fix

anhiga:
Quick and dirty fix to #205.

Without the fix, this code, with orbits with different epochs:
```Python
from poliastro import plotting
from poliastro import examples

frame = plotting.OrbitPlotter()

examples.iss.epoch
Out: <Time object: scale='utc' format='iso' value=2013-03-18 12:00:00.000>

examples.molniya.epoch
Out: <Time object: scale='utc' format='jyear_str' value=J2000.000>

frame.plot(examples.iss, label='iss')
frame.plot(examples.molniya, label='molniya')
```
yields:
without_fix
where epoch is wrong for the first orbit (`ISS`)

With the fix, and adding `propagate=True`:
```Python
from poliastro import plotting
from poliastro import examples

frame = plotting.OrbitPlotter()

examples.iss.epoch
Out: <Time object: scale='utc' format='iso' value=2013-03-18 12:00:00.000>

examples.molniya.epoch
Out: <Time object: scale='utc' format='jyear_str' value=J2000.000>

frame.plot(examples.iss, label='iss')
frame.plot(examples.molniya, label='molniya', propagate=True)
```
with_fix
`Molniya` orbit propagates forwards to the `ISS` epoch.

### DIFF
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -57,6 +57,7 @@ class OrbitPlotter(object):
             _, self.ax = plt.subplots(figsize=(6, 6))
         self.num_points = num_points
         self._frame = None
+        self._states = []  # type: list
         self._attractor_radius = None
 
     def set_frame(self, p_vec, q_vec, w_vec):
@@ -116,6 +117,8 @@ class OrbitPlotter(object):
             self.set_attractor(orbit)
         elif new_radius < self._attractor_radius:
             self.set_attractor(orbit)
+
+        self._states.append(orbit)
 
         lines = []
 
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -57,7 +57,7 @@ class OrbitPlotter(object):
             _, self.ax = plt.subplots(figsize=(6, 6))
         self.num_points = num_points
         self._frame = None
-        self._states = []  # type: list
+        self._orbits = []  # type: list
         self._attractor_radius = None
 
     def set_frame(self, p_vec, q_vec, w_vec):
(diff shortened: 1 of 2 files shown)


src/poliastro/plotting.py
```diff
@@ -101,14 +102,19 @@ def set_attractor(self, orbit):
...
 
         """
         # TODO: This function needs a refactoring
         if not self._frame:
             self.set_frame(*orbit.pqw())
 
+        # Propagates backwards of forwards to the last added orbit epoch.
+        if propagate:
+            if self._orbits:
+                orbit = orbit.propagate(self._orbits[-1].epoch - orbit.epoch)
```
Juanlu001: If we are only using the epoch of the orbit (and in fact, we are only using the first epoch) I would rather store only this first epoch. I don't need to store the whole list of orbits. Perhaps you can do something similar to `set_frame`?

src/poliastro/plotting.py
```diff
@@ -117,6 +123,8 @@ def plot(self, orbit, osculating=True, label=None):
         elif new_radius < self._attractor_radius:
             self.set_attractor(orbit)
 
+        self._orbits.append(orbit)
```
Juanlu001: See above

src/poliastro/plotting.py
anhiga: Well, there are two different approachs. We could store only the first epoch, or we could store only the epoch from the last orbit added. What do you think? In any case, you are right, only one epoch is needed.

Juanlu001: What happens if we propagate some orbits, and leave as is some others? And also, are we allowing the user to propagate to any of the epochs that are in the `OrbitPlotter`?

I am not convinced that this is the right approach. Perhaps we can it tomorrow during the meeting?

anhiga:
Closing and moving to #230.

PR #215, opened 2017-08-24
Added first version of coordinates.py

anhiga:
First version of `coordinates` module, containing:

* ICRS to body-centered frame transformation.

* Body-centered frame to ICRS transformation.


Juanlu001:
Tests missing :) Review this afternoon.

src/poliastro/coordinates.py
```diff
@@ -0,0 +1,169 @@
...
+import astropy.time as time
+import astropy.coordinates as coordinates
+import astropy.units as u
+
+from poliastro.util import transform
+
+
+T = 0
+d = 0
+J2000 = time.Time('J2000', scale='tdb')
```
Juanlu001: J2000 is already defined elsewhere:

https://github.com/poliastro/poliastro/blob/f6676ef/src/poliastro/twobody/orbit.py#L18

I would rather move it to `constants.py`. It doesn't need to be an Astropy `Constant`, but it clearly does not belong there.

src/poliastro/coordinates.py
```diff
@@ -0,0 +1,169 @@
...
+import math
+
+import astropy.time as time
+import astropy.coordinates as coordinates
+import astropy.units as u
+
+from poliastro.util import transform
+
+
+T = 0
```
Juanlu001: Now I understand what you told me. I don't like this approach either.

I would rather define one function for each body, receiving the `Time` as an input and then returning a pair `(ra, dec, W)` with the coordinates of the north pole and the prime meridian.

anhiga: It's a very good idea, together with putting parameters inside each body.

src/poliastro/coordinates.py
anhiga: Completely right!

anhiga:
I have another problem regarding module testing. Should we use DE431 to check transformations, and how?

Juanlu001:
> I have another problem regarding module testing. Should we use DE431 to check transformations, and how?

I was expecting this question... I don't think it's feasible to download DE431 each time we want to do testing. I suggest pasting the numeric output, describe how it's done (including a link to the ephemerides) and have the user "trust" it.

src/poliastro/coordinates.py
```diff
@@ -11,72 +11,285 @@
...
+    -------
+    ra, dec, W: tuple (~astropy.units.Quantity)
+        Right ascension and declination of north pole, and angle of the prime meridian.
+
+    """
+
+    T = (epoch.tdb - J2000).to('day').value / 36525
+    d = (epoch.tdb - J2000).to('day').value
+
+    ra = (281.01 - 0.033 * T) * u.deg
```
Juanlu001: @anhiga and I already clarified in private chat that, although putting these constants inside the `Body` definitions, things get hairy for Jupiter, the Moon and other bodies... So they are fine here.

src/poliastro/coordinates.py
```diff
@@ -11,72 +11,285 @@
...
+        Epoch, default to J2000.
+
+    Returns
+    -------
+    ra, dec, W: tuple (~astropy.units.Quantity)
+        Right ascension and declination of north pole, and angle of the prime meridian.
+
+    """
+
+    d = (epoch.tdb - J2000).to('day').value
```
Juanlu001: The code for computing `d` and `T` is repeated every time. I suggest adding a decorator instead:

```
@time_magnitudes
def xxx_rot_elements(T, d):
    # ...

xxx_rot_elements(epoch)  # The decorator changes the signature
```

It's the idea I applied here: https://github.com/poliastro/poliastro/blob/v0.6.0/src/poliastro/twobody/decorators.py

src/poliastro/coordinates.py
```diff
@@ -11,72 +11,285 @@
...
+
+    ra = 272.76 * u.deg
+    dec = 67.16 * u.deg
+    W = (160.20 - 1.4813688 * d) * u.deg
+
+    return ra, dec, W
+
+
+def earth_rot_elements(epoch=J2000):
```
Juanlu001: If it is not too magical, perhaps we can use a decorator for adding the docstring as well, or use some kind of template? Such as:

```
def xxx_():
def yyy_():

xxx_.__doc__ = DOC_TPL.format(planet="xxx")
yyy_.__doc__ = DOC_TPL.format(planet="yyy")
```

I don't want loops, deep introspection or very magical things, but there are too many identical docstrings in this module.

src/poliastro/coordinates.py
```diff
@@ -11,72 +11,285 @@
...
+    d = (epoch.tdb - J2000).to('day').value
+
+    ra = (317.68143 - 0.1061 * T) * u.deg
+    dec = (52.88650 - 0.0609 * T) * u.deg
+    W = (176.630 + 350.89198226 * d) * u.deg
+
+    return ra, dec, W
+
+
+def jupiter_rot_elements(epoch=J2000):
```
Juanlu001: Also, instead of having all these functions here and then the one below that maps body names to functions, perhaps we can have just methods in `Body`.

```
class Body:
    def __init__(self, parent, ...):
        # ...

    @time_magnitudes
    def rotational_elements(self, T, d):
        # ...
        return ra, dec, W
```

What do you think?

src/poliastro/coordinates.py
```diff
@@ -100,22 +313,16 @@ def body_centered_to_icrs(r, v, source_body, epoch=None, rotate_meridian=False):
...
-        epoch = time.Time('J2000', scale='tdb')
-
-    T = (epoch.tdb - J2000).to('day').value / 36525
-    d = (epoch.tdb - J2000).to('day').value
-
+    ra, dec, W = rot_elements_from_body(source_body, epoch)
     if rotate_meridian:
-        r = transform(r, -bodies_orientations[source_body.name]['W'], 'z', u.deg)
-        v = transform(v, -bodies_orientations[source_body.name]['W'], 'z', u.deg)
+        r = transform(r, -W.value, 'z', W.unit)
```
Juanlu001: Are you sure that `transform` does not work with `Quantity` objects? Last time I used it, it did. Perhaps you can simplify the calls here.

src/poliastro/coordinates.py
```diff
@@ -124,10 +331,10 @@ def body_centered_to_icrs(r, v, source_body, epoch=None, rotate_meridian=False):
 
     # icrs_coord = coordinates.ICRS(x=r_f[0], y=r_f[1], z=r_f[2], v_x=v_f[0], v_y=v_f[1],
     #                              v_z=v_f[2], representation='cartesian')
-    return r_f, v_f
+    return r_f.to(r.unit), v_f.to(v.unit)
```
Juanlu001: Good catch!

src/poliastro/coordinates.py
anhiga: Really interesting idea, I will take a close look at it.

src/poliastro/coordinates.py
anhiga: I don't really see how this could be implemented. Shouln't be a different `rotational_elements(self, T, d) for every different body?

src/poliastro/coordinates.py
Juanlu001: You can add the docstring in the decorator and matar dos pájaros de un tiro! :P

src/poliastro/coordinates.py
Juanlu001: You are right... And that might involve creating one subclass pero Body. Not nice.

We can discuss alternatives tomorrow.

src/poliastro/coordinates.py
anhiga: Your're right, it works with `Quantities` seamlessly.

src/poliastro/bodies.py
```diff
@@ -19,17 +19,56 @@
 
 Data references can be found in :py:mod:`~poliastro.constants`
 """
+import math
+from functools import wraps
```
Juanlu001: If you are not using the decorator anymore, I would remove it along with this import!

src/poliastro/bodies.py
```diff
@@ -89,6 +144,15 @@ class _Venus(_Body):
     symbol = u"\u2640"
     R = constants.R_venus
 
+    @staticmethod
+    def _rot_elements_at_epoch(T, d):
+
```
Juanlu001: I think you can remove the white lines under the method definition (present only in some of them)

src/poliastro/constants.py
```diff
@@ -23,6 +23,9 @@
 """
 
 from astropy.constants import Constant
+from astropy import time
+
+J2000 = time.Time('J2000', scale='tdb')
```
Juanlu001: Now that J2000 is defined here, what about removing it from `orbit.py`?

Juanlu001:
When PEP8 passes I will merge this :)

anhiga:
@Juanlu001 It is ready, but I can squash some commits, if 19 are too many.

Juanlu001:
> @Juanlu001 It is ready, but I can squash some commits, if 19 are too many.

No need to do that, I think it's fine as it is :) There are several duplicated code issues but I think they are worth keeping. Improvements in this regard can go in other pull requests. Merging this!

PR #217, opened 2017-08-25
Alternative bodies implementation

Juanlu001:
This structure is a bit contrived but allows for more flexibility. @anhiga I think this would help in adding rotating elements for #215, but I admit it is strange to use the classes like that. Please review it and merge it if you think it is going to be useful.

### DIFF
--- a/src/poliastro/bodies.py
+++ b/src/poliastro/bodies.py
@@ -23,8 +23,16 @@ from astropy import units as u
 from poliastro import constants
 
 
-class Body(object):
-    """Class to represent a body of the Solar System.
+class _Body(object):
+    def __str__(self):
+        return u"{0} ({1})".format(self.name, self.symbol)
+
+    def __repr__(self):
+        return self.__str__()
+
+
+class Body(_Body):
+    """Class to represent a generic body of the Solar System.
 
     """
     def __init__(self, parent, k, name, symbol=None, R=0 * u.km):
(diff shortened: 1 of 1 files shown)


Juanlu001:
To clarify: the `Body` class remains for the user to create generic bodies, inherit it, or whatever. The rest of the classes are all private so only the specific instances can be seen from the outside.

Juanlu001:
AppVeyor got stuck for 20 minutes... Restarting.

Juanlu001:
And also I approved the repeated code warnings in Codeclimate because I prefer explicit, repeated code in this case (and also because it annoys me)

Juanlu001:
http://status.appveyor.com/incidents/2tjbzx1mtnz8

I already subscribed.

anhiga:
I'm looking for another alternatives, but it appears to be the only way to have rotating elements in `bodies` module...
Also, although code could be a little contrived, it is easy to read it and understand it.

anhiga:
The only other approach I can think of right now is working with lambdas, so there is no need of initializing `T` and `d` until we use them. Maybe something like this:
```Python
Mercury = Body.from_parameters(Sun, k=constants.GM_mercury, name="Mercury",
symbol=u"\u263F", R=constants.R_mercury, ra=lambda T, d: 281.01 - 0.033 * T,
dec=lambda T, d: 61.45 - 0.005 * T, W=lambda T, d: 329.548 + 6.1385025 * d)

Mercury.ra(0, 30)
Out: 281.01
```

Or we could have a dictionary of lambdas:
```Python
bodies_orientations = {
    "Mercury": {
        "ra": lambda T, d: 281.01 - 0.033 * T,
        "dec": lambda T, d: 61.45 - 0.005 * T,
        "W": lambda T, d: 329.548 + 6.1385025 * d
    },
```
that could be inside `bodies`, or even inside `constants`.

Juanlu001:
Have you tried the lambdas approach? Because I am not sure if Python will
bound a `self` argument to them. The dictionary of lambdas can work but how
are you doing it with Jupiter for example?

anhiga:
I have tried implementing it, and it works (apparently), but I've never seen lambdas been using like this (I'm quite new to Python, there are probably many things I haven't seen) so I'm not sure if would be a good idea...

anhiga:
Despite of my comments, the lambdas solution would be probably a more confusing solution that the one you provided @Juanlu001, so there isn't another idea, I think we could merge this.

Juanlu001:
We checked locally the lambdas way for complex bodies and it's quite messy. We're going this route.

PR #221, opened 2017-08-28
Pip packaging

Juanlu001:
Try using only pip everywhere. Numba is no longer a special dependency. If this works, I will remove the conditional JIT in the same pull request.


AlexS12:
Seems great that you can install everything from pip now!

Juanlu001:
Yes! I will probably keep using conda for convenience, but it feels good not to be attached to it :) Plus more people will find poliastro easier to install.

PR #223, opened 2017-08-31
Add notebook studying Florence asteriod

Juanlu001:
Many, many issues appeared along the way. Anyway, I am so proud of what we have done @anhiga :blush: Please take a look and review it!


anhiga:
Epoch should be in TDB scale if comparing to Horizons values: `EPOCH = Time("2017-09-01 12:00", scale='tdb')`
Results still don't match, although.

docs/source/examples/Catch that asteroid!.ipynb
```diff
@@ -0,0 +1,890 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
+    "import matplotlib.pyplot as plt\n",
+    "plt.style.use(\"pybonacci\")\n",
```
anhiga: I would remove this.

Juanlu001:
It's finished! :heart_eyes: 

florence

Notice that this needs a development version of Astropy until this pull request is merged:

https://github.com/astropy/astropy/pull/6508

(you can do `pip install git+https://github.com/Juanlu001/astropy.git@ecliptic-precision)

This is ready for a final review before merging. Tomorrow I will add a cell stating the information above.

docs/source/examples/Catch that asteroid!.ipynb
Juanlu001: Yes, you're right... I opened an issue years ago to ship custom styles in matplotlib:

https://github.com/matplotlib/matplotlib/issues/4781

But until it's implemented, the only way is pointing the users to a URL and making them download the file in a specified folder... A complete mess. I will remove it (even though I love my color cycle...)

docs/source/examples/Catch that asteroid!.ipynb
```diff
@@ -0,0 +1,804 @@
...
+       "<Quantity 7066700.8455871325 km>"
+      ]
+     },
+     "execution_count": 16,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "norm(florence_icrs.r - earth.r)"
```
anhiga: ESA has given Florence distance from Earth surface, maybe we could substract Earth radius

Juanlu001: Good idea! Perhaps that difference of ~6700 km will take us even closer to the announced value :)

Juanlu001:
This is complete, so I am merging :) Cannot wait to see it on the docs!

Juanlu001:
In all its glory: http://docs.poliastro.space/en/latest/examples/Catch%20that%20asteroid!.html

PR #228, opened 2017-09-01
Added inertial_body_frame_to_pqw

anhiga:


Juanlu001:
No complains on my side :)

Just to clarify: this took a while because @anhiga invested some time trying to implement the inverse transformation as well, from perifocal to inertial. However, one needs some information about the orbit to perform the transformation, and in the end we decided it is not worth it to have such a function. We have mentioned somewhere the use case of reprojecting a plot, but that can be achieved through simpler means.

Thanks again and merging!

Juanlu001:
And by the way, the Travis failures are unrelated to this.

Issue #227, opened 2017-09-01
Always plot osculating orbit?


Juanlu001:
After looking at this plot by @anhiga:

![29966416-17dd66ce-8f12-11e7-9348-9c85fd11ac29](https://user-images.githubusercontent.com/316517/29970422-ca520c6c-8f24-11e7-8ece-d7bc2e46a6cb.png)

I suddenly wondered if it makes sense to plot a point _without the osculating orbit_. That point could have an elliptic, or circular, or hyperbolic orbit... And there is no way to know without at least plotting a fraction of the orbit. No usage of `osculating=False` in the user guide nor the examples either. I think we should consider removing that feature.

anhiga:
I think we could keep that feature as long as there is no problems with it.
It can be useful when plotting many bodies, `Atira NEOs` for example:
![atiras2](https://user-images.githubusercontent.com/18404378/29979889-928169b8-8f47-11e7-83b7-d1c0572735b2.png)

![atiras_without_osculating](https://user-images.githubusercontent.com/18404378/29979977-e94c56a4-8f47-11e7-970e-319c0d8a839f.PNG)

Even if the first is much more beautiful, maybe someone could prefer the second one 😆

Juanlu001:
This is a tie then :) I'll leave this open in case more people want to weigh in.

Juanlu001:
This feature was removed in #230, so closing.

PR #226, opened 2017-09-01
Epoch and plotting changes

anhiga:
Epoch removed from title, and added to a new `epoch legend` (optional). If these two legends are automatically located by Matplotlib in the 'best' place, they overlap. Possible solutions:
- [ ] Fixing one legend to a corner, and the other legend to another corner. Problem:
    * Legends could overlap with orbit plotting.
- [ ] Let `matplotlib` locate the first legend, and locate the second legend in a different place. Problem:
    * Same as above.
    * I haven't been able to get `loc` property of the first legend.
- [x] Making the two legends draggable.

Alternative solutions are welcome.

PR #230, opened 2017-09-01
Legend plotting

anhiga:
Another plotting PR 😆 .

Now label and epoch are in a figure legend.

Issues to discuss:
* Should epoch without label be showed in this case: `frame.plot(orbit)`?
* What format should epoch have? (Currently, `YYYY-MM-DD HH:MM`)
* What title should legend have?

Problems:
* When plotting several orbits, legend can overlap axes:
planets

### DIFF
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -103,7 +103,7 @@ class OrbitPlotter(object):
 
         self.ax.add_patch(mpl.patches.Circle((0, 0), radius, lw=0, color=color))
 
-    def plot(self, orbit, osculating=True, label=None):
+    def plot(self, orbit, label=None):
         """Plots state and osculating orbit in their plane.
 
         """
@@ -147,10 +147,9 @@ class OrbitPlotter(object):
                           'o', mew=0)
         lines.append(l)
 
-        if osculating:
-            l, = self.ax.plot(x.to(u.km).value, y.to(u.km).value,
-                              '--', color=l.get_color())
-            lines.append(l)
+        l, = self.ax.plot(x.to(u.km).value, y.to(u.km).value,
+                          '--', color=l.get_color())
+        lines.append(l)
 
         if label:
             # This will apply the label to either the point or the osculating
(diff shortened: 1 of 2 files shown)


anhiga:
@Juanlu001 I'm closing related PRs (#226, #214), if you agree.

Juanlu001:
Sure, do!

Juanlu001:
The legend title is fine for me. I would actually show 

YYYY-MM-DD HH:MM (Label)

or

YYYY-MM-DD HH:MM

Regarding overlap... This happens in the notebook, right? Let's leave it like that for now, we've spent far too much time in this :/

Juanlu001:
I mean: either we can manually adjust the figure size from the outside, or there is some magic matplotlib function that we are not aware of which properly fits everything into the figure.

Juanlu001:
Is this ready to merge?

anhiga:
Well, I have noticed another problem. It is a little confusing:

The only way to get handles and labels from a figure legend requires using `legends` property to obtain the legend, and then Legend methods `get_lines()` and `get_texts()`. In the case of axes legend, this can be achieved using Axes method `get_legend_handles_labels()`.
However, there is a subtle difference between these two ways: axes method keep handles (which in this case are `matplotlib.lines.Lines2D`) marker (should be `'o'`), while legend methods don't.

After the explanation, a visual example of what this difference mean:

Axes legend:
figure_1

Figure legend:
figure_2

When `osculating=False`, marker is only keeped in the first case (dashed lines are not a marker, so they are keeped anyway)

The only solutions I can think of are:
* Creating a new OrbitPlotter property `handles_labels` and store handles and labels there: I don't really like this one.
* Retract my comments and agree with @Juanlu001 in #227, removing non-osculating orbits.

Juanlu001:
I think this deserves some discussion with the matplotlib devs, perhaps there is something that we are missing. Would you please open an issue in https://github.com/matplotlib/matplotlib/issues? That way they can triage it appropriately.

Juanlu001:
ref: https://github.com/matplotlib/matplotlib/issues/9155

PR #231, opened 2017-09-02
Added function to compute SOI's radius

PabloRdrRbl:
As a result of Issue #150, in this Pull  Request an implementation of the computation of the radius of the Sphere of Influence of a body is proposed. The _approximate radius of the Laplace sphere of influence_ is computed using the formula presented by Bate, Mueller and White "Fundamentals of Astrodynamics" in equation 7.4-1.

To obtain the semimajor axis of the body two different approaches are used:
- Semimajor axis is introduced by the user.
- Semimajor axis is computed using **poliastro** for the ephemerides J2000 (default).

Some comments on the current status of the code:

- At the moment is not possible to compute the semimajor axis for Pluto since the following error arises when using **poliastro**'s ephemerides:
`KeyError: "pluto's position and velocity cannot be calculated with the 'builtin' ephemeris."`
- The results where validated using data from table A.2. from Curtis "Orbital Mechanics for Engineering Students". The computed radius from the code is bigger than the tabulated data, however as far as I know the source does not indicate which method is used to compute such data. There exit other approximations which may provide more accurate results (see [Wkipedia: Sphere of influence (astrodynamics). Increased accuracy on the SOI](https://en.wikipedia.org/wiki/Sphere_of_influence_(astrodynamics)#Increased_accuracy_on_the_SOI)).
- I tried to implement an exception for the case where no parent body exists, but it may be possible to do it better.
- It would be desirable to write tests.

### DIFF
--- a/src/poliastro/patched_conics.py
+++ b/src/poliastro/patched_conics.py
@@ -1,0 +1,28 @@
+# coding: utf-8
+"""Patched Conics Computations
+
+Contains methods to compute interplanetary trajectories approximating the three
+body problem with Patched Conics.
+"""
+
+from astropy import units as u
+from poliastro.twobody import Orbit
+from poliastro.constants import J2000
+
+
+@u.quantity_input(a=u.m)
+def compute_soi(body, a=None):
+    """Computes the approximated radius of the Laplace Sphere of Influence (SOI)
+ for a body (:py:class:`~Body` class).
+    """
+    # Compute semimajor axis at epoch J2000 for the body if it was not
+    # introduced by the user
+    if a == None:
+        ss = Orbit.from_body_ephem(body, J2000)
+        a = ss.a
+
+    try:
+        r_soi = a * (body.k / body.parent.k)**(2 / 5)
+        return r_soi.decompose()
+    except AttributeError:
+        print('Body has no parent body assigned.')


Juanlu001:
Thanks for this pull request @PabloRdrRbl! The Travis failure is related to a PEP8 issue, go read it :)

For the validation, you should add a `test_patched_conics.py` which checks the numerical results of the function. You can draw inspiration from `test_angles.py` for example.

I will do a code review as soon as I can.

src/poliastro/patched_conics.py
```diff
@@ -0,0 +1,28 @@
+# coding: utf-8
+"""Patched Conics Computations
+
+Contains methods to compute interplanetary trajectories approximating the three
+body problem with Patched Conics.
```
Juanlu001: Nice!

src/poliastro/patched_conics.py
```diff
@@ -0,0 +1,28 @@
...
+"""
+
+from astropy import units as u
+from poliastro.twobody import Orbit
+from poliastro.constants import J2000
+
+
+@u.quantity_input(a=u.m)
+def compute_soi(body, a=None):
+    """Computes the approximated radius of the Laplace Sphere of Influence (SOI)
```
Juanlu001: I would remove the class from here, and add the `Parameters` section.

src/poliastro/patched_conics.py
```diff
@@ -0,0 +1,28 @@
...
+ for a body (:py:class:`~Body` class).
+    """
+    # Compute semimajor axis at epoch J2000 for the body if it was not
+    # introduced by the user
+    if a == None:
```
Juanlu001: PEP8 gave you the hint here ;)

src/poliastro/patched_conics.py
```diff
@@ -0,0 +1,28 @@
...
+        ss = Orbit.from_body_ephem(body, J2000)
+        a = ss.a
+
+    try:
+        r_soi = a * (body.k / body.parent.k)**(2 / 5)
+        return r_soi.decompose()
+    except AttributeError:
+        print('Body has no parent body assigned.')
```
Juanlu001: The non catched issue here is something like "Body has no parent attribute", right? I think it's informative enough, I would not capture and just let it raise.

src/poliastro/patched_conics.py
PabloRdrRbl: Corrected!

src/poliastro/patched_conics.py
PabloRdrRbl: Done!

src/poliastro/patched_conics.py
PabloRdrRbl: I am not sure but, do you mean to pass to the function the different parameters instead of the whole object?

Juanlu001: I mean, having a shorter summary in the first line of the docstring and adding the "parameters" section as explained here:

https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections

Juanlu001:
The test is well written, but it's failing. When you get it to work I will merge this :)

src/poliastro/patched_conics.py
PabloRdrRbl: It should be done now 👍

src/poliastro/patched_conics.py
```diff
@@ -12,8 +12,20 @@
 
 @u.quantity_input(a=u.m)
 def compute_soi(body, a=None):
-    """Computes the approximated radius of the Laplace Sphere of Influence (SOI)
- for a body (:py:class:`~Body` class).
+    """Approximated radius of the Laplace Sphere of Influence (SOI) for a body.
+
+    Parameters
```
Juanlu001: Nice!

src/poliastro/tests/test_patched_conics.py
```diff
@@ -0,0 +1,34 @@
...
+        (Pluto, 3.08e9)
+    ]
+    for row in data:
+
+        body, expected_r_SOI = row
+        expected_r_SOI = expected_r_SOI * u.m
+
+        r_SOI = compute_soi(body)
+
+        assert_quantity_allclose(r_SOI, expected_r_SOI, rtol=1e-6)
```
Juanlu001: I suggest `1e-1` :wink:

Juanlu001: Oops, you just done that! :sweat_smile:

Juanlu001:
Regarding the error of the Moon: you either set the JPL ephemeris (big download... I would not recommend it) or just skip it.

PabloRdrRbl:
> Regarding the error of the Moon: you either set the JPL ephemeris (big download... I would not recommend it) or just skip it.

But in case Moon and Pluto are skipped in the test, the code wouldn't be still able to produce de radius of the SOI for this two bodies if the user does not provide the semimajor axis or we include the JPL alternative. @Juanlu001.

Juanlu001:
> But in case Moon and Pluto are skipped in the test, the code wouldn't be still able to produce de radius of the SOI for this two bodies if the user does not provide the semimajor axis or we include the JPL alternative. @Juanlu001.

You are right. The idea of skipping them was just for the purposes of testing.

Regarding the users we can do two things:

* Leave it like that, and expect them to understand that they have to do 

```
>>> from astropy.coordinates import solar_system_ephemeris
>>> solar_system_ephemeris.set("jpl")
```

* Catch the exception in `from_body_ephem` and raise a new one with a better error message that tells the user "do this import".

PabloRdrRbl:
> You are right. The idea of skipping them was just for the purposes of testing.
> 
> Regarding the users we can do two things:
> 
> Leave it like that, and expect them to understand that they have to do
> >>> from astropy.coordinates import solar_system_ephemeris
> >>> solar_system_ephemeris.set("jpl")
> Catch the exception in from_body_ephem and raise a new one with a better error message that tells the user "do this import".

I will try the second! I looks much pro, doesn't it?

Juanlu001:
Totally :) I expect the tests to pass now, ping me when this is ready (I don't get notifications on new commits).

src/poliastro/patched_conics.py
```diff
@@ -0,0 +1,46 @@
...
+        try:
+            ss = Orbit.from_body_ephem(body, J2000)
+            a = ss.a
+
+            r_SOI = a * (body.k / body.parent.k)**(2 / 5)
+
+            return r_SOI.decompose()
+
+        except KeyError:
+            print("To compute the semimajor axis for Moon",
```
Juanlu001: Last comment (I promise): instead of printing and passing, do `raise RuntimeError("...")`. In Python land it's better to fail as early as possible, instead of returning `None` and finding below a `None object has no attribute X` or something like that :)

Juanlu001:
Manually merged to make it in time for the release. Thanks a lot for this contribution @PabloRdrRbl! :tada:

Issue #224, opened 2017-08-31
Add from_relative to Body?


jzmiller1:
There was some discussion about adding a from_relative method to the Body class on a recently closed PR thread: https://github.com/poliastro/poliastro/pull/153

Is that something that there's some interest in seeing in a new PR?

Juanlu001:
You're right @jzmiller1, I oversaw that. Please go ahead :) I think this comment summarizes more or less what you wanted, right?

https://github.com/poliastro/poliastro/pull/153#issuecomment-284218548

jzmiller1:
Yes, something like that should work well.  I'll put a PR for it together in the next week unless you already had the code for it somewhere?

Juanlu001:
I don't have it, so please do. Also, I encourage you to join our chat, so we can guide you more quickly:

https://riot.im/app/#/room/#poliastro:matrix.org

Juanlu001:
This was closed in #235.

PR #235, opened 2017-09-04
Add from_relative to Body class

jzmiller1:
Add the from_relative method of creating a Body to allow easy creation
of bodies for exoplanets and other bodies that are often described with
mass and radius relative to Sun, Jupiter or Earth.  The from_relative
method will take any object as a reference as long as it implements the
attributes (k and R) and does not require the reference object to be a
poliastro Body object.

Updated the docstring on Body to remove reference to the Solar System
as bodies may be exoplanets, stars or other extrasolar objects.

Allowed for pass through of additional data to newly created bodies
using kwargs.

### DIFF
--- a/src/poliastro/bodies.py
+++ b/src/poliastro/bodies.py
@@ -58,10 +58,10 @@ class _Body(object):
 
 
 class Body(_Body):
-    """Class to represent a generic body of the Solar System.
+    """Class to represent a generic body.
 
     """
-    def __init__(self, parent, k, name, symbol=None, R=0 * u.km):
+    def __init__(self, parent, k, name, symbol=None, R=0 * u.km, **kwargs):
         """Constructor.
 
         Parameters
@@ -83,11 +83,19 @@ class Body(_Body):
         self.name = name
         self.symbol = symbol
         self.R = R
+        self.kwargs = kwargs
 
     @classmethod
     @u.quantity_input(k=u.km ** 3 / u.s ** 2, R=u.km)
-    def from_parameters(cls, parent, k, name, symbol, R):
-        return cls(parent, k, name, symbol, R)
+    def from_parameters(cls, parent, k, name, symbol, R, **kwargs):
+        return cls(parent, k, name, symbol, R, **kwargs)
+
+    @classmethod
+    def from_relative(cls, reference, parent=None, k=None, name=None,
+                      symbol=None, R=None, **kwargs):
+        k = k * reference.k
+        R = R * reference.R
+        return cls(parent, k, name, symbol, R, **kwargs)
 
 
 class _Sun(_Body):
--- a/src/poliastro/coordinates.py
+++ b/src/poliastro/coordinates.py
@@ -129,4 +129,3 @@ def inertial_body_centered_to_pqw(r, v, source_body):
     v_pqw = (np.array([-sin(nu), (ecc + cos(nu)), 0]) * sqrt(k / p)).T * u.km / u.s
 
     return r_pqw, v_pqw
-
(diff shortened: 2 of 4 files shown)


jzmiller1:
This is related to: https://github.com/poliastro/poliastro/issues/224

Juanlu001:
Thanks for this pull request @jzmiller1! I'm going to do brief code review so we can discuss the changes, and after one iteration I think this will be ready to merge :):

src/poliastro/bodies.py
```diff
@@ -58,10 +58,10 @@ def _rot_elements_at_epoch(T, d):
 
 
 class Body(_Body):
-    """Class to represent a generic body of the Solar System.
```
Juanlu001: Fair enough! :)

src/poliastro/bodies.py
```diff
@@ -58,10 +58,10 @@ def _rot_elements_at_epoch(T, d):
...
+    """Class to represent a generic body.
 
     """
-    def __init__(self, parent, k, name, symbol=None, R=0 * u.km):
+    def __init__(self, parent, k, name, symbol=None, R=0 * u.km, **kwargs):
```
Juanlu001: Can you think of any use case for the extra `**kwargs`?

src/poliastro/bodies.py
```diff
@@ -83,11 +83,22 @@ def __init__(self, parent, k, name, symbol=None, R=0 * u.km):
...
     @u.quantity_input(k=u.km ** 3 / u.s ** 2, R=u.km)
-    def from_parameters(cls, parent, k, name, symbol, R):
-        return cls(parent, k, name, symbol, R)
+    def from_parameters(cls, parent, k, name, symbol, R, **kwargs):
+        return cls(parent, k, name, symbol, R, **kwargs)
+
+    @classmethod
+    def from_relative(cls, reference, parent=None, k=None, name=None,
+                      symbol=None, R=None, **kwargs):
+        if hasattr(reference, 'k') and hasattr(reference, 'R'):
```
Juanlu001: I would apply duck typing here instead: I agree with the idea that `reference` does not need to be a `Body`, but do not try to catch the issue. If the `reference` variable does not have `k` or `R`, it is going to raise an `AttributeError` anyway, which I think it's informative enough.

src/poliastro/bodies.py
```diff
@@ -83,11 +83,22 @@ def __init__(self, parent, k, name, symbol=None, R=0 * u.km):
...
+            k = k * reference.k
+            R *= R * reference.R
```
Juanlu001: Isn't this `R = R * reference.R` or `R *= reference.R`? In any case be consistent with `k` (unless I have misunderstood something).

src/poliastro/bodies.py
jzmiller1: Any time you'd like to put extra data on a Body you can pass it as a keyword argument and use it from kwargs.  This way you won't need to add some attribute after creation or subclass body to add your own parameters.

Juanlu001: Sure, but this applies to any kind of object. What kind of extra parameters would you add to Bodies that are not in the constructor?

jzmiller1: Suppose you wanted to add a comment attribute to each body or some kind of additional value you are using for a custom calculation.  It would be easier to allow people to pass in any keywords they want rather than add a bunch of new parameters for whatever use case may come along?

Juanlu001: Fair enough, let's leave it like this then :)

Juanlu001:
There are another two blocks of code that I commented. If you have the time to address them, I would love to merge this before the 0.7.0 release, which is due next Thursday :)

jzmiller1:
I made an update to this.

Juanlu001:
This is good to go now! Next time you don't have to amend and force push :) I fixed a small PEP8 issue that slipped from another pull request (my bad). When the CI finishes, I'm merging.

Juanlu001:
The codecov failure is related to the last improperly merged pull request. Merging!

Juanlu001:
Thanks @jzmiller1! :tada:

PR #243, opened 2017-09-13
Add ecliptic frame to poliastro

### LINKED ISSUE #222: Coordinates from NEOS are in some ecliptic reference frame
...whereas planetary ephemerides are in ICRS. Unfortunately, Astropy 2.0.1 does not support velocities for ecliptic coordinate frames:

https://github.com/astropy/astropy/issues/6494

We should, at the very least, document this issue because it can potentially produce misguided plots (as we have been doing all along...)

Juanlu001:
Fix #222. Also, my Astropy fork is no longer needed to run the Florence notebook.

By the way @anhiga: the plots look *much* better now in the notebook: there were lots of horizontal space on the right! It took long, but I am now very happy with the result :+1:

Would you mind reviewing this?


Juanlu001:
Thanks! Yeah, I didn't add any new tests... Shame on me. We'll pay the price one day, but today is not that day!

PR #245, opened 2017-09-15
Neos ecliptic fix

anhiga:
Thanks to @Juanlu001, `poliastro` has now a module with transformations between Horizons ecliptic frame and ICRS. Given that DASTCOM5 and NeoWS both provides their data in the Horizons frame, corresponding transformations are necessary in order to get the orbits in ICRS.
This is related to #243 and #222.


Juanlu001:
Thanks for this @anhiga! I think it is a good first step, since it makes the NEOS functions return consistent results with all the rest.

However, notice that the core issue is still there: that `Orbit` objects have an implicit reference frame.

Regarding this pull request, the code looks good but I think there are two notebooks and perhaps parts of the user guide that need changes :)

Juanlu001:
Totally forgot about this, sorry! Reminder that the notebooks need fixing for this to get merged.

anhiga:
I fixed and rerun pertinent notebooks. Florence-Earth distance changed from 300 to 600km, but I think it is due to difference between first propagating and then transforming; and first transforming and then propagating.

anhiga:
* Timeout error in notebook execution (in Python 3.6)
* neows values test is also failing, only in OSX (?).

Juanlu001:
I asked @newlawrence to kindly review this. We can hold it back until him or someone with a Mac sheds some light :)

newlawrence:
Hi guys, I'm sorry I'm a little bit late. Currently I'm reviewing this issue you have. I haven't find an answer of what's going yet but I've discovered a clue that might help us to guess where the issue is. 

Comparing the Travis log with the results on my local machine, I've found that we aren't throwing the same results... while I have no idea where the value for the eccentricity under Travis comes, I can tell you that on my machine... the `orbit_from_spk_id` function simply does nothing!!! Meaning that its return value is that of **mock_orbital_data**'s dict without modification.

Since I don't know anything about poliastro code, I'll try to debug the function step by step both in **Windows** (where it does seem to work) and **macOs** side by side, to find out where the diverging point is.

The only thing I can think about is some weird issue related to state so... let's join our hands and pray to Cthulhu.

Juanlu001:
Thanks a lot Alberto!! Keep us posted about your progress, let's see if we
can release the fixes this Thursday/Friday :)

Juanlu001:
I'm updating this branch to see if the error happens again.

Juanlu001:
Folks, a brief summary on this pull request: @newlawrence has been trying to reproduce the Travis failure on two OS X laptops, installing *exactly* the same package versions, and didn't succeed. It's probably some subtle issue with dependencies, but we will have to narrow down the issue before reporting it upstream. On the other hand, I'm in a bit of a hurry to make the release, so I propose the following:

1. Release 0.8.0 without this in.
2. Wait until we submit a bug report in Travis CI. This might take some time, because in the ideal case, it should be very simple to showcase the buggy behavior without having to install poliastro and all its dependencies.
3. When we have it, conditionally disable the test in OS X and include this in master, so it can be released as part of poliastro 0.9.0.

I also considered including this in 0.8.1, but I am not sure about this. Thoughts?

Juanlu001:
Some months have passed, and I wonder if there has been any difference. I'm closing and reopening this.

Juanlu001:
Oops, now the failure is different. I suspect it's a change in Astropy, poliastro (and this PR) needs an update.

Juanlu001:
Well, Astropy 3.0 is almost there, so I will first update poliastro to reflect that and then we can take care of this. Anyway, this branch needs a rebase. @anhiga do you still have time to do it?

Juanlu001:
Uff nothing worse than solving git conflicts on Jupyter notebooks. For them we will have to start from scratch I'm afraid.

newlawrence:
Can it be, the [nbdime](https://github.com/jupyter/nbdime) tool, of any help?

Juanlu001:
Oh, maybe! http://nbdime.readthedocs.io/en/stable/merging.html

Juanlu001:
@newlawrence Back to the misterious OS X bug: do you think that the difference in NumPy version might be the source of the XPASS?

https://github.com/poliastro/poliastro/issues/358

newlawrence:
I'm pretty sure I've replicated the full development environment matching every single library version. Anyway, I'll take care of #358, to check if upgrading **pytest** and **NumPy** versions for the **linux** builds reproduce the problem on that platform.

Juanlu001:
This will be superseded by https://github.com/poliastro/poliastro/pull/439, so I'm closing.

PR #248, opened 2017-09-29
Warn for non-TDB time scales in from_body_ephem

### LINKED ISSUE #234: Add warning if input time for ephemeris is not TDB
Those users that want utter precision might want to check against HORIZONS or SPICE. These systems use TDB as their input, so in the absence of time scale an implicit conversion is made, which makes use of the "wrong" time:

https://space.stackexchange.com/q/22901

Example code:

```
In [17]: EPOCH = Time("2017-09-01 12:05:50")

In [18]: Orbit.from_body_ephem(Venus, EPOCH)
Out[18]: 1 x 1 AU x 24.4 deg orbit around Sun (☉)

In [19]: _.epoch
Out[19]: <Time object: scale='utc' format='iso' value=2017-09-01 12:05:50.000>

In [20]: _.tdb
Out[20]: <Time object: scale='tdb' format='iso' value=2017-09-01 12:06:59.183>

In [21]: _19 == _20
Out[21]: True
```

I would raise a warning if the input scale is not TDB to avoid this confusion, and also set the epoch to TDB to signal this fact. Thoughts? @poliastro/development

Juanlu001:
Fix #234.

### DIFF
--- a/src/poliastro/ephem.py
+++ b/src/poliastro/ephem.py
@@ -6,6 +6,10 @@ from astropy import units as u
 from astropy.coordinates import get_body_barycentric_posvel
 
 
+class TimeScaleWarning(UserWarning):
+    pass
+
+
 def get_body_ephem(body, epoch):
     """Position and velocity vectors of a given body at a certain time.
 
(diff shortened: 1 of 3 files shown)


anhiga:
Tested and worked flawlessly. If you don't mind, I'm merging a pull request for the first time 😄

Juanlu001:
Of course I don't mind :) Thanks!

PR #250, opened 2017-10-01
Add propagate to date

### LINKED ISSUE #236: Add propagation to a certain date
Right now the date arithmetic is left to the user, but there are some subtleties that can appear when handling different time scales. @anhiga already experimented with some of these issues, so adding a task here.

Juanlu001:
Waiting for official response for this Astropy issue:

https://github.com/astropy/astropy/issues/6638

This should fix #236.


Juanlu001:
Issue astropy/astropy#6638 has been identified as a (novice-friendly, easy to fix) bug. I will leave this open until we fix it there and decide whether this deserves a workaround or if we will require users to install latest Astropy.

Juanlu001:
...however, when a fix was attempted, we discovered that the issue was less trivial than expected https://github.com/astropy/astropy/pull/6653. I'm temporarily reducing the precision of the test as a workaround.

Juanlu001:
This is ready for review!

Juanlu001:
Thanks! Merging

PR #251, opened 2017-10-02
Jupyter in CI

### LINKED ISSUE #247: Run notebooks in CI?
Sometimes notebooks get outdated and the continuous integration process doesn't notice (see for instance #245). We should investigate if we can run them as part of the CI as well.

anhiga:
Trying to fix #247. First I want to test if this works in Travis


Juanlu001:
In case you're wondering: it's perfectly OK to add jupyter, ipywidgets and so on as dependencies. More and more I think we should build a "notebook-centered" tool focused on interactivity.

anhiga:
I was wondering exactly that. It will increase the build time significantly.

anhiga:
By the way, this PR has become a way of testing Travis and Jupyter (which I should have done in my fork, sorry for that). I will close this one and open another when it is all ok.

Juanlu001:
Perhaps, but now that you started this let's get a green build and see how the build time is affected.

We are doing many things in one build in a sequential way: code quality, tests, coverage, documentation, now notebooks. Perhaps we can have several Travis jobs (Linux, so they go fast) that are more lightway so the build is faster. But that should be a separate change.

Juanlu001:
> By the way, this PR has become a way of testing Travis and Jupyter (which I should have done in my fork, sorry for that). I will close this one and open another when it is all ok.

Don't worry for that :)

anhiga:
At the moment, PYTHON=3.5 builds (osx and Linux) are not working because _Going to Mars with Python using poliastro_ is using PEP 498 formatted string literals, which were added in 3.6:
```Python
f"Earth at launch\n({date_launch.datetime:%b %d})"
```
and 3.6 build was failing because _Propagation using Cowell's formulation_ has a cell with 6 `assert` which is really time-consuming, so increasing the MaxTimeout seems to solve the problem.

Juanlu001:
This helped us find that the notebooks do not support the Python version we claim to support. Awesome!

Builds are now one minute longer in Linux and two minutes slower in Windows. OS X times are too long now though, and those machines are scarce so I'd advocate for simplifying the OS X jobs and leaving the rest as they are. Plus, you are invited to fix the failing notebooks in this same PR :)

Juanlu001:
I think we should support Python 3.5 at least until 3.7 is out next June:

https://www.python.org/dev/peps/pep-0537/#release-schedule

anhiga:
I fixed _Going to Mars with Python using poliastro_, @Juanlu001  do you have any suggestion about _Propagation using Cowell's formulation_? It works now, but as I said it seems to need more than 30 seconds to execute.

anhiga:
Looking at the osx Travis log, it fails at the first notebook, because it needs more than 60s to execute it. I guess it depends on the build environment speed.

Juanlu001:
> I guess it depends on the build environment speed.

Perhaps let's just disable jupyter and docs builds in OS X?

> do you have any suggestion about `Propagation using Cowell's formulation`? It works now, but as I said it seems to need more than 30 seconds to execute.

I suggest commenting out the assert, so it won't run but the user can always uncomment the line. That way we don't have to touch the timeout.

Juanlu001:
Looks good to me. Merging!

PR #258, opened 2017-10-05
Added pip to RTD conf file

anhiga:
ReadTheDocs now builds poliastro docs using pip instead of conda,
and conda has been completely removed from poliastro CI, docs, etc.


anhiga:
I have tested this on my own RTD branch, and even though it fails (unrelated to this, as far as I know), installation with pip works perfectly.

Juanlu001:
I see! This is the failure:

https://readthedocs.org/projects/anhiga-poliastro/builds/6082645/

It's strange because the `set_first_last` method is defined here:

http://repo.or.cz/docutils.git/blob/refs/tags/docutils-0.14:/docutils/writers/html4css1/__init__.py#l213

I don't see this error reported anywhere, but I suspect it is some version mismatch between Sphinx, nbsphinx, and docutils.

Juanlu001:
The `set_first_last` method has been there since 2004 so I suspect there's a problem in nbsphinx.

http://repo.or.cz/docutils.git/blame_incremental/19aec1e70c4828d113ff64a12e24afe93df40b6c:/docutils/writers/html4css1.py#l368

Juanlu001:
I was wrong: nbsphinx uses `docutils.writers.html5_polyglot.HTMLTranslator` 

https://github.com/sphinx-doc/sphinx/blob/1.6.4/sphinx/writers/html5.py#L18-L37

which inherits from `docutils.writers._html_base.HTMLTranslator`:

http://repo.or.cz/docutils.git/blob/refs/tags/docutils-0.14:/docutils/writers/html5_polyglot/__init__.py#l148

which doesn't have a `set_first_last` method.

Juanlu001:
Sorry, I am totally dumb and restricted the wrong version of the wrong package. We can throw that commit to the thrash later... Still don't know what happens, working on it.

Juanlu001:
I suspect there are some problems when using the pip installation and trying to update Sphinx in Read the Docs. If this latest change doesn't work, we will have to report an issue.

Juanlu001:
I think this did the trick, but I don't understand the timeout. Let me check...

Juanlu001:
My build succeeded but it took 13.2 minutes, too close to the [limit of 15 minutes](http://docs.readthedocs.io/en/latest/builds.html):

https://readthedocs.org/projects/juanlu-poliastro/builds/6083401/

I nevertheless reported the issue upstream:

https://github.com/rtfd/readthedocs.org/issues/3148

Juanlu001:
Perhaps the longer build times are because we are running the notebooks in RTD. As we are already doing it in Travis, perhaps we could disable it.

Juanlu001:
Even after the last commit, not only did nbsphinx run but we got HTTP 429 errors, and for a good reason...

https://readthedocs.org/projects/anhiga-poliastro/builds/6083507/

Trying to temporarily set it to `never` and see what happens.

Juanlu001:
Good dammit, I did the reverse thing!!!! Never, ever program in a hurry...

Juanlu001:
Now it works! Sorry for the mess :pray:

anhiga:
I prettify the commit log a little bit. I think it is ready to be merged :)

Juanlu001:
Thanks a lot! Merging :grin:

PR #259, opened 2017-10-09
Update mybinder

Juanlu001:
Now we are using the beta version, and a `requirements.txt` that automatically downloads the latest version from master.


Juanlu001:
This is fairly trivial, so I'm merging it myself :grin:

Issue #218, opened 2017-08-25
Simplify 2D plotting


Juanlu001:
Historically there has been some code in the plotting functions that is doing lots of complicated things. I'm referring to

https://github.com/poliastro/poliastro/blob/1ec82f1/src/poliastro/plotting.py#L122-L141

This is

1. Sampling the orbit
2. ...in the perifocal frame (PQW)
3. Projecting to the plane of the orbit (`e` and `p` axis of PQW)
4. Projecting to the plane of a previous orbit (another PQW frame)

They are only 19 lines but, as you can see, very meaty.

I think we can simplify this by:

* Having a `.sample()` method in `Orbit` objects that accepts a discretization step, or an array of times, or true anomaly values... Something that returns an array of orbital elements of vectors.
* Having the ability to convert this vector, which is centered on `Orbit.attractor`, to PQW (see #108 and #215)
* Then select only the coordinates that belong to the plane

This probably deserves further discussion in the wiki and splitting into different issues, but let us put it here so we do not forget.

Juanlu001:
I'm missing some logic that relates to the projection in a different plane, but that is why we need discussion :)

Juanlu001:
The road ahead is more or less clear, but we should implement sampling first. Adding new issues, blocking this and postponing.

Juanlu001:
New issues: #236 #237.

Juanlu001:
Also, as noted in https://github.com/poliastro/poliastro/pull/228#issuecomment-327460712 and https://github.com/matplotlib/matplotlib/issues/9155#issuecomment-327458806, we should store all the orbits being plotted in their original reference frames, and then reproject. This will allow for reprojection and better legend handling.

Juanlu001:
Now that #263 has been implemented, this is no longer blocked.

Juanlu001:
Interestingly, now plotting in 3D is way easier than plotting in 2D!

![screenshot-2017-10-13 untitled](https://user-images.githubusercontent.com/316517/31534392-065ed8f0-aff7-11e7-8832-02117d886e37.png)

So I'm retargetting this as "simplify 2D plotting", which involves reprojection, selecting a plane, etc.

Juanlu001:
Assigning this to @anhiga, who volunteered to do it :)

Juanlu001:
Closed in #269!

Issue #117, opened 2016-02-23
Export to CZML


Juanlu001:
See https://github.com/poliastro/poliastro/issues/22 for previous ideas.

Juanlu001:
Relevant libraries:

* CesiumWidget http://mybinder.org/repo/petrushy/CesiumWidget
* cesiumpy http://mybinder.org/repo/sinhrks/cesiumpy

Juanlu001:
Postponing this to 0.7, when the exporting capabilities are more mature.

Juanlu001:
This might need a bit of refactoring after #22 is finally there. Postponing.

Juanlu001:
Retargetting this as 0.8 as part of #22.

Juanlu001:
Blocked on #237.

Juanlu001:
Now that #263 has been implemented, this is no longer blocked.

Juanlu001:
I have been doing a bit of research and I am not sure that exporting to CZML is worth it at all, since Cesium is mainly Earth-based (which is the main thing poliastro tries to avoid). Here is the email I have sent to an AGI representative:

> After some months, I am finally in a position where I can use poliastro with Cesium and CZML. However, I have been digging in the documentation and reading old google groups discussions and I could not find one feature that is essential for me: the ability to specify a central body that is *not* the Earth. That is mentioned in this outdated roadmap from 2015:
> 
> https://github.com/AnalyticalGraphicsInc/cesium/wiki/Architecture
> 
> > The CentralBody primitive represents the globe (in a future Cesium version, any central body such as the Moon and Mars will be supported).
> 
> 
> However, there are no further references since then.
> 
> I actually don't need high accuracy planetary imagery for each of the bodies of the Solar System, but a much simpler feature: specify a plain body with any color. Most of our work is based on interplanetary trajectories in the Solar System, so the sizes of the planets would be extremely small in comparison.
> 
> On the other hand, by reading the simple.czml example I see that nowhere it is specified which central body do the satellites orbit:
> 
> https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Specs/Data/CZML/simple.czml
> 
> This means that I can focus on exporting an orbit as a correct CZML and leave the rest to the user. But I fear that this will lead to incorrect results because there are many Earth-based assumptions in Cesium.

Juanlu001:
[...]

PR #263, opened 2017-10-11
Sample orbits

### LINKED ISSUE #237: Add method to sample an orbit
This is very useful for things like generating a table of elements (or `DataFrame`s, @Guillem-db!). One option would be receiving an array of `Time` values and returning a list of `Orbit` objects, but I do not have a strong opinion about this and it is open to discussion. Other options are:

* A `CartesianRepresentation`-like object, which can contain an array of values.
* An "interpolator-like" object - but this would be merely a proxy to `propagate`
* ?

Also, bear in mind that we should ensure that propagating backwards works.

Juanlu001:
Fix #237.

### DIFF
--- a/src/poliastro/util.py
+++ b/src/poliastro/util.py
@@ -4,6 +4,7 @@
 import numpy as np
 
 from astropy.coordinates import matrix_utilities
+from astropy.time import Time
 
 
 def circular_velocity(k, a):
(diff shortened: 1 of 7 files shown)


Juanlu001:
Floating point issues when manipulating astropy times strike back: https://github.com/astropy/astropy/issues/6638

Juanlu001:
I like how this is evolving :)

screenshot-2017-10-12 untitled

Juanlu001:
This pull request is ready for review.

Juanlu001:
This, by the way, unblocks #218 and #117.

anhiga:
I'll review this as soon as possible. By the way, Travis keep failing due to notebook execution.

Juanlu001:
You are right about the notebooks... I restarted the job and failed nonetheless. It's always the same notebook, so I'm going to disable it.

src/poliastro/util.py
```diff
@@ -71,3 +72,22 @@ def norm(vec):
 
     """
     return np.sqrt(vec.dot(vec))
+
+
+def time_range(start, *, periods=50, spacing=None, end=None):
+    """Generates range of astronomical times.
```
anhiga: This docstring should have `Parameters` and `Returns`.

Juanlu001: Good point, thanks! Will do.

Juanlu001:
Thanks @anhiga, as always! I improved the docstring as you suggested and did other minor doc additions. Will merge when the build is green.

PR #269, opened 2017-10-24
Simplify 2D plotting

anhiga:
After @Juanlu001 creating the `sample` function, simplification of 2d plotting got significantly easier.
The changes are the following:
* Now `set_frame` works also when a frame already exists, by removing all the lines and patches (not the legend), and replotting again.
* a list of tuples (orbit, legend) is stored to achieve the former.
* `sample` function is used in plotting, reducing code complexity.
Closely related to #218


src/poliastro/plotting.py
```diff
@@ -9,10 +9,8 @@
 from astropy import units as u
 from astropy.coordinates.angles import Angle
 
-from poliastro.twobody.classical import rv_pqw
```
Juanlu001: I *love* to see this go away!

src/poliastro/plotting.py
```diff
@@ -57,29 +55,36 @@ def __init__(self, ax=None, num_points=100):
         self.num_points = num_points
         self._frame = None
         self._attractor_radius = None
+        self.orbits = list(tuple())  # type: List[Tuple[poliastro.twobody.orbit.Orbit], str]
```
Juanlu001: Thanks for adding the type string!

src/poliastro/plotting.py
```diff
@@ -118,20 +125,7 @@ def plot(self, orbit, label=None, color=None):
...
-                          orbit.ecc.value,
-                          nu_vals.value)
-        r_pqw = r_pqw * u.km
-
-        # Express on inertial frame
-        e_vec, p_vec, h_vec = orbit.pqw()
-        p_vec = np.cross(h_vec, e_vec) * u.one
-        rr = (r_pqw[:, 0, None].dot(e_vec[None, :]) +
-              r_pqw[:, 1, None].dot(p_vec[None, :]))
+        rr = orbit.sample(self.num_points).get_xyz().transpose()
```
Juanlu001: As simple as this! :heart_eyes:

src/poliastro/plotting.py
```diff
@@ -57,29 +55,36 @@ def __init__(self, ax=None, num_points=100):
...
-                raise ValueError("Vectors must be mutually orthogonal.")
-            else:
-                self._frame = p_vec, q_vec, w_vec
+        old_frame = False
+        if self._frame is not None:
+            old_frame = True
+
+        if not np.allclose([norm(v) for v in (p_vec, q_vec, w_vec)], 1):
+            raise ValueError("Vectors must be unit.")
+        if not np.allclose([p_vec.dot(q_vec),
```
Juanlu001: `elif`?

src/poliastro/plotting.py
```diff
@@ -57,29 +55,36 @@ def __init__(self, ax=None, num_points=100):
...
+                            q_vec.dot(w_vec),
+                            w_vec.dot(p_vec)], 0):
+            raise ValueError("Vectors must be mutually orthogonal.")
         else:
-            raise NotImplementedError
+            self._frame = p_vec, q_vec, w_vec
+
+        if old_frame:
```
Juanlu001: The complexity in this method is too high. Suggestions:

* Move `for artist ... artist.remove()` to a `_clear()` method
* Move `if old_frame ... self.plot(orbit, label)` to a `_redraw()`, `_reproject()` or `_replot()` method
* Try to get rid of the `old_frame` variable completely, for instance instead of `if old_frame` do `if self.orbits`, which will check if there are orbits to replot

src/poliastro/plotting.py
Juanlu001: On the other hand: shouldn't this be `self._orbits`, with optionally a `@property` to give read-only access?

Juanlu001:
Also, there is one mypy failure and one test failure in appveyor, check them out.

src/poliastro/plotting.py
anhiga: You are righ! Changing.

src/poliastro/plotting.py
anhiga: The main problem here is how the figure size is set. Now, figure size depend on whether there is a legend or not. Otherwise, `self.ax.clear()` could be used...
I'm trying to find a better solution, comments are welcome.

anhiga:
Notebook execution is failing in all platforms. I'm trying to know why.

Juanlu001:
The problem lies in sampling an hyperbolic orbit. Right now the `.sample` method assumes a closed orbit, and that's a bug. Previously, the logic for sampling hyperbolic orbits was here:

https://github.com/anhiga/poliastro/blob/bde55913ae3d3315e57ff1f553c9bd63de214ebe/src/poliastro/plotting.py#L170-L185

Which, by the way, should go away as well.

Juanlu001:
And if that code goes away, then `num_points` should go away too!

Juanlu001:
I tried to take over but there are a handful of errors, let's see what we can do.

Juanlu001:
Well! I made some key changes and things look better now. Please take a look, and be so kind to add some tests :grin:

Juanlu001:
Oops, conflicts appeared after merging #271.

PR #271, opened 2017-10-24
Add 3D plotting functions

### LINKED ISSUE #22: 3D plotting
Summary:
- ~~Mayavi~~ https://github.com/enthought/mayavi is now having a second life, Python 3 support and Jupyter notebook exporting capabilities, but used to be buggy in the past
- ~~vispy~~ https://github.com/vispy/vispy seems to have stalled
- ~~PyOpenGL~~ http://pyopengl.sourceforge.net/ is a bit low level, seems more appropriate to render volumes
- ~~matplotlib~~ https://github.com/matplotlib/matplotlib has very limited 3D capabilities
- ~~Cesium~~ https://github.com/AnalyticalGraphicsInc/cesium seems more appropriate to render low-Earth orbits with realistic detail
- ~~GR~~ https://github.com/jheinen/gr does not work properly on Python 3 yet, does not solve matplotlib problems when used as a backend, is poorly documented, a tiny community and has lower quality results
- plotly https://github.com/plotly/plotly.py looks like the way to go!
- Bokeh just added extensions, which allow 3D plotting capabilities through JS libraries https://demo.bokehplots.com/apps/surface3d
- pythreejs https://github.com/jovyan/pythreejs

Juanlu001:
* No tests
* Still pending some refactoring in plotting
* Should we add a 3D equivalent to OrbitPlotter?

Fix #22

This is the result:

screenshot-2017-10-24 plotting in 3d


anhiga:
It looks really amazing!. And quite easy to understand, thanks to that `sample` function :+1:

Juanlu001:
Oops, I almost forgot: no attractor! 😅 Will do this or next week.

Juanlu001:
One interesting issue appeared while using `OrbitPlotter3D`:

Until now, `OrbitPlotter` has been used as an "immutable" object, adding plots in one Jupyter cell, which shows the figure. However, if we add `set_view` methods (or, as it is being done in #269, `set_frame` methods) it makes sense to plot some things in one cell, then changing the properties of the plot in the next cell.

However, when using `%matplotlib inline` ([docs](http://ipython.readthedocs.io/en/stable/interactive/plotting.html#id1)], the figure is only shown when it is created (first cell), and not when it is edited (second cell). Moreover, if we decide to return the figure, if it is created in the same cell it will be shown twice.

...And while I was reading that piece of documentation, I discovered that `%matplotlib inline` is not needed in recent versions of IPython! Which means I will have to review not only the poliastro documentation, but all the courses and articles I usually write...

Juanlu001:
In fact, I'm thinking whether it's a good idea to "reinvent" matplotlib semantics in this respect (i.e. figures are only shown once, in principle). Still have to reflect a bit more.

Juanlu001:
I am so far very dissatisfied with 3d plots in matplotlib, and also there are a number of important issues related to them that nobody is working on.

On the contrary, plotly looks promising:

https://gist.github.com/Juanlu001/bf6a63ebb55ae4e9f76930cc1ebef22c

newplot

Juanlu001:
Going fine! Pending:

* [ ] Plot objects other than orbits
  - Objects with `sample()`? -> Probably work already
  - Objects already sampled? -> Use case: non keplerian orbits from planets (Workaround: use osculating orbits)
* [ ] First show is shown, correct this
* [ ] Access figure or layout from the outside
* [ ] Implement set_view with elevation and azimut
* [ ] Programmatically export static image

Juanlu001:
I have been thinking lately that the output of `orbit.sample()` has not enough information to send it to `plot`, since the attractor is lost. Options:

* Introduce a new object, such as a `Trajectory` or something, which combines a `CartesianRepresentation` (a Reference Frame in the future) with a `Body`.
* If the object has no attractor, simply don't plot it, or plot a gray sphere without name, or assume it's the already configured attractor of the frame.

This is complex enough to deserve its own pull request, so I am deferring the implementation for next week or the following, if we're in time.

Juanlu001:
As a provisional solution to my concerns above, I think I'm implementing a `plot_trajectory` method that simply receives a `CartesianRepresentation`. An attractor must be set first, so I will make the `_set_attractor` method public.

Juanlu001:
I am now satisfied with the general implementation, and for the rest I am focusing on writing tests, documentation and changing the existing notebooks -- including the venerable MSL notebook, which has remained mostly the same for more than three years since I first presented it at PyData London :heart:

Juanlu001:
In the meanwhile, @anhiga would you be so kind to review it?

Juanlu001:
It's not clear though what's the status of interactive plots, see my comment in https://github.com/plotly/plotly.py/pull/641#issuecomment-342276064 and the linked issues.

src/poliastro/plotting.py
```diff
@@ -182,3 +199,173 @@ def _generate_vals(self, state):
             nu_vals[nu_invalid] = np.nan
 
         return nu_vals
+
+
+def _generate_label(orbit, label):
```
anhiga: I think this could also be used in the current `OrbitPlotter` class.

Juanlu001: You are right, but I didn't want to touch a line of `OrbitPlotter` to avoid conflicts. We can of course move it!

Juanlu001:
Thanks a lot @anhiga for the review. I'm aware that the coverage decreased but we're in a bit of a hurry so I'm merging this now. You can then merge or rebase #269, or combine some changes if you find them appropriate, or leave it as it stands. I will work on improving documentation and coverage in a separate PR.

Issue #275, opened 2017-10-26
Implement hyperbolic anomaly conversion functions


Juanlu001:
🐞 **Problem**

We should have hyperbolic anomaly conversion functions, in the same way we have elliptic anomaly functions here:

https://github.com/poliastro/poliastro/blob/v0.7.0/src/poliastro/twobody/angles.py

Right now, using `nu_to_M` for eccentricity values higher than one gives NaN:

```
In [2]: from poliastro.twobody.angles import *

In [3]: from astropy import units as u

In [4]: nu_to_M(0.1 * u.rad, 1.5 * u.one)
/home/juanlu/.miniconda36/envs/py36/lib/python3.6/site-packages/astropy/units/quantity.py:641: RuntimeWarning: invalid value encountered in sqrt
  *arrays, **kwargs)
Out[4]: <Quantity nan rad>
```

🎯 **Goal**

This is important since `Orbit.sample` uses the angle conversion functions to derive appropriate time values for propagation, and right now it doesn't work for hyperbolic orbits. 

💡 **Possible solutions**

We should implement these equations:

![hyperbolic](https://user-images.githubusercontent.com/316517/32038805-f3d8f26c-ba2b-11e7-94d3-68065d5ad5e6.png)

(from Curtis "Orbital Mechanics for Engineering Students")

See also: https://en.wikibooks.org/wiki/Astrodynamics/Time_of_Flight

However, there are limitations in the true anomaly values, so care should be taken:

![screenshot-2017-10-26 fundamentals of astrodynamics and applications](https://user-images.githubusercontent.com/316517/32038835-12416a86-ba2c-11e7-99d9-78ee00ccc7ae.png)

(from Vallado "Fundamentals of Astrodynamics")

Finding references for validation proved to be a bit difficult, so I posted a question on Physics.SE here https://physics.stackexchange.com/q/365117/7641

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
I'm labeling this as a bug to fix in 0.8.

anhiga:
Closed by #276

PR #276, opened 2017-10-26
Add conversion between different anomalies in hyperbolic orbits

anhiga:
Current anomaly conversions (mean -> true, true -> eccentric, ...)
are only valid when eccentricity < 1, not for hyperbolic orbits.

### DIFF
--- a/src/poliastro/twobody/angles.py
+++ b/src/poliastro/twobody/angles.py
@@ -109,8 +109,9 @@ def F_to_nu(F, ecc):
         True anomaly (rad).
 
     """
-    nu = 2 * np.arctan((np.exp(F.value) * np.sqrt(ecc + 1) - np.sqrt(ecc + 1)) /
-                       (np.exp(F.value) * np.sqrt(ecc - 1) + np.sqrt(ecc - 1)))
+    with u.set_enabled_equivalencies(u.dimensionless_angles()):
+        nu = 2 * np.arctan((np.exp(F) * np.sqrt(ecc + 1) - np.sqrt(ecc + 1)) /
+                           (np.exp(F) * np.sqrt(ecc - 1) + np.sqrt(ecc - 1)))
     return nu
 
 
--- a/src/poliastro/twobody/angles.py
+++ b/src/poliastro/twobody/angles.py
@@ -16,6 +16,14 @@ def _kepler_equation_prime(E, M, ecc):
     return 1 - ecc * np.cos(E)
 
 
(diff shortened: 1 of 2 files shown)


anhiga:
You are right, all the functions in the `angles` module accept float instead of quantities, but in the corresponding tests, they all are used with quantities, without problems.

However, some of the new functions can't accept both types. Should they accept float then? Or should we change all the functions to work with quantities?

src/poliastro/twobody/angles.py
```diff
@@ -60,6 +93,27 @@ def E_to_nu(E, ecc):
...
+    ecc : float
+        Eccentricity (>1).
+
+    Returns
+    -------
+    nu : float
+        True anomaly (rad).
+
+    """
+    nu = 2 * np.arctan((np.exp(F.value) * np.sqrt(ecc + 1) - np.sqrt(ecc + 1)) /
```
Juanlu001: Here is the problem with the units of `F`, right?

Juanlu001:
All the tests pass with this diff:

```diff
diff --git a/src/poliastro/twobody/angles.py b/src/poliastro/twobody/angles.py
index ccb6ed3..3f50412 100644
--- a/src/poliastro/twobody/angles.py
+++ b/src/poliastro/twobody/angles.py
@@ -109,8 +109,8 @@ def F_to_nu(F, ecc):
         True anomaly (rad).
 
     """
-    nu = 2 * np.arctan((np.exp(F.value) * np.sqrt(ecc + 1) - np.sqrt(ecc + 1)) /
-                       (np.exp(F.value) * np.sqrt(ecc - 1) + np.sqrt(ecc - 1)))
+    nu = 2 * np.arctan((np.exp(F) * np.sqrt(ecc + 1) - np.sqrt(ecc + 1)) /
+                       (np.exp(F) * np.sqrt(ecc - 1) + np.sqrt(ecc - 1)))
     return nu
 
 
```

So I'm not sure what the problems are.

Juanlu001:
Oh, I tried to implement `test_mean_to_true_hyperbolic` and now I see. Working on it.

Juanlu001:
Done! Please review my changes to your changes :sweat_smile:

anhiga:
The problem was that `np.exp()` raised an error when used with quantities, but I can't try it now, and tests are passing now, so...

Juanlu001:
Exactly, the error was fixed using the "angles are dimensionless" trick as it's used in other functions.

anhiga:
Great then, merging and closing #275

PR #289, opened 2017-11-28
Speed up vallado.lambert implementation

pablogsal:
This PR speeds up between **9% and 20%** the (optionally) jit compiled version of Vallado's lambert solver. Results up front using the already available benchmarking code:

| Name (time in us) | Min| Max |  Mean|  StdDev | Median | IQR | Outliers | OPS (Kops/s)|           Rounds | Iterations |
| --------| --------| --------| --------| --------| --------| --------| --------| --------| --------| --------| 
test_lambert_single_rev[lambert0] (NOW)  |            229.7380 (1.0)  |    302.0810 (1.0)   |   247.2138 (1.0) |     31.0362 (1.02)  |   232.8300 (1.0)   |   26.6762 (1.20)   |       1;1    |    4.0451 (1.0)      |     5    |       1|
test_lambert_single_rev[lambert0] (0004_oldmeth)  |   256.5990 (1.12) |    335.1490 (1.11) |    282.1900 (1.14)   |  30.4832 (1.0)   |   271.9400 (1.17) |    22.1830 (1.0)   |        1;1        |3.5437 (0.88)   |       5      |     1|

Depending on the vectorization capabilities of the CPU and the compiler optimization level (and the compiler itself) different improvements are produced in the range of (1.08x - 1.21x).

The improvement in performance was produced by:

-  Factor out common terms that are recalculated in the loop body (`norm_r0*norm_r` and `norm_r0+norm_r`).
- Removed redundant condition in the outermost if condition that precedes the second while.
- Removed branch of the `else` clause for the break statement.
- Removed branch in the bisection check.

The impact in legibility is low, given a good trade-off between speed improvement and legibility IMHO.

Further speed will be trivially gained given type information to the jit decorator. For example:
```python
@nb.jit(nb.types.UniTuple(nb.float64[:],2)(nb.float64,nb.float64[:],nb.float64[:],nb.float64,nb.boolean,nb.float64,nb.float64))
```
Still, there is a lot of python overhead for the compiled version of this function. The deepest python call reported by `Perf` is 45 stacks deep in and out the Python interpreter layer, going through the numpy API instead of working with the numpy buffers themselves. 

### Unrelated

Here are some good random lyrics: :notes: 

> Don't listen to the words they say 
They could never believe you anyway 
Just let it go another day 
I believe that love is gonna last forever 
And it's all within my mind, and it's all within my mind 

WolfMother - Far Away

### DIFF
--- a/src/poliastro/iod/vallado.py
+++ b/src/poliastro/iod/vallado.py
@@ -61,7 +61,10 @@ def _lambert(k, r0, r, tof, short, numiter, rtol):
 
     norm_r0 = np.dot(r0, r0)**.5
     norm_r = np.dot(r, r)**.5
-    cos_dnu = np.dot(r0, r) / (norm_r0 * norm_r)
+    norm_r0_times_norm_r = norm_r0 * norm_r
+    norm_r0_plus_norm_r = norm_r0 + norm_r
+
+    cos_dnu = np.dot(r0, r) / norm_r0_times_norm_r
 
     A = t_m * (norm_r * norm_r0 * (1 + cos_dnu))**.5
 
@@ -73,16 +76,17 @@ def _lambert(k, r0, r, tof, short, numiter, rtol):
     psi_up = 4 * np.pi
 
     count = 0
+
     while count < numiter:
-        y = norm_r0 + norm_r + A * (psi * c3(psi) - 1) / c2(psi)**.5
-        if A > 0.0 and y < 0.0:
+        y = norm_r0_plus_norm_r + A * (psi * c3(psi) - 1) / c2(psi)**.5
+        if A > 0.0:
             # Readjust xi_low until y > 0.0
             # Translated directly from Vallado
             while y < 0.0:
                 psi_low = psi
                 psi = (0.8 * (1.0 / c3(psi)) *
-                       (1.0 - (norm_r0 + norm_r) * np.sqrt(c2(psi)) / A))
-                y = norm_r0 + norm_r + A * (psi * c3(psi) - 1) / c2(psi)**.5
+                       (1.0 - norm_r0_times_norm_r * np.sqrt(c2(psi)) / A))
+                y = norm_r0_plus_norm_r + A * (psi * c3(psi) - 1) / c2(psi)**.5
 
         xi = np.sqrt(y / c2(psi))
         tof_new = (xi**3 * c3(psi) + A * np.sqrt(y)) / np.sqrt(k)
@@ -90,14 +94,13 @@ def _lambert(k, r0, r, tof, short, numiter, rtol):
         # Convergence check
         if np.abs((tof_new - tof) / tof) < rtol:
             break
-        else:
-            count += 1
-            # Bisection check
-            if tof_new <= tof:
-                psi_low = psi
-            else:
-                psi_up = psi
-            psi = (psi_up + psi_low) / 2
+        count += 1
+        # Bisection check
+        condition = (tof_new <= tof)
+        psi_low = psi_low + (psi - psi_low) * condition
+        psi_up = psi_up + (psi - psi_up) * (not condition)
+
+        psi = (psi_up + psi_low) / 2
     else:
         raise RuntimeError("Maximum number of iterations reached")
 


Juanlu001:
Hi @pablogsal! Thanks a lot for contributing to poliastro :heart: 

The diff looks great to me, and I have no objections to merging it. I would like to ask a couple of questions before:

* What was your approach to optimize this function? I see that the code changes are fairly contained. Does this mean that you first tried to scratch the surface? Or by contrast you already tried more aggressive optimizations and could not make it faster because the implementation is already good enough?
* I've been advised by the numba developers several times to not add types information to the JIT declarations ([one example](https://github.com/numba/numba/issues/1755#issuecomment-203464304), [another example](https://github.com/numba/numba/issues/1691#issuecomment-181365121)). However, if I understand correctly the performance impact should be significant only on the first call, when numba infers the types. Correct?
* Can you expand "there is a lot of python overhead for the compiled version of this function"? Did you detect more hotspots? Is there something we can do about it? Perhaps some strange NumPy function being called?

pablogsal:
- My approach is just scratching the surface, but I am not sure on the extent of the possible optimizations that can be performed after this without looking at the llvm intermediate representation. I usually look at the perf stats for the generated function and even taking into account that numba generates llvm in a black box fashion things like branch mispredictions and the time performance of different basic operation still applies (but I always check my assumptions using perf or similar approach). That is the reason I knew that removing the branches will be more performant. Regarding the refactoring of the common terms, it depends on how well numba does [CSE](https://en.wikipedia.org/wiki/Common_subexpression_elimination) but implicit calls to the numpy API is usually out of scope (as you can check overloading `__add__` and repeating the addition of 2 numpy arrays in the code and checking that `__add__` is called always). 
- Numba has evolved very good at inferring the types, but is just a matter of checking that the inferred types are ok with what you want. But if the inferred types are the correct ones, as you said, it will only impact the first call as far as I understand.
- With "there is a lot of python overhead for the compiled version of this function" I mean that there is a lot of calls through the numpy API and the C API of python because there are Pythonisms (raise, tuple unpack in the return, other jitted functions) and numpy operations. This has the disadvantage of some function call overhead (even if the numeric operations are very similar in nature). I have tried to use Cython to produce a **naive** Python-free version:

```python

cimport numpy as np
from libc.math cimport M_PI as pi, sqrt, abs, pow, cos, sin, cosh, sinh

cdef extern from "math.h":
    double gamma "gamma" (double)

import cython


@cython.cdivision(True)
@cython.boundscheck(False)
cdef double c2(double psi):
    r"""Second Stumpff function.

    For positive arguments:

    .. math::

        c_2(\psi) = \frac{1 - \cos{\sqrt{\psi}}}{\psi}

    """
    cdef double eps,res,delta,k

    eps = 1.0
    if psi > eps:
        res = (1 - cos(sqrt(psi))) / psi
    elif psi < -eps:
        res = (cosh(sqrt(-psi)) - 1) / (-psi)
    else:
        res = 1.0 / 2.0
        delta = (-psi) / gamma(2 + 2 + 1)
        k = 1
        while res + delta != res:
            res = res + delta
            k += 1
            delta = (-psi) ** k / gamma(2 * k + 2 + 1)

    return res


@cython.cdivision(True)
@cython.boundscheck(False)
cdef double c3(double psi):
    r"""Third Stumpff function.

    For positive arguments:

    .. math::

        c_3(\psi) = \frac{\sqrt{\psi} - \sin{\sqrt{\psi}}}{\sqrt{\psi^3}}

    """
    cdef double eps,res,delta,k

    eps = 1.0
    if psi > eps:
        res = (sqrt(psi) - sin(sqrt(psi))) / (psi * sqrt(psi))
    elif psi < -eps:
        res = (sinh(sqrt(-psi)) - sqrt(-psi)) / (-psi * sqrt(-psi))
    else:
        res = 1.0 / 6.0
        delta = (-psi) / gamma(2 + 3 + 1)
        k = 1
        while res + delta != res:
            res = res + delta
            k += 1
            delta = (-psi) ** k / gamma(2 * k + 3 + 1)

    return res

@cython.cdivision(True)
@cython.boundscheck(False)
cdef void _lambert(double k, double [:]r0,double[:] r, double tof, int short_, int numiter, double rtol,
        double [:]v0, double [:]v):
    cdef int t_m
    if short_:
        t_m = +1
    else:
        t_m = -1

    cdef double norm_r0 = sqrt(r0[0]*r0[0]+r0[1]*r0[1]+r0[2]*r0[2])
    cdef double norm_r = sqrt(r[0]*r[0]+r[1]*r[1]+r[2]*r[2])
    cdef double norm_r0_times_norm_r = norm_r0 * norm_r
    cdef double norm_r0_plus_norm_r = norm_r0 + norm_r

    cdef double cos_dnu = (r[0]*r0[0]+r[1]*r0[1]+r[2]*r0[2])  / norm_r0_times_norm_r

    cdef double A = t_m * sqrt((norm_r * norm_r0 * (1 + cos_dnu)))

    if A == 0.0:
        return

    cdef double psi = 0.0
    cdef double psi_low = -4 * pi
    cdef double psi_up = 4 * pi

    cdef int count = 0

    cdef double y,xi,tof_new
    cdef int condition

    while count < numiter:
        y = norm_r0_plus_norm_r + A * (psi * c3(psi) - 1) / sqrt(c2(psi))
        if A > 0.0:
            # Readjust xi_low until y > 0.0
            # Translated directly from Vallado
            while y < 0.0:
                psi_low = psi
                psi = (0.8 * (1.0 / c3(psi)) *
                       (1.0 - norm_r0_times_norm_r * sqrt(c2(psi)) / A))
                y = norm_r0_plus_norm_r + A * (psi * c3(psi) - 1) / sqrt(c2(psi))

        xi = sqrt(y / c2(psi))
        tof_new = (pow(xi,3) * c3(psi) + A * sqrt(y)) / sqrt(k)

        # Convergence check
        if abs((tof_new - tof) / tof) < rtol:
            break
        count += 1
        # Bisection check
        condition = (tof_new <= tof)
        psi_low = psi_low + (psi - psi_low) * condition
        psi_up = psi_up + (psi - psi_up) * (not condition)

        psi = (psi_up + psi_low) / 2
    else:
        return

    cdef double f,g,gdot

    f = 1 - y / norm_r0
    g = A * sqrt(y / k)

    gdot = 1 - y / norm_r

    v0[0]= (r[0] - f * r0[0]) / g
    v0[1]= (r[1] - f * r0[1]) / g
    v0[2]= (r[2] - f * r0[2]) / g
    v[0] = (gdot * r[0]- r0[0]) / g
    v[1] = (gdot * r[1]- r0[1]) / g
    v[2] = (gdot * r[2] - r0[2]) / g
```

These are the results:

| Name (time in us) | Min| Max |  Mean|  StdDev | Median | IQR | Outliers | OPS (Kops/s)|           Rounds | Iterations |
| --------| --------| --------| --------| --------| --------| --------| --------| --------| --------| --------| 
test_lambert_single_rev[cython]  |   229.7730 (1.0)   |   298.5470 (1.0)   |   247.4544 (1.0)   |   29.1872 (2.11)  |   233.8640 (1.0)   |   12.9770 (4.03)      |    1;0    |    4.0411 (1.0)      |     5           | 1|
test_lambert_single_rev[numba]    | 255.5220 (1.11)  |   878.3950 (2.94)  |   263.8604 (1.07)     |13.8067 (1.0)  |    262.1620 (1.12)   |   3.2002 (1.0)      |  60;303   |     3.7899 (0.94)    |   2977         |  1|

This cython implementation lacks a lot of manual tweaking (like `-O3` compilation instruction level parallelism, CPU vectorization, memap, refactors...etc) but it give us two things:

1. Numba is doing a very good job.
2. A lot of very detailed tweaking is needed to further optimize **this algorithm** with this implementation as optimizations start to be highly non-trivial.
3. Cache hits are starting to be important so a cache-friendlyness study is needed as well.

Juanlu001:
Thanks again for your detailed remarks @pablogsal. You know much more about performance optimization than me so I'm sure that you will find in poliastro a nice playground :)

For the moment, I'm merging this! :rocket:

PR #291, opened 2017-12-08
Update index.rst

MattEttus:
Include import of matplotlib and set up plotting before call to plot.


Juanlu001:
Thanks @MattEtus! Merging :shipit:

Issue #300, opened 2018-02-07
Add an easy way to plot the Solar System


Juanlu001:
🎯 **Goal**

In the same way that there are example orbits, it would be nice to be able to plot the whole Solar System in an easy way, with appropriate colors and labels.

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
Some thoughts on the API taken from [the chat](https://riot.im/app/#/room/#poliastro:matrix.org): for instance, you can do

```
from poliastro.examples import molniya
plot(molniya)
```

but one cannot do `plot(solar_system)`, since that is made up of several Orbit objects.

On the other hand,

```
from poliastro.bodies import solar_system

frame = OrbitPlotter()
for planet in solar_system: 
    frame.plot(planet)
```

is not very straightforward to me.

Another option would be

```
from poliastro.plotting.helpers import plot_solar_system
plot_solar_system()
```

where plot_solar_system(epoch=None) will plot today's osculating orbits by default.

shreyasbapat:
Is there any problem in taking this approach?
```
from poliastro.bodies import solar_system

frame = OrbitPlotter()
for planet in solar_system: 
    frame.plot(planet)
```

Juanlu001:
My problem with the first approach is that `solar_system` would be a list of `Orbit` objects, while the rest of the objects in that module are `Body`es. The second looks cleaner to me.

shreyasbapat:
Please Review the Pull Request #303

Juanlu001:
Closed in #304

PR #303, opened 2018-02-09
Added an Easy Way to plot the Solar System

shreyasbapat:
Addresses Issue #300


Juanlu001:
Hi @shreyasbapat! Thank you for your pull request 🙏 

The first thing you have to achieve is a green build. If you see the checks that triggered when you opened the pull request, some of them are red. In particular, there are some PEP8 failures:

https://travis-ci.org/poliastro/poliastro/jobs/339665480#L779

When you manage to fix those, the tests will run and will tell us whether this is mergeable or not.

Another check that will trigger is the coverage. You didn't add new tests for it, so it will likely go down. So for it to increase, you should add some sort of test that covers the function that you're creating. For instance, you can `assert` that the length of `frame.orbits` is in fact 8, the number of planets in the Solar System. You can draw inspiration from the rest of the tests of the library.

And finally, I recommend you to replace the repeated `from_body_ephem`, `plot` calls with a `for` loop 😉

shreyasbapat:
I thought to do it but as the code was really small! Anyways the for loops are not a big deal. I have look for these red checks. Thanks @Juanlu001

shreyasbapat:
Thanks @Juanlu001  . I have fixed the PEP8 failures and added test cases to it. But I am facing issues related to AstroPy. Please tell what to do.

PR #304, opened 2018-02-11
Updating to AstroPy Version 2.0.3 and Adding New Features

shreyasbapat:
1) Updated to 2.0.3
2) Issue #300  also solved


src/poliastro/plotting.py
```diff
@@ -3,22 +3,21 @@
 """
 import os.path
 from itertools import cycle
-
-import numpy as np
```
Juanlu001: Can you keep the imports as they are to reduce the number of changed lines? This will reduce the noise for the reviewer :)

src/poliastro/plotting.py
```diff
@@ -375,3 +374,19 @@ def savefig(self, filename, **layout_kwargs):
             image=ext[1:], image_filename=basename,
             show_link=False,  # Boilerplate
         )
+
+
+def plot_solar_system(epoch=None):
+    orbits = []
+    orbits.append(Orbit.from_body_ephem(Earth, epoch))
```
Juanlu001: Good use of a `for` loop to call `op.plot` repeatedly. Can yo udo the same for `from_body_ephem`? There's a way to write this function in four or five lines :wink:

src/poliastro/tests/test_plotting.py
```diff
@@ -66,3 +66,9 @@ def test_color():
     assert ax.get_legend().get_lines()[0].get_c() == c
     for element in ax.get_lines():
         assert element.get_c() == c
+
+
+def test_plot_solar_system():
```
Juanlu001: Good test!

src/poliastro/plotting.py
shreyasbapat: Actually Yes! But then I have to import classes.I will do that!

src/poliastro/tests/test_plotting.py
shreyasbapat: Thanks :)

src/poliastro/plotting.py
shreyasbapat: Okay! Working on that!

shreyasbapat: I used isort tool to resolve the problems. I guess that shifted the line below others.

src/poliastro/plotting.py
```diff
@@ -375,3 +378,12 @@ def savefig(self, filename, **layout_kwargs):
             image=ext[1:], image_filename=basename,
             show_link=False,  # Boilerplate
         )
+
+
+def plot_solar_system(epoch=None):
```
Juanlu001: Just one last thing: can you add a docstring here so the new function gets properly documented? :)

shreyasbapat: Okay :)

shreyasbapat:
@Juanlu001  Done 👍

Juanlu001:
Awesome! Merging 👏

Issue #286, opened 2017-11-21
Allow Orbit.sample to receive a different propagator


Juanlu001:
🐞 **Problem**

Right now the default propagator is used in `Orbit.sample`:

https://github.com/poliastro/poliastro/blob/v0.8.0/src/poliastro/twobody/orbit.py#L312
https://github.com/poliastro/poliastro/blob/v0.8.0/src/poliastro/twobody/orbit.py#L238-L254

It would be nice to have the ability to use other propagators.

🎯 **Goal**

This will pave the way for the moment where we have other objects we can `sample`, and also help with testing the new propagators (#253, #285). 

💡 **Possible solutions**

Just add another parameter to `sample` and `propagate`, with sensible defaults.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

PR #311, opened 2018-02-16
Orbit.sample can now recieve a different propagator

shreyasbapat:
Issue #286 Addressed

### DIFF
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -8,7 +8,7 @@ from astropy.coordinates import CartesianRepresentation, get_body_barycentric_po
 
 from poliastro.constants import J2000
 from poliastro.twobody.angles import nu_to_M
-from poliastro.twobody.propagation import propagate
+from poliastro.twobody.propagation import propagate, cowell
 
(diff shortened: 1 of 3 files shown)


shreyasbapat:
I will resolve these failing CI tests tonight!

Juanlu001:
Before you go on, I think this needs a rebase again :)

src/poliastro/twobody/orbit.py
```diff
@@ -8,7 +8,7 @@
 
 from poliastro.constants import J2000
 from poliastro.twobody.angles import nu_to_M
-from poliastro.twobody.propagation import propagate
+from poliastro.twobody.propagation import propagate, cowell
```
Juanlu001: Where are you using the `cowell` function? And on the other hand, do you plan to add it for testing?

Juanlu001:
Never mind, I'm merging this :) Thanks!

src/poliastro/twobody/orbit.py
shreyasbapat: Yeah! I plan to do so. Currently, not using it

Issue #287, opened 2017-11-21
Return times in Orbit.sample


Juanlu001:
🐞 **Problem**

Sampling an orbit was introduced in #263, and it just returns a `CartesianRepresentation` object:

https://github.com/poliastro/poliastro/blob/v0.8.0/src/poliastro/twobody/orbit.py#L315

It would be nice to return the time values as well.

🎯 **Goal**

Just sampling the positions is enough for things like plotting, but any application that needs to analyze the results would benefit from having the times as well.

💡 **Possible solutions**

* Return a tuple `times, positions`
* Return a new type of object (`Trajectory`?)

Notice that both of these break the API. Any solution that returns different things depending on an optional parameter won't be accepted.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

ritiek:
I don't know much about poliastro's code base but another solution could be creating a new method (say `get_times()`) on an `Orbit` object that only returns `times`? Something like this:
```python
    class Orbit(object):
    ...
    def get_times(self, time_values):
        values = np.zeros((len(time_values), 3)) * self.r.unit
        for ii, epoch in enumerate(time_values):
            rr = self.propagate(epoch).r
            values[ii] = rr
        return values

    def _sample(self, time_values):
        values = self.get_times(time_values)
        return CartesianRepresentation(values, xyz_axis=1)
    ...
```

The good thing is that it won't break the API as users could just call method `Orbit.get_times()` to get times. But this won't be as efficient as we would have to compute `times` in each method call.

Juanlu001:
@ritiek notice that you swapped the time values and the position values. Can you give it a second look in case you can apply that idea using the correct code? :)

ritiek:
@Juanlu001 Yep, my bad. Ignore my previous implementation. How about something like this (and leaving everything else untouched):

```python
class Orbit(object):
    ...
    def sample(self, values=None):
        time_values = self.get_times(values)
        return self._sample(time_values)

    def get_times(self, values=None):
        if values is None:
            return self.get_times(100)

        elif isinstance(values, int):
            if self.ecc < 1:
                nu_values = np.linspace(0, 2 * np.pi, values) * u.rad
            else:
                # Select a sensible limiting value for non-closed orbits
[...]

PR #312, opened 2018-02-18
Return times in Orbit.sample()

ritiek:
Addresses #287. This PR returns a tuple `(Time, CartesianRepresentation)` in `Orbit.sample()`.

I've update existing and added a small test [here](https://github.com/ritiek/poliastro/blob/2e6ff3e6d87a35a808648107917679912f5e06fb/src/poliastro/tests/tests_twobody/test_orbit.py#L170-L172) that checks the value count. If there any other tests we need, please let me know!

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -169,7 +169,7 @@ def test_sample_numpoints():
     ss = Orbit.from_classical(_body, _d, _, _a, _a, _a, _a)
     times, positions = ss.sample(values=50)
 
-    assert len(positions) == 50 and len(times) == 50
+    assert len(positions) == len(times) == 50
 
 
 def test_sample_with_time_value():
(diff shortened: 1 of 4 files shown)


src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -167,8 +167,9 @@ def test_sample_numpoints():
     _a = 1.0 * u.deg  # Unused angle
     _body = Sun  # Unused body
     ss = Orbit.from_classical(_body, _d, _, _a, _a, _a, _a)
+    times, positions = ss.sample(values=50)
 
-    assert len(ss.sample(values=50)) == 50
+    assert len(positions) == 50 and len(times) == 50
```
Juanlu001: What about `len(positions) == len(times) == 50`? :)

ritiek: Yep, that is indeed a better way to check this. Gotta fix this real quick.

Juanlu001:
Merging, thanks!

Issue #294, opened 2017-12-20
Default steps 2D plots are too visible


Juanlu001:
🐞 **Problem**

With the default number of points, the steps in the pericenter are too visible:

| Before                                                                                                       | After                                                                                                         |
|--------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------|
| ![index](https://user-images.githubusercontent.com/316517/34196128-ff6eb412-e560-11e7-892b-0d452ccfdda4.png) | ![index2](https://user-images.githubusercontent.com/316517/34196136-0571c99e-e561-11e7-8a8b-6d7b204b1200.png) |

This came after the propagation changes in plotting and the different way to compute the data points.

🎯 **Goal**

Make the plots smooth again.

💡 **Possible solutions**

* Increase the default `num_points`
* Other?

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
Another side effect:

![index4](https://user-images.githubusercontent.com/316517/34196364-15a25076-e562-11e7-8d9a-9abb34aacdb2.png)

This is a slightly different issue, namely that the pericenter is not always included in the propagation.

Juanlu001:
Notice that, while fixing the previous one, one might step into #265.

nikita-astronaut:
Started working on this issue

Juanlu001:
Thanks @nikita-astronaut, keep us posted! Notice that the propagation might fail and distract you, see the last comment in this issue.

nikita-astronaut:
I'll be prepared. Could you please provide code to reproduce this issue?

Juanlu001:
This is a good moment to say that just increasing the number of points is not going to be the preferred solution. In the report ["The generalized Sundman transformation for propagation of high-eccentricity elliptical orbits"](http://www.dtic.mil/dtic/tr/fulltext/u2/a605040.pdf) we can read:

> Because of the large number of satellites, attention must be paid to the total time of computation.
For very eccentric orbits, satisfactory accuracy can be maintained with a larger step size near apogee
than is used at perigee

Take a look at this plot:

[...]

PR #320, opened 2018-02-20
changed default num_points in OrbitPlotter class

nikita-astronaut:
Addresses #294 
I simple changed the default num_points in OrbitPlotter class, now orbits in the example look smooth.
Still, this is not the best solution. Nonsmoothness appears in the vicinity of perihelion, cause the speed there is the highest, but the points in time are chosen uniformly, so the difference in position is the largest. 

What I think could be done next is to place points non uniformly but with density ~1/current_speed, so that between every two points the position change is the same.

What do you think?

P.S. I am actually looking to joining your project later in the GSoC18.


Juanlu001:
Hi @nikita-astronaut, thanks for your interest in poliastro!

As you say, just increasing the number of points is not the best solution. I added more information in the original issue, so please take your time to read it.

nikita-astronaut:
Hi @Juanlu001 
I was surprised (but why?) that the functions of transitions between anomalies are already present in the package, so I used them. Now the angle is sampled uniformly from eccentric anomaly, not from the true anomaly. Seems like now the orbit looks smooth for only `num_points = 150` instead of `num_points = 300` I used previously.

before
after_1
after_2

I attached three images. The first is how it was before (true anomaly + 100 points), second: eccentric anomaly + 100 points (the problem is ameliorated but nonsmoothness however is still seen), third: eccentric anomaly + 150 points (orbit looks smooth).

Now, I think, I'll try to go further and work with s-integrator you've mentioned.

docs/source/examples/Using NEOS package.ipynb
```diff
@@ -29,7 +29,9 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {},
+   "metadata": {
```
Juanlu001: Can you restore the changes in this notebook?

src/poliastro/plotting.py
```diff
@@ -59,15 +59,15 @@ class OrbitPlotter(object):
...
+    def __init__(self, ax=None, num_points=150):
         """Constructor.
 
         Parameters
         ----------
         ax : ~matplotlib.axes.Axes
             Axes in which to plot. If not given, new ones will be created.
         num_points : int, optional
-            Number of points to use in plots, default to 100.
+            Number of points to use in plots, default to 300.
```
Juanlu001: Update docstring :)

src/poliastro/twobody/orbit.py
```diff
@@ -4,10 +4,13 @@
 
 from astropy import units as u
 from astropy import time
+
 from astropy.coordinates import CartesianRepresentation, get_body_barycentric_posvel
 
 from poliastro.constants import J2000
 from poliastro.twobody.angles import nu_to_M
```
Juanlu001: What about importing all these functions in the same line?

src/poliastro/twobody/orbit.py
```diff
@@ -294,12 +297,17 @@ def sample(self, values=None, function=propagate):
 
         elif isinstance(values, int):
             if self.ecc < 1:
-                nu_values = np.linspace(0, 2 * np.pi, values) * u.rad
+                # first sample eccenric anomaly, then transform into true anomaly
```
Juanlu001: s/eccenric/eccentric/g

Also, what about adding a reference to the report in the comment? "Why sample uniformly on the eccentric anomaly to minimize error in the apocenter, see ..."

src/poliastro/twobody/orbit.py
```diff
@@ -294,12 +297,17 @@ def sample(self, values=None, function=propagate):
...
+                E_values = np.linspace(0, 2 * np.pi, values) * u.rad
+                nu_values = E_to_nu(E_values, self.ecc)
+                
             else:
                 # Select a sensible limiting value for non-closed orbits
                 # This corresponds to r = 3p
                 nu_limit = np.arccos(-(1 - 1 / 3.) / self.ecc)
-                nu_values = np.linspace(-nu_limit, nu_limit, values)
+                E_limit = nu_to_E(nu_limit, self.ecc) / u.rad
```
Juanlu001: Not sure if the eccentric anomaly makes sense in the context of hyperbolic orbits. I wouldn't be surprised if this fails.

nikita-astronaut:
I think I fixed everything you were asking)

Juanlu001:
Awesome! Thanks for your contribution :grin:

Issue #295, opened 2018-01-06
Fix position in hyperbolic plots


Juanlu001:
🐞 **Problem**

The position is not plotted correctly for hyperbolic orbits, see https://github.com/poliastro/poliastro/issues/111#issuecomment-355731620.

💡 **Possible solutions**

One option: add an offset of `+nu_limit` to `nu_values`:

https://github.com/poliastro/poliastro/blob/44dbc6e26f2d0a6b395a910c39c9e8131c8946b2/src/poliastro/twobody/orbit.py#L296-L300

Other option: remove the assumption that the first value from `sample` is the current position:

https://github.com/poliastro/poliastro/blob/44dbc6e26f2d0a6b395a910c39c9e8131c8946b2/src/poliastro/plotting.py#L158-L159

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

shreyasbapat:
Working on it!

Issue #111, opened 2016-01-24
Allow hyperbolic orbits to stay within the bounds of a plot


Juanlu001:
After #19 was fixed plotting only one hyperbolic orbit looks okay, but there can be cases when one wants it to do something like this:

![new_horizons](https://cloud.githubusercontent.com/assets/316517/12539950/2ef2ac8e-c2fe-11e5-8d10-3a3067e6f3f1.png)

This plot needed a patch in `_generate_vals` (hence tagging it as a bug), but it only works for cases where there's a previous closed orbit plotted. The logic to handle both cases has to be more complex.

Juanlu001:
I'm postponing this for the next release too.

Juanlu001:
Asking the algorithm to `_adjust_to_bigger_bounded_orbit()` magically is probably too much. I should give more control to the user. Postponing.

Juanlu001:
Blocked on #218.

Juanlu001:
This is no longer blocked.

sudnadja:
When plotting certain hyperbolic paths, the position plots of the objects appear incorrect. 
For example, in July of 2015, New Horizons flew past pluto. Taking position data from horizons for both Pluto and New Horizons with 500@0 origin and "frame" as the fundamental plane, those objects are placed into poliastro with: 

PlutoM = Orbit.from_vectors(Sun, [1.197648971483503E+09, -4.443903841678363E+09, -1.747653049536265E+09] * u.km , [5.382941983259522E+00,  8.185732351064959E-01, -1.389415786284875E+00] * u.km/u.s, Time('2015-07-14 07:59',scale='tdb') )
NewHorizonsM = Orbit.from_vectors(Sun, [1.197659243752796E+09, -4.443716685978071E+09, -1.747610548576734E+09] * u.km , [5.540549267188614E+00, -1.251544669134140E+01, -4.848892572767733E+00] * u.km/u.s, Time('2015-07-14 07:59',scale='tdb') )

New Horizons is nearly on top of Pluto at that point, however the plot places New Horizons much earlier on its path: 


![new horizons pluto encounter](https://user-images.githubusercontent.com/35163155/34637945-041e3858-f276-11e7-86ee-be2757538d0d.png)

And a notebook exhibiting this behavior: 

[New+Horizons+Pluto+Encounter.zip](https://github.com/poliastro/poliastro/files/1608594/New.Horizons.Pluto.Encounter.zip)

Juanlu001:
[...]

PR #321, opened 2018-02-21
Positions in Hyperbolic Plots

shreyasbapat:
I have added an offset of ``+nu_limit`` to ``nu_values`` . Issue #295  is addressed.

### DIFF
--- a/docs/source/examples/Revisiting Lambert's problem in Python.ipynb
+++ b/docs/source/examples/Revisiting Lambert's problem in Python.ipynb
@@ -1172,7 +1172,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.1"
+   "version": "3.6.3"
   }
  },
  "nbformat": 4,
--- a/docs/source/examples/Visualizing the SpaceX Tesla Roadster trip to Mars.ipynb
+++ b/docs/source/examples/Visualizing the SpaceX Tesla Roadster trip to Mars.ipynb
@@ -79,20 +79,9 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
    "metadata": {},
-   "outputs": [
-    {
-     "data": {
-      "text/plain": [
-       "1 x 2 AU x 24.2 deg orbit around Sun (☉)"
-      ]
-     },
-     "execution_count": 3,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
    "source": [
     "roadster = get_roadster_orbit(EPOCH)\n",
     "roadster"
@@ -6561,7 +6550,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.4"
+   "version": "3.6.3"
   }
  },
  "nbformat": 4,
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -6,6 +6,8 @@ from astropy import units as u
 from astropy.tests.helper import assert_quantity_allclose
 
 from astropy import time
+from astropy.time import Time
+from astropy.coordinates import CartesianRepresentation
 
 from poliastro.bodies import Sun, Earth
 from poliastro.twobody import Orbit
@@ -197,3 +199,14 @@ def test_sample_with_nu_value():
     r = positions.get_xyz().transpose()
 
     assert_quantity_allclose(r, expected_r, rtol=1.e-7)
+
+
+def test_nu_value_check():
+    _d = [1.197659243752796E+09, -4.443716685978071E+09, -1.747610548576734E+09] * u.km
+    _v = [5.540549267188614E+00, -1.251544669134140E+01, -4.848892572767733E+00] * u.km / u.s
+    ss = Orbit.from_vectors(Sun, _d, _v, Time('2015-07-14 07:59', scale='tdb'))
+    values, positions = ss.sample(100)
+
+    assert isinstance(positions, CartesianRepresentation)
+    assert isinstance(values, Time)
+    assert len(positions) == len(values) == 101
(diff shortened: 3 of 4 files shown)


Juanlu001:
Could you send a couple of screenshots with the before and the after behavior? Like this: https://github.com/poliastro/poliastro/issues/111#issuecomment-355731620

Juanlu001:
Please rebase!

shreyasbapat:
There is a problem in this approach. When we add an offset of "+nu_limit" to "nu_values" the values in "nu_values" become form 0 to 2*nu_limit. Now this goes to a function nu_to_F which in turn computes a logarithmic value, which does not get computed when we give 0 to 2*nu_limit

Juanlu001:
I thought about this carefully. The error is a bit obscure, but probably due to the fact that there are NaN values appearing in the array:

```
In [3]: Time.now() + 1 * u.h
Out[3]: <Time object: scale='utc' format='datetime' value=2018-02-23 11:06:07.789386>

In [4]: (0 * u.h) / (0 * u.h)
/home/juanlu/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/quantity.py:641: RuntimeWarning: invalid value encountered in true_divide
  *arrays, **kwargs)
Out[4]: <Quantity nan>

In [5]: Time.now() + (0 * u.h) / (0 * u.h)
/home/juanlu/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/quantity.py:641: RuntimeWarning: invalid value encountered in true_divide
  *arrays, **kwargs)
UnitConversionError                       Traceback (most recent call last)
~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/time/core.py in __init__(self, val, val2, format, scale, copy)
   1534                 try:
-> 1535                     val = val.to(u.day)
   1536                     if val2 is not None:

~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/quantity.py in to(self, unit, equivalencies)
    843         unit = Unit(unit)
--> 844         return self._new_view(self._to_value(unit, equivalencies), unit)
    845 

~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/quantity.py in _to_value(self, unit, equivalencies)
    815         return self.unit.to(unit, self.view(np.ndarray),
--> 816                             equivalencies=equivalencies)
    817 

~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/core.py in to(self, other, value, equivalencies)
    978         """
--> 979         return self._get_converter(other, equivalencies=equivalencies)(value)
    980 

~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/core.py in _get_converter(self, other, equivalencies)
    912 
--> 913             raise exc
    914 

~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/core.py in _get_converter(self, other, equivalencies)
    898             return self._apply_equivalencies(
--> 899                 self, other, self._normalize_equivalencies(equivalencies))
    900         except UnitsError as exc:

~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/core.py in _apply_equivalencies(self, unit, other, equivalencies)
    882             "{0} and {1} are not convertible".format(
--> 883                 unit_str, other_str))
    884 

UnitConversionError: '' (dimensionless) and 'd' (time) are not convertible

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/time/core.py in __add__(self, other)
   1373             try:
-> 1374                 other = TimeDelta(other)
   1375             except Exception:

~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/time/core.py in __init__(self, val, val2, format, scale, copy)
   1540                                      'be used to initiate {0} instances .'
-> 1541                                      .format(self.__class__.__name__))
   1542                 format = 'jd'

ValueError: Only Quantities with Time units can be used to initiate TimeDelta instances .

During handling of the above exception, another exception occurred:

OperandTypeError                          Traceback (most recent call last)
<ipython-input-5-19cbc565acb7> in <module>()
----> 1 Time.now() + (0 * u.h) / (0 * u.h)

~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/time/core.py in __add__(self, other)
   1374                 other = TimeDelta(other)
   1375             except Exception:
-> 1376                 raise OperandTypeError(self, other, '+')
   1377 
   1378         # Tdelta + something is dealt with in TimeDelta, so we have

OperandTypeError: Unsupported operand type(s) for +: 'Time' and 'Quantity'
```

The reason why we have NaN values is because we are going above the maximum true anomaly for which the hyperbolic orbit makes sense. Therefore, we have to discard the offset solution (I'm updating the original issue with this information).

The current code "selects a sensible limiting value":

https://github.com/poliastro/poliastro/blob/44dbc6e26f2d0a6b395a910c39c9e8131c8946b2/src/poliastro/twobody/orbit.py#L296-L300

But this value might be lower than the actual anomaly of the point we're plotting!

https://github.com/poliastro/poliastro/issues/111#issuecomment-355746579

Therefore, the sensible value has to be the maximum between `r = 3p` and `r = current_r`. @shreyasbapat could you please try this approach instead?

Juanlu001:
And also, we have to remove the assumtion that the first point is the current point in the plotting code, as I said in #295.

shreyasbapat:
Then what would be the first point? Won't we specify some point?

shreyasbapat:
I will try this approach tonight itself :)

Juanlu001:
On second thought, perhaps we can do something like this:

```
nu_limit = # ...
nu_values = np.linspace(-nu_limit, nu_limit, values)
nu_values = np.insert(nu_values, 0, current_nu)
```

what do you think?

shreyasbapat:
What is this current_nu?

Juanlu001:
The `nu` of the orbit that we're plotting.

shreyasbapat:
Okay! Let's give it a try!

shreyasbapat:
screenshot-2018-2-23 new horizons pluto encounter
Something like this!

Juanlu001:
Nice! Now the plot appears where it is supposed to appear. Still it's not the output that I would like, but I will take care of that in a future pull request :)

This needs at least one test to pass.

shreyasbapat:
I will take care of the test cases!

shreyasbapat:
Yeah! It's not very exact :/

Juanlu001:
According to codecov, this code is not being covered:

https://codecov.io/gh/poliastro/poliastro/compare/36379a98f19dda2e037de6dc2547084450fec93a...0146c12ac973c86116894afad6c550a85f6f7b18

And it makes sense, because for tests to run, the function name has to start with `test_` :)

However, I went to the Travis build and saw that the test names are not there. I will fix that myself on another PR.

PR #322, opened 2018-02-22
added draft for mean_motion propagator

nikita-astronaut:
Hello @Juanlu001 

There is a lot to correct, I believe, but there's a `mean_motion` function. Now `iss` and `halleys[0]` propagate without falling and results of propagation agree with the `kepler` propagator where is doesn't fall.

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_propagation.py
+++ b/src/poliastro/tests/tests_twobody/test_propagation.py
@@ -1,4 +1,5 @@
 from pytest import approx
+import pytest
 
 import numpy as np
 from numpy.testing import assert_allclose
@@ -10,12 +11,13 @@ from astropy.tests.helper import assert_quantity_allclose
 from poliastro.constants import J2000
 from poliastro.bodies import Sun, Earth
 from poliastro.twobody import Orbit
-from poliastro.twobody.propagation import cowell
+from poliastro.twobody.propagation import cowell, kepler, mean_motion
 
 from poliastro.util import norm
 
 
-def test_propagation():
+@pytest.mark.parametrize('method', [kepler, mean_motion])
+def test_propagation(method):
     # Data from Vallado, example 2.4
     r0 = [1131.340, -2282.343, 6672.423] * u.km
     v0 = [-5.64305, 4.30333, 2.42879] * u.km / u.s
@@ -24,7 +26,7 @@ def test_propagation():
 
     ss0 = Orbit.from_vectors(Earth, r0, v0)
     tof = 40 * u.min
-    ss1 = ss0.propagate(tof)
+    ss1 = ss0.propagate(tof, method=method)
 
     r, v = ss1.rv()
 
--- a/src/poliastro/tests/tests_twobody/test_propagation.py
+++ b/src/poliastro/tests/tests_twobody/test_propagation.py
@@ -8,15 +8,16 @@ from astropy import units as u
 from astropy import time
 from astropy.tests.helper import assert_quantity_allclose
 
+from poliastro.twobody.rv import rv2coe
 from poliastro.constants import J2000
 from poliastro.bodies import Sun, Earth
 from poliastro.twobody import Orbit
-from poliastro.twobody.propagation import cowell, kepler, mean_motion
+from poliastro.twobody.propagation import cowell, kepler, mean_motion, cowell
 
 from poliastro.util import norm
 
 
-@pytest.mark.parametrize('method', [kepler, mean_motion])
+@pytest.mark.parametrize('method', [kepler, mean_motion, cowell])
 def test_propagation(method):
     # Data from Vallado, example 2.4
     r0 = [1131.340, -2282.343, 6672.423] * u.km
@@ -65,6 +66,23 @@ def test_propagation_hyperbolic():
 
     assert_quantity_allclose(norm(r), expected_r_norm, rtol=1e-4)
     assert_quantity_allclose(norm(v), expected_v_norm, rtol=1e-3)
+
+
+def test_propagation_mean_motion_parabolic():
+    # example from Howard Curtis, section 3.5, problem 3.15
+    p = 13200.0 * u.km
+    _a = 0.0 * u.deg
+    orbit = Orbit.parabolic(Earth, p, _a, _a, _a, _a)
+    orbit = orbit.propagate(0.8897 * 3600.0 / 2.0 * u.s, method=mean_motion)
+
+    _, _, _, _, _, nu0 = rv2coe(Earth.k.to(u.km**3 / u.s**2).value,
+                                orbit.r.to(u.km).value,
+                                orbit.v.to(u.km / u.s).value)
+    assert_quantity_allclose(nu0, np.deg2rad(90.0), rtol=1e-4)
+
+    orbit = Orbit.parabolic(Earth, p, _a, _a, _a, _a)
+    orbit = orbit.propagate(36.0 * 3600.0 * u.s, method=mean_motion)
+    assert_quantity_allclose(np.sqrt(np.sum(np.array(orbit.r.to(u.km).value) ** 2)), 304700.0, rtol=1e-4)
 
 
 def test_propagation_zero_time_returns_same_state():
--- a/src/poliastro/tests/tests_twobody/test_propagation.py
+++ b/src/poliastro/tests/tests_twobody/test_propagation.py
@@ -69,11 +69,11 @@ def test_propagation_hyperbolic():
 
 
 def test_propagation_mean_motion_parabolic():
-    # example from Howard Curtis, section 3.5, problem 3.15
-    p = 13200.0 * u.km
+    # example from Howard Curtis (3rd edition), section 3.5, problem 3.15
+    p = 2.0 * 6600 * u.km
     _a = 0.0 * u.deg
     orbit = Orbit.parabolic(Earth, p, _a, _a, _a, _a)
-    orbit = orbit.propagate(0.8897 * 3600.0 / 2.0 * u.s, method=mean_motion)
+    orbit = orbit.propagate(0.8897 / 2.0 * u.h, method=mean_motion)
 
     _, _, _, _, _, nu0 = rv2coe(Earth.k.to(u.km**3 / u.s**2).value,
                                 orbit.r.to(u.km).value,
@@ -81,8 +81,8 @@ def test_propagation_mean_motion_parabolic():
     assert_quantity_allclose(nu0, np.deg2rad(90.0), rtol=1e-4)
 
     orbit = Orbit.parabolic(Earth, p, _a, _a, _a, _a)
-    orbit = orbit.propagate(36.0 * 3600.0 * u.s, method=mean_motion)
-    assert_quantity_allclose(np.sqrt(np.sum(np.array(orbit.r.to(u.km).value) ** 2)), 304700.0, rtol=1e-4)
+    orbit = orbit.propagate(36.0 * u.h, method=mean_motion)
+    assert_quantity_allclose(norm(orbit.r.to(u.km).value), 304700.0, rtol=1e-4)
 
 
 def test_propagation_zero_time_returns_same_state():
--- a/src/poliastro/tests/tests_twobody/test_sample.py
+++ b/src/poliastro/tests/tests_twobody/test_sample.py
@@ -7,7 +7,7 @@ from astropy.time import Time
 from poliastro.bodies import Earth
 from poliastro.twobody import Orbit
 from poliastro.twobody.propagation import propagate
-
+import numpy as np
 
 def test_sample_angle_zero_returns_same():
     # Data from Vallado, example 2.4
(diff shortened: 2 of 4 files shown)


nikita-astronaut:
I must admit I haven't tested the case `ecc == 1`. How can I do this? How to import such an orbit?

nikita-astronaut:
I must admit I haven't tested the case `ecc == 1`. How can I do this? How to import such an orbit?

Juanlu001:
> Now iss and halleys[0] propagate without falling and results of propagation agree with the kepler propagator where is doesn't fail.

Wow, impressive! I will try to take a closer look this weekend.

For the `ecc == 1` case, you have a `parabolic` classmethod:

docs.poliastro.space/en/latest/api.html#poliastro.twobody.orbit.Orbit.parabolic

nikita-astronaut:
The code falls, not all the tests can be passed. I faced the problem that `nu_to_M` function returns `M` in radians or dimensionless. I don't understand, why so. Will try to find out.

Juanlu001:
> The code falls, not all the tests can be passed.

The code fails because of PEP8 errors, we still don't know whether the tests are fine. Perhaps we should split that part as well.

>  I faced the problem that nu_to_M function returns M in radians or dimensionless.

You will probably need to do something like this:

https://github.com/poliastro/poliastro/blob/0492f60646c11b5ae556a3df12906da691cbeafd/src/poliastro/twobody/angles.py#L182

Juanlu001:
This needs a rebase on current master.

nikita-astronaut:
Hey, @Juanlu001 

Now the code fails because it doesn't pass the `test_sample_angle_zero_returns_same()` test. The reason is the following: now the `propagate` functions transfers `nu` to `M` and back. Because of these conversions, the error occurs. For instance, the `x` coordinate is in one case `1131.3399999999997` (before propagation), in the other `1131.34` (after propagation). And the assertion fails. How do you think it would be best to fix it?

Juanlu001:
I see! I think the best thing is to use [`assert_quantity_allclose`](http://docs.astropy.org/en/stable/api/astropy.tests.helper.assert_quantity_allclose.html):

```
In [1]: from astropy.tests.helper import assert_quantity_allclose

In [2]: from astropy import units as u

In [3]: assert_quantity_allclose(1 * u.km, 1000 * u.m)

In [4]: assert_quantity_allclose([1] * u.km, [1000] * u.m)

In [5]: assert_quantity_allclose([1, 2] * u.km, [1000, 2000] * u.m)
```

nikita-astronaut:
Ok! So you suggest me to also change tests files?

Juanlu001:
Of course I will examine those changes closely, but in this case I think it's necessary.

Juanlu001:
Also, I will require you to add tests that fail with the current propagator and that succeed with the propagator that you're adding. You can use the Halley and ISS orbits as you already saw.

nikita-astronaut:
@Juanlu001 , hey! I faced the problem I cant really explain.
In function `test_sample_angle_zero_returns_same()` I make (for debug) 
`print(np.allclose([(ss0.r[0] / u.km).value], [(rr[0].get_xyz()[0] / u.km).value]))
    assert np.testing.assert_allclose([(ss0.r[0] / u.km).value], [(rr[0].get_xyz()[0] / u.km).value])`

Print returns `True`, but assert fails. Can it be some trick of dimension wrappers and checkers that are everywhere?

Juanlu001:
You can try using the Astropy function I suggested instead

On Feb 23, 2018 20:02, "Nikita Astrakhantsev" <notifications@github.com>
wrote:

@Juanlu001 <https://github.com/juanlu001> , hey! I faced the problem I cant
really explain.
In function test_sample_angle_zero_returns_same() I make (for debug)
print(np.allclose([(ss0.r[0] / u.km).value], [(rr[0].get_xyz()[0] /
u.km).value]))
assert np.testing.assert_allclose([(ss0.r[0] / u.km).value],
[(rr[0].get_xyz()[0] / u.km).value])

Print returns True, but assert fails. Can it be some trick of dimension
wrappers and checkers that are everywhere?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/poliastro/poliastro/pull/322#issuecomment-368107130>,
or mute the thread

Juanlu001:
Fixed :) When you add some tests and all checks pass, I will do a complete review of the code so you can make some modifications as needed, and then this is ready to merge.

Juanlu001:
Specifically, we need some tests for the parabolic case, as you can see:

https://codecov.io/gh/poliastro/poliastro/pull/322/diff?src=pr&el=tree#diff-c3JjL3BvbGlhc3Ryby90d29ib2R5L3Byb3BhZ2F0aW9uLnB5

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -24,7 +26,7 @@ def test_propagation():
 
     ss0 = Orbit.from_vectors(Earth, r0, v0)
     tof = 40 * u.min
-    ss1 = ss0.propagate(tof)
+    ss1 = ss0.propagate(tof, method=method)
```
Juanlu001: Awesome!

src/poliastro/tests/tests_twobody/test_sample.py
```diff
@@ -6,7 +6,8 @@
 
 from poliastro.bodies import Earth
 from poliastro.twobody import Orbit
-from poliastro.twobody.propagation import propagate
+from poliastro.twobody.propagation import propagate, kepler, mean_motion
```
Juanlu001: Is `propagate` still needed here?

src/poliastro/tests/tests_twobody/test_sample.py
```diff
@@ -6,7 +6,8 @@
...
+import numpy as np
```
Juanlu001: Same for this

src/poliastro/tests/tests_twobody/test_sample.py
```diff
@@ -16,17 +17,18 @@ def test_sample_angle_zero_returns_same():
...
     _, rr = ss0.sample(nu_values)
 
-    assert (rr[0].get_xyz() == ss0.r).all()
+    assert_quantity_allclose(rr[0].get_xyz(), ss0.r)
 
 
-@pytest.mark.parametrize("time_of_flight,function", [
-    (1 * u.min, propagate),
-    (40 * u.min, propagate),
+@pytest.mark.parametrize("time_of_flight,method", [
```
Juanlu001: Prefer stacking `@pytest.mark.parametrize` here to get all the combinations, as explained here: https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions (bottom of the section)

src/poliastro/tests/tests_twobody/test_sample.py
```diff
@@ -36,18 +38,20 @@ def test_sample_one_point_equals_propagation_small_deltas(time_of_flight, functi
...
+    _, rr = ss0.sample(sample_times, method)
 
-    assert (rr[0].get_xyz() == expected_ss.r).all()
+    assert_quantity_allclose(rr[0].get_xyz(), expected_ss.r)
 
 
-@pytest.mark.parametrize("time_of_flight,function", [
-    (6 * u.h, propagate),
-    (2 * u.day, propagate),
+@pytest.mark.parametrize("time_of_flight,method", [
```
Juanlu001: Same as above

src/poliastro/twobody/propagation.py
```diff
@@ -103,6 +106,60 @@ def cowell(k, r0, v0, tof, rtol=1e-10, *, ad=None, callback=None, nsteps=1000):
...
+    if ecc != 1.0:
+        a = p / (1.0 - ecc ** 2)
+        # given the initial mean anomaly, calculate mean anomaly
+        # at the end, mean motion (n) equals sqrt(mu / |a^3|)
+        with u.set_enabled_equivalencies(u.dimensionless_angles()):
+            M = M0 + tof * np.sqrt(k / np.abs(a ** 3)) * u.rad
+            nu = M_to_nu(M, ecc)
+
+    # parabolic orbit
+    else:
```
Juanlu001: I think this is a good addition, but needs testing.

src/poliastro/twobody/propagation.py
```diff
@@ -103,6 +106,60 @@ def cowell(k, r0, v0, tof, rtol=1e-10, *, ad=None, callback=None, nsteps=1000):
...
+    increases mean anomaly and performs inverse transformation to get final \vec{r}, \vec{v}
+    The logic is based on formulae (4), (6) and (7) from http://dx.doi.org/10.1007/s10569-013-9476-9
+    """
+
+    # get the initial true anomaly and orbit parameters that are constant over time
+    p, ecc, inc, raan, argp, nu0 = rv2coe(k, r0, v0)
+    # get the initial mean anomaly
+    M0 = nu_to_M(nu0, ecc)
+    # elliptic or hyperbolic orbits
+    if ecc != 1.0:
```
Juanlu001: Just for curiosity: what is the asymptotic behavior of this when the eccentricity is very close to 1.0? We should add a `# TODO: ` here to remind us that this needs better validation. Perhaps in the original paper there's some information.

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -10,12 +11,13 @@
 from poliastro.constants import J2000
 from poliastro.bodies import Sun, Earth
 from poliastro.twobody import Orbit
-from poliastro.twobody.propagation import cowell
+from poliastro.twobody.propagation import cowell, kepler, mean_motion
```
Juanlu001: Also, can you add a test that demonstrates that the new `mean_motion` propagator works with the Halley orbit, and another one marked as `@pytest.mark.xfail` with the `kepler` propagator?

src/poliastro/twobody/propagation.py
```diff
@@ -103,6 +106,61 @@ def cowell(k, r0, v0, tof, rtol=1e-10, *, ad=None, callback=None, nsteps=1000):
...
+    The logic is based on formulae (4), (6) and (7) from http://dx.doi.org/10.1007/s10569-013-9476-9
+    """
+
+    # get the initial true anomaly and orbit parameters that are constant over time
+    p, ecc, inc, raan, argp, nu0 = rv2coe(k, r0, v0)
+
+    # get the initial mean anomaly
+    M0 = nu_to_M(nu0, ecc)
+    # elliptic or hyperbolic orbits
+    if not np.isclose(ecc, 1.0, rtol=1e-06):
```
Juanlu001: Can you explain why the parabolic test failed before this line was changed?

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -65,6 +68,23 @@ def test_propagation_hyperbolic():
     assert_quantity_allclose(norm(v), expected_v_norm, rtol=1e-3)
 
 
+def test_propagation_mean_motion_parabolic():
+    # example from Howard Curtis, section 3.5, problem 3.15
```
Juanlu001: Please specify 3rd edition

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -65,6 +68,23 @@ def test_propagation_hyperbolic():
...
+    p = 13200.0 * u.km
```
Juanlu001: Can you explicitly write `2 * 6600` here?

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -65,6 +68,23 @@ def test_propagation_hyperbolic():
...
+    _a = 0.0 * u.deg
+    orbit = Orbit.parabolic(Earth, p, _a, _a, _a, _a)
+    orbit = orbit.propagate(0.8897 * 3600.0 / 2.0 * u.s, method=mean_motion)
```
Juanlu001: What about `0.8897 * u.h / 2`? :)

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -65,6 +68,23 @@ def test_propagation_hyperbolic():
...
+
+    _, _, _, _, _, nu0 = rv2coe(Earth.k.to(u.km**3 / u.s**2).value,
+                                orbit.r.to(u.km).value,
+                                orbit.v.to(u.km / u.s).value)
+    assert_quantity_allclose(nu0, np.deg2rad(90.0), rtol=1e-4)
+
+    orbit = Orbit.parabolic(Earth, p, _a, _a, _a, _a)
+    orbit = orbit.propagate(36.0 * 3600.0 * u.s, method=mean_motion)
```
Juanlu001: Same as above, use `36 * u.h`

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -65,6 +68,23 @@ def test_propagation_hyperbolic():
...
+    assert_quantity_allclose(np.sqrt(np.sum(np.array(orbit.r.to(u.km).value) ** 2)), 304700.0, rtol=1e-4)
```
Juanlu001: You can use `from poliastro.util import norm` here

Juanlu001:
This is good to go. Thanks a lot for this pull request!

Issue #274, opened 2017-10-25
Allow plotting segments of orbits in 2D


Juanlu001:
🐞 **Problem**

Right now, the `plot` function and the `OrbitPlotter.plot` method decide which points to plot, making it impossible to customize.

🎯 **Goal**

Plotting a complete orbit is easy but plotting a fraction of an orbit is extremely hard. This can lead to too much clutter when complex maneuvers are being plotted.

💡 **Possible solutions**

This refactoring #269 paves the way for this feature: we can pass custom parameters to the `sample` method calls, so instead of a fixed number of points we can specify a range of times or true anomaly values.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
The new 3D plotting functions already allow this thanks to the `plot_trajectory` method. It would be nice to have it in 2D plots as well.

Juanlu001:
However, this will need a little refactoring in `OrbitPlotter` that won't arrive in time for 0.8.0.

Juanlu001:
This functionality can be a workaround for hyperbolic cases while we figure out what to do: https://github.com/poliastro/poliastro/issues/111

shreyasbapat:
Working on it !

shreyasbapat:
Addressed in PR #332

Issue #204, opened 2017-08-11
Problem when plotting two attractors


anhiga:
When there are two (or more) different attractors in a plot, all of them are located in `(0, 0)`, what leads to:
![iss_eros](https://user-images.githubusercontent.com/18404378/29219763-488c5514-7eb9-11e7-8d44-29c337b3c119.png)

Juanlu001:
Can you upload a minimal version of the code as well?

anhiga:
I'm sorry, here it is 😄 :
```Python
from poliastro.neos import neows
from poliastro.plotting import OrbitPlotter

# Attractor: Earth
from poliastro.examples import iss

#Attractor: Sun
eros = neows.orbit_from_name('eros')

frame = OrbitPlotter()
frame.plot(iss, label='iss')
frame.plot(eros, label='eros)
```

anhiga:
Well, actually the first plot is previous to #199 fix. The problem is still the same, altough:
![figure_1](https://user-images.githubusercontent.com/18404378/29220551-ac9667dc-7ebb-11e7-975b-8001561357a0.png)

Juanlu001:
The first `plot` sets the frame _and_ the attractor. We are projecting the following orbits to the correct frame, but not centering them in the correct attractor. With better reference frame handling (#218) and better hyperbolic handling (#111, itself blocked on the former) this should be possible.

Juanlu001:
Won't arrive in time for November, postponing.

Juanlu001:
In https://github.com/poliastro/poliastro/pull/332 we added a `NotImplementedError`, so at least now the result is not garbage. We still have to address #111 before working on this one.

shreyasbapat:
Is there any reason this should be blocked now?

Juanlu001:
Thanks @shreyasbapat, this is not blocked anymore.

astrojuanlu:
Modern way to reproduce this:

```python
from poliastro.twobody import Orbit
from poliastro.plotting.static import StaticOrbitPlotter

# Attractor: Earth
from poliastro.examples import iss

# Attractor: Sun
eros = Orbit.from_sbdb("eros")

frame = StaticOrbitPlotter()
frame.plot(iss, label="iss")
frame.plot(eros, label="eros")
```

PR #332, opened 2018-03-03
Plotting segments of Orbits in 2D as well

shreyasbapat:
Refactored the ```OrbitPlotter``` . Now it can plot segments of Orbits in 2D as well due to the ```plot_trajectory``` method. Addresses Issue #274


Juanlu001:
Awesome! Can you add some screenshot of how the plot_trajectory method looks like? For instance, plot some Hohmann transfer with only the traversed parts of the orbit.

shreyasbapat:
Sure! I will do that in night

shreyasbapat:
@Juanlu001  , Do you have some data about the hohmann transfer?

shreyasbapat:
Do you have any idea on how to plot this?
screenshot from 2018-03-03 23-43-58

Juanlu001:
I'm being able to plot in your branch without problems:

screenshot-2018-3-4 jupyterlab

Using this notebook:

https://gist.github.com/Juanlu001/e574886ccf8bff10c8c08459981a674f

Please review the reasons why it does not work for you. Leaving some extra comments.

src/poliastro/plotting.py
```diff
@@ -112,6 +112,14 @@ def _redraw(self):
         self.ax.relim()
         self.ax.autoscale()
 
+    def plot_trajectory(self, trajectory, color, label=None):
```
Juanlu001: This is inconsistent with `OrbitPlotter3D.plot_trajectory`:

https://github.com/poliastro/poliastro/pull/332/files#diff-be441b39adc37d0e0a9887217a253560R346

The signatures should be the same. That's the point of unifying the codebases, right? :wink:

src/poliastro/plotting.py
```diff
@@ -137,6 +145,14 @@ def _project(self, rr):
         y = rr_proj.dot(self._frame[1])
         return x, y
 
+    def verify_attractor(self, orbit):
```
Juanlu001: Does it make sense to have a public method here? In `OrbitPlotter3D` we're doing `self.set_attractor` and then `self._redraw_attractor`. Can we do a similar thing here?

https://github.com/poliastro/poliastro/pull/332/files#diff-be441b39adc37d0e0a9887217a253560R330

src/poliastro/plotting.py
shreyasbapat: True!

src/poliastro/plotting.py
shreyasbapat: Yeah Sure! Now addressing this!

shreyasbapat: Okay! So, basically, we would never merge these two classes. We would just remove the ambiguity for the user about the names of the methods and what they do... Am I right?

shreyasbapat:
Rebased! :)

src/poliastro/plotting.py
```diff
@@ -112,31 +113,53 @@ def _redraw(self):
...
+        ----------
+        attractor : ~poliastro.bodies.Body
+            Central body.
+
+        """
+        if self._attractor is None:
+            self._attractor = attractor
+
+        elif attractor is not self._attractor:
+            raise NotImplementedError("Attractor has already been set to {}.".format(self._attractor.name))
```
Juanlu001: This also alleviated https://github.com/poliastro/poliastro/issues/204 :)

Juanlu001:
This is looking better :) Now we can improve notebooks like this one without cluttering the final plot:

http://docs.poliastro.space/en/latest/examples/Going%20to%20Jupiter%20with%20Python%20using%20Jupyter%20and%20poliastro.html

Merging!

shreyasbapat:
True:)

PR #334, opened 2018-03-03
added propagation to true_anomaly + a test

nikita-astronaut:


nikita-astronaut:
@Juanlu001 can you please help me understand why the test failed? What it is actually about?

shreyasbapat:
PEP8 issues. Use autopep8 tool to fix these :)

nikita-astronaut:
Strange: coverage test says I don't cover new function in `orbit`, but this test is present...

shreyasbapat:
The name of the test function need to be started with "test_"

Juanlu001:
Thanks @shreyasbapat! The PEP8 failures were in the Travis CI build:

https://travis-ci.org/poliastro/poliastro/jobs/348605680#L842

However you went a bit too far, since we relax the line length limit:

https://github.com/poliastro/poliastro/blob/53ebfc2268108836d415fe0fcedd290e122d37f9/setup.cfg#L10

I'm adding more comments.

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -203,8 +218,10 @@ def test_sample_with_nu_value():
 
 def test_hyperbolic_nu_value_check():
     # A custom hyperbolic orbit
-    r = [1.197659243752796E+09, -4.443716685978071E+09, -1.747610548576734E+09] * u.km
-    v = [5.540549267188614E+00, -1.251544669134140E+01, -4.848892572767733E+00] * u.km / u.s
+    r = [1.197659243752796E+09, -4.443716685978071E+09, -
```
Juanlu001: These lines do not need splitting

src/poliastro/twobody/orbit.py
```diff
@@ -104,7 +104,8 @@ def from_classical(cls, attractor, a, ecc, inc, raan, argp, nu, epoch=J2000):
 
         """
         if ecc == 1.0 * u.one:
-            raise ValueError("For parabolic orbits use Orbit.parabolic instead")
+            raise ValueError(
```
Juanlu001: This line does not need splitting

src/poliastro/twobody/orbit.py
```diff
@@ -375,7 +385,8 @@ def delegated_(self_):
             delegated = property(delegated_)
 
         else:
-            raise AttributeError("'{}' object has no attribute '{}'".format(self.__class__, item))
+            raise AttributeError(
```
Juanlu001: This line does not need splitting

src/poliastro/twobody/orbit.py
```diff
@@ -237,6 +238,15 @@ def __str__(self):
     def __repr__(self):
         return self.__str__()
 
+    def set_true_anomaly(self, target_nu):
```
Juanlu001: I don't like this approach very much. On the one hand, the `set_*` seems to suggest that we are changing the value of an element _in place_, which we don't do. On the other hand, seems a bit heavy to do all the elements conversion to change only one value.

shreyasbapat:
@Juanlu001  Welcome! But the ```autopep8``` does not change the line lengths in any case. I always use it :)

Issue #315, opened 2018-02-18
OrbitPlotter does not implement show()


Juanlu001:
🐞 **Problem**

While `OrbitPlotter3D` does implement a `show()` method that is useful when redrawing, `OrbitPlotter` does not. Therefore, there is no easy way to do a `set_frame` and then show the resulting plot, or in general reuse a `frame`.

🎯 **Goal**

To be consistent with `OrbitPlotter3D` and improve reusability of a plotting frame. 

💡 **Possible solutions**

* Easy way: just add a `show` method. Is it doable in matplotlib?
* Difficult, but rewarding, way: refactor the `plotting` module and try to unify interfaces.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
I think the way matplotlib works in the notebook is essentially incompatible with how plotly does. @shreyasbapat tried to tackle this issue in #336 but there were still some usability problems, and also now we have a `OrbitPlotter2D` class. I'm closing this issue.

PR #336, opened 2018-03-08
OrbitPlotter now implements show()

shreyasbapat:
Issue #315  Addressed

### DIFF
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -72,7 +72,7 @@ class OrbitPlotter(object):
         """
         self.ax = ax
         if not self.ax:
-            _, self.ax = plt.subplots(figsize=(6, 6))
+            self.fig, self.ax = plt.subplots(figsize=(6, 6))
         self.num_points = num_points
         self._frame = None
         self._attractor = None
@@ -128,6 +128,9 @@ class OrbitPlotter(object):
         a, = self.ax.plot(x.to(u.km).value, y.to(u.km).value, '--', color=color, label=label)
         lines.append(a)
         return lines
+
+    def show(self):
+        return self.fig
 
     def set_attractor(self, attractor):
         """Sets plotting attractor.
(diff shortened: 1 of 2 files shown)


shreyasbapat:
@Juanlu001  Please Review :)

Juanlu001:
The code looks good, but I have a problem with this: if you use it, the plot shows twice.

screenshot-2018-3-9 jupyterlab

The issue is probably tricky to fix, since it involves the display mechanism of Jupyter, which is not consistent. Please make sure to try out the changes that you push so you find this kind of problems earlier.

Juanlu001:
The more I find these issues, the more I think we should use Plotly also for 2D plots. However, I don't feel OK dropping the matplotlib interface, because probably it will produce better static plots. This means that we should refactor the whole plotting module in a way that lets you select "backends", and start with two: an interactive backend using Plotly for the notebook (2D and 3D) and a "static" backend that works using matplotlib and only 2D. I added a new issue for this (https://github.com/poliastro/poliastro/issues/338), so please @shreyasbapat think about it, edit your proposal accordingly and let's defer this PR for a moment when we have a clear plan.

shreyasbapat:
Oh! Well I thought we just have to show what we are getting. Okay! I got your point. Basically, now I got the point what you suggested. I will work on it. Thanks a lot. I think this PR can be changed after we refactor our modules as it will be useful to reuse the frame. So I am not closing it Is it fine @Juanlu001 ?

Juanlu001:
Agreed!

shreyasbapat:
Now, should we try to attempt this?

Juanlu001:
I think the aspect ratio issues shown here are much more important https://github.com/poliastro/poliastro/issues/345

Juanlu001:
I closed #315, so I think the best thing is to close this and open a new PR for #338. Is that okay?

shreyasbapat:
True!

shreyasbapat:
Closing this!

PR #339, opened 2018-03-13
Hyperbolic Plot Fix

shreyasbapat:
Now the hyperbolic plot looks much cleaner. 
```
Orbit.from_vectors(
    Sun,
    [ -9.77441841e+07,  1.01000539e+08,  4.37584668e+07] * u.km,
    [ 23.75936985,-43.09599568, -8.7084724 ] * u.km / u.s,
)
```
screenshot from 2018-03-13 16-45-32

Earlier, we used to get a plot like this: 
plot

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -1,5 +1,6 @@
 import pytest
 
+import numpy as np
 from numpy.testing import assert_allclose
 
 from astropy import units as u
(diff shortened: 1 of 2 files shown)


src/poliastro/twobody/orbit.py
```diff
@@ -319,7 +319,7 @@ def sample(self, values=None, method=mean_motion):
                 # the arc cosine, which is in the range [0, 180)
                 # Start from -nu_limit
                 wrapped_nu = self.nu if self.nu < 180 * u.deg else self.nu - 360 * u.deg
-                nu_limit = max(np.arccos(-(1 - 1 / 3.) / self.ecc), wrapped_nu)
+                nu_limit = max(np.arccos(-(1 - 1 / 3.) / self.ecc), abs(wrapped_nu))
```
Juanlu001: Good catch! :tada:

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -211,3 +211,21 @@ def test_hyperbolic_nu_value_check():
     assert isinstance(positions, CartesianRepresentation)
     assert isinstance(values, Time)
     assert len(positions) == len(values) == 100
+
+
+def test_hyperbolic_modulus_wrapped_nu():
```
Juanlu001: This test is a bit complicated though. What about testing that the first `position` of `a.sample(100)` is exactly the position of `a`, for this particular case?

src/poliastro/twobody/orbit.py
shreyasbapat: Thanks!

src/poliastro/tests/tests_twobody/test_orbit.py
shreyasbapat: Yeah! I will do that ! Thanks :)

Juanlu001: With your permission, I simplified it even further :)

shreyasbapat: I didn't knew much about AstroPy. I will work on it too! Thanks :)

Issue #299, opened 2018-02-07
Animated plots


Juanlu001:
🎯 **Goal**

Create an easy API for animations, where the evolution in time of one or more orbits can be shown.

💡 **Possible solutions**

Let's first settle on producing animations with a fixed camera view.

This can be challenging since the code for 2D and 3D visualization is quite different. Also, matplotlib is not very well suited for animations in general, and perhaps we would have to switch to Plotly for everything.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

aeronau:
I have written this proof of concept for animated plots of 2d orbits to understand the inner workings of poliastro. It is very basic and I am trying to improve it. 

I make use of matplotlib's [`FuncAnimation'](https://matplotlib.org/api/_as_gen/matplotlib.animation.FuncAnimation.html) to render frame-by-frame the points that make up the orbit. Note that the save option may depend on ffmpeg. Simple 2d animations come out 

The next step would be to add an `animate` property for the `Orbit` object or an argument in `plot` so as to trigger the steps to create and output the animation.

Animation of 3d trajectories may be much more complicated as it would be optimal that the viewport followed the body without losing sight of the other orbits---which act as references for the viewer. I usually dislike the viewer annoyingly using the cursor constantly to follow the object in 3d animations by default.

I think matplotlib should be used for 3d animations considering the state of the project. I would not switch to Plotly [yet, if anything].

Juanlu001:
Awesome! Let me attach a high-speed GIF for more impressive results:

![ezgif com-optimize](https://user-images.githubusercontent.com/316517/37906591-f035400e-3102-11e8-980a-c4a7a7046985.gif)

The question, as you say, is the API that we want for these animations to work.

Also, plotly animations are not easy to export standalone:

https://github.com/plotly/plotly.py/issues/664

@aeronau do you want to work on this during GSOC/SOCIS, perhaps your free time?

aeronau:
If I understand this correctly, are you suggesting that I work on an idea that is not in the [list](http://openastronomy.org/gsoc/gsoc2018/#/projects)? Or that I tackle my GSOC proposal and this subject at the same time?

[...]

Issue #338, opened 2018-03-09
Refactor plotting to use backends and a consistent notebook experience


Juanlu001:
🐞 **Problem**

The experience with 2D and 3D plots in the notebook is inconsistent, due to the inconsistencies in the rendering between matplotlib and plotly. The transition from Jupyter notebook to JupyterLab is making things more difficult, as the documentation is scattered and now the users need to install JavaScript extensions as well.

Also, `OrbitPlotter` and `OrbitPlotter3D` do not share a common parent class and API, which is also a problem since most of the functionality is shared.

Related: #315 and #316.

🎯 **Goal**

1. [ ] Refactor the `plotting` module so it allows the user to configure a "backend", via an environment variable or an IPython magic
2. [ ] Unify the `OrbitPlotter` and `OrbitPlotter3D` classes, putting common functionality in a parent superclass
3. [ ] Use Plotly for the notebook, both in 2D and 3D, and leave a matplotlib backend for those that want static, publication-quality 2D

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

shreyasbapat:
Started Working on this!

Juanlu001:
I think there is a fair share of code duplication now in `plotting.py`, but still I think #345 should be fixed first.

shreyasbapat:
Yeah! True.

shreyasbapat:
Do you mean to say that we have ``OrbitPlotter2D`` , ``OrbitPlotter3D`` and ``BaseOrbitPlotter`` all three at a same time?

Juanlu001:
for the moment yes, although `BaseOrbitPlotter` should perhaps be private (so `_BaseOrbitPlotter`) and we might think of just having one class and switching the behavior using a flag, or something like that

shreyasbapat:
Oh! Yes that's quite possible! Atleast it would reduce the clutter

shreyasbapat:
@Juanlu001  Did we complete the objective here?

Juanlu001:
[...]

PR #340, opened 2018-03-26
Issue #299  -- Animated plot of an orbit 

aeronau:
Addresses #299 
This is an implementation of the work done as part of [Issue 299](https://github.com/poliastro/poliastro/issues/299).
Something funny is going on and I do not know why it does not show an animation when i.e. `op.animate(ss_i)` is called. Maybe someone with experience can review the work done on the plotting file.
This is my GSOC mandatory pull request so feel free to accept or reject this pull request or tag it with _in progress_ and _help wanted_.


src/poliastro/plotting.py
```diff
@@ -59,7 +59,7 @@ class OrbitPlotter(object):
 
     """
 
-    def __init__(self, ax=None, num_points=150):
+    def __init__(self, fig=None, ax=None, num_points=150):
```
Juanlu001: What if `ax` is not `None` but `fig` is? Perhaps instead of passing the figure, you can do

```
if ax:
    self.ax = ax
    self.fig = ax.figure
else:
    self.fig, self.ax = plt.subplots(...)
```

src/poliastro/plotting.py
```diff
@@ -198,6 +199,55 @@ def plot(self, orbit, label=None, color=None):
 
         return lines
 
+    def set_lines_up_to_instant(frame, self, x, y, lines):
```
Juanlu001: Maybe this should be private?

src/poliastro/plotting.py
```diff
@@ -198,6 +199,55 @@ def plot(self, orbit, label=None, color=None):
...
+        a, = self.ax.plot([], [], '--', color=color, label=label)
+
+        ani = FuncAnimation(self.fig, self.set_lines_up_to_instant, frames=np.linspace(0, len(x) - 1, len(x)), fargs=(self, x.to(u.km).value, y.to(u.km).value, a), blit=True, save_count=len(x))
+
+        lines.append(a)
+
+        self.ax.set_xlabel("$x$ (km)")
+        self.ax.set_ylabel("$y$ (km)")
+        self.ax.set_aspect(1)
+        return lines
```
Juanlu001: Why not returning the `ani` instead?

And also, since matplotlib 2.1.0 we can display animations in notebooks! https://matplotlib.org/users/prev_whats_new/whats_new_2.1.0.html#interactive-js-widgets-for-animation

src/poliastro/plotting.py
```diff
@@ -198,6 +199,55 @@ def plot(self, orbit, label=None, color=None):
...
+            self._orbits.append((orbit, label))
+
+        self.set_attractor(orbit.attractor)
+        self._redraw_attractor(orbit.r_p * 0.15)  # Arbitrary Threshhold
+
+        _, positions_in_tuples = orbit.sample(self.num_points)
+        positions = positions_in_tuples.get_xyz().transpose()
+        x, y = self._project(positions)
+
+        # The following should be in a init_animation function, but I [aeronau] still don't know how to pass the arguments to the init function
```
Juanlu001: As far as I understand, the `init_func` has no arguments. https://matplotlib.org/api/_as_gen/matplotlib.animation.FuncAnimation.html I don't know whether there's a better way to do this, perhaps we should read some tutorial.

Juanlu001:
And by the way, configure your git client correctly so all your commits appear with your name and email :)

https://help.github.com/articles/setting-your-commit-email-address-in-git/

src/poliastro/plotting.py
Juanlu001: In the future, maybe it makes sense to have a different object, that inherits some common functionality from OrbitPlotter. But this requires coordination with #338 and @shreyasbapat, so let's keep it simple for now.

aeronau:
I have animatted ss_n after plotting ss_i using the lines below. Note that currently **plots have to be called before the animation** of the orbit as `animate` makes use of `plt.show`, which was what was missing before this commit. This means that only **one orbit can be animated per plot**. I am working on this but I do not really know how to go around it at the moment.
```
import os
os.chdir("/home/arnau/Desktop/poliastro/src")
from poliastro.plotting import *
ss_i = Orbit.circular(Earth, alt=700 * u.km)
ss_f = Orbit.circular(Earth, alt=1000 * u.km)
ss_n = Orbit.circular(Earth, alt=1500 * u.km)
op = OrbitPlotter()
op.plot(ss_i)
op.animate(ss_n)
```
basic animation

aeronau:
Okey, I'll have a look at the style guide and fix it as soon as possible... I reckon that we are quite close to having it plot multiple orbits and coexist with other plots.

Juanlu001:
If you think we're close, then go on! Just don't try to accomplish a lot in
one pull request, it's better to keep them simple :)

aeronau:
After dealing with matplotlib animation I have come to a conclusion that the plotting module requires a great deal of modifications in order to support animations properly without making visualization overcomplicated. I prefer to keep the PEP8 compliant version of the last commit and push for a more adavanced plotting module in the future.

Juanlu001:
> After dealing with matplotlib animation I have come to a conclusion that the plotting module requires a great deal of modifications in order to support animations properly without making visualization overcomplicated. I prefer to keep the PEP8 compliant version of the last commit and push for a more adavanced plotting module in the future.

Perfect! I suggest you to use an editor that warns you against pep8 violations. If you use some automated converter such as autopep8 instead, be careful not to commit changes in other files that are not related to this pull request.

If you need help, ping us :wink:

shreyasbapat:
Yes! ``autopep8 filename.py --in-place`` does that

Juanlu001:
The tests are failing because we are using a dummy string for the axes instead of a proper Mock object:

https://travis-ci.org/poliastro/poliastro/jobs/368556292#L1037

Can you try to edit those ones drawing inspiration from https://github.com/poliastro/poliastro/blob/050339c63c6eab6a1a58adc286f423f5265e0cac/src/poliastro/tests/test_plotting3d.py#L25-L37 ?

Juanlu001:
Also, this branch got a bit behind and will need a rebase.

aeronau:
I think it would be best if this pull is closed. I'd really like to work on this but I currently do not have much time and I am not satisfied with what I have. I think it has been open for too long (my fault) and I do not want to incur time delays. I will look into it as soon as possible but I will tackle the problem differently.

Juanlu001:
No problem @aeronau, and thanks for the effort anyway. I think it's difficult to get this right in a general way, and also matplotlib can be challenging at times, so this PR is very appreciated. Hope to see you contributing soon :)

PR #341, opened 2018-04-03
WIP:J2adding

nikita-astronaut:


Juanlu001:
This looks really promising :) However, I guess it's a work in progress since the J2 constant of the Earth is hardcoded there, instead of allowing for a different J2. If that's the case, please edit the PR title to start with `WIP: `.

Perhaps the most sensible place to put known J2 constants would be `bodies.py`. I hesitate whether to add an extra `J2` parameter, or use the extra `kwargs` introduced in https://github.com/poliastro/poliastro/pull/235/. Your call.

Also, I acknowledge that it will be difficult to find validation cases for bodies different than the Earth. That's why we need to validate against SPICE, STK or GMAT.

nikita-astronaut:
@Juanlu001 , yes, indeed in progress) Please have a look at matrix. I am thinking on how to best change the signature for allowing the other J2

nikita-astronaut:
Is there a point to validate for the other bodies if the validation for Earth passes? Well, what type of mistake could have been remained unnoticed in this case?

Juanlu001:
In this case, the mistake is obvious: with the current implementation, if you try with Mars the validation would fail because the J2 value is hardcoded.

However, it's unlikely that you will find textbook examples with other bodies, or in general other values of J2. I added some discussion on the topic here:

https://github.com/poliastro/poliastro/wiki/Validation

This simple change is a good opportunity to try some of these validation approaches.

Juanlu001:
@AunSiro - keep an eye on this pull request, I might need your help :)

src/poliastro/twobody/perturbations.py
```diff
@@ -0,0 +1,38 @@
...
+    R: float
+        attractor radius
+
+    Notes
+    -----
+    The J2 accounts for the obliteness of the attractor. The formula is given in
+    Howard Curtis, (12.30)
+
+    """
+
```
Juanlu001: Extra whitespace

src/poliastro/twobody/perturbations.py
```diff
@@ -0,0 +1,38 @@
...
+    r_vec = state[:3]
+    r = norm(r_vec)
+
+    factor = (3.0 / 2.0) * k * J2 * (R.to(u.km).value ** 2) / (r ** 5)
+
+    a_x = 5.0 * r_vec[2] ** 2 / r ** 2 - 1
```
Juanlu001: The equations look good to me.

src/poliastro/twobody/perturbations.py
```diff
@@ -0,0 +1,38 @@
...
+    -----
+    The J2 accounts for the obliteness of the attractor. The formula is given in
+    Howard Curtis, (12.30)
+
+    """
+
+    r_vec = state[:3]
+    r = norm(r_vec)
+
+    factor = (3.0 / 2.0) * k * J2 * (R.to(u.km).value ** 2) / (r ** 5)
```
Juanlu001: I think we should keep the perturbations units-agnostic, at least for now. The API would look something like this:

```
r, v = cowell(orbit, tof, ad=J2_perturbation, J2=Earth.J2.value, R=Earth.R.to(u.km).value)
```

I prefer to start with this (ugly) API and then improve when we figure out than having to regret adding it and go back because the performance is bad.

Juanlu001:
Thanks for the addition @nikita-astronaut! This is the beginning of something big :artificial_satellite:

Issue #347, opened 2018-04-20
Propagation failure for highly eccentric hyperbolic orbits with mean_motion


Juanlu001:
🐞 **Problem**

The last cell of the [Florence notebook](http://docs.poliastro.space/en/latest/examples/Catch%20that%20asteroid!.html) fails with these warnings and errors:

```
/home/juanlu/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/quantity.py:639: RuntimeWarning:

overflow encountered in cosh

/home/juanlu/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/quantity.py:639: RuntimeWarning:

overflow encountered in sinh

/home/juanlu/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/units/quantity.py:639: RuntimeWarning:

invalid value encountered in true_divide

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-33-ccd86024eefa> in <module>()
      2 
      3 # This first plot sets the frame
----> 4 frame.plot(florence_hyper, label="Florence")
      5 
      6 # And then we add the Moon

~/Development/poliastro/poliastro-library/src/poliastro/plotting.py in plot(self, orbit, label, color)
    169         self.set_attractor(orbit.attractor)
    170         self._redraw_attractor(orbit.r_p * 0.15)  # Arbitrary Threshhold
--> 171         _, positions = orbit.sample(self.num_points)
    172 
    173         x0, y0 = self._project(orbit.r[None])

~/Development/poliastro/poliastro-library/src/poliastro/twobody/orbit.py in sample(self, values, method)
    331                 nu_values = np.linspace(-nu_limit, nu_limit, values)
    332 
--> 333             return self.sample(nu_values, method)
    334 
    335         elif hasattr(values, "unit") and values.unit in ('rad', 'deg'):

~/Development/poliastro/poliastro-library/src/poliastro/twobody/orbit.py in sample(self, values, method)
    336             values = self._generate_time_values(values)
    337 
--> 338         return values, self._sample(values, method)
    339 
    340     def _sample(self, time_values, method=mean_motion):

~/Development/poliastro/poliastro-library/src/poliastro/twobody/orbit.py in _sample(self, time_values, method)
    341         values = np.zeros((len(time_values), 3)) * self.r.unit
    342         for ii, epoch in enumerate(time_values):
[...]

PR #350, opened 2018-04-20
Added an option to choose propagator in OrbitPlotter

shreyasbapat:
Now we can choose appropriate propagator in ``OrbitPlotter`` while calling ``plot()``. Solves #347

### DIFF
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -21,6 +21,7 @@ from poliastro.util import norm
 from poliastro.bodies import (Earth, Jupiter, Mars, Mercury, Neptune, Saturn,
                               Uranus, Venus)
 from poliastro.twobody.orbit import Orbit
+from poliastro.twobody.propagation import mean_motion, kepler
 
 
 BODY_COLORS = {
@@ -157,7 +158,7 @@ class OrbitPlotter(object):
 
         self.ax.add_patch(mpl.patches.Circle((0, 0), radius, lw=0, color=color))
 
-    def plot(self, orbit, label=None, color=None):
+    def plot(self, orbit, label=None, color=None, method=mean_motion):
         """Plots state and osculating orbit in their plane.
         """
         if not self._frame:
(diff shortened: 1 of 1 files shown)


shreyasbapat:
screenshot from 2018-04-20 18-02-40
Just for reference

Juanlu001:
It's great that we have so much flexibility, but perhaps it's too much. We are already setting the number of sampling points in the `OrbitPlotter` constructor, so what do you think of keeping the method as a plotter property as well?

shreyasbapat:
I think it's fine as then issues like the one you faced won't happen and the user will be able to use various propagators in OrbitPlotter as well..

shreyasbapat:
Should I add it to OrbitPlotter  constructor?

Juanlu001:
But if we're setting the method per-plot instead of per-frame, what about the number of points? Shouldn't it be set per-plot as well? My concern here is consistency. Four options:

* We leave things as they are in this pull request. I think it's inconsistent, unless someone convinces me of the contrary.
* We move `num_points` to each `plot` call. I think it's bad because it makes the signature too long and, in any case, if someone is so concerned about it they can already `sample` in any way they want and use `plot_trajectory`.
* We move `method` to the `OrbitPlotter` constructor. This prevents the user from selecting the method per-plot. I ask: is this so important?
* We do nothing, and tell the user "if you want to change the propagator, just use `plot_trajectory`".

Thoughts?

shreyasbapat:
I don't have much idea about consistency, but ``num_points`` is something that affects the quality of plot. Changing it can either introduce steps in the plot. 
I think if seems inconsistent, we can move it to the constructor. 
We can leave it too :P  
I am not much clear about this.

Juanlu001:
I just remembered that, in STK, you can indeed select a different propagator for each orbit in one scenario:

frame

So I've changed my mind to accept this. ~@shreyasbapat would you like to change the propagator in the last orbit of the Florence notebook to fix it? (See the errors here http://docs.poliastro.space/en/latest/examples/Catch%20that%20asteroid!.html)~ I remember you told me you were on exams, so I'm going to merge this and take care of the rest myself.

shreyasbapat:
Thanks a lot @Juanlu001 . I will be actively up after 1st May Again! Between this, whenever I get time, I will try to do something!

Issue #349, opened 2018-04-20
2D legend not updated if last plot is trajectory


Juanlu001:
🐞 **Problem**

While running the [Jupiter notebook](), I noticed that this displays the correct legend:

```
op = OrbitPlotter()

op.plot(ss_e0, label="Earth")
op.plot_trajectory(ic1.sample(np.linspace(0 * u.deg, 180 * u.deg, 50))[-1], label="Inner Cruise 1")
#op.plot(ss_efly)
op.plot(ss_aph_post, label="Back to Earth");
```

![index](https://user-images.githubusercontent.com/316517/39050165-a8fcd2fa-44a4-11e8-9888-83c9d2c5d585.png)

but, if the last plot is just a trajectory, then the trajectories are not shown:

```
op = OrbitPlotter()

op.plot(ss_e0, label="Earth")
op.plot_trajectory(ic1.sample(np.linspace(0 * u.deg, 180 * u.deg, 50))[-1], label="Inner Cruise 1")
#op.plot(ss_efly)
op.plot_trajectory(ss_aph_post.sample(np.linspace(180 * u.deg, 360 * u.deg, 50))[-1], label="Back to Earth");
```

![index2](https://user-images.githubusercontent.com/316517/39050220-df94705c-44a4-11e8-9493-3ad86a4d6ade.png)

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

shreyasbapat:
Maybe we should give out labels in plot_trajectory and not in plot. Then it would be easy to solve :)

Juanlu001:
Can you elaborate? You mean setting the legend in plot_trajectory?

shreyasbapat:
I am not too sure about it, but yes!

shreyasbapat:
The link in the Jupiter Notebook leads nowhere...:/

shreyasbapat:
Here I propose 2-3 solutions :grimacing: 
I tried out the first one ->
1) Make plot trajectory accept orbit as well as the CartesianRepresentation. Something like this: 
`    def plot_trajectory(self, trajectory, *, label=None, color=None):
        """Plots a precomputed trajectory.

        Parameters
        ----------
[...]

PR #352, opened 2018-04-21
plot_trajectory does not miss labels

shreyasbapat:
Even if you do like this, labels are printed :grinning: 
screenshot from 2018-04-21 14-50-51
Solves #349

### DIFF
--- a/src/poliastro/tests/test_plotting.py
+++ b/src/poliastro/tests/test_plotting.py
@@ -6,6 +6,10 @@ import matplotlib.pyplot as plt
 
 from poliastro.examples import iss
 
+from poliastro.bodies import Earth, Mars
+
+from poliastro.twobody.orbit import Orbit
+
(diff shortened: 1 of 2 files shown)


src/poliastro/tests/test_plotting.py
```diff
@@ -75,3 +79,14 @@ def test_color():
...
+
+def test_plot_trajectory_sets_label():
+    op = OrbitPlotter()
+    earth = Orbit.from_body_ephem(Earth)
+    mars = Orbit.from_body_ephem(Mars)
+    _, trajectory = earth.sample()
+    op.plot(mars, label="Mars")
+    op.plot_trajectory(trajectory, label="Earth")
+    legend = plt.gca().get_legend()
+    assert legend.get_texts()[1].get_text() == "Earth"
```
Juanlu001: Good catch!

shreyasbapat: Thanks :)

PR #356, opened 2018-04-24
Update ISSUE_TEMPLATE.md

jmerskine1:
Added the terminal commands for astropy and poliastro version HOWEVER only works if you have pip installed, may want to rewrite my section with some instructions of how to determine current version if this is not the case?

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -14,12 +14,6 @@
 
 `pip freeze | grep poliastro`
 
-<!--- If the "pip freeze" commands don't work, you likely dont have pip installed. -->
-<!--- In that case, please include your current version of astropy & poliastro here-->
-
-* poliastro version => 
-* astropy version =>
-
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -10,6 +10,15 @@
 
 `python -c "import poliastro.testing; poliastro.testing.test()"`
 
+`pip freeze | grep astropy`
+
+`pip freeze | grep poliastro`
+
+<!--- If the "pip freeze" commands don't work, you likely dont have pip installed. -->
```
Juanlu001: Do you think there's a case where the user might not have `pip` installed?

jmerskine1: To be honest, most likely not.

Juanlu001: Then if you could please remove that part, I'll go ahead and merge this :)

PR #362, opened 2018-04-27
added tests, changed rtol and numiter in kepler

nikita-astronaut:
The `kepler` method suffered from not convergence on large propagation times, when `xi` (universal formulation parameter) is high. 

It turns out, that the method converges in a physical sense, but numerically the requirement `rtol < 1e-10` is not satisfied but to the floating point operations limitation.

After relaxing the requirement `xi - xi_new` to `1e-7`, the method converges and provides physically accurate results. The other thing one had to do is to increase the maximum number of iterations done by the Newton method. The convergence check is done after each iteration, so this would not increase the propagation time for easy orbits, but will make the convergence of hard ones possible.

I added tests, checking that the orbits stated in the old issue converge (iss for 100 years and halleys for 100 years) and that their orbital elements conserve, meaning that the propagation is done right.

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_propagation.py
+++ b/src/poliastro/tests/tests_twobody/test_propagation.py
@@ -265,7 +265,10 @@ def test_long_propagations_kepler_agrees_mean_motion():
     assert_quantity_allclose(r_mm, r_k, rtol=1e-6)
     assert_quantity_allclose(v_mm, v_k, rtol=1e-6)
 
-    halleys = dastcom5.orbit_from_name('1P')[0]
+    r_halleys = [-9018878.63569932, -94116054.79839276, 22619058.69943215]  # km
+    v_halleys = [-49.95092305, -12.94843055, -4.29251577]  # km/s
+    halleys = Orbit.from_vectors(Sun, r_halleys * u.km, v_halleys * u.km / u.s)
+
     r_mm, v_mm = halleys.propagate(tof, method=mean_motion).rv()
     r_k, v_k = halleys.propagate(tof, method=kepler).rv()
     assert_quantity_allclose(r_mm, r_k, rtol=1e-6)
--- a/src/poliastro/tests/tests_twobody/test_propagation.py
+++ b/src/poliastro/tests/tests_twobody/test_propagation.py
@@ -262,8 +262,8 @@ def test_long_propagations_kepler_agrees_mean_motion():
     tof = 100 * u.year
     r_mm, v_mm = iss.propagate(tof, method=mean_motion).rv()
     r_k, v_k = iss.propagate(tof, method=kepler).rv()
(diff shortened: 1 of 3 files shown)


src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -254,3 +256,17 @@ def test_propagate_long_times_keeps_geometry(method):
...
+    tof = 100 * u.year
+    r_mm, v_mm = iss.propagate(tof, method=mean_motion).rv()
+    r_k, v_k = iss.propagate(tof, method=kepler).rv()
+    assert_quantity_allclose(r_mm, r_k, rtol=1e-6)
+    assert_quantity_allclose(v_mm, v_k, rtol=1e-6)
+
+    halleys = dastcom5.orbit_from_name('1P')[0]
+    r_mm, v_mm = halleys.propagate(tof, method=mean_motion).rv()
+    r_k, v_k = halleys.propagate(tof, method=kepler).rv()
+    assert_quantity_allclose(r_mm, r_k, rtol=1e-6)
```
Juanlu001: Better to hardcore the r, v values (from the days of DASTCOM5)

src/poliastro/twobody/propagation.py
```diff
@@ -248,7 +246,7 @@ def _kepler(k, r0, v0, tof, numiter, rtol):
         xi_new = xi + (sqrt_mu * tof - xi * xi * xi * c3_psi -
                        dot_r0v0 / sqrt_mu * xi * xi * c2_psi -
                        norm_r0 * xi * (1 - psi * c3_psi)) / norm_r
-        if abs(np.divide(xi_new - xi, xi_new)) < rtol or abs(xi_new - xi) < rtol:
+        if abs(xi_new - xi) < 1e-6:
```
Juanlu001: Why hardcoding the tolerance here?

nikita-astronaut:
Hello! Thanks for the review. I would manually set 1e-7 to the `_kepler` function. This value is perfect: if it is smaller, say 1e-8, then we run into machine numerical problem and the algorithm won't converge. If it is less, then the method precision is not high enough. The value 1e-7 allows to reach the maximum precision in kepler allowed by the machine floating point boundaries.

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -15,6 +15,8 @@
 from poliastro.twobody.propagation import cowell, kepler, mean_motion
 from poliastro.examples import iss
 
+from poliastro.neos import dastcom5
```
Juanlu001: This import is no longer needed?

Juanlu001:
I added a minor style comment but anyway I will merge this. Thanks!

PR #368, opened 2018-05-06
WIP:trying the new solvers

nikita-astronaut:
Hello! This is the PR to show how changing the API would affect poliastro.

1) The new API does not support callbacks as the previous one. Before it was the function called at every step, now it requires some event (something becomes zero). So for now I removed callbacks stuff.

2) In `test_propagate_to_date_has_proper_epoch` I removed cowell, because maintaining the `rtol=1e-10` makes run of this test fantastically long.

3) In `test_sample.py` I had to mark two cowell tests xfail, because somehow the required precision 1e-7 is not achieved. Perhaps I'll manage to fix that at some point.

4) In the new API, the `jac` parameter is only required by some solvers, and RK45 is not one of them, so it is now used anywhere. But this is a useful function for later.

5) I could not manage to fix the precision problem by changing the integration method.

To my opinion, if we need to refactor and use the new API, we could say that the Cowell method is used only for perturbation stuff, and `mean_motion` is used for analytical gravitational orbits.


nikita-astronaut:
Hello! I rolled back a bit. Seems like using `method='Radau'` resolves all the issues. I returned tests back and removed some `xfail`s. Seems a lot better now.

Juanlu001:
Sorry, I made the change myself :) Now you will have to `git stash && git fetch origin && git merge origin/RK45vsDORPRI835 && git stash pop` before you go on, otherwise you will have ugly error messages on push.

src/poliastro/twobody/propagation.py
```diff
@@ -87,18 +101,11 @@ def cowell(orbit, tof, rtol=1e-10, *, ad=None, callback=None, nsteps=1000, **ad_
...
-        rr.set_solout(callback)
-
-    # Make integration step
-    rr.integrate(tof)
-
-    if rr.successful():
-        r, v = rr.y[:3], rr.y[3:]
+    f_with_ad = functools.partial(func_twobody, k=k, ad=ad, ad_kwargs=ad_kwargs)
+    jac_with_k = functools.partial(gravitational_jac, k=k)
+    rr = solve_ivp(f_with_ad, (0, tof), u0, t_eval=[tof], rtol=rtol, method='Radau')
```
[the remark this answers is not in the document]
nikita-astronaut: `RK45` doesn't get enough `rtol` in a couple of tests run in `test_sample.py`. But I don't know why using `Radau` helps, it looks like magic here

src/poliastro/twobody/propagation.py
```diff
@@ -87,18 +101,11 @@ def cowell(orbit, tof, rtol=1e-10, *, ad=None, callback=None, nsteps=1000, **ad_
...
-    if callback:
-        rr.set_solout(callback)
-
-    # Make integration step
-    rr.integrate(tof)
-
-    if rr.successful():
-        r, v = rr.y[:3], rr.y[3:]
+    f_with_ad = functools.partial(func_twobody, k=k, ad=ad, ad_kwargs=ad_kwargs)
+    jac_with_k = functools.partial(gravitational_jac, k=k)
```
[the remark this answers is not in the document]
nikita-astronaut: No, the reason is that this Jacobian is gravitational only. For other petrurbations I will add later, when everything works here. 

So I need to a check whether ad is provided, and if no, use gravitational_jac.

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -240,7 +218,6 @@ def test_propagate_to_date_has_proper_epoch():
 @pytest.mark.filterwarnings('ignore::UserWarning')
 @pytest.mark.parametrize('method', [
     mean_motion,
-    pytest.param(cowell, marks=pytest.mark.xfail),
```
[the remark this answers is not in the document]
nikita-astronaut: No, it just runs very-very long with Cowell and new API. I will investigate. 100 years seem to be too much

Juanlu001:
Also, it would be nice to have a precision comparison with the old solver. For example, using this snippet:

```
import numpy as np

from poliastro.twobody.propagation import cowell

np.set_printoptions(precision=20)

r0 = [-2384.46, 5729.01, 3050.46] * u.km
v0 = [-7.36138, -2.98997, 1.64354] * u.km / u.s

initial = Orbit.from_vectors(Earth, r0, v0)

def accel(t0, state, k):
    v_vec = state[3:]
    norm_v = (v_vec * v_vec).sum() ** .5
    return 1e-5 * v_vec / norm_v

final = initial.propagate(3 * u.day, method=cowell, ad=accel)
print(final.r)
print(final.v)
```

I get:

```
>>> print(final.r)
[ 13179.39566663877121754922 -13026.25123408228319021873
  -9852.66213692844394245185] km
>>> print(final.v)
[ 2.78170542314378943516  3.21596786944631274352  0.16327165546278937791] km / s
```

Juanlu001:
Regarding the poor performance of Radau that we discussed privately, I found an interesting plot:

screenshot-2018-5-7 untitled - 3817ff090acce3164049f11182e827fc8bdb pdf

(from https://ieeexplore.ieee.org/document/5641330/)

So we are not the only ones to notice a slowdown when using an implicit scheme for a non-stiff problem.

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -65,3 +65,37 @@ def test_atmospheric_drag():
...
+    v0 = [-7.36138, -2.98997, 1.64354] * u.km / u.s
+
+    initial = Orbit.from_vectors(Earth, r0, v0)
+
+    def accel(t0, state, k):
+        v_vec = state[3:]
+        norm_v = (v_vec * v_vec).sum() ** .5
+        return 1e-6 * v_vec / norm_v
+
+    final = initial.propagate(3 * u.day, method=cowell, ad=accel)
```
Juanlu001: This test has no assertions?

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -65,3 +65,37 @@ def test_atmospheric_drag():
...
+        norm_v = (v_vec * v_vec).sum() ** .5
+        return 1e-5 * v_vec / norm_v
+
+    final = initial.propagate(3 * u.day, method=cowell, ad=accel)
+
+    assert_quantity_allclose(final.r, r_expected)
+    assert_quantity_allclose(final.v, v_expected)
+
+
+def test_integrator_converges_with_small_perturbations():
```
Juanlu001: Could we try with `0 * v_vec / norm_v` instead?

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -65,3 +65,37 @@ def test_atmospheric_drag():
     r, v = cowell(orbit, tof, ad=atmospheric_drag, R=R, C_D=C_D, A=A, m=m, H0=H0, rho0=rho0)
 
     assert_quantity_allclose(norm(r) - norm(r0), dr_expected, rtol=1e-2)
+
+
+def test_new_integrator_works_with_small_perturbations():
```
Juanlu001: No need to call it "new_integrator", just "cowell"

src/poliastro/integrators.py
```diff
@@ -0,0 +1,185 @@
...
+        if i % 2 == 0:
+            coeffs[i + 1] = coeffs[i] * s
+        else:
+            coeffs[i + 1] = coeffs[i] * s_back
+    return np.array(coeffs)
+
+
+class DOP835DenseOutput(DenseOutput):
+    def __init__(self, t_old, t_new, y_old, interpolation):
+        super(DOP835DenseOutput, self).__init__(t_old, t_new)
```
Juanlu001: In Python 3 you can use `super().__init__(t_old, t_new)`, please change

src/poliastro/integrators.py
```diff
@@ -0,0 +1,185 @@
...
+                                     support_complex=True)
+        self.y_old = None
+        self.max_step = validate_max_step(max_step)
+        self.beta_stabilizer = validate_beta_stabilizer(beta_stabilizer)
+        self.max_nsteps = validate_max_nsteps(max_nsteps)
+        self.safety_factor = validate_safety_factor(safety_factor)
+        self.rtol, self.atol = validate_tol(rtol, atol, self.n)
+        self.min_step_change = min_step_change
+        self.max_step_change = max_step_change
+        self.order = 7
```
Juanlu001: 7?

src/poliastro/integrators.py
```diff
@@ -0,0 +1,185 @@
+from scipy.integrate import OdeSolver, DenseOutput
```
Juanlu001: This whole module is just impressive. Good job!

src/poliastro/integrator_params.py
```diff
@@ -0,0 +1,100 @@
+import numpy as np
+
+A = []
```
Juanlu001: Could we do

```
A = [
    np.array([5.26001519587677318785587544488e-2]),
    np.array([1.97250569845378994544595329183e-2, 5.91751709536136983633785987549e-2]),
    ...
]
```

instead? Same for the rest of the lists.

Also, is there a reason they have to be arrays? Can we just do

```
A = [
    [5.26001519587677318785587544488e-2],
    [1.97250569845378994544595329183e-2, 5.91751709536136983633785987549e-2],
    [2.95875854768068491816892993775e-2, 0, 8.87627564304205475450678981324e-2]
    ...
]
```

?

src/poliastro/tests/tests_twobody/test_perturbations.py
nikita-astronaut: yes, the idea of the test is that the integrator simply converges and does not fail) I have no data to compare to in this case. But perhaps I could make up an analytical solution to compare. Necessary?

Juanlu001:
Also, before merging this:

* Rebase the branch on current master: `git fetch upstream && git rebase upstream/master && git push -f origin RK45vsDORPRI835`
* Write a brief wiki page explaining the options that we considered and their problems, so we don't forget

And then I'll hit the button. Let's see if we can accelerate this before the next release, but we will likely need https://github.com/numba/numba/issues/2952 fixed for that.

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -65,3 +65,37 @@ def test_atmospheric_drag():
...
+    v0 = [-7.36138, -2.98997, 1.64354] * u.km / u.s
+
+    initial = Orbit.from_vectors(Earth, r0, v0)
+
+    def accel(t0, state, k):
+        v_vec = state[3:]
+        norm_v = (v_vec * v_vec).sum() ** .5
+        return 0.0 * v_vec / norm_v
+
+    final = initial.propagate(3 * u.day, method=cowell, ad=accel)
```
Juanlu001: What about checking that `final` is approximately equal to `initial`?

src/poliastro/integrators.py
nikita-astronaut: thanks) that was a great thing to me to recall how all these finite-difference suffisticated methods work

Juanlu001:
Aaaaaaaaaaaaaaand I'm merging! :tada: Thanks @nikita-astronaut!

Juanlu001:
Opened discussion to contribute this code directly to SciPy: https://github.com/scipy/scipy/pull/9290

PR #371, opened 2018-05-11
Fix pickling

Juanlu001:
I will likely backport this quick fix to 0.9.1 and test more thoroughly for 0.10.0.


Juanlu001:
Reminder: https://docs.scipy.org/doc/numpy/dev/gitwash/development_workflow.html#backporting

Issue #377, opened 2018-05-21
exception in "Propagation using Cowell's formulation.ipynb" example


enritoomey:
In cell 10 of  docs/source/examples/Propagation using Cowell's formulation.ipynb example, when integrate method of scipy ode module is called,  the follwing exeption ocurs:
  
---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
<ipython-input-10-b6cb24b30f44> in <module>()
     10 
     11 while rr.successful() and rr.t + dt < t[-1]:
---> 12     rr.integrate(rr.t + dt)
     13     res[ii] = rr.y
     14     ii += 1

~/pyvenvs/poliastro/lib/python3.5/site-packages/scipy/integrate/_ode.py in integrate(self, t, step, relax)
    430             self._y, self.t = mth(self.f, self.jac or (lambda: None),
    431                                   self._y, self.t, t,
--> 432                                   self.f_params, self.jac_params)
    433         except SystemError:
    434             # f2py issue with tuple returns, see ticket 1187.

~/pyvenvs/poliastro/lib/python3.5/site-packages/scipy/integrate/_ode.py in run(self, f, jac, y0, t0, t1, f_params, jac_params)
   1170     def run(self, f, jac, y0, t0, t1, f_params, jac_params):
   1171         x, y, iwork, istate = self.runner(*((f, t0, y0, t1) +
-> 1172                                           tuple(self.call_args) + (f_params,)))
   1173         self.istate = istate
   1174         if istate < 0:

error: failed in processing argument list for call-back fcn.

This is not the only exception I encountered when running the notebook. Also, in cell 13 of same notebook, the following exception occurs.

thrust
0.10

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/pyvenvs/poliastro/lib/python3.5/site-packages/ipywidgets/widgets/interaction.py in update(self, *args)
    249                     value = widget.get_interact_value()
    250                     self.kwargs[widget._kwarg] = value
--> 251                 self.result = self.f(**self.kwargs)
    252                 show_inline_matplotlib_plots()
    253                 if self.auto_display and self.result is not None:

<ipython-input-12-9f1061196c69> in plot_iss(thrust, mass)
     17     r, v = r0, v0
     18     for ii in range(1, len(t)):
---> 19         r, v = cowell(k, r, v, t[ii] - t[ii - 1], ad=ad)
     20         x, y, z = r
     21         vx, vy, vz = v

[...]

PR #323, opened 2018-02-22
Seprated the build of Notebooks in Travis CI

shreyasbapat:
Addresses Issue #318

PR #378, opened 2018-05-21
Fix notebooks, run in CI

Juanlu001:
After we removed the notebooks from the continuous integration mandatory builds due to timeouts (see https://github.com/poliastro/poliastro/pull/323) we apparently missed some updates (see #377). This is an attempt to run the notebooks in Circle CI, a separated service, so we have proper warnings again when they get outdated.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
           name: build docs
           command: |
             . venv/bin/activate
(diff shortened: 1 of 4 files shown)


Juanlu001:
Awesome! I can see the artifacts (docs) here: https://circleci.com/gh/poliastro/poliastro/4#artifacts/containers/0 Merging!

Issue #376, opened 2018-05-16
Setup benchmarks


Juanlu001:
We should setup proper benchmarks now that we're making changes to the algorithmic parts of poliastro again.

* Starting point: https://github.com/poliastro/poliastro-benchmarks
* Information on astropy-benchmarks: https://groups.google.com/d/msg/astropy-dev/wcWWac4PF0U/eToGXLnrAgAJ
* How to run asv: https://docs.google.com/document/d/1AoPBAbD8DiDVEM6HuOtPKekN3phtcCF4Qk6pxZ0ID-w/edit?usp=sharing

From the astropy-dev mail:

> One tricky part about performance benchmarks is that running them on
continuous integration does not give very accurate results and instead
using dedicated hardware is preferable.

So the first task on my side is to setup a dedicated machine to run these benchmarks. Will try some Cloud VPS providers.

PR #379, opened 2018-05-21
optimizing orbit.sample

nikita-astronaut:
Hello, @Juanlu001 

Did I understand correctly how you proposed it to be? Perhaps the current version seems quite ugly, though uses dense_output. 

The problem is that `cowell` function can be called from `propagate` and from `sample` methods, that's why we need at some places first wrap into `[]` and then unwrap `[0]`.

### DIFF
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -9,7 +9,7 @@ from astropy.coordinates import CartesianRepresentation, get_body_barycentric_po
 
 from poliastro.constants import J2000
 from poliastro.twobody.angles import nu_to_M, E_to_nu
-from poliastro.twobody.propagation import propagate, mean_motion
+from poliastro.twobody.propagation import propagate, mean_motion, cowell
 
 from poliastro.twobody import rv
 from poliastro.twobody import classical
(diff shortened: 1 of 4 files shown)


Juanlu001:
> Did I understand correctly how you proposed it to be? Perhaps the current version seems quite ugly, though uses dense_output.

You did it exactly the way I thought of it, and it's uglier than I thought :)

To justify the ugliness, can you do a simple `%timeit` with both versions to see how faster is the dense output one? If it's a really noticeable performance increase, let's get this merged and I will propose an alternative object design.

src/poliastro/twobody/propagation.py
```diff
@@ -100,17 +86,29 @@ def cowell(orbit, tof, rtol=1e-10, *, ad=None, jac=None, **ad_kwargs):
...
 
-    return r, v
+    rrs = []
+    vvs = []
+    for i in range(len(tof)):
+        y = result.sol(tof[i])
+        rrs.append(y[:3])
+        vvs.append(y[3:])
+
+    if len(tof) == 1:
```
Juanlu001: If the user propagates with a list of times, I would always expect a list, even if the input list has only one element. So I would not make a special case here.

src/poliastro/twobody/orbit.py
```diff
@@ -334,11 +334,17 @@ def sample(self, values=None, method=mean_motion):
...
     def _sample(self, time_values, method=mean_motion):
-        values = np.zeros((len(time_values), 3)) * self.r.unit
-        for ii, epoch in enumerate(time_values):
-            rr = self.propagate(epoch, method).r
-            values[ii] = rr
-
+        # if use cowell, propagate to max_time and use other values as intermediate (dense output)
+        if method == cowell:
+            values, _ = cowell(self, (time_values - self.epoch).to(u.s).value)
+            if len(time_values) == 1:
```
Juanlu001: See below

src/poliastro/twobody/propagation.py
nikita-astronaut: But we have to! The problem is, there are functions that propagate `cowell` not with the list, but with single time value (for instance come tests). And the output is expected to be, once again, single orbit. So, we need to track whether user passed list or single value to cowell and act accordingly.

What I wrote above is not exactly this, I admit: if user gives list, we should return list. But we have to, anyway, work with the special case when user passed just one single value not in the list.

Juanlu001: I know, and that's why we have to distinguish between scalar value and list. But still you added a third case, which is converting a list of one element to the scalar case. As you said, if the user gives a list we should return a list.

Juanlu001:
With your permission @nikita-astronaut, I'm pasting here your screenshot :) This change makes `cowell` sampling an order of magnitude faster. Until we have proper benchmarks (#376), this is more than enough to merge, even if the code is not extremely beautiful.

image

Merging!

PR #381, opened 2018-05-31
WIP:adding 3rd body

nikita-astronaut:
Hello! I am adding the third body. To make the discussion public, I make a PR though the tests are not passing yet.

The test with the 3rd-body perturbation itself gives the wrong answer and I started to think where the mistake can be.

I wrote the check of coordinates test. In Cowell, there is the example number 12.10, where the position of Moon in xyz is given at certain time. Then this position is used to compute the perturbations.

The code that I am currently using to get the coordinates 
`epoch = Time(2456498.8333, format='jd', scale='tdb')
    moon = Orbit.from_body_ephem(Moon, epoch)
    moon = change_frame(moon, ICRS, GCRS)`
gives the y and z components with the wrong sign. Changing these signs manually in 3rd-body tests does not make it pass, but still, the coordinates issue should be resolved... 

@Juanlu001 , perhaps you have an idea?

### DIFF
--- a/src/poliastro/coordinates.py
+++ b/src/poliastro/coordinates.py
@@ -133,9 +133,9 @@ def inertial_body_centered_to_pqw(r, v, source_body):
     return r_pqw, v_pqw
 
 
-def change_frame(orbit, frame_orig, frame_dest):
+def transform(orbit, frame_orig, frame_dest):
     """Transforms Orbit from one frame to another.
-    
+
     Parameters
     ----------
     orbit : ~poliastro.bodies.Orbit
--- a/src/poliastro/coordinates.py
+++ b/src/poliastro/coordinates.py
@@ -140,9 +140,9 @@ def transform(orbit, frame_orig, frame_dest):
     ----------
     orbit : ~poliastro.bodies.Orbit
         Orbit to transform
-    frame_orig : ~astropy.coordinates.BaseRADecFrame
+    frame_orig : ~astropy.coordinates.BaseCoordinateFrame
         Initial frame
-    frame_dest : ~astropy.coordinates.BaseRADecFrame
+    frame_dest : ~astropy.coordinates.BaseCoordinateFrame
         Final frame
 
     Returns
@@ -161,6 +161,6 @@ def transform(orbit, frame_orig, frame_dest):
     orbit_dest.representation = CartesianRepresentation
 
     return Orbit.from_vectors(orbit.attractor,
-                              [orbit_dest.x, orbit_dest.y, orbit_dest.z] * u.km,
-                              [orbit_dest.v_x, orbit_dest.v_y, orbit_dest.v_z] * (u.km / u.s),
+                              orbit_dest.data.xyz,
+                              orbit_dest.data.differentials['s'].d_xyz,
                               epoch=orbit.epoch)
--- a/src/poliastro/coordinates.py
+++ b/src/poliastro/coordinates.py
@@ -9,11 +9,12 @@ from math import sin, cos, sqrt
 import numpy as np
 
 import astropy.units as u
-from astropy.coordinates import get_body_barycentric_posvel
+from astropy.coordinates import get_body_barycentric_posvel, CartesianRepresentation, CartesianDifferential
 
 from poliastro.constants import J2000
 from poliastro.twobody.rv import rv2coe
 from poliastro.util import transform
+from poliastro.twobody.orbit import Orbit
 
 
 def body_centered_to_icrs(r, v, source_body, epoch=J2000, rotate_meridian=False):
@@ -130,3 +131,36 @@ def inertial_body_centered_to_pqw(r, v, source_body):
     v_pqw = (np.array([-sin(nu), (ecc + cos(nu)), 0]) * sqrt(k / p)).T * u.km / u.s
 
     return r_pqw, v_pqw
+
+
+def change_frame(orbit, frame_orig, frame_dest):
+    """Transforms Orbit from one frame to another.
+    
+    Parameters
+    ----------
+    orbit : ~poliastro.bodies.Orbit
+        Orbit to transform
+    frame_orig : ~astropy.coordinates.BaseRADecFrame
+        Initial frame
+    frame_dest : ~astropy.coordinates.BaseRADecFrame
+        Final frame
+
+    Returns
+    -------
+    orbit: ~poliastro.bodies.Orbit
+        Orbit in the new frame
+
+    """
+
+    orbit_orig = frame_orig(x=orbit.r[0], y=orbit.r[1], z=orbit.r[2],
+                            v_x=orbit.v[0], v_y=orbit.v[1], v_z=orbit.v[2],
+                            representation=CartesianRepresentation,
+                            differential_type=CartesianDifferential)
+
+    orbit_dest = orbit_orig.transform_to(frame_dest(obstime=orbit.epoch))
+    orbit_dest.representation = CartesianRepresentation
+
+    return Orbit.from_vectors(orbit.attractor,
+                              [orbit_dest.x, orbit_dest.y, orbit_dest.z] * u.km,
+                              [orbit_dest.v_x, orbit_dest.v_y, orbit_dest.v_z] * (u.km / u.s),
+                              epoch=orbit.epoch)
--- a/src/poliastro/coordinates.py
+++ b/src/poliastro/coordinates.py
@@ -13,7 +13,7 @@ from astropy.coordinates import get_body_barycentric_posvel, CartesianRepresenta
 
 from poliastro.constants import J2000
 from poliastro.twobody.rv import rv2coe
-from poliastro.util import transform
+from poliastro.util import transform as transform_vector
 from poliastro.twobody.orbit import Orbit
 
 
@@ -43,14 +43,14 @@ def body_centered_to_icrs(r, v, source_body, epoch=J2000, rotate_meridian=False)
 
     ra, dec, W = source_body.rot_elements_at_epoch(epoch)
     if rotate_meridian:
-        r = transform(r, -W, 'z')
-        v = transform(v, -W, 'z')
+        r = transform_vector(r, -W, 'z')
+        v = transform_vector(v, -W, 'z')
 
-    r_trans1 = transform(r, -(90 * u.deg - dec), 'x')
-    r_trans2 = transform(r_trans1, -(90 * u.deg + ra), 'z')
+    r_trans1 = transform_vector(r, -(90 * u.deg - dec), 'x')
+    r_trans2 = transform_vector(r_trans1, -(90 * u.deg + ra), 'z')
 
-    v_trans1 = transform(v, -(90 * u.deg - dec), 'x')
-    v_trans2 = transform(v_trans1, -(90 * u.deg + ra), 'z')
+    v_trans1 = transform_vector(v, -(90 * u.deg - dec), 'x')
+    v_trans2 = transform_vector(v_trans1, -(90 * u.deg + ra), 'z')
 
     icrs_frame_pos_coord, icrs_frame_vel_coord = get_body_barycentric_posvel(source_body.name, time=epoch)
 
@@ -88,16 +88,16 @@ def icrs_to_body_centered(r, v, target_body, epoch=J2000, rotate_meridian=False)
     icrs_frame_pos_coord, icrs_frame_vel_coord = get_body_barycentric_posvel(target_body.name, time=epoch)
 
     r_trans1 = r - icrs_frame_pos_coord.xyz
-    r_trans2 = transform(r_trans1, (90 * u.deg + ra), 'z')
-    r_f = transform(r_trans2, (90 * u.deg - dec), 'x')
+    r_trans2 = transform_vector(r_trans1, (90 * u.deg + ra), 'z')
+    r_f = transform_vector(r_trans2, (90 * u.deg - dec), 'x')
 
     v_trans1 = v - icrs_frame_vel_coord.xyz
-    v_trans2 = transform(v_trans1, (90 * u.deg + ra), 'z')
-    v_f = transform(v_trans2, (90 * u.deg - dec), 'x')
+    v_trans2 = transform_vector(v_trans1, (90 * u.deg + ra), 'z')
+    v_f = transform_vector(v_trans2, (90 * u.deg - dec), 'x')
 
     if rotate_meridian:
-        r_f = transform(r_f, W, 'z')
-        v_f = transform(v_f, W, 'z')
+        r_f = transform_vector(r_f, W, 'z')
+        v_f = transform_vector(v_f, W, 'z')
 
     return r_f.to(r.unit), v_f.to(v.unit)
 
--- a/src/poliastro/tests/test_patched_conics.py
+++ b/src/poliastro/tests/test_patched_conics.py
@@ -39,13 +39,13 @@ def test_compute_soi():
 
         assert_quantity_allclose(r_SOI, expected_r_SOI, rtol=1e-1)
 
-
+'''
 @pytest.mark.parametrize("missing_body", [Moon, Pluto])
 def test_compute_missing_body_soi_raises_error(missing_body):
     with pytest.raises(RuntimeError) as excinfo:
         compute_soi(missing_body)
     assert "To compute the semimajor axis for Moon and Pluto use the JPL ephemeris" in excinfo.exconly()
-
+'''
 
 def test_compute_soi_given_a():
     parent = Body(None, 1 * u.km ** 3 / u.s ** 2, "Parent")
--- a/src/poliastro/tests/test_patched_conics.py
+++ b/src/poliastro/tests/test_patched_conics.py
@@ -40,13 +40,12 @@ def test_compute_soi():
         assert_quantity_allclose(r_SOI, expected_r_SOI, rtol=1e-1)
 
 
-'''
+@pytest.mark.xfail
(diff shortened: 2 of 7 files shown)


src/poliastro/coordinates.py
```diff
@@ -130,3 +131,36 @@ def inertial_body_centered_to_pqw(r, v, source_body):
     v_pqw = (np.array([-sin(nu), (ecc + cos(nu)), 0]) * sqrt(k / p)).T * u.km / u.s
 
     return r_pqw, v_pqw
+
+
+def change_frame(orbit, frame_orig, frame_dest):
```
Juanlu001: I'd rather call this `transform`, to be consistent with Astropy method `transform_to` :)

src/poliastro/coordinates.py
```diff
@@ -130,3 +131,36 @@ def inertial_body_centered_to_pqw(r, v, source_body):
...
+    """Transforms Orbit from one frame to another.
+    
+    Parameters
+    ----------
+    orbit : ~poliastro.bodies.Orbit
+        Orbit to transform
+    frame_orig : ~astropy.coordinates.BaseRADecFrame
```
Juanlu001: This is in fact `BaseCoordinateFrame` http://docs.astropy.org/en/stable/api/astropy.coordinates.BaseCoordinateFrame.html

src/poliastro/coordinates.py
```diff
@@ -130,3 +131,36 @@ def inertial_body_centered_to_pqw(r, v, source_body):
...
+    orbit_orig = frame_orig(x=orbit.r[0], y=orbit.r[1], z=orbit.r[2],
+                            v_x=orbit.v[0], v_y=orbit.v[1], v_z=orbit.v[2],
+                            representation=CartesianRepresentation,
+                            differential_type=CartesianDifferential)
+
+    orbit_dest = orbit_orig.transform_to(frame_dest(obstime=orbit.epoch))
+    orbit_dest.representation = CartesianRepresentation
+
+    return Orbit.from_vectors(orbit.attractor,
+                              [orbit_dest.x, orbit_dest.y, orbit_dest.z] * u.km,
```
Juanlu001: I know you copied this from the example, but can you try using `orbit_dest.xyz` instead? http://docs.astropy.org/en/stable/api/astropy.coordinates.CartesianRepresentation.html#astropy.coordinates.CartesianRepresentation.xyz

src/poliastro/coordinates.py
```diff
@@ -130,3 +131,36 @@ def inertial_body_centered_to_pqw(r, v, source_body):
...
+                              [orbit_dest.v_x, orbit_dest.v_y, orbit_dest.v_z] * (u.km / u.s),
```
Juanlu001: And here I guess it would be `v_xyz` http://docs.astropy.org/en/stable/api/astropy.coordinates.CartesianDifferential.html#astropy.coordinates.CartesianDifferential.d_xyz (not `d_xyz`, see https://github.com/astropy/astropy/pull/6569/files#diff-1e15736771f2d9c393cd740f79ae039f)

src/poliastro/coordinates.py
```diff
@@ -130,3 +131,36 @@ def inertial_body_centered_to_pqw(r, v, source_body):
...
+    frame_dest : ~astropy.coordinates.BaseRADecFrame
+        Final frame
+
+    Returns
+    -------
+    orbit: ~poliastro.bodies.Orbit
+        Orbit in the new frame
+
+    """
+
```
Juanlu001: Remove extra whitespace

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -101,3 +107,63 @@ def accel(t0, state, k):
...
+    initial = Orbit.from_classical(Earth, a, ecc, inc, raan, argp, nu, epoch=epoch)
+
+    r, v = cowell(initial, tof, rtol=1e-8, ad=third_body, 
+                  k_third=Moon.k.to(u.km**3 / u.s**2).value, third_body=moon_dense)
+    _, _, _, raan, _, _ = rv2coe(Earth.k.to(u.km**3 / u.s**2).value, r, v)
+
+    assert_quantity_allclose((raan * u.rad).to(u.deg) - 360 * u.deg, -0.06 * u.deg, rtol=1e-2)
+
+
+def test_moon_at_right_position():
```
Juanlu001: This test is here twice?

src/poliastro/twobody/orbit.py
```diff
@@ -162,9 +162,12 @@ def from_body_ephem(cls, body, epoch=None):
             warn("Input time was converted to scale='tdb' with value "
                  "{}. Use Time(..., scale='tdb') instead."
                  .format(epoch.tdb.value), TimeScaleWarning)
-
-        r, v = get_body_barycentric_posvel(body.name, epoch)
-        return cls.from_vectors(body.parent, r.xyz.to(u.km), v.xyz.to(u.km / u.day), epoch)
+        if get_velocity:
```
Juanlu001: Instead of hardcoding the JPL ephemeris here, the user should be responsible of setting the correct ephemeris:

```
In [3]: Orbit.from_body_ephem(Moon)
KeyError                                  Traceback (most recent call last)
<ipython-input-3-c571ef3eb3af> in <module>()
----> 1 Orbit.from_body_ephem(Moon)

~/Development/poliastro/poliastro-library/src/poliastro/twobody/orbit.py in from_body_ephem(cls, body, epoch)
    164                  .format(epoch.tdb.value), TimeScaleWarning)
    165 
--> 166         r, v = get_body_barycentric_posvel(body.name, epoch)
    167         return cls.from_vectors(body.parent, r.xyz.to(u.km), v.xyz.to(u.km / u.day), epoch)
    168 

~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/coordinates/solar_system.py in get_body_barycentric_posvel(body, time, ephemeris)
    328 
    329     """
--> 330     return _get_body_barycentric_posvel(body, time, ephemeris)
    331 
    332 

~/.miniconda36/envs/poliastro36/lib/python3.6/site-packages/astropy/coordinates/solar_system.py in _get_body_barycentric_posvel(body, time, ephemeris, get_velocity)
    223             if get_velocity:
    224                 raise KeyError("the Moon's velocity cannot be calculated with "
--> 225                                "the '{0}' ephemeris.".format(ephemeris))
    226             return calc_moon(time).cartesian
    227 

KeyError: "the Moon's velocity cannot be calculated with the 'builtin' ephemeris."

In [4]: from astropy.coordinates import solar_system_ephemeris
   ...: solar_system_ephemeris.set("jpl")
   ...: 
   ...: 
Out[4]: <ScienceState solar_system_ephemeris: 'jpl'>

In [5]: Orbit.from_body_ephem(Moon)
Out[5]: 151032073 x -151032947 km x 23.5 deg orbit around Earth (♁)
```

so I would remove this change altogether.

src/poliastro/twobody/perturbations.py
```diff
@@ -76,3 +78,23 @@ def atmospheric_drag(t0, state, k, R, C_D, A, m, H0, rho0):
...
+
+    Parameters
+    ----------
+    t0 : float
+        Current time (s)
+    state : numpy.ndarray
+        Six component state vector [x, y, z, vx, vy, vz] (km, km/s).
+    k : float
+        gravitational constant, (km^3/s^2)
+    third_body: ~OdeSolution object
```
Juanlu001: Why it has to be an `OdeSolution` object? Any callable that receives a `t0` (time elapsed since start of the propagation) and returns a state vector should work, right?

src/poliastro/twobody/perturbations.py
```diff
@@ -76,3 +78,23 @@ def atmospheric_drag(t0, state, k, R, C_D, A, m, H0, rho0):
...
+        third body that causes the perturbation
+    """
+
```
Juanlu001: Remove extra white line

Juanlu001:
Also, this will need a rebase on current master:

screenshot-2018-6-1 poliastro poliastro

Juanlu001:
We've found an errata in Curtis, haven't we? :slightly_smiling_face: 

screenshot-2018-6-1 horizons web-interface

(data from https://ssd.jpl.nasa.gov/horizons.cgi)

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -101,3 +107,63 @@ def accel(t0, state, k):
...
+    # based on Cowell, example 12.10
+    epoch = Time(2456498.8333, format='jd', scale='tdb')
+    moon = Orbit.from_body_ephem(Moon, epoch)
+    moon = change_frame(moon, ICRS, GCRS)
+    r_expected = np.array([340958.0, 137043.0, 27521.3]) * u.km
```
Juanlu001: These numbers are wrong, see HORIZONS:

screenshot-2018-6-1 horizons web-interface

and Curtis:

screenshot from 2018-06-01 09-52-51

Juanlu001:
Added one extra comment about the Moon position tests. Notice that it should be crystal clear whether those positions are with respect to the Earth-Moon-Barycenter (EMB) or the Earth center.

nikita-astronaut:
I made `h = 0.25` in `poliastro.ephem`, cause otherwise testing would take too long. For 60 days, there are around `2000` time points to sample from ephem, and this takes too long. I checked, even for `h = 0.5` the results are the same (even here no dependence).

nikita-astronaut:
Hey, @Juanlu001 

Now for every orbit from Curtis I check three parameters. You can see, for GEO and LEO, there is a mismatch in 1 out of 3 parameters, for HEO all the parameters fail to agree. 

Do you know any other sources to validate the function?)

Juanlu001:
Your code seems to be using the JPL ephemeris (via the default `get_velocity=True`). However, the Curtis example uses the Meeus formulas, which are available in Astropy via the `builtin` ephemeris. So I would not try to validate the algorithms using two different ways of computing the Moon position.

I understand that one cannot directly create an `Orbit` object for the Moon using the `builtin` ephemeris because it does not return the velocity. However, for this particular example we are only interested in the position time history, which is something that can be obtained using `astropy.coordinates.get_body_barycentric` and the `builtin` ephemeris:

http://docs.astropy.org/en/stable/api/astropy.coordinates.get_body_barycentric.html

So I would try two things first:

1. Rollback the `get_velocity=True` changes in `Orbit.from_body_ephem` and use `get_body_barycentric` directly for these tests, making sure that the `builtin` ephemeris is used. This way the computation of the position of the Moon will match the one done by Curtis.
2. Run the MATLAB D.49 script to retrieve the true values instead of reading them from the plot.

Two problems might arise here:

1. Curtis is not using DOP853, but Runge-Kutta 4-5. We might have to propagate with the same algorithm if we want true accurate validation.
2. Astropy returns the Moon coordinates in ICRS, after stating that ["Meeus algorithm gives GeocentricTrueEcliptic coordinates"](https://github.com/astropy/astropy/blob/a9300cc3d68f288f47674e00530afb53611ee1eb/astropy/coordinates/orbital_elements.py#L241). However, the `*TrueEcliptic` frames of Astropy are no longer `True` in this release (see looooooong discussion in https://github.com/astropy/astropy/pull/6508) and therefore there might be accuracy problems with the frame conversion there because of the nutation choice.

Lastly, as general coding advice:

1. Rebase & force push early so you don't face conflicts later.
2. Use `pytest.mark.parametrize` for the 3rd body tests, because all the code is shared and only the data is different.

src/poliastro/coordinates.py
nikita-astronaut: done

src/poliastro/coordinates.py
nikita-astronaut: 'GCRS' object has no attribute 'xyz'

src/poliastro/coordinates.py
nikita-astronaut: same thing

src/poliastro/coordinates.py
Juanlu001: I still see it there?

src/poliastro/coordinates.py
Juanlu001: Sorry, it should be `orbit_dest.data.xyz`.

src/poliastro/coordinates.py
Juanlu001: In this case it's more complex, but it's still there:

```
In [31]: dv = GCRS(x=1 * u.km, y=0 * u.km, z=0 * u.km, v_x=0*u.km/u.s, v_y=1*u.k
    ...: m/u.s, v_z=0*u.km/u.s, representation=CartesianRepresentation, differen
    ...: tial_type=CartesianDifferential)

In [32]: dv.data.xyz
Out[32]: <Quantity [1., 0., 0.] km>

In [33]: dv.data.differentials['s'].d_xyz
Out[33]: <Quantity [0., 1., 0.] km / s>
```

src/poliastro/tests/test_patched_conics.py
```diff
@@ -40,11 +40,13 @@ def test_compute_soi():
         assert_quantity_allclose(r_SOI, expected_r_SOI, rtol=1e-1)
 
 
+'''
 @pytest.mark.parametrize("missing_body", [Moon, Pluto])
```
Juanlu001: Why did you have to comment out this test? Also, in this case you can use `pytest.mark.xfail` or `pytest.mark.skip` https://docs.pytest.org/en/latest/skipping.html#skipping-test-functions

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -1,13 +1,22 @@
...
+from poliastro.ephem import build_ephem_interpolant
 from astropy import units as u
 from poliastro.util import norm
-from poliastro.twobody.perturbations import J2_perturbation, atmospheric_drag
-from poliastro.bodies import Earth
+from poliastro.twobody.perturbations import J2_perturbation, atmospheric_drag, third_body
+from poliastro.bodies import Earth, Moon, Sun
 from astropy.tests.helper import assert_quantity_allclose
 from poliastro.twobody import Orbit
+# from poliastro.coordinates import transform
```
Juanlu001: Can we remove this commented out import?

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -101,3 +110,66 @@ def accel(t0, state, k):
...
+    epoch = Time(j_date, format='jd', scale='tdb')
+    initial = Orbit.from_classical(Earth, *test_params['orbit'], epoch=epoch)
+    r, v = cowell(initial, np.linspace(0, tof, 400), rtol=1e-10, ad=third_body,
+                  k_third=body.k.to(u.km**3 / u.s**2).value, third_body=body_r)
+
+    incs, raans, argps = [], [], []
+    for ti, ri, vi in zip(np.linspace(0, tof, 400), r, v):
+        _, _, inc, raan, argp, _ = rv2coe(Earth.k.to(u.km**3 / u.s**2).value, ri, vi)
+
+        # fighting %2pi
```
Juanlu001: I guess this is done to put argp, raan and inc in the range [-pi, pi]? In that case, you could use http://docs.astropy.org/en/stable/api/astropy.coordinates.Angle.html

```
In [45]: a = Angle([-20.0, 150.0, 350.0] * u.deg)

In [46]: a.wrap_at(180 * u.deg)
Out[46]: <Angle [-20., 150., -10.] deg>

In [47]: import numpy as np

In [48]: a.wrap_at(np.pi * u.rad)
Out[48]: <Angle [-20., 150., -10.] deg>
```
```

File: src/poliastro/tests/tests_twobody/test_perturbations.py [hunk 14]
```diff
@@ -101,3 +110,66 @@ def accel(t0, state, k):
...
+        argps.append((argp + np.pi) % (2 * np.pi) - np.pi)
+        raans.append((raan + np.pi) % (2 * np.pi) - np.pi)
+        incs.append((inc + np.pi) % (2 * np.pi) - np.pi)
+
+    # averaging in order to get agreement with Curtis
+    inc_f, raan_f, argp_f = np.mean(incs[-5:]), np.mean(raans[-5:]), np.mean(argps[-5:])
```
Juanlu001: Why do you use the last 5 values only? This deserves a comment

src/poliastro/twobody/orbit.py
```diff
@@ -162,9 +162,12 @@ def from_body_ephem(cls, body, epoch=None):
             warn("Input time was converted to scale='tdb' with value "
                  "{}. Use Time(..., scale='tdb') instead."
                  .format(epoch.tdb.value), TimeScaleWarning)
+        if get_velocity:
```
Juanlu001: I still would remove this functionality, and make the user responsible of setting the correct ephemerides from the outside.

src/poliastro/tests/test_patched_conics.py
nikita-astronaut: I couldn't figure out the meaning of this test and wanted to discuss it with you. after my changes it does not pass anymore

Juanlu001:
Merged manually :) Thanks a lot @nikita-astronaut! This was a tough one :muscle:

nikita-astronaut:
Before merge, we had to make a very important comment!

The reason why tests disagreed was than in Curtis, they claimed that argp_initial is given in `deg`, but actually in their simulations they forgot to transform and took value on `rad`, which caused huge discrepancy. As we looked at their matlab code and used the value of `argp` in `rad`, everything agreed perfectly.

Issue #385, opened 2018-06-11
Sun Incorrectly plotted in plot_solar_system


shreyasbapat:
The radius of the plotted sun exceeds the radii of some orbits of inner planets. 
🐞 **Problem**

See this figure:
![screenshot from 2018-06-11 17-23-59](https://user-images.githubusercontent.com/24257914/41230207-8c8137ee-6d9c-11e8-8513-bdba9ed660ab.png)

🖥 **Please paste the output of following commands**

`conda info -a`

`python -c "import poliastro.testing; poliastro.testing.test()"`

`pip freeze | grep astropy`

`pip freeze | grep poliastro`

🎯 **Goal**

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
Thanks @shreyasbapat! This must be a problem in the method that redraws the attractor, or the minimum radius. Do you want to try fixing it?

shreyasbapat:
Yeah Sure! I will try to fix this.

Juanlu001:
Closed in #387.

PR #387, opened 2018-06-11
Attractor Size is fixed

shreyasbapat:
See this: 
screenshot from 2018-06-12 00-36-50
Solves #385

### DIFF
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -596,46 +596,6 @@ class OrbitPlotter2D:
         )
 
 
-class _PlotlyOrbitPlotter:
-
-    def __init__(self, dim=2):
-        if dim == 3:
-            self._layout = Layout(
-                autosize=True,
-                scene=dict(
-                    xaxis=dict(
-                        title="x (km)",
-                    ),
-                    yaxis=dict(
-                        title="y (km)",
-                    ),
-                    zaxis=dict(
-                        title="z (km)",
-                    ),
-                    aspectmode="data",  # Important!
-                ),
-            )
-        else:
-            self._layout = Layout(
-                autosize=True,
-                xaxis=dict(
-                    title="x (km)",
-                    constrain="domain",
-                ),
-                yaxis=dict(
-                    title="y (km)",
-                    scaleanchor="x",
-                ),
-            )
-        self._data = []  # type: List[dict]
-        self._attractor = None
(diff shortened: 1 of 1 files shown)


src/poliastro/plotting.py
```diff
@@ -160,12 +160,14 @@ def _project(self, rr):
     def _redraw_attractor(self, min_radius=0 * u.km):
         radius = max(self._attractor.R.to(u.km).value, min_radius.to(u.km).value)
         color = BODY_COLORS.get(self._attractor.name, "#999999")
-        self._attractor_radius = radius
 
         for attractor in self.ax.findobj(match=mpl.patches.Circle):
             attractor.remove()
 
-        self.ax.add_patch(mpl.patches.Circle((0, 0), radius, lw=0, color=color))
+        if radius < self._attractor_radius:
```
Juanlu001: Shouldn't this have a unit, as it's done in the other two cases (`np.inf * u.km`)? I'm just asking, if the code works like this then it's fine for me.

shreyasbapat: Although it was working, but then it seems more uniform. Thanks for the suggestion.

Juanlu001:
Interesting:

```
>>> 1 < np.inf
True
>>> 1 * u.km < np.inf
True
>>> 1 * u.km < np.inf * u.km
True
>>> 1 < np.inf * u.km
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/juanlu/.miniconda36/envs/py36/lib/python3.6/site-packages/astropy/units/quantity.py", line 620, in __array_ufunc__
    converters, unit = converters_and_unit(function, method, *inputs)
  File "/home/juanlu/.miniconda36/envs/py36/lib/python3.6/site-packages/astropy/units/quantity_helper.py", line 556, in converters_and_unit
    .format(function.__name__))
astropy.units.core.UnitsError: Can only apply 'greater' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan)
```

Juanlu001:
Thanks @shreyasbapat!

shreyasbapat:
Indeed! Why is this happening? Any clue?

Juanlu001:
I would expect `1 * u.km < np.inf` to fail. The thing is that everything is less than `np.inf`, so I'm not sure how would I fix this. In any case, this is an `astropy.units` issue, feel free to report if you want :)

PR #388, opened 2018-06-14
added solar drag

nikita-astronaut:
Seems like everything's working for now

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_perturbations.py
+++ b/src/poliastro/tests/tests_twobody/test_perturbations.py
@@ -1,6 +1,7 @@
 import pytest
 import functools
 import numpy as np
+import cProfile
 
 from astropy.time import Time
 from poliastro.twobody.propagation import cowell
@@ -173,11 +174,14 @@ def test_3rd_body_Curtis(test_params):
 
 solar_pressure_checks = [{'t_days': 200, 'deltas_expected': [3e-3, -8e-3, -0.035, -80.0]},
                          {'t_days': 400, 'deltas_expected': [-1.3e-3, 0.01, -0.07, 8.0]},
-                         {'t_days': 600, 'deltas_expected': [7e-3, 0.03, -0.10, -80.0]},
+                         {'t_days': 600, 'deltas_expected': [7e-3, 0.03, -0.10, -80.0]}
+                         ]
+'''
                          {'t_days': 800, 'deltas_expected': [-7.5e-3, 0.02, -0.13, 1.7]},
                          {'t_days': 1000, 'deltas_expected': [6e-3, 0.065, -0.165, -70.0]},
                          {'t_days': 1095, 'deltas_expected': [0.0, 0.06, -0.165, -10.0]},
                          ]
+'''
 
 
 def normalize_to_Curtis(t0, sun_r):
@@ -190,7 +194,7 @@ def test_solar_pressure():
     solar_system_ephemeris.set("jpl")
 
     j_date = 2438400.5
-    tof = 1095
+    tof = 600
     sun_r = build_ephem_interpolant(Sun, 365, (j_date, j_date + tof), rtol=1e-2)
     epoch = Time(j_date, format='jd', scale='tdb')
     drag_force_orbit = [10085.44 * u.km, 0.025422 * u.one, 88.3924 * u.deg,
--- a/src/poliastro/tests/tests_twobody/test_perturbations.py
+++ b/src/poliastro/tests/tests_twobody/test_perturbations.py
@@ -8,7 +8,7 @@ from poliastro.twobody.rv import rv2coe
 from poliastro.ephem import build_ephem_interpolant
 from astropy import units as u
 from poliastro.util import norm
-from poliastro.twobody.perturbations import J2_perturbation, atmospheric_drag, third_body, solar_pressure
+from poliastro.twobody.perturbations import J2_perturbation, atmospheric_drag, third_body, radiation_pressure
 from poliastro.bodies import Earth, Moon, Sun
 from astropy.tests.helper import assert_quantity_allclose
 from poliastro.twobody import Orbit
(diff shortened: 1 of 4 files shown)


src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -168,3 +169,57 @@ def test_3rd_body_Curtis(test_params):
...
+    nu_ini = 343.4268 * u.deg
+    initial = Orbit.from_classical(Earth, a_ini, ecc_ini, inc_ini, raan_ini, argp_ini, nu_ini, epoch=epoch)
+    # in Curtis, the mean distance to Sun is used. In order to validate against it, we have to do the same thing
+    sun_normalized = functools.partial(normalize_to_Curtis, sun_r=sun_r)
+
+    r, v = cowell(initial, np.linspace(0, (tof * u.day).to(u.s).value, 4000), rtol=1e-8, ad=solar_pressure,
+                  R_Earth=Earth.R.to(u.km).value, C_R=2.0, A=2e-4, m=100, Wdivc_s=Sun.Wdivc.value, sun=sun_normalized)
+
+    delta_as, delta_eccs, delta_incs, delta_raans, delta_argps, delta_hs = [], [], [], [], [], []
+    for ri, vi in zip(r, v):
```
Juanlu001: Wow, so this is the required code to get a time history of orbital elements... There should be an easier way to do it. Not for this pull request, of course.

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -168,3 +169,57 @@ def test_3rd_body_Curtis(test_params):
                               (argp_f * u.rad).to(u.deg) - test_params['orbit'][4]],
                              [test_params['raan'], test_params['inc'], test_params['argp']],
                              rtol=1e-1)
+
+
+solar_pressure_checks = [{'t_days': 200, 'deltas_expected': [3e-3, -8e-3, -0.035, -80.0]},
```
Juanlu001: Can't we use `pytest.mark.parametrize` instead?

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -168,3 +169,54 @@ def test_3rd_body_Curtis(test_params):
...
+                         {'t_days': 400, 'deltas_expected': [-1.3e-3, 0.01, -0.07, 8.0]},
+                         {'t_days': 600, 'deltas_expected': [7e-3, 0.03, -0.10, -80.0]},
+                         {'t_days': 800, 'deltas_expected': [-7.5e-3, 0.02, -0.13, 1.7]},
+                         {'t_days': 1000, 'deltas_expected': [6e-3, 0.065, -0.165, -70.0]},
+                         {'t_days': 1095, 'deltas_expected': [0.0, 0.06, -0.165, -10.0]},
+                         ]
+
+drag_force_orbit = [10085.44 * u.km, 0.025422 * u.one, 88.3924 * u.deg,
```
Juanlu001: Shouldn't this be defined inside the test function?

src/poliastro/twobody/perturbations.py
```diff
@@ -99,3 +99,65 @@ def third_body(t0, state, k, k_third, third_body):
     body_r = third_body(t0)
     delta_r = body_r - state[:3]
     return k_third * delta_r / norm(delta_r) ** 3 - k_third * body_r / norm(body_r) ** 3
+
+
+def shadow_function(r_sat, r_sun, R_Earth):
```
Juanlu001: Does this apply only for the Earth? Or in fact any attractor around any star?

src/poliastro/twobody/perturbations.py
```diff
@@ -99,3 +99,65 @@ def third_body(t0, state, k, k_third, third_body):
...
+    """Determines whether the satellite is in Earth's shadow,
+       uses algorithm 12.3 from Howard Curtis
+
+    Parameters
+    ----------
+    r_sat : numpy.ndarray
+        position of the satellite in the Earth frame (km)
```
Juanlu001: Same as above: we should avoid the `Earth/Sun` terminology and use `attractor/star` or similar. This should work for satellites around Mars, or around planets that orbit stars other than the Sun.

src/poliastro/twobody/perturbations.py
```diff
@@ -99,3 +99,65 @@ def third_body(t0, state, k, k_third, third_body):
...
+    r_sun_norm = norm(r_sun)
+
+    theta = np.arccos(np.dot(r_sat, r_sun) / r_sat_norm / r_sun_norm)
+    theta_1 = np.arccos(R_Earth / r_sat_norm)
+    theta_2 = np.arccos(R_Earth / r_sun_norm)
+
+    return theta < theta_1 + theta_2
+
+
+def solar_pressure(t0, state, k, R_Earth, C_R, A, m, Wdivc_s, sun):
```
Juanlu001: Rename to `radiation_pressure`?

src/poliastro/twobody/perturbations.py
```diff
@@ -99,3 +99,65 @@ def third_body(t0, state, k, k_third, third_body):
...
+    This function provides the acceleration due to solar pressure. We follow
+    Howard Curtis, section 12.9
+
+    """
+
+    r_sun = sun(t0)
+    r_sat = state[:3]
+    P_s = Wdivc_s / (norm(r_sun) ** 2)
+
+    nu = float(shadow_function(r_sat, r_sun, R_Earth))
```
Juanlu001: Smart!

Juanlu001:
Also, this needs a rebase on current master.

src/poliastro/tests/tests_twobody/test_perturbations.py
nikita-astronaut: No, we can't, because we first calculate trajectory and then check it on some special time values. If we calculate trajectory multiple times, testing would take tooooo long

Juanlu001: OK, I understand. I think this could be achieved more cleanly using `xunit` setup functions https://docs.pytest.org/en/latest/xunit_setup.html or fixtures https://docs.pytest.org/en/latest/fixture.html#scope-sharing-a-fixture-instance-across-tests-in-a-class-module-or-session but let's not block the pull request because of this.

src/poliastro/twobody/perturbations.py
```diff
@@ -101,6 +102,7 @@ def third_body(t0, state, k, k_third, third_body):
     return k_third * delta_r / norm(delta_r) ** 3 - k_third * body_r / norm(body_r) ** 3
 
 
+@jit
```
Juanlu001: Just as a clarification, `norm` is not jit-eable because it's meant to preserve units of quantities.

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 import functools
 import numpy as np
+import cProfile
```
Juanlu001: Oops!

Juanlu001:
Merged! Thanks @nikita-astronaut :clap:

PR #389, opened 2018-06-17
Added a dark mode to OrbitPlotter

shreyasbapat:
Looks like this:
screenshot from 2018-06-17 17-29-38

### DIFF
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -65,7 +65,7 @@ class OrbitPlotter(object):
 
     """
 
-    def __init__(self, ax=None, num_points=150):
+    def __init__(self, ax=None, num_points=150, dark=False):
         """Constructor.
 
         Parameters
@@ -76,6 +76,7 @@ class OrbitPlotter(object):
             Number of points to use in plots, default to 150.
         """
         self.ax = ax
+        plt.style.use('dark_background') if dark else plt.style.use('default')
         if not self.ax:
             _, self.ax = plt.subplots(figsize=(6, 6))
         self.num_points = num_points
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -74,6 +74,8 @@ class OrbitPlotter(object):
             Axes in which to plot. If not given, new ones will be created.
         num_points : int, optional
             Number of points to use in plots, default to 150.
+        dark : bool, optional
+            If set as True, plots the orbit in Dark mode.
         """
         self.ax = ax
         if not self.ax:
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -76,9 +76,12 @@ class OrbitPlotter(object):
             Number of points to use in plots, default to 150.
         """
         self.ax = ax
-        plt.style.use('dark_background') if dark else plt.style.use('default')
         if not self.ax:
-            _, self.ax = plt.subplots(figsize=(6, 6))
+            if dark:
+                with plt.style.context('dark_background'):
+                    _, self.ax = plt.subplots(figsize=(6, 6))
+            else:
+                _, self.ax = plt.subplots(figsize=(6, 6))
         self.num_points = num_points
         self._frame = None
         self._attractor = None
--- a/src/poliastro/tests/test_plotting.py
+++ b/src/poliastro/tests/test_plotting.py
@@ -90,3 +90,10 @@ def test_plot_trajectory_sets_label():
     op.plot_trajectory(trajectory, label="Earth")
     legend = plt.gca().get_legend()
     assert legend.get_texts()[1].get_text() == "Earth"
+
+
+def test_dark_mode_plots_dark_plot():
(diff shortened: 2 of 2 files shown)


shreyasbapat:
Didn't really get what happened with this! @Juanlu001 Do you have any idea?

Juanlu001:
There are some slow tests, and the job was stopped because of that. I restarted it, hope it helps :crossed_fingers: but we will probably need an alternative here.

Juanlu001:
@nikita-astronaut this test failed: https://travis-ci.org/poliastro/poliastro/jobs/393287342#L1193

Juanlu001:
Also @shreyasbapat this change will affect the global matplotlib state, so for example will override any style the user has chosen. Do you think you could enclose the plotting commands in a `with` block like this?

https://matplotlib.org/users/style_sheets.html#temporary-styling

shreyasbapat:
I will try!

shreyasbapat:
By the way, the change will just affect the particular OrbitPlotter. If we
plot another orbit, without dark mode on, it will plot a normal plot.

nikita-astronaut:
Hey everyone! Concerning the test that failed, the log suggests that there is some problem in
`>       solar_system_ephemeris.set("jpl")`:
`E           urllib.error.URLError: <urlopen error The read operation timed out>`

Perhaps, one needs to just rerun tests? If in doesn't work, @Juanlu001 , how this could be fixed?

Juanlu001:
@nikita-astronaut we can use `de432s` as the ephemeris if the years of the Moon orbit are between 1950 and 2050:

http://docs.astropy.org/en/stable/api/astropy.coordinates.solar_system_ephemeris.html#astropy.coordinates.solar_system_ephemeris

Juanlu001:
@shreyasbapat if we continue working on this, you will need to rebase after https://github.com/poliastro/poliastro/pull/390 is merged.

On the other hand, remember that, as a workaround while we implement this, the user can do

```
with plt.style.context('dark_background'):
    frame = OrbitPlotter()
    frame.plot(...)
```

This does not work for plotly-based `OrbitPlotter` classes though.

shreyasbapat:
Doing that willl enable the dark mode for that particular ``OrbitPlotter`` right?

Juanlu001:
> Doing that willl enable the dark mode for that particular OrbitPlotter right?

Correct!

shreyasbapat:
Added the ``with`` block

shreyasbapat:
@Juanlu001  Please look at it, if you approve, I will add some tests.

Juanlu001:
Very interesting! Then go ahead and add some tests :+1:

shreyasbapat:
I need help with some tests as I googled a lot but couldn't find any way to assert the changes that happen in the matplotlib. Please tell if you know some, else, I will have to research

Juanlu001:
Here you have some inspiration:

https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/tests/test_style.py

Maybe you can test the color of the background.

shreyasbapat:
True! I was thinking about the same. Thanks! :)

shreyasbapat:
I almost forgot about this xD ! Sorry!

src/poliastro/plotting.py
```diff
@@ -82,8 +82,10 @@ def __init__(self, ax=None, num_points=150, dark=False):
             if dark:
                 with plt.style.context('dark_background'):
                     _, self.ax = plt.subplots(figsize=(6, 6))
+                    self.edgecolor = mpl.rcParams['figure.edgecolor']
```
Juanlu001: Could we do this without adding an extra property to `OrbitPlotter`?

Juanlu001:
No problem! It would be better to find an alternative approach though. I understand that we can't test the matplotlib rcParams this way, since we are using a context manager, but perhaps we can get the figure background color with `fig.get_facecolor()` and check that it's black or dark?

src/poliastro/plotting.py
shreyasbapat: We have to do this because we are using temporary styling. And I could not get any property of the plot outside the ``OrbitPlotter`` itself. Do you have any idea? Because what we can get outside is a ``plt.subplots`` object which does not help :/

shreyasbapat:
Okay ! Let me try!

src/poliastro/plotting.py
Juanlu001: You could also get `plt.gcf()` (Get Current Figure) and use `plt.gcf().get_facecolor()` there

shreyasbapat:
Hi! 
Thanks for your suggestion, this might be a better option!

src/poliastro/plotting.py
shreyasbapat: Yes! I did almost a similar thing!

shreyasbapat:
@Juanlu001  We need to fix this ASAP now :/

Juanlu001:
Thanks! 👏

PR #392, opened 2018-06-22
initial changes

nikita-astronaut:
Currently our API allows only one perturbation acceleration. However, in case of J3/J4 perturbations, their effect is usually accounted together with, say, J2 perturbation. So it is reasonable to allow passing multiple perturbation forces at once.


Juanlu001:
And also, if we try to be very general and play too much with `partial` objects, I fear we might have a hard time in the future accelerating these with numba. Now we're blocked by https://github.com/numba/numba/issues/2952, but I suppose that this will be corrected in the future.

Juanlu001:
Let's better close this and follow a different approach.

PR #393, opened 2018-06-23
WIP : Added the __BaseOrbitPlotter Class

shreyasbapat:
Can we discuss on this? This might be terribly wrong! @Juanlu001  please have a look and tell what needs to be changed. This is completely based on my perception. I will change the tests later. 
Addresses #338 

Things are in working condition!

### DIFF
--- a/src/poliastro/plotting.py
+++ b/src/poliastro/plotting.py
@@ -230,17 +230,6 @@ def _generate_sphere(radius, center, num=20):
     return xx, yy, zz
 
 
-def _plot_sphere(radius, color, name, center=[0, 0, 0] * u.km):
-    xx, yy, zz = _generate_sphere(radius, center)
-    sphere = Surface(
-        x=xx.to(u.km).value, y=yy.to(u.km).value, z=zz.to(u.km).value,
-        name=name,
-        colorscale=[[0, color], [1, color]],
-        cauto=False, cmin=1, cmax=1, showscale=False,  # Boilerplate
-    )
-    return sphere
-
-
 class _BaseOrbitPlotter:
     """
     Parent Class for the 2D and 3D OrbitPlotter Classes based on Plotly.
@@ -273,6 +262,17 @@ class _BaseOrbitPlotter:
 
         elif attractor is not self._attractor:
             raise NotImplementedError("Attractor has already been set to {}.".format(self._attractor.name))
+        
+    def _redraw_attractor(self, min_radius=0 * u.km):
+        # Select a sensible value for the radius: realistic for low orbits,
+        # visible for high and very high orbits
+        radius = max(self._attractor.R.to(u.km), min_radius.to(u.km))
+        if radius < self._attractor_radius:
+            # If the resulting radius is smaller than the current one, redraw it
+            shape = self._plot_sphere(radius, BODY_COLORS.get(self._attractor.name, "#999999"), self._attractor.name)
+            # Overwrite stored properties
+            self._attractor_radius = radius
+            self._attractor_data = shape
 
     def plot_trajectory(self, trajectory, *, label=None, color=None):
         """Plots a precomputed trajectory.
@@ -294,6 +294,30 @@ class _BaseOrbitPlotter:
             self._redraw_attractor(trajectory.norm().min() * 0.15)  # Arbitrary threshold
 
         self._plot_trajectory(trajectory, str(label), color, False)
+
+    def plot(self, orbit, *, label=None, color=None):
+        """Plots state and osculating orbit in their plane.
+
+        Parameters
+        ----------
+        orbit : ~poliastro.twobody.orbit.Orbit
+        label : string, optional
+        color : string, optional
+        """
+        if color is None:
+            color = next(self._color_cycle)
+
+        self.set_attractor(orbit.attractor)
+        self._redraw_attractor(orbit.r_p * 0.15)  # Arbitrary threshold
+
+        label = _generate_label(orbit, label)
+        _, trajectory = orbit.sample()
+
+        self._plot_trajectory(trajectory, label, color, True)
+        # Plot required 2D/3D shape in the position of the body
+        radius = min(self._attractor_radius * 0.5, (norm(orbit.r) - orbit.attractor.R) * 0.3)  # Arbitrary thresholds
+        shape = self._plot_sphere(radius, color, label, center=orbit.r)
+        self._data.append(shape)
 
     def _prepare_plot(self, **layout_kwargs):
         # If there are no orbits, draw only the attractor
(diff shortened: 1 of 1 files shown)


shreyasbapat:
One more change : I removed the handle with care phrase from the docstring of OrbitPlotter2D. Do let me know if we plan to keep it.

src/poliastro/plotting.py
```diff
@@ -241,205 +241,48 @@ def _plot_sphere(radius, color, name, center=[0, 0, 0] * u.km):
...
+    """
+    Parent Class for the 2D and 3D OrbitPlotter Classes based on Plotly.
     """
 
-    def __init__(self):
-        self._layout = Layout(
-            autosize=True,
-            scene=dict(
+    def __init__(self, dim3=True):
+        if dim3:
```
Juanlu001: Instead of this boolean parameter, why not override the `__init__` method in each subclass? You can leave the common stuff here, call `super().__init__()` in the subclasses and avoid having the `dim3` parameter.

src/poliastro/plotting.py
```diff
@@ -480,24 +323,39 @@ def _redraw_attractor(self, min_radius=0 * u.km):
 
         # If the resulting radius is smaller than the current one, redraw it
         if radius < self._attractor_radius:
-            circle = self._plot_circle(
-                radius, BODY_COLORS.get(self._attractor.name, "#999999"), self._attractor.name)
+            if self.dim:
```
Juanlu001: I guess this was intended to be `dim3`? In any case, I also think this should not be here, and instead controlled in the subclass.

src/poliastro/plotting.py
```diff
@@ -480,24 +323,39 @@ def _redraw_attractor(self, min_radius=0 * u.km):
...
-            x=trajectory.x.to(u.km).value, y=trajectory.y.to(u.km).value,
-            name=label,
-            line=dict(
-                color=color,
-                width=2,
-                dash='dash' if dashed else 'solid',
-            ),
-            mode="lines",  # Boilerplate
-        )
+        if self.dim:
```
Juanlu001: Same as above

src/poliastro/plotting.py
```diff
@@ -535,33 +414,14 @@ def plot(self, orbit, *, label=None, color=None):
...
-        trajectory : ~astropy.coordinates.CartesianRepresentation
-            Trajectory to plot.
-        label : string, optional
-        color : string, optional
-
-        """
-        if self._attractor is None:
-            raise ValueError("An attractor must be set up first, please use "
-                             "set_attractor(Major_Body).")
+        if self.dim:
```
Juanlu001: Same as above

src/poliastro/plotting.py
```diff
@@ -572,20 +432,11 @@ def _prepare_plot(self, **layout_kwargs):
             self._layout.update(layout_kwargs)
 
     def show(self, **layout_kwargs):
```
Juanlu001: Why removing this docstring?

src/poliastro/plotting.py
```diff
@@ -572,20 +432,11 @@ def _prepare_plot(self, **layout_kwargs):
...
-        """Shows the plot in the Notebook.
-        """
         self._prepare_plot(**layout_kwargs)
 
         iplot(self.figure)
 
     def savefig(self, filename, **layout_kwargs):
-        """Function for saving the plot locally.
```
Juanlu001: Same as above

src/poliastro/plotting.py
```diff
@@ -596,6 +447,50 @@ def savefig(self, filename, **layout_kwargs):
...
+
+
+class OrbitPlotter2D(_BaseOrbitPlotter):
+    """OrbitPlotter2D class.
+
+    .. versionadded:: 0.9.0
+    """
+
+    def __init__(self):
+        _BaseOrbitPlotter.__init__(self, dim3=False)
```
Juanlu001: Use `super()`!

src/poliastro/plotting.py
shreyasbapat: Somewhere in our discussions, we decided to use a flag. Okay I will remove it

Juanlu001: I recall I told you about a flag, but I did not predict it would be with this intent. See, I'm bad at discussing without seeing the code :)

shreyasbapat: Oh! Sorry. I am just correcting things. Would commit the changes in a while. :)

src/poliastro/plotting.py
shreyasbapat: Added back!

shreyasbapat:
Sorry that I was out from the work almost 2-3 days! I have worked on your review. This is what I got! :) Please tell if any other changes are to be made.

shreyasbapat:
@Juanlu001  Please review :)

src/poliastro/plotting.py
```diff
@@ -399,6 +327,95 @@ def savefig(self, filename, **layout_kwargs):
...
+            }
+        })
+
+    def _redraw_attractor(self, min_radius=0 * u.km):
+        # Select a sensible value for the radius: realistic for low orbits,
+        # visible for high and very high orbits
+        radius = max(self._attractor.R.to(u.km), min_radius.to(u.km))
+        if radius < self._attractor_radius:
+            # If the resulting radius is smaller than the current one, redraw it
+            shape = _plot_sphere(radius, BODY_COLORS.get(self._attractor.name, "#999999"), self._attractor.name)
```
Juanlu001: This is almost exactly equal as `OrbitPlotter2D._redraw_attractor`, except for `_plot_sphere` vs `self._plot_circle`. Maybe we can have a `self._plot_sphere` in `OrbitPlotter3D`, rename `OrbitPlotter2D._plot_circle` to `OrbitPlotter2D._plot_sphere` (yes, it's a 2D sphere, right?) and we can move `_redraw_attractor` to the base class. What do you think?

src/poliastro/plotting.py
```diff
@@ -399,6 +327,95 @@ def savefig(self, filename, **layout_kwargs):
...
+            x=trajectory.x.to(u.km).value, y=trajectory.y.to(u.km).value, z=trajectory.z.to(u.km).value,
+            name=label,
+            line=dict(
+                color=color,
+                width=5,
+                dash='dash' if dashed else 'solid',
+            ),
+            mode="lines",  # Boilerplate
+        )
+        self._data.append(trace)
```
Juanlu001: What about moving `self._data.append(trace)` to the base class and call `super()` at the end of `_plot_trajectory` for both subclasses?

src/poliastro/plotting.py
```diff
@@ -399,6 +327,95 @@ def savefig(self, filename, **layout_kwargs):
...
+        self._redraw_attractor(orbit.r_p * 0.15)  # Arbitrary threshold
+
+        label = _generate_label(orbit, label)
+        _, trajectory = orbit.sample()
+
+        self._plot_trajectory(trajectory, label, color, True)
+
+        # Plot required 2D/3D shape in the position of the body
+        radius = min(self._attractor_radius * 0.5, (norm(orbit.r) - orbit.attractor.R) * 0.3)  # Arbitrary thresholds
+        shape = _plot_sphere(radius, color, label, center=orbit.r)
```
Juanlu001: Exactly the same case: both `plot` methods are identical except for one function. Renaming `_plot_sphere` to `OrbitPlotter3D._plot_sphere`, we could move `plot` to the base class.

src/poliastro/plotting.py
shreyasbapat: Yeah! True. It would be much better

src/poliastro/plotting.py
shreyasbapat: Yes True! Done!

src/poliastro/plotting.py
shreyasbapat: Apart from this, resolved everything! Is this necessary? As it would anyways increase number of lines

Juanlu001: Nah, let's leave it like this.

Juanlu001:
There's one PEP8 issue left though :)

shreyasbapat:
Fixed!

PR #394, opened 2018-06-27
added parabolic region approach

nikita-astronaut:
Hello! I added parabolic region approach and tests to it. Seems like everything works.

However, in hyperbolic region, if we take their initial guess, it fails for high ecc.

So, it indeed works faster for small eccs, but for ecc like 3200, Newton can not converge with their initial guess. SO I decided not to add it, at least for now.

### DIFF
--- a/src/poliastro/twobody/angles.py
+++ b/src/poliastro/twobody/angles.py
@@ -24,6 +24,126 @@ def _kepler_equation_prime_hyper(F, M, ecc):
     return ecc * np.cosh(F) - 1
 
 
+def _kepler_equation_parabolic(D, M, ecc):
+    return M_parabolic(ecc, D) - M
+
+
+def _kepler_equation_prime_parabolic(D, M, ecc):
+    return M_parabolic_prime(ecc, D)
+
+
+def M_parabolic(ecc, D, tolerance=1e-16):
+    """Computes the Kepler equation r.h.s. in near-parabolic regime
+
+    Parameters
+    ----------
+    D : float
+        Eccentric anomaly (rad).
+    ecc : float
+        Eccentricity,
+    tolerance : float (optional)
+        smallness of the last term in series
+    Returns
+    -------
+    M_parabolic : float
+        kepler equation r.h.s.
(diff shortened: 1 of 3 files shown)


src/poliastro/twobody/angles.py
```diff
@@ -24,6 +24,126 @@ def _kepler_equation_prime_hyper(F, M, ecc):
...
+    tolerance : float (optional)
+        smallness of the last term in series
+    Returns
+    -------
+    M_parabolic : float
+        kepler equation r.h.s.
+
+    Note
+    -----
+    Taken from the paper <<Robust resolution of Kepler’s equation in all eccentricity regimes>>,
```
Juanlu001: Could you change the citation style here to something more standard? For example (taken from scholar.google.com, Chicago style):

```
.. [1] Farnocchia, Davide, Davide Bracali Cioci, and Andrea Milani.
   "Robust resolution of Kepler’s equation in all eccentricity regimes."
   Celestial Mechanics and Dynamical Astronomy 116, no. 1 (2013): 21-34.
```

(as explained in https://numpydoc.readthedocs.io/en/latest/format.html#sections, section 13)

Juanlu001: Also, rename `Note` to `Notes` (and adjust the `-----`).

nikita-astronaut:
Hello! Thanks for the review)

I added initial guess from the paper for hyperbolic region, but for non-high eccenricies. Weens like for ecc = 3200, their initial guess fails, but ours is robust, stable and works. But for smaller eccs, their initial guess works faster, so I added it. The threshold of 10 is chosen at random)

Juanlu001:
Now there are some convergence failures though

nikita-astronaut:
With their initial guess, there are always tests not converging. The problem is with large \Delta t and ecc, cause it was not included into the paper tests. So, our initial guess, as compared to the paper, is around 1 iteration slower, but at least stable.

Juanlu001:
I think this is good to go!

Issue #391, opened 2018-06-21
2 Failures during testing after successful install on Windows


zbaum1:
🐞 **Problem**

🖥 **Please paste the output of following commands**

`conda info -a`

`python -c "import poliastro.testing; poliastro.testing.test()"`

```
C:\>python -c "import poliastro.testing; poliastro.testing.test()"
============================= test session starts =============================
platform win32 -- Python 3.6.5, pytest-3.5.1, py-1.5.3, pluggy-0.6.0
rootdir: C:\, inifile:
plugins: remotedata-0.2.1, openfiles-0.3.0, doctestplus-0.1.3, arraydiff-0.2
collected 374 items

Users\zachary.baum\AppData\Local\Continuum\anaconda3\lib\site-packages\poliastro\tests\test_bodies.py . [  0%]
.....                                                                    [  1%]
Users\zachary.baum\AppData\Local\Continuum\anaconda3\lib\site-packages\poliastro\tests\test_coordinates.py . [  1%]
..                                                                       [  2%]
Users\zachary.baum\AppData\Local\Continuum\anaconda3\lib\site-packages\poliastro\tests\test_hyper.py . [  2%]
..........                                                               [  5%]
Users\zachary.baum\AppData\Local\Continuum\anaconda3\lib\site-packages\poliastro\tests\test_iod.py . [  5%]
........                                                                 [  7%]
Users\zachary.baum\AppData\Local\Continuum\anaconda3\lib\site-packages\poliastro\tests\test_jit.py . [  8%]
..                                                                       [  8%]
Users\zachary.baum\AppData\Local\Continuum\anaconda3\lib\site-packages\poliastro\tests\test_maneuver.py . [  8%]
.....                                                                    [ 10%]
Users\zachary.baum\AppData\Local\Continuum\anaconda3\lib\site-packages\poliastro\tests\test_patched_conics.py . [ 10%]
.                                                                        [ 10%]
Users\zachary.baum\AppData\Local\Continuum\anaconda3\lib\site-packages\poliastro\tests\test_plotting.py . [ 10%]
........                                                                 [ 13%]
Users\zachary.baum\AppData\Local\Continuum\anaconda3\lib\site-packages\poliastro\tests\test_plotting2d.py . [ 13%]
......F                                                                  [ 15%]
Users\zachary.baum\AppData\Local\Continuum\anaconda3\lib\site-packages\poliastro\tests\test_plotting3d.py . [ 15%]
.......F                                                                 [ 17%]
[...]

PR #396, opened 2018-06-28
Saving figures in temporary directory

shreyasbapat:
Solves #391 
Ran the notebook = "Plotting in 3D" with one change. 
screenshot from 2018-06-28 22-27-00
Changed this to : 
screenshot from 2018-06-28 22-29-43
As now, we don't allow to plot two bodies with different attractors in the same plot.


src/poliastro/tests/test_plotting2d.py
```diff
@@ -92,7 +96,10 @@ def test_savefig_calls_prepare_plot():
     m = OrbitPlotter2D()
     earth = Orbit.from_body_ephem(Earth)
     m.plot(orbit=earth, label="Obj")
+    pwd = os.getcwd()
```
Juanlu001: Could you use

```
with tempfile.NamedTemporaryFile(...) as fp:
    m.savefig(filename=fp.name)
```

instead? I think it's much clearer

src/poliastro/tests/test_plotting3d.py
```diff
@@ -106,7 +110,10 @@ def test_savefig_calls_prepare_plot():
     m = OrbitPlotter3D()
     earth = Orbit.from_body_ephem(Earth)
     m.plot(orbit=earth, label="Object")
+    pwd = os.getcwd()
```
Juanlu001: Same as above

src/poliastro/tests/test_plotting2d.py
shreyasbapat: Oh yes! 
And sorry I didn't know that the cell was on purpose :sweat_smile:

src/poliastro/tests/test_plotting3d.py
shreyasbapat: Ok

src/poliastro/tests/test_plotting2d.py
shreyasbapat: Should I add a markup cell explaining the same thing? That the cell is on purpose? :sweat:

Juanlu001:
Merged manually! Thanks @shreyasbapat :pray:

shreyasbapat:
Sorry for the trouble!

Issue #397, opened 2018-06-29
Adapt to Plotly 3.0


Juanlu001:
Plotly 3.0 is about to be released and it has a ton of interesting things! :tada: https://medium.com/@plotlygraphs/introducing-plotly-py-3-0-0-7bb1333f69c6

(for the moment, as explained in the installation instructions, only a release candidate is available)

In particular:

* > Just like every other object in Jupyter. plotly.offline.init_notebook_mode and plotly.offline.iplot are no longer necessary.
  - In fact, I just tried and it just works!
* A new `FigureWidget` class has been introduced, which allows for interactivity, animations...
  - Anyhow we should focus on the `Figure` class (we're implicitly using it in the form of a dictionary) because it's easier to share and persist in the notebook.
  - We should probably make the `OrbitPlotter` classes more flexible and don't hardcode a particular type of object, thus allowing the user to create a `Figure` or a `FigureWidget`
* New imperative functions, more similar to how matplotlib works
  - These might make figure updates easier
* Static image export
  - This will require installing node anyway, so for easy, publication-quality static plots we should still offer a matplotlib backend, see #338 

There's a work in progress migration guide:

https://github.com/plotly/plotly.py/blob/ipyplotly_integration/migration-guide.md

This release is going to be huge, so I propose we just add `plotly>=3.0.0` when it's on PyPI and start working from there.

Juanlu001:
As a side comment, I think this will enable a lot of interesting things:

* Animations! @aeronau perhaps you will want to take a look :wink: 
* Better workflow! We will no longer need the `from plotly.offline import init_notebook_mode; init_notebook_mode(connected=True)` part, therefore not making Plotly explicit

(We would have to check whether #281 would be fixed)

Now that the world is moving towards Jupyter Lab, we should make sure that:

* poliastro works nicely for early Jupyter Lab adopters (see #337),
* poliastro keeps working in Jupyter Notebook, which is going to be around for years, and
* poliastro is still useful in non interactive workflows, including static plotting generation

shreyasbapat:
So, my vague question, what do we start with? Should we start moving to Plotly 3.0 Pre releases?

Juanlu001:
[...]

PR #399, opened 2018-07-05
Added the plotly 3.0 Compatibility to OrbitPlotter2D

shreyasbapat:
Tested. It Works! We may need to bump the required plotly version in setup.py to get the CI build correctly.


shreyasbapat:
Addresses #397

shreyasbapat:
screenshot from 2018-07-06 12-21-51
Is there something which I am missing? Coz you can see the plots in Jupyter Lab and I can't :/

shreyasbapat:
Do I need to install the plotly extension like this: ``jupyter labextension install @jupyterlab/plotly-extension`` ? But this essentially requires user to install it too and it has prerequisites such as nodejs  and npm.

shreyasbapat:
Installing the extension worked for me. But probably we should mention this somewhere in the docs. Or install these dependencies automatically while shipping poliastro?

shreyasbapat:
I think there is a similar issue in documentation and there is some missing js extension which needs to be installed

shreyasbapat:
The plots will now be visible in the docs!

Juanlu001:
> But probably we should mention this somewhere in the docs.

Totally! At least, refer the people to the proper link. Related: #338 

>  But this essentially requires user to install it too and it has prerequisites such as nodejs and npm.

Yes... I currently use conda for that, so maybe I should update the conda packages, and for pip users we should add more info to the docs. https://github.com/conda-forge/poliastro-feedstock/issues/17

> Or install these dependencies automatically while shipping poliastro?

I don't know how to do that, for the moment let's focus on explaining the manual steps.

shreyasbapat:
So, specifically where should I add the note?

Juanlu001:
I see that we're having problems displaying the output:

https://shreyas-poliastro.readthedocs.io/en/rerun-notebooks/examples/Plotting%20in%203D.html

> Data type cannot be displayed: application/vnd.plotly.v1+json

I found a comment on an unrelated project that might be interesting:

https://github.com/fsprojects/IfSharp/issues/142#issuecomment-327116272

Could you please try again with Plotly 3.0 and the `init_notebook_mode` with `connected` both `True` and `False`? Bear in mind that you cannot refer to specific builds in Read the Docs, so take screenshots.

shreyasbapat:
Okay! On it!

shreyasbapat:
Meanwhile I found this! I don't know at what level it is relatable, but I also installed the same extension and problem occured.
https://www.npmjs.com/package/@jupyterlab/plotly-extension/v/0.16.0

shreyasbapat:
init_notebook_mode(connected=True) does the job for us!

Juanlu001:
Does `init_notebook_mode(connected=False)` work also?

shreyasbapat:
Yes, that works too!

shreyasbapat:
screencapture-shreyas-poliastro-readthedocs-io-en-rerun-notebooks-examples-plotting-in-3d-html-2018-07-07-01_59_59
This is not done :confused:

shreyasbapat:
I mean it's so inconsistent. Sometimes they load, sometimes they don't. 
Refresh the same page multiple times and you will see the plots appearing and disappearing:
https://shreyas-poliastro.readthedocs.io/en/rerun-notebooks/examples/Plotting%20in%203D.html

shreyasbapat:
One more thing. It's just my hypothesis. Works well on FIrefox, Not on chrome? Some browser extension is needed?

Juanlu001:
> Works well on FIrefox, Not on chrome? Some browser extension is needed?

I can confirm... Perhaps there's something wrong with this approach? We'll have to investigate more.

Juanlu001:
There's an intermittent error "Uncaught TypeError: Plotly.newPlot is not a function". I suspect it has to do with how the default Read the Docs HTML theme loads some asynchronous JavaScript stuff...

Juanlu001:
Just to discard the low hanging fruit, could you try if `init_notebook_mode(connected=False)` produces fewer errors?

Juanlu001:
Another idea: can you add another JS library to the first cell?

https://github.com/plotly/plotly.py/issues/1033#issuecomment-402894617

shreyasbapat:
No changes in the state by changing Connected to False. 
I will try to add another script.

shreyasbapat:
LOL! 
Adding that library removed every other plot that was being plotted a while ago! 
https://shreyas-poliastro.readthedocs.io/en/rerun-notebooks/examples/Plotting%20in%203D.html

shreyasbapat:
Though now I can see this: ``ReferenceError: require is not defined[Learn More]``

shreyasbapat:
I tried adding some other scripts as well. It doesn't work. I will investigate further.

shreyasbapat:
BTW @Juanlu001  I saw somewhere that changing the theme would do the job. What exactly will the theme change?

shreyasbapat:
Oh! We won't change the theme. Sorry

Juanlu001:
Well, at least let's see if that fixes the issue, and then we can decide what to do.

Juanlu001:
This needs a rebase on current master.

Juanlu001:
I took this PR and did some modifications (will submit them soon). It's sad that Read the Docs can't display Plotly plots correctly, but we can't forget that the important issue is to attain Plotly 3.0 compatibility. The change in the tuple arithmetic does that, and I also added `require.js` to all notebooks with Plotly plots so at least they show properly in Firefox (not in Chrome), so let's leave like that. Depending on how much time does RTD need to fix these issues, we might need to change our workflow and publish Sphinx artifacts ourselves.

Juanlu001:
Superseded by https://github.com/poliastro/poliastro/pull/399.

shreyasbapat:
Great!😄

PR #409, opened 2018-07-16
WIP:jitting Newton

nikita-astronaut:
Hello, @Juanlu001 

I jist copied the scipy Newton into our angles.py file and everything worked nice. But as I added @jit everywhere, it screwed units and now test_angles.py miserably fails. I investigated but seems like your help is needed, as you know numba better...

### DIFF
--- a/src/poliastro/jit.py
+++ b/src/poliastro/jit.py
@@ -6,6 +6,7 @@ decorator instead.
 """
 import warnings
 import inspect
+import astropy.units as u
 
 
 def ijit(first=None, *args, **kwargs):
@@ -30,3 +31,16 @@ except ImportError:
                   "algorithms will be slow. Consider installing numba to "
                   "boost performance.")
     jit = ijit
+
+
+def accel_angles(func):
+    func_fast = jit(func)
+
+    def wrapper(angle, ecc):
+        if hasattr(angle, "unit"):
+            angle = angle.to(u.rad).value
+        if hasattr(ecc, "unit"):
+            ecc = ecc.to(u.one).value
+        return func_fast(angle, ecc) * u.rad
+
+    return wrapper
--- a/src/poliastro/twobody/angles.py
+++ b/src/poliastro/twobody/angles.py
@@ -4,36 +4,7 @@
 import numpy as np
 
 from astropy import units as u
-from poliastro.jit import jit
-
-
-@jit
-def newton(regime, x0, args=(), tol=1.48e-08, maxiter=50):
-    if tol <= 0:
-        raise ValueError("tol too small (%g <= 0)" % tol)
-    if maxiter < 1:
-        raise ValueError("maxiter must be greater than 0")
-    # Multiply by 1.0 to convert to floating point.  We don't use float(x0)
-    # so it still works if x0 is complex.
-    p0 = 1.0 * x0
-    for iter in range(maxiter):
-        if regime == 'parabolic':
-            fder = _kepler_equation_parabolic(p0, *args)
-            fval = _kepler_equation_prime_parabolic(p0, *args)
-        elif regime == 'hyperbolic':
-            fder = _kepler_equation_hyper(p0, *args)
-            fval = _kepler_equation_prime_hyper(p0, *args)
-        else:
-            fder = _kepler_equation(p0, *args)
-            fval = _kepler_equation_prime(p0, *args)
-
-        newton_step = fval / fder
-        p = p0 - newton_step
-        if abs(p - p0) < tol:
-            return p
-        p0 = p
-    msg = "Failed to converge after %d iterations, value is %s" % (maxiter, p)
-    raise RuntimeError(msg)
+from poliastro.jit import jit, accel_angles
 
 
 @jit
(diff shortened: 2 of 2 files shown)


Juanlu001:
(First of all: PEP8 issues!)

There are two problems with this approach, both related to numba limitations:

1. Writing our own `newton` method and passing a function to optimize will be awfully slow because of https://github.com/numba/numba/issues/2952
2. numba functions won't respect array subclasses, and that includes `astropy.units.Unit`.

There are no easy solutions for either:

1. We can't have a separate `newton` function, so we will have to **hardcode** the iteration, like it's done here:

https://github.com/poliastro/poliastro/blob/946a46220269ae526da435c0c2607c9f1be9f6da/src/poliastro/iod/izzo.py#L254-L308

2. As done with `norm` and other functions, we will need a "fast" version and a high level version for everything to work.

Alternatives:

* Study a PhD in type inference and [extend numba to support a restricted version of some astropy types](http://numba.pydata.org/numba-doc/0.39.0/extending/interval-example.html).
* Rewrite some parts in Cython.
* Rewrite some parts in C, C++ or FORTRAN.
* Admit that Python (and therefore poliastro) will never be fast.

The first three are extremely complex and the last is not an option :stuck_out_tongue_winking_eye:

Juanlu001:
What do you think about this idea?

```python
In [29]: def accel(func):
    ...:     func_fast = jit(func)
    ...:     def wrapper(angle, ecc):
    ...:         return func_fast(angle.to(u.rad).value, ecc) * u.rad
    ...: 
    ...:     return wrapper
    ...: 

In [30]: @accel
    ...: def nu_to_E(nu, ecc):
    ...:     return 2 * np.arctan(np.sqrt((1 - ecc) / (1 + ecc)) * np.tan(nu / 2))
    ...: 

In [31]: nu_to_E((1 * u.rad).to(u.deg), 0.1 * u.one)
Out[31]: <Quantity 0.91791204 rad>

In [32]: nu_to_E((1 * u.rad), 0.1 * u.one)
Out[32]: <Quantity 0.91791204 rad>

In [33]: nu_to_E((1 * u.rad), 0.1)
Out[33]: <Quantity 0.91791204 rad>

In [34]: nu_to_E(1, 0.1)
AttributeError                            Traceback (most recent call last)
<ipython-input-34-349da992ed75> in <module>()
----> 1 nu_to_E(1, 0.1)

<ipython-input-29-d2463f20b6c7> in wrapper(angle, ecc)
      2     func_fast = jit(func)
      3     def wrapper(angle, ecc):
----> 4         return func_fast(angle.to(u.rad).value, ecc) * u.rad
      5 
      6     return wrapper

AttributeError: 'int' object has no attribute 'to'

In [35]: from poliastro.twobody.angles import nu_to_E as nu_to_E_slow

In [36]: %timeit nu_to_E_slow((1 * u.rad).to(u.deg), 0.1 * u.one)
239 µs ± 12.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [37]: %timeit nu_to_E((1 * u.rad).to(u.deg), 0.1 * u.one)
56.6 µs ± 840 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Pros: We get an order of magnitude improvement.
Cons: The function now requires units, does not work with scalars.

Juanlu001:
These are the results:

```
$ asv run 9740e4a^!
[...]
$ asv compare 19ed5ff 9740e4a

All benchmarks:

       before           after         ratio
     [19ed5ff ]       [9740e4a ]
-      69.7±0.2ms       56.7±0.1ms     0.81  examples.time_propagate_iss_one_period

```

So now `propagate(method=mean_motion)` is a 20 % faster! :tada: 

However, it's still going to be > 2x slower than `kepler`:

```
$ asv compare d32f3ab8 9740e4a

All benchmarks:

       before           after         ratio
     [d32f3ab8]       [9740e4a ]
+     24.3±0.06ms       56.7±0.1ms     2.33  examples.time_propagate_iss_one_period

```

So @nikita-astronaut I give you two options:

* You spend some more time playing with the new numba `inspect_types(pretty=True)` to see if there's something that you can optimize a bit more.
* You say "I'll optimize this further in a separate pull request", and I'll do a review and after that we can merge.

nikita-astronaut:
Let's do it in the other PR because I also want to speed up forward functions (the ones that don't use Newton)

Juanlu001:
OK, I'm merging this as an starting point, but I have some reservations. In particular, it seems that the wrapper returned by `accel_angles` cannot be used inside a numba function, which stops us from jitting more functions above it. Perhaps the approach of trying to work well with everything (i.e. `hasattr(angle, "unit")`) is the wrong approach, and we have to try something like this:

```
@jit
def M_to_nu_fast(M, ecc, delta=1e-2):
    ...

M_to_nu = wrap_units(M_to_nu_fast)

@jit
def M_to_E_fast(M, ecc):
    ...

M_to_nu = wrap_units(M_to_nu_fast)

...

def _wrap_units(func):  # inside angles.py
    @wraps(func)
    def wrapper(angle, ecc, **kwargs):
        return func(angle.to(u.rad).value, ecc.value, **kwargs)) * u.rad

```

And the `_fast` version should be used inside numba code.

Issue #382, opened 2018-06-05
Orbit.from_body_ephem returns wrong orbit for the Moon


Juanlu001:
🐞 **Problem**

The method `Orbit.from_body_ephem` returns the wrong orbit for bodies not orbiting the Sun, in particular for the Moon:

```
In [4]: Orbit.from_body_ephem(Moon)
Out[4]: 150821918 x -150822832 km x 23.5 deg orbit around Earth (♁)
```

This is because `from_body_ephem` uses `get_body_barycentric`, which returns ICRS coordinates, but then sets the `.parent` to the Earth:

https://github.com/poliastro/poliastro/blob/b659a947d9960e13ca844e7d049206e680fb0b8a/src/poliastro/twobody/orbit.py#L166-L167

🎯 **Goal**

At least for the bodies included in `poliastro.bodies`, `Orbit.from_body_ephem` should always return a correct result.

💡 **Possible solutions**

Setting the `.parent` to the Sun in the case of the Moon would be absurd, and give incorrect results for the propagation. Therefore, we have to look for other alternatives:

* Just do not allow bodies that do not orbit the Sun. `get_body_barycentric` assumes bodies from the Solar System, so this would not even work for other bodies, unless a proper ephemerides is provided.
* Make a special case for the Moon, and convert from ICRS to GCRS before returning the `Orbit`. This would increase the number of possible frames returned by poliastro to three: ICRS for most bodies, false ecliptic for NEOS, and GCRS for the Moon. Still, it's the only correct thing to do.
* ?

In the long term, it's clear that we need strong coupling between `bodies`, `Orbit.from_body_ephem`, and reference frames (see #257, #109).

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
> In the long term...

I will provide a short term fix for this next week, after the beta release.

shreyasbapat:
http://docs.astropy.org/en/stable/api/astropy.coordinates.get_body_barycentric_posvel.html

> The velocity cannot be calculated for the Moon. To just get the position, use get_body_barycentric().

Here, it is writter that we can not calculate the velocity of Moon using ``get_body_barycentric_posvel``
[...]

PR #410, opened 2018-07-17
Returning Correct Orbit of Moon (GCRS)

shreyasbapat:
`358331 x 414655 km x 20.8 deg orbit around Earth (♁)`
I have some doubts before this. Where should we include ``solar_system_ephemeris.set("jpl")``? Is it safe to keep in poliastro? Or the user must use it on their own?
Solves #382

### DIFF
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -4,7 +4,6 @@ import numpy as np
 
 from astropy import units as u
 from astropy import time
-from astropy.utils.data import conf
 
 from astropy.coordinates import CartesianRepresentation, get_body_barycentric_posvel, get_body_barycentric, \
     ICRS, GCRS, CartesianDifferential, solar_system_ephemeris
@@ -23,8 +22,6 @@ from ._base import BaseState
 
 
 ORBIT_FORMAT = "{r_p:.0f} x {r_a:.0f} x {inc:.1f} orbit around {body}"
-
-conf.remote_timeout = 10000
 
 
 class TimeScaleWarning(UserWarning):
@@ -168,8 +165,6 @@ class Orbit(object):
             warn("Input time was converted to scale='tdb' with value "
                  "{}. Use Time(..., scale='tdb') instead."
                  .format(epoch.tdb.value), TimeScaleWarning)
-        if body == Moon:
-            solar_system_ephemeris.set("jpl")
         r, v = get_body_barycentric_posvel(body.name, epoch)
         if body == Moon:
             moon_icrs = ICRS(x=r.x, y=r.y, z=r.z, v_x=v.x, v_y=v.y, v_z=v.z,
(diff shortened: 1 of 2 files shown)


shreyasbapat:
If not, I feel we must include this somewhere in the docs. I more thing that is missing to be included in docs is the plotly jupyterlab extension.

src/poliastro/twobody/orbit.py
```diff
@@ -162,7 +168,19 @@ def from_body_ephem(cls, body, epoch=None):
             warn("Input time was converted to scale='tdb' with value "
                  "{}. Use Time(..., scale='tdb') instead."
                  .format(epoch.tdb.value), TimeScaleWarning)
+        if body == Moon:
+            solar_system_ephemeris.set("jpl")
```
Juanlu001: We should don't do this: it affects the whole program without noticing, as you saw by the test failures.

src/poliastro/twobody/orbit.py
```diff
@@ -162,7 +168,19 @@ def from_body_ephem(cls, body, epoch=None):
...
         r, v = get_body_barycentric_posvel(body.name, epoch)
+        if body == Moon:
+            moon_icrs = ICRS(x=r.x, y=r.y, z=r.z, v_x=v.x, v_y=v.y, v_z=v.z,
+                             representation=CartesianRepresentation, differential_type=CartesianDifferential
+                             )
+            moon_gcrs = moon_icrs.transform_to(GCRS(obstime=epoch))
+            moon_gcrs.representation = CartesianRepresentation
+            return cls.from_vectors(
```
Juanlu001: You can avoid having two `return` statements by carefully setting the attractor, the position and velocity inside this `if` branch.

src/poliastro/twobody/orbit.py
shreyasbapat: Yeah Sure!

src/poliastro/twobody/orbit.py
shreyasbapat: Yes, I did that! Thanks for pointing this out!

shreyasbapat:
That's why, I saw multiple tests failing xD. Sorry for the trouble

shreyasbapat:
Is it still me xD?

Juanlu001:
Yes, but in this case the tests need to be update, since they rely on the old behavior I'd say.

Juanlu001:
@nikita-astronaut thoughts?

shreyasbapat:
I didn't actually understand what this means:

> the tests need to be update, since they rely on the old behavior

Can you elaborate a bit?

Juanlu001:
At least the orbit of the Moon makes sense now!

```
In [5]: solar_system_ephemeris.set("jpl")
Out[5]: <ScienceState solar_system_ephemeris: 'jpl'>

In [6]: Orbit.from_body_ephem(Moon)
Out[6]: 360195 x 409425 km x 20.8 deg orbit around Earth (♁)
```

Juanlu001:
The problem is here:

https://github.com/poliastro/poliastro/blob/da6bd151588eac259fc28dd2380c5ed467e6f915/src/poliastro/ephem.py#L10-L19

You have to make another special case there for the Moon, I think.

Juanlu001:
Also, this will need a rebase on current master.

shreyasbapat:
Something of this kind?
screenshot from 2018-07-20 01-37-23

Juanlu001:
Circle CI is having an outage https://status.circleci.com/ Will look at this again in some hours

Juanlu001:
The coverage increased even without adding a new test, but that's because the perturbations tests are already covering this. Thanks @shreyasbapat and merging!

PR #414, opened 2018-07-21
Cleanup, change structure

### LINKED ISSUE #412: Consider putting accelerated functions into separate package
🐞 **Problem**

From a coding perspective and using the tools we've used so far, it's impossible to provide functions that at the same time are fast and safe in the sense of physical units. Also, if one starts from the safe functions and decorates it to create the fast function, these can't be currently reused in other accelerated functions. See discussion in https://github.com/poliastro/poliastro/pull/409 for specific examples.

🎯 **Goal**

To simplify for future developers the creation of fast and safe functions, effectively translating to the project structure the architecture that we have been presenting recent talks about poliastro:

screenshot_2018-07-18 notebook on nbviewer

💡 **Possible solutions**

One possibility is to create a `poliastro.core` package, in a similar fashion as the layered structure of pypa/cryptography with the `hazmat` package for the "dangerous" algorithms.

```
.
├── core
│   ├── angles.py
│   ├── iod.py
│   └── [...]
├── iod
├── neos
├── tests
├── twobody
```

This package would somehow replicate the "safe" layer. The user facing functions would be imported from poliastro, whereas the functions from `poliastro.core` would be all unsafe, accelerated with numba, and ready to be called from other numba functions. For example, in `poliastro.core.angles`:

```
from poliastro.core.jit import jit  # Move it from poliastro.jit

@jit
def M_to_nu(...):
    ...

```

and then in `poliastro.twobody.angles`:

```
from poliastro.core.angles import M_to_nu as M_to_nu_fast

def M_to_nu(...):
    return M_to_nu_fast(M.to(u.rad).value, ecc.value) * u.rad

```

And the same for critical functions such as propagators, element transformations... Open questions:

* Should `poliastro._core` be private? I think it shouldn't, since the user might want to use these functions in their own fast code.
* Does `poliastro.twobody` still make sense? A good moment to think about it.
* Should we write some developer documentation as well? I think we should, but maybe it's not critical for the reorganization itself.

Other suggestions?

Juanlu001:
Fix #412:

* Clearly separate low level functions into a new `poliastro.core` package
* Make our PyPy-compatible `@jit` wrapper private

Also:

*  `@jit` some extra functions. Wohoo!
* Do a general whitespace and imports order cleanup


Juanlu001:
Life's too short to wait for docs to build at 4 AM.

PR #428, opened 2018-07-31
documenting new features

nikita-astronaut:


src/poliastro/twobody/orbit.py
```diff
@@ -281,7 +281,7 @@ def propagate(self, value, method=mean_motion, rtol=1e-10, **kwargs):
 
             return propagate(self, time_of_flight, method=method, rtol=rtol, **kwargs)
 
-    def sample(self, values=None, method=mean_motion):
+    def sample(self, values=None, method=mean_motion, dense_output=False):
```
Juanlu001: Instead of exposing `dense_output` to the user, I propose doing something else, since this is a very low level detail and can lead to confusion. For instance, they could do

```
sample(..., method=kepler, dense_output=True)
```

and it's not clear what should happen in this case. We can either:

* Make `dense_output=True/False` a property of the propagator. The simplest, although a bit ugly way, would be doing something like

```
def cowell(...):
    ...

cowell.dense_output = True
kepler.dense_output = False
```

we could also use some dictionary somewhere:

```
DENSE_OUTPUT_METHODS = [cowell]

...
if method in DENSE_OUTPUT_METHODS:
    ...
```

or turn the propagators into classes:

```
class Cowell(Propagator):
    dense_output = True
    def __call__(self, ...):
        ...

...
orbit.sample(..., method=Cowell())
```

but I find this last option the least appealing if we're going this way.

* Move the propagation loop to the propagator code, instead of having it in `Orbit._sample`. This could also be done in several ways: in particular, `cowell` already accepts a list of values in `tof` (something that the other propagators don't) so perhaps we should generalize this behavior and, if `kepler` gets a list of `tof`, do the for loop inside.

src/poliastro/core/_jit.py
```diff
@@ -6,6 +6,7 @@
 """
 import warnings
 import inspect
+import astropy.units as u
```
Juanlu001: This is not needed, right?

Juanlu001:
The tests failed because of a naming issue in the notebook

Juanlu001:
Wow, something happened and some tests are not running:

https://circleci.com/gh/poliastro/poliastro/385

Juanlu001:
Confirmed:

```
$ pytest -vvv
============================================= test session starts ==============================================
platform linux -- Python 3.6.5, pytest-3.6.3, py-1.5.4, pluggy-0.6.0 -- /home/juanlu/.miniconda36/envs/poliastro36/bin/python
cachedir: .pytest_cache
rootdir: /home/juanlu/Development/poliastro/poliastro-library, inifile: setup.cfg
plugins: cov-2.5.1
collected 421 items
[...]
$ pip install -U pytest
[...]
$ pytest -vvv
============================================= test session starts ==============================================
platform linux -- Python 3.6.5, pytest-3.7.1, py-1.5.4, pluggy-0.7.1 -- /home/juanlu/.miniconda36/envs/poliastro36/bin/python
cachedir: .pytest_cache
rootdir: /home/juanlu/Development/poliastro/poliastro-library, inifile: setup.cfg
plugins: cov-2.5.1
collected 79 items                                                  
```

Issue #434, opened 2018-08-12
Allow the user to define an Orbit in any reference frame


Juanlu001:
🐞 **Problem**

As has been pointed out many times, it's unclear which reference frame is poliastro using for `Orbit` objects, and the answer is that it's implicitly assuming an inertial reference frame centered in the `attractor` (otherwise `propagate` wouldn't work). 

🎯 **Goal**

The use of reference frames in poliastro should be more explicit and configurable. Reference frame transformations are needed all the time, and now poliastro does not make things easy on this front.

💡 **Possible solutions**

See https://github.com/poliastro/poliastro/wiki/Orbits-and-reference-frames for further explanation and ideas.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
Perhaps I don't really need to define all the frames before I implement this. It's a boring task that's mentally blocking me from making progress. So I will probably proceed like this:

1. Add the ability to use reference frames labeled as "experimental" in 0.11,
2. Add the rest of the frames and consolidate the API in 0.12

Also notice that I already requested in Astropy to have these kind of frames, see https://github.com/astropy/astropy/issues/5898.

Juanlu001:
On second thought, not adding the frames might be more difficult than adding them.

Juanlu001:
> I tried to implement frames parallel to ICRS centered in other bodies, but I was stopped by #7770.

Juanlu001:
Although we started working on this in #439, the real API for defining an `Orbit` in _any_ frame will have to wait for the next release. At least, the question "which frame does this use" should be settled.

hrishikeshgoyal:
Hi.. I am interested in looking into this.

Juanlu001:
Hi @hrishikeshgoyal, thanks for your interest! All the information is in the wiki and the issue itself. Where do you plan to start?

hrishikeshgoyal:
Hi @Juanlu001, I am new to astrodynamics.. so first exploring some related concepts.. Is there a prescribed book, blog and/or website which can help me in wrap up this?

shreyasbapat:
You can follow a book titled "Orbital Mechanics for Engineering Students"

Juanlu001:
[...]

PR #437, opened 2018-08-19
Add function to compute flybys

Juanlu001:
See discussion in #150.

I would like to request @PabloRdrRbl, @veronicasaz and @AunSiro review on the algorithm here. Also, @PabloRdrRbl @veronicasaz with this function we can find out the correct theta angle that the Parker Solar Probe uses in each flyby using some root finding method, in case you want to try.

This PR is still missing documentation and probably more tests.


Juanlu001:
I added a draft notebook to showcase how to use this functionality with the case of the Parker Solar Probe. In particular:

> The two solutions have different inclinations. We still have to find out which is the good one. We could do this by computing the inclination over the ecliptic or visualizing the geometry of the flyby, since both are known.

This means that we would benefit from some tool to visualize the flyby (as stated in #150), better reference frame conversions for `Orbit` objects (#434), or both.

Juanlu001:
Also, I **really** need to fix LaTeX output. https://github.com/astropy/astropy/pull/7420

Juanlu001:
Nice to have:

* Having a `Orbit.transform_to` method
* Have `Orbit.rv` return a `CartesianRepresentation` with differentials?
* Make plotter objects frame-aware, so the user never makes the mistake of mixing frames again
  - Idea: `OrbitPlotter.set_frame` does proper `transform_to` of the orbits
  - With some limitations: only frames in the inertial mapping are accepted for the moment

Juanlu001:
Also, @AunSiro voiced some concerns in person:

1. It's not clear what's the scope of the algorithm and its precision limits. We should document those.
2. The user can enter a periapsis distance that is smaller than the planet radius. However, the function is so low level that checking that outside of it is trivial.

Issue #370, opened 2018-05-10
Review poliastro Orbit design


Juanlu001:
After discussing #369, @pablogsal has suggested reviewing the design of `Orbit` objects, in particular the delegation of properties to `BaseState` subclasses.

The previous design consisted in one `State` object and nothing else (which is the equivalent to today's `Orbit`). `State.__init__` only accepted cartesian vectors, a non-singular, intuitive way of initializing the orbits. Therefore `State.from_elements` converted the elements to (r, v) and then created the object.

However, this is an utter waste of resources, since in the case I wanted to do

```
ss = State.from_elements(a, ecc, inc, ...)  # Remember: now it would be Orbit
ss.ecc
```

this would convert the elements to (r, v), and then convert back to (a, ecc, inc, ...) to retrieve `ss.ecc`. These problems were described in the wiki:

https://github.com/poliastro/poliastro/wiki/Different-element-sets

And in #17. The proposed solution was to split `State` in subclasses that would know how to convert among themselves and retrieve the different properties. This was accomplished in #61, but suffered from circular imports and other limitations. The rest of the story is better described here:

https://github.com/poliastro/poliastro/wiki/Simplify-State-objects

After this, in #135 I created the `Orbit` class and made it contain a subclass of `BaseState` to separate functionality and avoid circular imports.

The remaining problem is that, to make `Orbit.ecc` possible without doing `Orbit.state.ecc`, I delegate all the properties to `State`:

https://github.com/poliastro/poliastro/blob/34a9e2c83cd77e918feb0182d2fa162ba06cbd07/src/poliastro/twobody/orbit.py#L379-L399

This is less than ideal because, apart from being the root cause for #369, it makes the code magical and, in @pablogsal 's opinion, breaks the semantic relation between `Orbit`, `BaseState`, and their properties.

I open this issue to continue discussion and look for alternative designs.

Juanlu001:
Another problem derived from this design: `Orbit` properties are not discoverable, they don't autocomplete nor appear in the docs #435

Juanlu001:
@pablogsal it was not that difficult :) Feel free to comment on the result! #552

PR #439, opened 2018-08-29
Frames in Orbit objects

Juanlu001:
This is the first step in addressing https://github.com/poliastro/poliastro/issues/434. What I did here was to create reference frames centered in each of the bodies available in poliastro that are parallel to ICRS (sort of equivalents for HCRS and GCRS for the rest of the bodies) and add a `Orbit.frame` property that can compute this frame for a given `Orbit`. If the attractor is a custom body, for the moment it will raise an error.

Missing:

* [x] Validation of reference frame transformations themselves.
* [x] Proper frame in `Orbit.sample`.
* [x] Documentation.

Some notes:

* For orbits coming from ephemerides, it will set them to ICRS, so no changes on the Moon until we have a transform method.
* We still don't need to touch propagation, since we cannot set a custom frame using the public API.
* I realized that flake8 has a lot to complain about poliastro. Those fixes will go in a separate pull request to not clutter this one and help reviewers.
* I was hesitating whether to touch the `BaseState` code, but in the end I didn't because of https://github.com/poliastro/poliastro/issues/370.
* We had to do the work ourselves because of https://github.com/astropy/astropy/issues/7770.


Juanlu001:
I suspect the base report is missing because of recent Codecov incidences: https://status.codecov.io/incidents/t9tlhn0pfxl1

Juanlu001:
Coverage publishing failed:

https://circleci.com/gh/poliastro/poliastro/526

> Error: HTTP 400
Unable to locate build via CircleCI API. Please upload with the Codecov repository upload token to resolve issue.

Juanlu001:
The testing code is affected by the changes in `sample`, since we were using that to plot trajectories. I think the safest thing for now is to keep accepting a `BaseRepresentation` and give a warning.

Juanlu001:
I've been waiting for so long to have this, so I'm merging! :rocket:

hrishikeshgoyal:
Does this make sense to attache a new reference frame to `Orbit` in addition to the existing `self._frame` that can act as a communicating frame? 
For example if user needs to create an orbit using ephemerides given in some frame(say foo) other than the frame(say bar) parallel to ICRS centred at attractor and subsequently needs to call propagate. In that case following can be a sequence of steps to follow:
a. Take user inputs in foo
b. Transform inputs given in foo to bar
c. Create the orbit in the frame bar.
c. Perform all required operations in frame bar (e.g propagate). It might also require to transform (i.e. from foo to bar) the parameters passed to the operation.
d. Transform the output of the operations performed in the frame bar to foo.
e. Output the results in foo to user.

Juanlu001:
Answered in https://github.com/poliastro/poliastro/issues/434#issuecomment-455673516

Issue #442, opened 2018-09-04
Orbital state, longitude and latitude


cSchubes:
🐞 The state is completely hidden from the user, so I have no idea what it entails. What I need is to be able to propagate the orbit, and obtain the longitude and latitude of the satellite over the surface of the Earth at any given time in the orbit.

🖥 
`python -c "import poliastro.testing; poliastro.testing.test()"`

`pip freeze | grep astropy`

`pip freeze | grep poliastro`

🎯 This is important to me for my simulation work, especially since I don't know much about orbital dynamics (I'm an electrical engineer). Also, making the state clear in the docs would be super helpful.

💡 **Possible solutions**
Add state documentation and if possible, make the longitude/latitude available in that state.

📋  N/A

cSchubes:
sorry, forgot the outputs.

`pip freeze | grep astropy` :
astropy==3.0.4
`pip freeze | grep poliastro`:
poliastro==0.10.0

Tests are currently running.

cSchubes:
On this link: [http://docs.poliastro.space/en/latest/user_guide.html#defining-the-orbit-orbit-objects] in the 'From classical orbital elements' section it says 'To see a complete list of properties, check out the poliastro.twobody.orbit.Orbit class on the API reference.'

But when I follow that link to [http://docs.poliastro.space/en/latest/safe.html#poliastro.twobody.orbit.Orbit](http://docs.poliastro.space/en/latest/safe.html#poliastro.twobody.orbit.Orbit) there is not a list of what the state object contains.

Juanlu001:
This is a duplicate of https://github.com/poliastro/poliastro/issues/435 then - and I must admit I didn't remember that sentence in the user guide, it's misleading. I'm closing this one, feel free to comment further if you have more questions.

Issue #257, opened 2017-10-05
Try creating a ClassicalElementsRepresentation


Juanlu001:
By creating a `ClassicalElementsRepresentation` we could integrate better with the `astropy.coordinates` machinery, including transformation from/to other representations.

Here is the guide on how to do it:

http://docs.astropy.org/en/v2.0.2/coordinates/representations.html#creating-your-own-representations

Some problems that might arise:

* Different units for each element
* Singularities
* ?

anhiga:
It's a really good idea, but how could we deal with the gravitational parameter? It would be impossible to have a straight `from_cartesian()` transformation without knowing `k`.

Juanlu001:
One could create a `ClassicalElementsRepresentation` based on a specific `k`:

```
ThisRep = ClassicalElementsRepresentation.from_k(...)
```

Then `ThisRep` would store `k` internally and allow conversions. What do you think?

Juanlu001:
I'm postponing this until we have better validation of reference frames, see #109.

shreyasbapat:
What I understand is: 
We have to use ``CartesianRepresentation`` class to return the positions, but we are already doing that as ``_sample()``  already returns a ``CartesianRepresentation`` object. So, what are we talking about here? :confused:

shreyasbapat:
Or, probably, we are planning to create a new class which could represent  positions and name it ``ClassicalElementRepresentation`` ?

Juanlu001:
The thing is that now you can only sample position and velocity, it would be nice to sample any other thing (the true anomaly for example) and be able to convert that to a trajectory. So the idea would be to create a class that holds classical elements in an array, in the same way that `CartesianRepresentation` holds position and velocity in arrays.

Juanlu001:
Let's move this to the next milestone. @eteq said on chat that it would be very nice to see this integrated on Astropy, although the question on how to treat velocities is still open.

Juanlu001:
Postponing this to the next milestone. There's some discussion that should happen though, so it's not clear we'll go this way.

Juanlu001:
We don't really need validation of reference frames to do this: the _representation_ is contained in a _frame_, but can live independently.

Juanlu001:
The problem is that `coe2r` works, but `r2coe` doesn't: we need the complete information of the orbit to retrieve the orbital parameters.

<details>
A one-way representation would be:

```python
class ClassicalElementsRepresentation(BaseRepresentation):
[...]

PR #446, opened 2018-09-05
Frames improvements

Juanlu001:
* [x] Rework `Orbit` constructors to allow different plane and improve the experience for ecliptic orbits
* [x] Workaround for upstream HCRS conversion bug (cherry pick https://github.com/poliastro/poliastro/pull/437/commits/7ee45b7cc3c9e6eab83388eddf02c6a870fab598)
* [x] Implement `represent_as`, see https://github.com/poliastro/poliastro/issues/442#issuecomment-418410454
* [ ] ~~Make plotter objects *at least* give a warning if the frames are inconsistent~~
* [ ] ~~Make creating objects from a `CartesianRepresentation` much easier, see https://nbviewer.jupyter.org/github/poliastro/poliastro/blob/f32050a82/docs/source/examples/Analyzing%20the%20Parker%20Solar%20Probe%20flybys.ipynb#5.-Exit-orbit~~

Things that we will not do:

* `Orbit.transform_to(frame)`: it will have to wait until we have proper checks for propagation
  - But we could have a function that changes the plane instead!
* Have `Orbit.rv()` return a `CartesianRepresentation` with differentials
  - To be consistent, we would need #257 for `Orbit.coe()`
* Make plotter objects fully frame-aware: it will have to wait until we have perifocal frames


Juanlu001:
Perturbation tests are failing probably because I changed the Moon frame back to GCRS.

Juanlu001:
I ran out of time to do the rest before the beta, so I'm merging this!

PR #452, opened 2018-09-08
Miscellaneous fixes

Juanlu001:
In the absence of better tests, nothing like rerunning the notebooks to catch a couple of bugs! I also added some functions that make certain common operations easier.


Juanlu001:
Sorry about the coverage, but I'm merging this

PR #457, opened 2018-09-17
Added test for Lagrange points function

danlub:
As a continuation of previous work, I add a test for `lagrange_points_vec` in `tests/tests_threebody`.

I used the values that can be found in Figure 2.36 in Curtis, H. *Orbital Mechanics for Engineering Students*. 3rd Edition, which is the one that served as an example in the original pull request (#173 )

Figure 2.36

Also, I added the `u.quantity_input()` decorator to validate the units and corrected some small details in `restricted.py`.


Juanlu001:
Thanks for this! I don't have further comments: all the quality checks have passed and the coverage has increased, so this is going in :)

Issue #453, opened 2018-09-12
Debian packaging of poliastro 0.10.0/0.11b1


olebole:
Hi,

as one of the affiliated packages, I am currently [trying to package poliastro](https://bugs.debian.org/908659) for Debian (from there it will migrate to Ubuntu, Mint etc.). For this, I have some questions, which don't fit into your template (so I took the freedom to not use it ;-) )

The Debian package will have its own [git repository](https://salsa.debian.org/debian-astro-team/poliastro.git), which basically has the latest poliastro release, and a `debian/` subdirectory containing all Debian specific things (build and install instructions, patches etc.). I already created the first files there for initial tests.

* Do I really need numba? Numba seems to be a flaky in Debian (has a [number of RC bugs](https://bugs.debian.org/cgi-bin/pkgreport.cgi?archive=no&pend-exc=pending-fixed&pend-exc=fixed&pend-exc=done&sev-inc=critical&sev-inc=grave&sev-inc=serious&repeatmerged=no&src=numba)), so I would llike to avoid it as a strong dependency. Without numba, I get the following error (both 0.10 and 0.11b1):
```
_________________________ test_no_numba_emits_warning __________________________

recwarn = WarningsRecorder(record=True)

    def test_no_numba_emits_warning(recwarn):
>       with _fake_numba_import():

poliastro/tests/test_jit.py:54: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3.7/contextlib.py:112: in __enter__
    return next(self.gen)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    @contextmanager
    def _fake_numba_import():
[…]
>       del sys.modules['numba']
E       KeyError: 'numba'

poliastro/tests/test_jit.py:28: KeyError
```

* Do I need jplephem? This is not in Debian yet, and it seems to be used only in `test_pertubations.py` (indirectly, via `astropy.coordinates.solar_system_ephemeris`). Removing the `solar_system_ephemeris('de432s')` calls just results in three failures due to the missing ~~~Moon~~~ Pluto *(in 0.11b1)* in the default approximation (which I could easily disable).

* I get the following white space failure (both 0.10 and 0.11b1):
[...]

PR #458, opened 2018-09-19
Use default ephemeris in tests

olebole:
This is just for demonstration in #453 ...

Pluto is not part of the astropy builtin ephemeris calculation and therefore would require jplephem for astropy. This is my patch to delete them.


Juanlu001:
I might actually want to merge this, since it's probably adding testing time somewhere. Not a big deal, ~10 MB, but still...

Juanlu001:
Thanks @olebole!

PR #471, opened 2018-10-09
Define absolute tolerance when comparing with 0.0

olebole:
By default, assert_quantity_allclose() allows only a relative tolerance (1e-7), while the absolute tolerance is zero. For comparison with 0.0, this is however useless, and leads to failure on some CPU types ([arm64](https://tests.reproducible-builds.org/debian/rbuild/buster/arm64/poliastro_0.11.0-3.rbuild.log.gz), [i386](https://tests.reproducible-builds.org/debian/rbuild/buster/i386/poliastro_0.11.0-3.rbuild.log.gz)).

The value 1e-12 is chosen ad-hoc to let the test pass; it may need an adjustment.


Juanlu001:
Thanks @olebole, good catch! Merging

Issue #281, opened 2017-11-15
Plotly graphs do not show in docs


Juanlu001:
See https://github.com/spatialaudio/nbsphinx/issues/128

I wonder if there's some sort of dirty workaround we can have in place prior the release...

Juanlu001:
I'd love to have this in 0.8.0, but I doubt it will arrive in time.

Juanlu001:
In fact, this is a problem of the default RTD theme: https://github.com/rtfd/sphinx_rtd_theme/issues/328 I'm not switching the theme to accomodate the notebooks, so we will either include `require.js` in the quick 'n' dirty way or wait for an upstream fix.

Juanlu001:
This pull request fixes the issue https://github.com/rtfd/sphinx_rtd_theme/pull/467

Juanlu001:
Reopening, as it's the case again.

Juanlu001:
Being tracked at https://github.com/rtfd/sphinx_rtd_theme/pull/545, but it will probably take a while to ship.

shreyasbapat:
The PR is still open. Do you think it is a local issue and we can solve it in poliastro?

Juanlu001:
It's indeed an upstream issue that will take a while, but in the meanwhile we have a workaround, which is adding some extra cells at the top (and running them!)

```python
# Temporary hack, see https://github.com/poliastro/poliastro/issues/281
from IPython.display import HTML
HTML('<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.10/require.min.js"></script>')
```

shreyasbapat:
Oh! Okay. I will add these lines and run the notebooks again

shreyasbapat:
I thought it was not really an issue. Even I experienced this multiple times. That the plots were showing fine locally, while building in RTD stopped them. However, even Firefox does not load the plots all the time 😅

Juanlu001:
This is still an issue :( Conversations restarted on the sphinx_rtd_theme, which is where jQuery should go. An alternative would be to change Plotly to load jQuery asynchronously. In any case, I want to fix this upstream.

Juanlu001:
I spent some time setting up a test case to display the behavior, see my comment at

https://github.com/rtfd/readthedocs.org/issues/4367#issuecomment-423814712

> My current workarounds are:
> 
> * Use widgets, because they seem to work better.
[...]

Issue #480, opened 2018-10-28
OrbitPlotter2D has not method set_frame


Juanlu001:
`OrbitPlotter2D` should be as consistent as possible with `OrbitPlotter`, but it's lacking a `set_frame` method. I noticed this while trying to write a `plot_solar_system` function that uses the Plotly backend instead.

shreyasbapat:
I will do it :)

Issue #448, opened 2018-09-07
Unify plotting tests


Juanlu001:
We now have a testing module for each `OrbitPlotter` class and lots of repeated tests and it's annoying. We should parametrize them somehow (see also #447). This can be done independently of #338.

shreyasbapat:
@Juanlu001  Is this anywhere related to #481 ? Can I take this up?

Juanlu001:
Sure, if #481 becomes a blocker, I'll close it and reopen when I'm ready.

shreyasbapat:
So, we will have just one test script for all the OrbitPlotters or Plotly OrbitPlotters?

Juanlu001:
On second thought, let's leave this one because I will try to do the plotting refactor over the winter holidays, and we might repeat work unnecessarily.

Issue #447, opened 2018-09-07
Consider using fixtures in tests


Juanlu001:
There are a lot of repeated code in the tests, mainly sample data for orbits. I have been experimenting with pytest fixtures lately and I think they are a great idea. They can even be parametrized, which might help repeating the same test with different kinds of orbits.

https://docs.pytest.org/en/3.8.0/fixture.html

Juanlu001:
While reviewing this, I realized that perhaps we could take https://github.com/poliastro/poliastro/blob/master/src/poliastro/tests/tests_twobody/test_angles.py and replace `for` loops with `pytest.mark.parametrize`. And also remove this leftover `print`:

https://github.com/poliastro/poliastro/blob/5dfc7a1715243a4a9130f63493d781b80e2a2cd9/src/poliastro/tests/tests_twobody/test_angles.py#L121

Juanlu001:
I don't see many places to _replace current tests_ with fixtures. I imagine we could think about cases to _improve_ or _expand_ the tests by testing more kinds of orbits, but I am not sure whether `pytest.mark.parametrize` would be more suitable to the task.

In fact, I am going to close this issue, because it's "a solution in search of a problem" and might push us in the wrong direction.

PR #481, opened 2018-10-28
WIP: Split plotting backends, default to Plotly

Juanlu001:
Addresses #338.

Now trying to import OrbitPlotter directly from poliastro.plotting
fails, because it's the "old" backend and users are encouraged to
use OrbitPlotter2D and OrbitPlotter3D instead, which use plotly
and not matplotlib behind the scenes.

This does not address:

* Updates in documentation
* Failing test because OrbitPlotter2D has no set_frame method (#480)
* Test simplification and refactoring (#448 and #447)

Also, notice that #281 should be fixed first.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -128,6 +128,11 @@ else:
 
 # -- Options for HTML output ----------------------------------------------
 
+
+def setup(app):
+    app.add_javascript("https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js")
+    app.add_javascript("https://unpkg.com/@jupyter-widgets/html-manager@*/dist/embed-amd.js")
+
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 import sphinx_rtd_theme
--- a/docs/source/embed/molniya.html
+++ b/docs/source/embed/molniya.html
@@ -1,0 +1,1143 @@
+<script type="application/vnd.jupyter.widget-state+json">
+{
+    "version_major": 2,
+    "version_minor": 0,
+    "state": {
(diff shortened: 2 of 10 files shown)


Juanlu001:
The `OrbitPlotter*` code is so ugly (entirely my fault) that I can't wait to refactor it all :) However, for that I'd like to wait for better reference frame support: #434, which should arrive for the next release.

`OrbitPlotter2D` has incomplete redraw support (a `set_frame` call will fail with a `NotImplementedError`) but at least it now plots things properly, which it wasn't doing before (fix #483).

This PR still needs to update the documentation to be mergeable.

Juanlu001:
Also, I might want to use this opportunity to _just_ use `FigureWidget`s everywhere, forget about `init_notebook_mode`, fix #281 by accident, and start a new family of problems because of widgets handling in Jupyter notebook and lab...

Juanlu001:
Rebase after https://github.com/poliastro/poliastro/pull/487 is merged.

shreyasbapat:
There has to be lot of updates to the documentation and other places. Will you be doing it? or Can I help? :smile:

Juanlu001:
I did a test with plotly widgets with this PR and it's not ready yet: there are lots of workflows that fail, and updating the widget just does not happen. This requires some thought.

Juanlu001:
On the other hand, I tried embedding a widget in the docs inspired by ipyleaflet and it worked beautifully, at least in my local computer. However, there's a strange effect on hover:

hover-strange

Juanlu001:
> I promise this will be done before Christmas, but can't do any better.

Well... :sweat_smile: 

I added some notes about my current thinking at https://github.com/poliastro/poliastro/issues/338#issuecomment-449955855. I am closing this until I have a better idea of what to do.

PR #503, opened 2018-12-01
Fix legend layout

AntoniiaK:


Juanlu001:
Thanks a lot! Merged :rocket:

PR #504, opened 2018-12-06
Use fixtures in Tests - test_twobody and test_angles

shreyasbapat:
Let's do this step by step. I will make several different pull requests which will eventually enable usage of fixtures in every place where it is better for us to use.

### DIFF
--- a/src/poliastro/tests/test_twobody.py
+++ b/src/poliastro/tests/test_twobody.py
@@ -1,40 +1,32 @@
+import pytest
 import numpy as np
 from numpy.testing import assert_allclose
 
 from astropy import units as u
 
 from poliastro.bodies import Earth
-
 from poliastro.core.elements import coe2rv, coe2mee, rv2coe
 from poliastro.twobody.equinoctial import mee2coe
 
 
-def test_convert_between_coe_and_rv_is_transitive():
-    k = Earth.k.to(u.km**3 / u.s**2).value  # u.km**3 / u.s**2
+@pytest.fixture(scope="module")
+def expected_res():
     p = 11067.790  # u.km
     ecc = 0.83285  # u.one
     inc = np.deg2rad(87.87)  # u.rad
     raan = np.deg2rad(227.89)  # u.rad
     argp = np.deg2rad(53.38)  # u.rad
     nu = np.deg2rad(92.335)  # u.rad
-
     expected_res = (p, ecc, inc, raan, argp, nu)
+    return expected_res
 
+
+def test_convert_between_coe_and_rv_is_transitive(expected_res):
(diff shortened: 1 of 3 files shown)


Juanlu001:
If you want some early suggestions:

* Tell me about your plans, so we can discuss :)
* Start small. In the original fixtures issue it's said that "There are a lot of repeated code in the tests, mainly sample data for orbits" https://github.com/poliastro/poliastro/issues/447, so perhaps in `test_angles`, `test_twobody` or `test_propagation` there are also opportunities.
* Leave the plotting module for a separate PR, as I have some unfinished work that might affect all of it...

shreyasbapat:
Sure. I will work on the other tests. I was just trying out. I will discuss with you today. :)

shreyasbapat:
@Juanlu001 Please review whenever you get a chance. If you feel something is less than what it should be, please let me know, I will fix it. :)

src/poliastro/tests/test_twobody.py
```diff
@@ -1,40 +1,32 @@
...
 
 from poliastro.bodies import Earth
-
 from poliastro.core.elements import coe2rv, coe2mee, rv2coe
 from poliastro.twobody.equinoctial import mee2coe
 
 
-def test_convert_between_coe_and_rv_is_transitive():
-    k = Earth.k.to(u.km**3 / u.s**2).value  # u.km**3 / u.s**2
+@pytest.fixture(scope="module")
```
Juanlu001: Didn't know about `scope="module"`, nice!

https://docs.pytest.org/en/3.8.0/fixture.html#scope-sharing-a-fixture-instance-across-tests-in-a-class-module-or-session

However, I believe that they are used for expensive fixtures, so I'd rather remove it.

src/poliastro/tests/tests_twobody/test_angles.py
```diff
@@ -7,6 +7,24 @@
 from poliastro.twobody import angles
 
 
+@pytest.fixture(scope="module")
```
Juanlu001: In this particular case, let's use `ANGLES_DATA` just as a module variable, without using a fixture.

src/poliastro/tests/test_twobody.py
shreyasbapat: Sure! will do so!

src/poliastro/tests/tests_twobody/test_angles.py
shreyasbapat: Okay!

shreyasbapat:
@Juanlu001  I have rebased as well as made the suggested changes :)

Juanlu001:
Merged! Thanks for your patience :)

Issue #492, opened 2018-11-14
AssertionError in izzo library


Tarlan0001:
I got the following:
``` 
(f_v0, f_v), = izzo.lambert(Earth_k, r0, r, tof)
File "c:\users\...\site-packages\poliastro\iod\izzo.py", line 44, in lambert
    for v0, v in sols:
AssertionError
```
I printed the variables: `r0=<Quantity [-4396.85721438,  2117.6460576 ,  4400.14349892] km>`, `r=<Quantity [ 4732.15467662, -2279.13443768, -4735.69156808] km>`, `tof= 30.0 min`

Where is the issue?

Juanlu001:
Hi @Tarlan0001, thanks for your interest in poliastro! I will take a look and see what's going on, but in any case a better error should be returned.

Juanlu001:
Unfortunately I could not reproduce the issue:

```
from poliastro.bodies import Earth
from astropy import units as u
from poliastro.iod import izzo

r0 = [-4396.85721438, 2117.6460576 , 4400.14349892] * u.km
r = [ 4732.15467662, -2279.13443768, -4735.69156808]  *u.km
tof = 30 * u.min
(f_v0, f_v), = izzo.lambert(Earth.k, r0, r, tof)
print(f_v0)
print(f_v)
```

Result:

```
$ python /tmp/izzo.py 
[-1.77393825 -7.89553974 -2.80865135] km / s
[ 5.82836646  5.32284027 -1.57360116] km / s
```

Can you please give some more details about your environment? conda or pip, versions of numpy, numba and scipy, operative system, python version.

Tarlan0001:
Try this (the values for `r0` and `r` would be the same, strangely):

```
def keplerian2cartesian(kepler):

    a=(kepler[0]+kepler[1]+Req*2)*1000/2
    e=(kepler[0]-kepler[1])/(kepler[0]+kepler[1]+Req*2)
    i=kepler[2]
    OMEGA=kepler[3]
    omega=kepler[4]
    tetta=kepler[5]

    gamma=math.atan((e*math.sin(tetta))/(1+e*math.cos(tetta)))
  
    r=(a*(1-e**2))/(1+e*math.cos(tetta))
    v=math.sqrt(Earth_k.value*(2/r-1/a))
   
    X=r*( math.cos(tetta+omega)*math.cos(OMEGA) - math.sin(tetta+omega)*math.cos(i)*math.sin(OMEGA) )
[...]

PR #505, opened 2018-12-07
Add collinearity checking in Lambert solution

AntoniiaK:


Juanlu001:
Hi @AntoniiaK! I see you are fixing two issues of the ones reported in #492: collinearity and the corner case for the chord calculation. Great! Codeclimate complains about cyclomatic complexity being too high, but I don't think it's something we can avoid now, so I'm merging this.

Issue #476, opened 2018-10-22
Orbits around bodies without predefined reference frame do not have str representation


Juanlu001:
_Comes from #474, by @TimothySHamilton_

For example, for any orbit around `Moon`:

```pytb
In [32]: orbit_moon_enter = Orbit.from_vectors(Moon, r, v)

In [33]: orbit_moon_enter
Out[33]: 
In [34]: str(orbit_moon_enter)
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/.miniconda36/envs/poliastro37/lib/python3.7/site-packages/poliastro/frames.py in get_frame(attractor, plane, obstime)
    242     try:
--> 243         frames = _FRAME_MAPPING[attractor]
    244     except KeyError:

KeyError: Moon (☾)

During handling of the above exception, another exception occurred:

NotImplementedError                       Traceback (most recent call last)
<ipython-input-34-12bb16b34e33> in <module>()
----> 1 str(orbit_moon_enter)

~/.miniconda36/envs/poliastro37/lib/python3.7/site-packages/poliastro/twobody/orbit.py in __str__(self)
    357         return ORBIT_FORMAT.format(
    358             r_p=self.r_p.to(unit).value, r_a=self.r_a.to(unit), inc=self.inc.to(u.deg),
--> 359             frame=self.frame.__class__.__name__,
    360             body=self.attractor,
    361             epoch=self.epoch, scale=self.epoch.scale.upper(),

~/.miniconda36/envs/poliastro37/lib/python3.7/site-packages/poliastro/twobody/orbit.py in frame(self)
     86         """
     87         if self._frame is None:
---> 88             self._frame = get_frame(self.attractor, self._plane, self.epoch)
     89 
     90         return self._frame

~/.miniconda36/envs/poliastro37/lib/python3.7/site-packages/poliastro/frames.py in get_frame(attractor, plane, obstime)
    243         frames = _FRAME_MAPPING[attractor]
    244     except KeyError:
--> 245         raise NotImplementedError("Frames for orbits around custom bodies are not yet supported")
    246 
    247     try:

NotImplementedError: Frames for orbits around custom bodies are not yet supported
```

This should be easy to fix, having two templates for the string representation and falling back to a simpler one when there is no frame. And then, actually implementing frames for all sorts of bodies: https://github.com/poliastro/poliastro/issues/109.

AntoniiaK:
Hello! 
Do you want remove these exceptions
   ```
 try:
[...]

PR #506, opened 2018-12-15
fix orbits around bodies without predefined reference frame do not have str representation 

AntoniiaK:


Juanlu001:
Thanks! This fixes #476

PR #515, opened 2018-12-27
Update issue templates

Juanlu001:


Juanlu001:
This was only a test.

Issue #477, opened 2018-10-22
Propagator kepler crashes when propagating a hyperbolic orbit 0 seconds


Juanlu001:
_Comes from #474, by @TimothySHamilton_

There is an untested code path that would make the `kepler` propagator crash if the eccentricity is greater than 1 and the propagation time is 0:

https://github.com/poliastro/poliastro/blob/1f12764588adee6fb1bf45c625930be3312fa154/src/poliastro/core/propagation.py#L68-L71

This is similar to https://github.com/poliastro/poliastro/issues/328, but on a different level. It should be easy to fix, and requires us to add this code to `propagate`:

```python
if time_of_flight.value == 0:
    return orbit
```

We should strive to make this change for all the methods, as an efficiency measure.

aarribas:
Since it's an easy fix, may I take care of it as an entry point to contribute to the project?

Juanlu001:
Sure, go ahead! If you find any problems, just ask, here or in the chat. I
would expect a test if you are familiar with pytest, but if not we can
decide when you open a pull request.

On 24 Oct 2018 3:58 pm, "Andrés Arribas Flores" <notifications@github.com>
wrote:

Since it's an easy fix, may I take care of it as an entry point to
contribute to the project?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/poliastro/poliastro/issues/477#issuecomment-432668232>,
or mute the thread

Juanlu001:
@aarribas and I forgot to say: thanks so much! :)

aarribas:
I had in mind to include the test(s) too, of course.

aarribas:
I noticed the function documentation comments are not consistent (Kepler function has none) whereas mean_motion has some. The parameters are also inconsistently commented. May I also improve on that.  Incidentally, I had a look at  other .py in /core and noticed that a lot of doc comments where removed when the functions where "jitted". Was this intented? I found them useful.

Juanlu001:
@aarribas definitely - documentation is important to us, and if you noticed these inconsistencies we would be very happy if you fixed them as well (perhaps in separate commits though, so the review is easier). Also, we moved lots of functions to core in a rush, the day of the release, and at 3 AM local time, so I admit we probably missed lots of docstrings in the process.

Juanlu001:
Fixed in #527

Issue #474, opened 2018-10-22
Orbit propagate hangs for some r, v vectors around Earth; flips vectors around Moon


TimothySHamilton:
🐞 **Problem**
Orbit propagate hangs for some r, v vectors around Earth; flips vectors around Moon

When I propagate a orbit with certain initial r & v vectors around the Earth, the calculation hangs and will not let me stop it except by closing the terminal window.  It seems to happen for velocity vectors that are at low speed (maybe <1 km/s or so) and pointing generally inward towards the Earth.  For example, the following causes it to hang:
```    
import numpy as np
import math
from astropy import units as u
from poliastro.bodies import Earth, Moon
from poliastro.twobody import Orbit

r=[8.e3, 1.e3, 0.]*u.km
v=[-0.5, -0.5, 0.]*u.km/u.s
orbit1=Orbit.from_vectors(Earth,r,v)
orbit2=orbit1.propagate(1.*u.h)
```

On the other hand, putting similar r, v vectors around the Moon doesn't hang, but propagation turns them in different directions.  For example, given this r and v,
```
r= [ 61445.76498656  24827.93010168      0.        ] km
v= [-0.42581645 -0.18867869  0.        ] km / s
```
and defining an orbit around the moon:
```
orbit_moon_enter=Orbit.from_vectors(Moon,r,v)
r_confirm=orbit_moon_enter.r
v_confirm=orbit_moon_enter.v
print("r  =",r_confirm)
print("v  =",v_confirm)
```
gives
```
r  = [ 61445.76498656  24827.93010168      0.        ] km
v  = [-0.42581645 -0.18867869  0.        ] km / s
```
Now propagating it by time=0:
```
orbit_moon_pc_f=orbit_moon_enter.propagate(0.*u.h)
r_f=orbit_moon_pc_f.r
v_f=orbit_moon_pc_f.v
print("r=",r_f,"v=",v_f)
```
returns different vectors for r and v:
```
r= [ -6.59608255e+04   6.41698121e+03  -7.85853550e-13] km 
v= [  4.64795914e-01  -2.97331246e-02   3.64125759e-18] km / s
```
Interestingly, they have the same magnitude as the ones put into the orbit, but they point in different directions. 

[...]

PR #520, opened 2018-12-31
Solve propagation kepler for zero time of flight

IngenieroDeAviones:
I just tried to solve for #477. I added a time_of_flight checker in orbit.py and propagation.py so one can make sure that for time values of zero the actual orbit is returned.

I will also try to analyze in #474 in a couple of days.


IngenieroDeAviones:
Yep, you are right. Which file would you suggest me to change? The orbit.py one so we make sure that an unnecessary call to other function is done, maybe? 

I also added a test to check that the returned orbit is exactly the same as the one propagated.

Juanlu001:
I would leave `src/poliastro/twobody/orbit.py` as it is, and modify only `propagation.py`.

Also, there seems to be some sort of merge conflict. Let's fix it... next year :wink::tada:

PR #521, opened 2018-12-31
Add a id_type in from_horizons method and rewrite notebook

shreyasbapat:
Happy New Year! 
2019's first PR (xD)

I noticed that for major bodies, We had to pass `id_type` in `astroquery.jplhorizons.Horizons` . So I added that here.


src/poliastro/twobody/orbit.py
```diff
@@ -238,14 +238,18 @@ def from_body_ephem(cls, body, epoch=None):
         return ss
 
     @classmethod
-    def from_horizons(cls, name, epoch=None, plane=Planes.EARTH_EQUATOR):
+    def from_horizons(
+        cls, name, epoch=None, plane=Planes.EARTH_EQUATOR, id_type="smallbody"
```
Juanlu001: I don't like to expose this to the user, but the notebook doesn't work otherwise. A necessary evil, I guess!

Juanlu001:
Oops, I clicked the wrong button. I was about to merge, so I will force push to your branch myself.

src/poliastro/twobody/orbit.py
shreyasbapat: Not just notebook, user won't be able to get the orbit of `major body` if the parameter is not set.

shreyasbapat:
Cool! No issues :grin:

Juanlu001:
Merging! Thanks @shreyasbapat :grin:

Issue #514, opened 2018-12-27
New orbit creation methods


astrojuanlu:
Ideas:

* [x] Heliosynchronous orbits
* [x] Frozen orbits
* [x] Geosynchronous orbits
* [x] Geostationary orbits

Some of these need extra planetary parameters, such as J2.

hrishikeshgoyal:
@Juanlu001  I'd like to start with Geosynchronous and Geostationary orbits.

Juanlu001:
Related: #526

Juanlu001:
Geostationary orbits were added in #524. It would be nice to extend that method to allow for geosynchronous orbits as well.

ChanchalKumarMaji:
@Juanlu001 will I start working on this Issue ?????

Juanlu001:
@ChanchalKumarMaji sure, do you plan to work on which method in particular?

ChanchalKumarMaji:
I plan to start with heliosynchronous orbits. Any suggesions ?

Juanlu001:
@ChanchalKumarMaji That you take a look at https://github.com/poliastro/poliastro/pull/524 for inspiration, and find out what parameters will you need to compute that.

sashank27:
@Juanlu001 I would like to work on the Heliosynchronous orbits if no one is working on that. Also, what changes are required to add the module of geosynchronous orbits as well?

Juanlu001:
@sashank27 sure, all yours! The only changes we might need are adding new planetary parameters to `poliastro.constants`, the rest should all go to a new `Orbit` classmethod.

Juanlu001:
Sorry, I amend myself: for geosynchronous orbits we don't need a new method, just adding the inclination as parameter to geostationary.

Juanlu001:
Notice that frozen orbits were added by @Meuge at https://github.com/poliastro/poliastro/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22

Juanlu001:
(Fixed link)

Juanlu001:
Reconsidered: this issue is not at all easy.

hrishikeshgoyal:
> Sorry, I amend myself: for geosynchronous orbits we don't need a new method, just adding the inclination as parameter to geostationary.

Would we need eccentricity as well in the parameters?

MichaelRThompson:
> 
> 
[...]

PR #522, opened 2019-01-01
Added a classmethod to get geostationary orbit in Orbit class.

hrishikeshgoyal:
This addresses one of the enhancements mentioned under #514


shreyasbapat:
You might want to fix the code formatting errors. 
Run `black filename.py` to fix them and push again. :)

src/poliastro/constants.py
```diff
@@ -335,3 +335,12 @@
...
     system="si",
 )
+
+altitude_GEO = Constant(
+    "altitude_GEO",
+    "altitude of geosynchronous equatorial(geostationary) orbit",
+    3.5786e7,
+    "m",
+    1,
+    "Wikipedia",
```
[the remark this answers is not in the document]
hrishikeshgoyal: Sure, I will do that. Thanks for the review :)
`black`  Is that a package I need to install?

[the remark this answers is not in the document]
shreyasbapat: Yes! you may `pip install black` . And you may also consider a rebase to current master!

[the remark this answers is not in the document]
shreyasbapat: Also, I am not sure if we can use data from `Wikipedia` . (I have never seen any other reference to it) For that @Juanlu001 can tell. 
After this, you will also have to add a test for the method.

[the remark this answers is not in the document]
shreyasbapat: You can copy paste the code in notebook, and then call the function!

src/poliastro/constants.py
```diff
@@ -335,3 +335,13 @@
     "Howard Curtis",
     system="si",
 )
+
+altitude_GEO = Constant(
```
Juanlu001: Bear in mind that GEO altitude depends on the planet, so I would prefer not to include it as a constant.

src/poliastro/twobody/orbit.py
```diff
@@ -297,6 +297,13 @@ def circular(
             attractor, a, ecc, inc, raan, argp, arglat, epoch, plane
         )
 
+    @classmethod
+    def geostationary(cls):
```
Juanlu001: This method should allow the user to specify the attractor/body, and compute the geostationary orbit from there.

src/poliastro/constants.py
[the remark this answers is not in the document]
Juanlu001: @hrishikeshgoyal take a look at https://github.com/poliastro/poliastro/blob/master/CONTRIBUTING.rst#development-environment in case you didn't before, and use `pytest -k geostationary` to run only tests with "geostationary" in the name.

src/poliastro/constants.py
hrishikeshgoyal: Yes, to make it work for any given attractor, that cannot be a constant. Will work on it.

src/poliastro/twobody/orbit.py
hrishikeshgoyal: Definitely. Will make required changes.

hrishikeshgoyal: @Juanlu001  I believe that for this we would need to add a new attribute(or inside kwargs dict) to Body class for storing period of one complete rotation on its axis. With this additional information we can find out geostationary orbit of any given Body(if exists).

Juanlu001: OK, as long as we don't use `hasattr` or similar inside the geostationary method to check for the property. For a simpler alternative, we can pass it as an extra parameter.

hrishikeshgoyal: Okay.. I proceed with an extra parameter for now.

Juanlu001: `Body` objects should not have methods that create `Orbit`s. `Body` objects should resemble a dumb `nametuple` as much as possible - the only exception now being the rotational elements, which should go away when we improve the reference frame objects.

And, as a more practical note, it would introduce a circular dependency because `orbit.py` imports `bodies`.

hrishikeshgoyal: Yes, that makes more sense. Thanks for the clarification.

PR #523, opened 2019-01-01
Make *State objects private

Juanlu001:
One small step towards #370. These objects should have always been private.


Juanlu001:
Merging anyway.

PR #527, opened 2019-01-03
Added tof checker to propagate

IngenieroDeAviones:
Just added a few lines of code to impose the condition that if  time of flight is zero, 
 propagate function returns the same orbit to user.

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_propagation.py
+++ b/src/poliastro/tests/tests_twobody/test_propagation.py
@@ -149,10 +149,17 @@ def test_propagation_zero_time_returns_same_state():
     assert_allclose(r.value, r0.value)
     assert_allclose(v.value, v0.value)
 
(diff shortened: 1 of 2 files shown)


IngenieroDeAviones:
Sorry for the answer delay. I added the test for the hyperbolic  proposed case following the same structure of the test_tof_zero.

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -163,7 +163,6 @@ def test_propagation_hyperbolic_zero_time_returns_same_state():
     assert_allclose(r.value, r0.value)
     assert_allclose(v.value, v0.value)
 
-def test_apply_zero_maneuver_returns_equal_state():
```
Juanlu001: Oops!

PR #532, opened 2019-01-08
Simplify Orbit.sample and Orbit.propagate

AntoniiaK:


Juanlu001:
Hi @AntoniiaK, thanks for your pull request! I think this is going towards the right direction, but it needs formatting with `black` (the quality stage is failing) and probably a `git rebase` on current master. If you need a hand, ping me!

AntoniiaK:
I've done formatting.

Juanlu001:
The test collection is failing because `src/poliastro/plotting.py` is present again in the repository.

AntoniiaK:
How can I fix it?

AntoniiaK:
Need fix docs/source/examples/Natural and artificial perturbations.ipynb

src/poliastro/twobody/orbit.py
```diff
@@ -485,51 +484,45 @@ def propagate(self, value, method=mean_motion, rtol=1e-10, **kwargs):
...
+        return propagate(self, time_of_flight, method=method, rtol=rtol, **kwargs)
+
+    def propagate_to_anomaly(self, value):
+        """Propagates an orbit.
+
+        If value is true anomaly, propagate orbit to this anomaly and return the result.
+
+        Parameters
+        ----------
+        value : ~astropy.units.Quantity
```
Juanlu001:
```suggestion
        nu : ~astropy.units.Quantity
```

src/poliastro/twobody/orbit.py
```diff
@@ -485,51 +484,45 @@ def propagate(self, value, method=mean_motion, rtol=1e-10, **kwargs):
...
+            True anomaly values.
```
Juanlu001:
```suggestion
            True anomaly value.
```

src/poliastro/twobody/orbit.py
```diff
@@ -485,51 +484,45 @@ def propagate(self, value, method=mean_motion, rtol=1e-10, **kwargs):
...
         else:
-            if isinstance(value, time.Time) and not isinstance(value, time.TimeDelta):
-                time_of_flight = value - self.epoch
-            else:
-                time_of_flight = time.TimeDelta(value)
+            time_of_flight = time.TimeDelta(value)
 
-            return propagate(self, time_of_flight, method=method, rtol=rtol, **kwargs)
+        return propagate(self, time_of_flight, method=method, rtol=rtol, **kwargs)
+
```
Juanlu001: You could use `@u.quantity_input` here to specify that `nu` should have angle units.

src/poliastro/twobody/orbit.py
```diff
@@ -485,51 +484,45 @@ def propagate(self, value, method=mean_motion, rtol=1e-10, **kwargs):
...
+            self.v.to(u.km / u.s).value,
+        )
 
-    def sample(self, values=None, method=mean_motion):
+        # Compute time of flight for correct epoch
+        M = nu_to_M(self.nu, self.ecc)
+        new_M = nu_to_M(value, self.ecc)
+        time_of_flight = Angle(new_M - M).wrap_at(360 * u.deg) / self.n
+
+        return self.propagate(time_of_flight)
```
Juanlu001: Instead of calling `self.propagate` here, we already have the elements of the new orbit (`p, ecc, inc, raan, argp`, as well as the new `nu`) and the time of flight that you just computed, so we should use `from_classical` directly.

src/poliastro/twobody/orbit.py
```diff
@@ -485,51 +484,45 @@ def propagate(self, value, method=mean_motion, rtol=1e-10, **kwargs):
...
+
+    def sample(self, values=None):
```
Juanlu001: As we only have one possibility now, we can have a better default.

```suggestion
    def sample(self, values=100):
```

src/poliastro/twobody/orbit.py
```diff
@@ -550,25 +543,20 @@ def sample(self, values=None, method=mean_motion):
...
-        <GCRS Coordinate ...>
-        >>> iss.sample([0, 10, 20] * u.minute)  # doctest: +ELLIPSIS
-        <GCRS Coordinate ...>
-        >>> iss.sample([iss.epoch + iss.period / 2])  # doctest: +ELLIPSIS
-        <GCRS Coordinate ...>
 
         """
         if values is None:
-            return self.sample(100, method)
+            return self.sample(100)
```
Juanlu001: And therefore this `is values is None` block and the `else` that came after can be removed.

src/poliastro/twobody/orbit.py
```diff
@@ -550,25 +543,20 @@ def sample(self, values=None, method=mean_motion):
...
-        elif isinstance(values, int):
+        else:
             if self.ecc < 1:
                 # first sample eccentric anomaly, then transform into true anomaly
                 # why sampling eccentric anomaly uniformly to minimize error in the apocenter, see
                 # http://www.dtic.mil/dtic/tr/fulltext/u2/a605040.pdf
                 # Start from pericenter
                 E_values = np.linspace(0, 2 * np.pi, values) * u.rad
                 nu_values = E_to_nu(E_values, self.ecc)
+                time = self._generate_time_values(nu_values)
```
Juanlu001: We are repeating this line twice, perhaps we can move if after the conditional?

src/poliastro/twobody/orbit.py
```diff
@@ -578,25 +566,14 @@ def sample(self, values=None, method=mean_motion):
...
-            # A list of Times is assumed
-            values = [(value - self.epoch).sec for value in values] * u.s
-
-        return self._sample(values, method)
+            return self._sample(time)
 
-    def _sample(self, time_values, method=mean_motion):
-        positions = method(self, time_values.to(u.s).value)
+    def _sample(self, time_values):
+        positions = [self.propagate(time).r.value for time in time_values]
```
Juanlu001: This change shouldn't be necessary if we follow this:

> The propagation methods always expect a list of times

Also, calling `self.propagate` in a loop would probably be very slow.

PR #540, opened 2019-01-15
Fixed latex documentation

aOrionis:
LaTeX expressions didn't look right 
image
Modified the expressions to change that.


src/poliastro/threebody/soi.py
```diff
@@ -12,7 +12,7 @@
 The result is:
 
 .. math::
-    a\\times(\\frac{m}{M}^{\\frac{2}{5}})
+    a\\times(\\frac{m}{M})^{\\frac{2}{5}}
```
Juanlu001: While we are at it, can we remove `\times` and use `\left(` and `\right)` for the delimiters so the correct size is chosen?

src/poliastro/threebody/soi.py
```diff
@@ -23,7 +23,7 @@
 The result is:
 
 .. math::
-    a\\times(\\frac{m}{3\\times(M)}^{\\frac{1}{3}})
+    a\\times(\\frac{m}{3\\times(M)})^{\\frac{1}{3}}
```
Juanlu001: There are some useless parentheses around `M` as well.

src/poliastro/threebody/soi.py
Juanlu001: Current appearance:

aOrionis:
Are these changes ok?

Juanlu001:
> Are these changes ok?

Absolutely. Thanks again!

PR #541, opened 2019-01-17
Added Verner to integrators module

IngenieroDeAviones:
The algorithm works right in a separated script. I tried to implement it as an `OdeSolver` so it can be used easily with `scipy.integrate.solve_ivp` and therefore, with cowell's propagation algorithm.


IngenieroDeAviones:
You are right, it should be inherited from RungeKutta class. I also found more information on Tableu Methods and finally understood why all those coefficients ci, aij and bi :rofl:

Working hard on this, hope to solve it soon.

IngenieroDeAviones:
Since two methods were available on Verners web, I decided to implement the one called 'most efficient'.
I have  the following doubt:  the method is said to have n_stage = 10, this means that the maximum amount of coefficients in matrix b is 10, and also j_max = 10 for aij. But more coefficients are given: a11_j, a12_j, a13_j

Not sure how to deal with them.

Juanlu001:
Beware, because maybe the 10th order method is "too high" and it might require float96 or even symbolic arithmetic. I think 8th order is a good compromise, and also will be easier to compare with `DOP853`.

astrojuanlu:
This work continues, 3 years later, in https://github.com/scipy/scipy/pull/14956 :)

Issue #531, opened 2019-01-08
Plotting module lacks documentation


shreyasbapat:
https://docs.poliastro.space/en/latest/api/safe/plotting.html
This is completely empty! 
🐞 **Problem**

🖥 **Please paste the output of following commands**

* `conda info -a` (only if you have conda)
* `conda list` (only if you have conda)
* `pip freeze`

<details>
  
```bash
# Paste your output here:

```
</details>

* `python -c "import poliastro.testing; poliastro.testing.test()"`

<details>

```bash
# Paste your output here:

```
</details>

🎯 **Goal**

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

PR #551, opened 2019-01-28
Update plotting documentation

IngenieroDeAviones:
This PR tries to solve  #531, since there was an `rst` file that I forgot to remove in the last documentation update. It also adds the file associated to the porkchop function.


Juanlu001:
Thanks a lot! Will merge when tests pass :)

PR #556, opened 2019-01-30
Fix #549: Add a automatic issue closing bot

### LINKED ISSUE #549: Automate closing of support issues
There are some users that ask for support in the issue tracker about installation problems or general usage of the library. That is perfectly fine, but if those issues don't come from a poliastro bug or propose an enhancement, they are handled in a different way: we iterate with the user, and if they don't respond we close it.

This process should be automatic. It would be great to have a bot somewhere.

shreyasbapat:
Works when label: `more-info-needed` is present on an issue, and user does not reply in 10 days!

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,0 +1,14 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 10
+# Label requiring a response
+responseRequiredLabel: Support
+# Comment to post when closing an Issue for lack of response. Set to `false` to disable
+closeComment: >
+  This issue has been automatically closed because there has been no response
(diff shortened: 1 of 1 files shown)


shreyasbapat:
Behavior can be seen here: https://github.com/shreyasbapat/Test-No-Response/issues/1

Juanlu001:
If we could just change the label to `support` and the time to 14 days that would be great. The `more-info-needed` tag can be used for things that look like legitimate bugs, which is not exactly what I had in mind:

> There are some users that ask for support in the issue tracker about installation problems or general usage of the library. That is perfectly fine, but if those issues don't come from a poliastro bug or propose an enhancement, they are handled in a different way: we iterate with the user, and if they don't respond we close it.

(from https://github.com/poliastro/poliastro/issues/549)

Also, this will need a rebase.

shreyasbapat:
Done :+1:

Issue #335, opened 2018-03-06
Add Geocentric Solar Ecliptic (GSE) frame


Juanlu001:
🎯 **Goal**

Produce plots like this:

![screenshot-2018-3-6 esa ssa-neo coordination centre monthly newsletter - march 2018 - juanlu001 gmail com - gmail](https://user-images.githubusercontent.com/316517/37024534-6b096394-2129-11e8-8db0-aaac19aefa90.png)

And this:

https://twitter.com/planet4589/status/910720207701897221

💡 **Possible solutions**

Perhaps it's similar to this one, or even exactly this one! http://docs.astropy.org/en/stable/api/astropy.coordinates.GeocentricTrueEcliptic.html

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

shreyasbapat:
Do you think we would have to change things in our plotting module too?

Juanlu001:
If http://docs.astropy.org/en/stable/api/astropy.coordinates.GeocentricTrueEcliptic.html is really the frame that we need, this is a non issue. But it would be nice to have a `OrbitPlotter2D.set_frame` that accepts a frame and knows what to do with it, instead of the horrible `Orbit.pqw()` that we use now. Don't remember if there's an issue for that.

shreyasbapat:
> 2.1 Geocentric solar ecliptic (GSE)
This system has its X axis towards the Sun and its Z axis perpendicular to the plane of the Earth's orbit around the Sun (positive North). This system is fixed with respect to the Earth-Sun line. It is convenient for specifying magnetospheric boundaries. It has also been widely adopted as the system for representing vector quantities in space physics databases.

While, `GeocentricTrueEcliptic` says:

> Geocentric ecliptic coordinates. These origin of the coordinates are the geocenter (Earth), with the x axis pointing to the true (not mean) equinox at the time specified by the equinox attribute, and the xy-plane in the plane of the ecliptic for that date.

I don't think that's the same :/

shreyasbapat:
Also, if I look at it more closely, I don't see a difference between the code of `class BarycentricTrueEcliptic(BaseEclipticFrame)` and `class GeocentricTrueEcliptic(BaseEclipticFrame)`

[...]

PR #562, opened 2019-02-03
Add a GeocentricSolarEcliptic Frame 

shreyasbapat:
Addresses #335

### DIFF
--- a/src/poliastro/frames.py
+++ b/src/poliastro/frames.py
@@ -243,10 +243,12 @@ def _make_rotation_matrix_from_reprs(start_representation, end_representation):
     return matrix
 
 
-_EARTH_NORTH_POLE_GCRS = UnitSphericalRepresentation(lon=0 * u.deg, lat=66.5 * u.deg)
+_EARTH_ORBIT_PERPEN_POINT_GCRS = UnitSphericalRepresentation(
+    lon=0 * u.deg, lat=66.5 * u.deg
+)
 
 _EARTH_DETILT_MATRIX = _make_rotation_matrix_from_reprs(
-    _EARTH_NORTH_POLE_GCRS, CartesianRepresentation(0, 0, 1)
+    _EARTH_ORBIT_PERPEN_POINT_GCRS, CartesianRepresentation(0, 0, 1)
 )
 
 
--- a/src/poliastro/frames.py
+++ b/src/poliastro/frames.py
@@ -5,7 +5,7 @@ from enum import Enum
 from typing import Dict
 
 import numpy as np
-from astropy import _erfa, units as u
+from astropy import _erfa as erfa, units as u
 from astropy.coordinates import (
     GCRS,
     ICRS,
@@ -67,7 +67,7 @@ class HeliocentricEclipticJ2000(BaseEclipticFrame):
 
 def _ecliptic_rotation_matrix():
     jd1, jd2 = get_jd12(J2000, J2000.scale)
-    obl = _erfa.obl80(jd1, jd2) * u.radian
+    obl = erfa.obl80(jd1, jd2) * u.radian
     assert obl.to(u.arcsec).value == 84381.448
     return rotation_matrix(obl, "x")
 
@@ -244,13 +244,14 @@ def _make_rotation_matrix_from_reprs(start_representation, end_representation):
     return matrix
 
 
-_EARTH_ORBIT_PERPEN_POINT_GCRS = UnitSphericalRepresentation(
-    lon=0 * u.deg, lat=66.5 * u.deg
-)
-
-_EARTH_DETILT_MATRIX = _make_rotation_matrix_from_reprs(
-    _EARTH_ORBIT_PERPEN_POINT_GCRS, CartesianRepresentation(0, 0, 1)
-)
+def _obliquity_rotation_matrix(equinox):
+    """
+    Function to calculate obliquity of the earth.
+    This uses obl06 of erfa.
+    """
+    jd1, jd2 = get_jd12(equinox, "tt")
+    obl = erfa.obl06(jd1, jd2) * u.radian
+    return obl * u.deg
 
 
 class GeocentricSolarEcliptic(BaseEclipticFrame):
@@ -274,14 +275,22 @@ def gcrs_to_geosolarecliptic(gcrs_coo, to_frame):
             "To perform this transformation the coordinate"
             " Frame needs an obstime Attribute"
         )
+    _earth_orbit_perpen_point_gcrs = UnitSphericalRepresentation(
+        lon=0 * u.deg, lat=(90 * u.deg - -_obliquity_rotation_matrix(to_frame.obstime))
+    )
+
+    _earth_detilt_matrix = _make_rotation_matrix_from_reprs(
+        _earth_orbit_perpen_point_gcrs, CartesianRepresentation(0, 0, 1)
+    )
 
     sun_pos_gcrs = get_body("sun", to_frame.obstime).cartesian
     earth_pos_gcrs = get_body("earth", to_frame.obstime).cartesian
     sun_earth = sun_pos_gcrs - earth_pos_gcrs
 
-    sun_earth_detilt = sun_earth.transform(_EARTH_DETILT_MATRIX)
+    sun_earth_detilt = sun_earth.transform(_earth_detilt_matrix)
 
     x_axis = CartesianRepresentation(1, 0, 0)
+    # Earth-Sun Line in Geocentric Solar Ecliptic Frame
 
     if to_frame.obstime.isscalar:
         rot_matrix = _make_rotation_matrix_from_reprs(sun_earth_detilt, x_axis)
@@ -291,7 +300,7 @@ def gcrs_to_geosolarecliptic(gcrs_coo, to_frame):
         ]
         rot_matrix = np.stack(rot_matrix_list)
 
-    return matrix_product(rot_matrix, _EARTH_DETILT_MATRIX)
+    return matrix_product(rot_matrix, _earth_detilt_matrix)
 
 
 @frame_transform_graph.transform(DynamicMatrixTransform, GeocentricSolarEcliptic, GCRS)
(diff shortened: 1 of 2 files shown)


shreyasbapat:
After thinking about it! I have some questions in mind! 
In @Juanlu001  's pull request https://github.com/astropy/astropy/pull/6508 he has introduced a transformation matrix which is basically for `TrueEcliptic` or `MeanEcliptic` (That is still not clear). But transformation to GSE is very easy in way that Geocentric True Ecliptic and Geocentric solar ecliptic, both share their X-Y Axes in ecliptic plane. I just have to rotate the coordinate axes by an angle which is **angle between earth-sun and earth-equinox(true)**. 
First, how do I get that angle? 
If I don't have to get that angle, from where do I get the transformation matrix? Does ERFA has such a matrix? I assume no?

Juanlu001:
Okay, now I understand the problem.

> Does ERFA has such a matrix? I assume no?

I am not sure. I looked at it very quickly and didn't see it. We should look more closely.

> If I don't have to get that angle, from where do I get the transformation matrix?

Even if it's not in ERFA, I see some formulas here and there that use the "Sun ecliptic longitude":

https://www.spenvis.oma.be/help/background/coortran/coortran.html#T2

However, a random website is not enough and I'd prefer to look at them from some source.

shreyasbapat:
Oh! GEI is the same as Geocentric True Ecliptic! I also saw this!

Juanlu001:
This will need a rebase now

shreyasbapat:
Sorry for a looooong break. I will start working on it

shreyasbapat:
> https://www.spenvis.oma.be/help/background/coortran/coortran.html#T2

Should I go with this?

Juanlu001:
Take a look at https://docs.sunpy.org/en/stable/api/sunpy.coordinates.transformations.hcrs_to_hgs.html and https://github.com/sunpy/sunpy/blob/master/sunpy/coordinates/transformations.py#L304 first, and try to understand if we can reuse part of that transformation. Notice that we don't intend to depend on SunPy though.

shreyasbapat:
This is helpful!

shreyasbapat:
Back to it finally! Rebased and started exploring sunpy links!

src/poliastro/frames.py
```diff
@@ -220,6 +222,78 @@ class PlutoICRS(_PlanetaryICRS):
...
+    rotation_angle = -np.arccos(
+        A.dot(B) / (A.norm() * B.norm())
+    )  # negation is required
+
+    # This line works around some input/output quirks of Astropy's rotation_matrix()
+    matrix = np.array(rotation_matrix(rotation_angle, rotation_axis.xyz.value.tolist()))
+    return matrix
+
+
+_EARTH_NORTH_POLE_GCRS = UnitSphericalRepresentation(lon=0 * u.deg, lat=90 * u.deg)
```
shreyasbapat: I have some doubt here! 

Should this be: 

```
_EARTH_NORTH_POLE_GCRS = UnitSphericalRepresentation(lon=0*u.deg, lat=-66.5*u.deg)

_EARTH_DETILT_MATRIX = _make_rotation_matrix_from_reprs(_EARTH_NORTH_POLE_GCRS, CartesianRepresentation(0, 0, 1))

```
This is taking the angle from the axis that's perpendicular to orbit. But I think the first one is correct. Doing this introduces some negetive signs in the matrix.

shreyasbapat: 66.5 deg is 90- 23.5 (Tilt of the Earth Axis)

src/poliastro/frames.py
```diff
@@ -220,6 +226,76 @@ class PlutoICRS(_PlanetaryICRS):
...
+    rotation_angle = -np.arccos(
+        A.dot(B) / (A.norm() * B.norm())
+    )  # negation is required
+
+    # This line works around some input/output quirks of Astropy's rotation_matrix()
+    matrix = np.array(rotation_matrix(rotation_angle, rotation_axis.xyz.value.tolist()))
+    return matrix
+
+
+_EARTH_NORTH_POLE_GCRS = UnitSphericalRepresentation(lon=0 * u.deg, lat=66.5 * u.deg)
```
shreyasbapat: I figured out what is correct. Can you please verify this?

src/poliastro/frames.py
```diff
@@ -220,6 +226,78 @@ class PlutoICRS(_PlanetaryICRS):
...
+@frame_transform_graph.transform(DynamicMatrixTransform, GCRS, GeocentricSolarEcliptic)
+def gcrs_to_geosolarecliptic(gcrs_coo, to_frame):
+
+    if to_frame.obstime is None:
+        raise ValueError(
+            "To perform this transformation the coordinate"
+            " Frame needs an obstime Attribute"
+        )
+
+    sun_pos_icrs = get_body_barycentric("sun", to_frame.obstime)
```
shreyasbapat: And would changing the frame of this matter? As we are finding the difference vector

src/poliastro/frames.py
```diff
@@ -220,6 +227,78 @@ class PlutoICRS(_PlanetaryICRS):
...
+        A.dot(B) / (A.norm() * B.norm())
+    )  # negation is required
+
+    # This line works around some input/output quirks of Astropy's rotation_matrix()
+    matrix = np.array(rotation_matrix(rotation_angle, rotation_axis.xyz.value.tolist()))
+    return matrix
+
+
+_EARTH_ORBIT_PERPEN_POINT_GCRS = UnitSphericalRepresentation(
+    lon=0 * u.deg, lat=66.5 * u.deg
```
Juanlu001: This is called the "obliquity" and can be obtained from ERFA, see for example 

https://github.com/astropy/astropy/blob/v3.1.2/astropy/coordinates/builtin_frames/ecliptic_transforms.py#L22-L33

src/poliastro/frames.py
```diff
@@ -220,6 +227,78 @@ class PlutoICRS(_PlanetaryICRS):
...
+            " Frame needs an obstime Attribute"
+        )
+
+    sun_pos_gcrs = get_body("sun", to_frame.obstime).cartesian
+    earth_pos_gcrs = get_body("earth", to_frame.obstime).cartesian
+    sun_earth = sun_pos_gcrs - earth_pos_gcrs
+
+    sun_earth_detilt = sun_earth.transform(_EARTH_DETILT_MATRIX)
+
+    x_axis = CartesianRepresentation(1, 0, 0)
```
Juanlu001: What does this x axis represent, and in which reference frame is it written?

src/poliastro/frames.py
```diff
@@ -220,6 +227,78 @@ class PlutoICRS(_PlanetaryICRS):
...
+        raise ValueError(
+            "To perform this transformation the coordinate"
+            " Frame needs an obstime Attribute"
+        )
+
+    sun_pos_gcrs = get_body("sun", to_frame.obstime).cartesian
+    earth_pos_gcrs = get_body("earth", to_frame.obstime).cartesian
+    sun_earth = sun_pos_gcrs - earth_pos_gcrs
+
+    sun_earth_detilt = sun_earth.transform(_EARTH_DETILT_MATRIX)
```
Juanlu001: Isn't this essentially ecliptic coordinates? If I'm right, then this transformation could be made simpler by connecting it to 

https://github.com/poliastro/poliastro/blob/c8f316c362da7be277c08397b173a1f65009deab/src/poliastro/frames.py#L49-L58

src/poliastro/frames.py
shreyasbapat: So, I will try obtaining it fro ERFA!

src/poliastro/frames.py
shreyasbapat: This represents the earth-sun line in GSE

src/poliastro/frames.py
shreyasbapat: I will have a look asap

src/poliastro/frames.py
Juanlu001: Sure, let's add a code comment then :)

shreyasbapat: Sure!

src/poliastro/frames.py
shreyasbapat: I did the GCRS to GSE transformation, and I guess it's easier

src/poliastro/frames.py
Juanlu001: Shouldn't the comment be _above_ the line? :)

shreyasbapat:
Sure thing. Will do it tonight!

shreyasbapat:
Codeclimate will never pass :/ We will have to supress it.

Juanlu001:
According to Codecov, there are two untested branches:

https://codecov.io/gh/poliastro/poliastro/compare/e1b893c11764d301f584071cd50573e9a6947e93...c32b516398de55d9008c1c7d057ef3774c0e7ce1/diff

* `obstime=None`
* `obstime` not scalar

Please? :innocent:

src/poliastro/tests/test_frames.py
```diff
@@ -109,3 +112,20 @@ def test_icrs_body_position_to_planetary_frame_yields_zeros(body, frame):
         )
 
     assert_quantity_allclose(vector_result.xyz, [0, 0, 0] * u.km, atol=1e-7 * u.km)
+
+
+def test_round_trip_from_GeocentricSolarEcliptic_gives_same_results():
+    gcrs = SkyCoord("02h31m49.09s", "+89d15m50.8s", frame=GCRS(obstime=Time("J2000")))
```
Juanlu001: And also, I prefer that we avoid the use of `SkyCoord`, if possible.

shreyasbapat: Done :)

shreyasbapat:
> According to Codecov, there are two untested branches:
https://codecov.io/gh/poliastro/poliastro/compare/e1b893c11764d301f584071cd50573e9a6947e93...c32b516398de55d9008c1c7d057ef3774c0e7ce1/diff
obstime=None
obstime not scalar
Please?

Just realised, astropy infrastructure doesn't allow us to pass `None` anyways. It defaults to J2000 automatically. And neither does it allow to pass list.

Juanlu001:
:heart_eyes: Thanks a lot!

PR #568, opened 2019-02-07
Precise propagation

Juanlu001:
In #555 I introduced a minor precision loss, because we started doing arithmetic between `astropy.time.Time` and `Quantity` objects. This PR brings back the use of `astropy.time.TimeDelta` instead, which is more precise and has more explicit scale properties.

It also takes advantage of the new `Orbit.from_coords` method inside `Orbit.propagate` :muscle:


Juanlu001:
All checks pass and I'm impatient to merge, so let's get it. Doesn't bring many tests, and I suspect there are some bugs with the new functionality. Happy bug hunting :wink:

Issue #559, opened 2019-02-02
Release 0.12 checklist


Juanlu001:
(From: https://github.com/poliastro/poliastro/wiki/HOWTO-Release - Next time let's have a template, see #516)

## Preparing the release

* [x] Rerun all notebooks (pay special attention to new features)
* [x] Review documentation, optionally improving the user guide or adding new notebooks showcasing new functionality
* [x] Write the changelog and list of contributors
  - Commits since last release branch https://github.com/poliastro/poliastro/compare/0.11.x...master
  - Issues closed in this milestone https://github.com/poliastro/poliastro/milestone/12?closed=1 (note that there might be issues without milestone!)
  - Pull requests merged since last branching https://github.com/poliastro/poliastro/pulls?utf8=%E2%9C%93&q=is%3Apr+merged%3A%3E2018-11-05+base%3Amaster
* [x] Add 0.12.x branch to CI scripts

## Before the beta release

* [x] New branch
* [x] Bump version **to 0.12b1** in:
  - `README.rst`
  - `__init__.py`
  - Sphinx `conf.py`
  - AppVeyor CI script
* [x] Check that the release branch will be tested on CI
* [x] Check all the badges in `README` point to the appropriate git **branch** (replace `master` by new branch `sed -i 's/master/.../g' README.rst`)
* [x] Check that docs badges and URLs point to appropriate **tag** (replace `latest` by new tag, without sed!)
* [x] Commit
* [x] Generate sdist and bdist_wheel
* [x] `twine upload dist/* --repository-url https://test.pypi.org/legacy/`
* [x] Tag
* [x] Bump master to next development version

## Beta release

* [x] Push branch to GitHub **and tags**
* [x] Check on Test PyPI that the badges will work
* [x] Upload sdist and bdist_wheel to PyPI - *this step cannot be undone if the release is removed!* `twine upload dist/* --repository-url https://upload.pypi.org/legacy/`
* [x] Check Read the Docs (check https://docs.readthedocs.io/en/latest/webhooks.html#github first)

## Before final release

* [x] Backport any bugs
* [x] Bump version **to 0.12.0** in:
  - `README.rst`
  - `__init__.py`
  - Sphinx `conf.py`
  - AppVeyor CI script
* [x] Check that docs badges and URLs point to appropriate **tag** (replace `v0.12b1` by `v0.12.0`)
* [x] Commit
* [x] Generate sdist and bdist_wheel
[...]

PR #569, opened 2019-02-07
VHP porkchops and Jupyter Notebook example

jorgepiloto:
I added the VHP option for porkchops, which is related to the arrival velocity at target body. Also, I thought it would be interesting to have a Jupyter Notebook example on this new functionality before starting with #559.


Juanlu001:
Looking forward to rebase this one :)

jorgepiloto:
There we go! I still need to study in depth more plotting options related to `porkchops`. Also the idea for plotting this figures between other bodies such us asteroids may be interesting :rocket:

jorgepiloto:
There we go! I still need to study in depth more plotting options related to `porkchops`. Also the idea for plotting this figures between other bodies such us asteroids may be interesting :rocket:

Juanlu001:
Thanks @jorgepiloto !

PR #575, opened 2019-02-09
Changes in orbit_from_name in datacom5.py

pswaldia:
According to  `orbit_from_name`  function description
```python
"""Return :py:class:`~poliastro.twobody.orbit.Orbit` given a name.
Retrieve info from JPL DASTCOM5 database.
Parameters
----------
name : str
NEO name.
Returns
-------
orbit : list (~poliastro.twobody.orbit.Orbit)
NEO orbits.
"""
```
it returns poliastro.twobody.orbit.Orbit class . It actually returns a list of class objects that needs to be indexed while retrieving the orbit from that list in order to plot.
Plotting the orbit without indexing the value returned by orbit_from_name is producing following error. Traceback has been provided below.

Traceback (most recent call last):

*File "/Users/pswaldia1/poliastro/src/poliastro/neos/dastcom5.py",* line *595, in < module >*
*frame.plot(orbit_from_name('Atira'))*
*File "/anaconda3/envs/virtual_platform/lib/python3.5/site*
*packages/poliastro/plotting/static.py", line 216, in plot*
*self.set_frame(*orbit.pqw())*
*AttributeError: 'list' object has no attribute 'pqw'*

According to me there is no requirement of sending the list since we are only returning one orbit at a time . This pull request solves this by returning already indexed list containing a single orbit class (in the function iteself).This would be more convenient for the users as it was for me.

### DIFF
--- a/docs/source/examples/Using NEOS package.ipynb
+++ b/docs/source/examples/Using NEOS package.ipynb
@@ -302,8 +302,8 @@
     }
    ],
    "source": [
-    "atira = dastcom5.orbit_from_name(\"atira\")[0]  # NEO\n",
-    "wikipedia = dastcom5.orbit_from_name(\"wikipedia\")[0]  # Asteroid, but not NEO.\n",
+    "atira = dastcom5.orbit_from_name(\"atira\")  # NEO\n",
+    "wikipedia = dastcom5.orbit_from_name(\"wikipedia\")  # Asteroid, but not NEO.\n",
     "\n",
     "frame = OrbitPlotter2D()\n",
     "frame.plot(atira, label=\"Atira (NEO)\")\n",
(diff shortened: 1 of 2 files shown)


Juanlu001:
Hi @pswaldia, thanks for your contribution! Before reviewing anything else, one question: how do you reproduce this functionality?

```python
halleys = dastcom5.orbit_from_name("1P")

frame = OrbitPlotter2D()
frame.plot(halleys[0], label="Halley")
frame.plot(halleys[5], label="Halley")
frame.plot(halleys[10], label="Halley")
frame.plot(halleys[20], label="Halley")
frame.plot(halleys[-1], label="Halley")
```

pswaldia:
Ok ! So i didn't checked it for all the cases. A mistake on my side. I am closing this pull request.

Issue #534, opened 2019-01-09
Consider replacing NEOWS with astroquery


Juanlu001:
Ref: https://github.com/astropy/astroquery/pull/1325#issuecomment-452765939

Documentation: https://astroquery.readthedocs.io/en/latest/jplsbdb/jplsbdb.html

If all the examples in the [User Guide]() and the NEOS notebook work, with the SBDB interface, that would be great! Otherwise, there's a case for a certain degree of overlap and duplication in astroquery.

Juanlu001:
At first I felt bad about @anhiga work, but then I realized that it predated astroquery implementation by one year. We are pioneers :wink:

jorgepiloto:
Started working on this, hope to be completed at the end of the day :smile:

jorgepiloto:
This is the algorithm I came with:

```python
def from_astroquery(body_name):
    """Returns Orbit instance after query."""
    
    # Request to get body
    body = SBDB.query(body_name)    
    a = body['orbit']['elements']['a'].to(u.km) * u.km # Working 
    ecc = float(body['orbit']['elements']['e']) * u.dimensionless_unscaled # Working
    inc = body['orbit']['elements']['i'].to(u.deg) * u.deg
    raan = body['orbit']['elements']['om'].to(u.deg) * u.deg
    argp = body['orbit']['elements']['w'].to(u.deg) * u.deg
    nu = body['orbit']['elements']['ma'].to(u.deg) * u.deg
    epoch = Time(body['orbit']['epoch'].to(u.d), format='jd')

    return Orbit.from_classical(Sun, a, ecc, inc, raan, argp, nu, epoch=epoch)
```

Those weird transformations are due to transformation from the default class for each classical value (astropy CompositeUnits) to astropy Quantity.

Should this replace the whole `neos.neows` module?

Juanlu001:
Notice that you can use `u.one` instead of `u.dimensionless_unscaled`... It's my child :) https://github.com/astropy/astropy/pull/1980

Those transformations are indeed weird, but nothing too serious. I wonder if `SBDB.query` receives some extra parameters we should allow in `from_astroquery`, but apart from that this looks good.

Also, are you _sure_ that the plane of these orbits is the Earth Equator and not the Ecliptic?

shreyasbapat:
I think it will be better if we name it `from_sbdb` ? (We already have `from_horizons` which is also from astroquery)

Juanlu001:
[...]

Issue #475, opened 2018-10-22
Propagator mean_motion hangs for some r, v vectors around Earth


astrojuanlu:
_Comes from #474, by @TimothySHamilton_

The problem can be better seen by disabling Numba JIT, which shows some NumPy floating point warnings, and then turning them into errors:

<details>

```pytb
$ NUMBA_DISABLE_JIT=1 ipython --no-banner

In [1]: import numpy as np

In [2]: np.seterr(all="raise")
Out[2]: {'divide': 'warn', 'over': 'warn', 'under': 'ignore', 'invalid': 'warn'}

In [3]: import numpy as np
   ...: import math
   ...: from astropy import units as u
   ...: from poliastro.bodies import Earth, Moon
   ...: from poliastro.twobody import Orbit
   ...: 
   ...: r=[8.e3, 1.e3, 0.]*u.km
   ...: v=[-0.5, -0.5, 0.]*u.km/u.s
   ...: orbit1=Orbit.from_vectors(Earth,r,v)
   ...: orbit2=orbit1.propagate(1.*u.h)
   ...: 
   ...: 
---------------------------------------------------------------------------
FloatingPointError                        Traceback (most recent call last)
<ipython-input-3-08d7b74965c9> in <module>()
      8 v=[-0.5, -0.5, 0.]*u.km/u.s
      9 orbit1=Orbit.from_vectors(Earth,r,v)
---> 10 orbit2=orbit1.propagate(1.*u.h)

~/.miniconda36/envs/poliastro37/lib/python3.7/site-packages/poliastro/twobody/orbit.py in propagate(self, value, method, rtol, **kwargs)
    403                 time_of_flight = time.TimeDelta(value)
    404 
--> 405             return propagate(self, time_of_flight, method=method, rtol=rtol, **kwargs)
    406 
    407     def sample(self, values=None, method=mean_motion):

~/.miniconda36/envs/poliastro37/lib/python3.7/site-packages/poliastro/twobody/propagation.py in propagate(orbit, time_of_flight, method, rtol, **kwargs)
    177 
    178     """
--> 179     r, v = method(orbit, time_of_flight.to(u.s).value, rtol=rtol, **kwargs)
    180     return orbit.from_vectors(orbit.attractor, r * u.km, v * u.km / u.s, orbit.epoch + time_of_flight, orbit.plane)

[...]

PR #577, opened 2019-02-10
Astroquery SBDB for replacing NEOWS

jorgepiloto:
This PR tries to solve #534 by including a new `Orbit` constructor called `from_sbdb` that makes use of `astroquery` module to download orbital elements from JPL database.

Before including the tests, I just wanted to make sure that I placed the function in a proper location inside the project.

### DIFF
--- a/docs/source/examples/Catch that asteroid!.ipynb
+++ b/docs/source/examples/Catch that asteroid!.ipynb
@@ -120,44 +120,15 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": 7,
+   "cell_type": "raw",
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
-      "/home/juanlu/Development/poliastro/poliastro-library/src/poliastro/twobody/propagation.py:230: UserWarning:\n",
-      "\n",
-      "Frame <class 'astropy.coordinates.builtin_frames.icrs.ICRS'> does not support 'obstime', time values were not returned\n",
-      "\n"
-     ]
-    },
-    {
-     "data": {
-      "application/vnd.jupyter.widget-view+json": {
-       "model_id": "bce2f313793e4a43bddcd67f4bc61ebc",
-       "version_major": 2,
-       "version_minor": 0
-      },
-      "text/plain": [
-       "FigureWidget({\n",
-       "    'data': [{'hoverinfo': 'none',\n",
-       "              'line': {'color': 'rgb(31, 119, 180)', 'dash':…"
-      ]
-     },
-     "metadata": {},
-     "output_type": "display_data"
-    }
-   ],
    "source": [
     "earth.plot(label=Earth)"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 8,
+   "execution_count": 7,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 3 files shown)


src/poliastro/twobody/orbit.py
```diff
@@ -438,6 +439,40 @@ def from_horizons(
...
+        
+        a = body['orbit']['elements']['a'].to(u.km) * u.km 
+        ecc = float(body['orbit']['elements']['e']) * u.one 
+        inc = body['orbit']['elements']['i'].to(u.deg) * u.deg
+        raan = body['orbit']['elements']['om'].to(u.deg) * u.deg
+        argp = body['orbit']['elements']['w'].to(u.deg) * u.deg
+        nu = body['orbit']['elements']['ma'].to(u.deg) * u.deg
+        epoch = time.Time(body['orbit']['epoch'].to(u.d), format='jd')
+        
+        ss = cls.from_classical(Sun, a, ecc, inc, raan, argp, nu, epoch=epoch)
```
Juanlu001: Should the plane be `Planes.EARTH_ECLIPTIC`?

jorgepiloto:
Oh, it was my fault, sorry. I will remove the non-associated files, add the reference plane and check the jupyter example `Catch that Asteroid`

src/poliastro/twobody/orbit.py
```diff
@@ -443,6 +444,50 @@ def from_horizons(
...
+        """
+
+        obj = SBDB.query(name, full_precision=True, **kargs)
+
+        a = obj["orbit"]["elements"]["a"].to(u.AU) * u.AU
+        ecc = float(obj["orbit"]["elements"]["e"]) * u.one
+        inc = obj["orbit"]["elements"]["i"].to(u.deg) * u.deg
+        raan = obj["orbit"]["elements"]["om"].to(u.deg) * u.deg
+        argp = obj["orbit"]["elements"]["w"].to(u.deg) * u.deg
+        nu = obj["orbit"]["elements"]["ma"].to(u.deg) * u.deg
```
Juanlu001: I think this is the mean anomaly!

docs/source/examples/Catch that asteroid!.ipynb
```diff
@@ -194,17 +165,17 @@
...
-   "execution_count": 10,
+   "execution_count": 9,
    "metadata": {},
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "(<Quantity [-2.76132873e+08, -1.71570015e+08, -1.09377634e+08] km>,\n",
-       " <Quantity [13.17478674, -9.82584125, -1.48126639] km / s>)"
+       "(<Quantity [-9.80228537e+07, -2.29262614e+08, -1.01451789e+08] km>,\n",
```
Juanlu001: The data here changed a lot, see my previous comment for a possible reason.

src/poliastro/twobody/orbit.py
jorgepiloto: You are completely righ @Juanlu001, I will solve this in a moment. Sorry for the error, I'll pay more attention next time.

Juanlu001:
> Sorry for the error, I'll pay more attention next time.

Don't worry! I make these kind of errors _all the time_. I mean: *all the time*. And it makes me feel super stupid. In the end, it turns out that our unit test was not good enough, and didn't raise a red flag when we introduced this little bug. So, the only solution is to make better tests, and sometimes testing is way, way harder than implementing. I'm still learning myself :)

jorgepiloto:
Some time ago, I started hearing about Python tests. Most of the examples in the web are so simple (basic arithmetic operations most of them) and therefore I did not understood its power. Now I realize they are vital, otherwise a project may be constructed over a wrong script basis. Still working on this.

By the way: if we convert `M` to `nu`, may we had some error due to numerical computation for almost parabolic orbits? I think this can be related with #475

Juanlu001:
> if we convert M to nu, may we had some error due to numerical computation for almost parabolic orbits? I think this can be related with #475

Maybe :/ But we have to treat that issue separately

docs/source/examples/Catch that asteroid!.ipynb
jorgepiloto: I checked the Jupyter example after applying proper conversion to true anomaly and this solved immediately.

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -704,3 +707,27 @@ def test_from_coord_if_coord_is_not_of_shape_zero():
 
     assert_quantity_allclose(ss.r, pos * u.km, rtol=1e-5)
     assert_quantity_allclose(ss.v, vel * u.km / u.s, rtol=1e-5)
+
+
+def test_from_sbdb():
+
+    # Dictionary with structure: 'Obejct': [a, e, i, raan, argp, nu, epoch]
```
Juanlu001:
```suggestion
    # Dictionary with structure: 'Object': [a, e, i, raan, argp, nu, epoch]
```

Juanlu001:
Thanks a lot! This is a small step towards deprecating the poliastro.neos module.

PR #580, opened 2019-02-14
Improved documentation of core elements

jorgepiloto:
This small PR finally completes all the docstrings for the different functions available in module `core.elements`. I was studying conversions between elements, so why not `poliastro`?   ┐(￣ヮ￣)┌


Juanlu001:
Thanks a lot, merging as-is!

PR #582, opened 2019-02-18
Matrix badge

ElmerDivi:
First pull request.
Let me know if you have any feedback :)


ElmerDivi:
On a second look,i notice the use of 'flat-square' instead of 'flat' style in most of the badges.
Should I change it ?

PR #583, opened 2019-02-18
Changed badge style to flat-square

ElmerDivi:
Noticed the use of 'flat-square' instead of 'flat' style in most of the badges.
So, changed it.


Juanlu001:
Thanks very much! I'm in low connectivity so I admit I didn't do a proper review of the previous one and happily clicked the "merge" button :)

PR #584, opened 2019-02-20
Maneuver docstrings

jorgepiloto:
I am planning to add some extra manoeuvrers to this module such us non-coplanar ones and rendezvous. But for the moment, this PR add the necessary docstrings to the available `Maneuver` constructors.

### DIFF
--- a/src/poliastro/maneuver.py
+++ b/src/poliastro/maneuver.py
@@ -4,6 +4,8 @@
 import numpy as np
 from astropy import units as u
 
+from poliastro.core.elements import pqw2ijk, rv_pqw
+from poliastro.core.util import cross
 from poliastro.util import norm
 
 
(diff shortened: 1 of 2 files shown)


src/poliastro/maneuver.py
```diff
@@ -60,28 +56,101 @@ def __getitem__(self, key):
...
-        r_i = orbit_i.a
-        v_i = orbit_i.v
-        k = orbit_i.attractor.k
-        R = r_f / r_i
-        dv_a = ((np.sqrt(2 * R / (1 + R)) - 1) * v_i).decompose()
-        dv_b = (-(1 - np.sqrt(2 / (1 + R))) / np.sqrt(R) * v_i).decompose()
-        t_trans = (np.pi * np.sqrt((r_i * (1 + R) / 2) ** 3 / k)).decompose()
-        return cls((0 * u.s, dv_a), (t_trans, dv_b))
+
+        if orbit_i.ecc == 0:
```
Juanlu001: What happens with orbits that have an eccentricity of 1e-5? We could either:

* Accept some tolerance
* Provide a "circularize" maneuver

src/poliastro/maneuver.py
```diff
@@ -60,28 +56,101 @@ def __getitem__(self, key):
...
+            k = orbit_i.attractor.k
+            R = r_f / r_i
+            dv_a = ((np.sqrt(2 * R / (1 + R)) - 1) * v_i).decompose()
+            dv_b = (-(1 - np.sqrt(2 / (1 + R))) / np.sqrt(R) * v_i).decompose()
+            t_trans = (np.pi * np.sqrt((r_i * (1 + R) / 2) ** 3 / k)).decompose()
+
+            return cls((0 * u.s, dv_a), (t_trans, dv_b))
+
+        else:
+            raise ValueError(
```
Juanlu001: Apart from my other comment, this new branch we'll need a test.

src/poliastro/maneuver.py
jorgepiloto: Sure! I will add the variable for `tolerance`.  When you say 'circularize' you mean to return a `Maneuver` object that:
1. Changes `Orbit.ecc` to zero
2. After `ecc` correction, applies Hohmann maneuver

Juanlu001: I would say that circularize would:

1. Propagate until pericenter (or apocenter?)
2. Apply tangential burn

And then we would combine it with the Hohmann transfer. How does that sound?

jorgepiloto:
@Juanlu001 I have written the Hohmann manoeuvrer in terms of an initial elliptical as:

<a href="https://www.codecogs.com/eqnedit.php?latex=\Delta&space;v_{a}&space;=&space;\sqrt{\frac{2\mu}{r_{initial}}&space;-&space;\frac{\mu}{a_{trans}}}&space;-&space;v_{initial}\\&space;\Delta&space;v_{b}&space;=&space;\sqrt{\frac{\mu}{r_{final}}}&space;-&space;\sqrt{\frac{2\mu}{r_{final}}&space;-&space;\frac{\mu}{a_{trans}}}" target="_blank"></a>

Following Vallado's notation. However, when I execute the tests to make sure computations are consistent. I am getting the following exception:

```python
    def __init__(self, *args):
        r"""Constructor.
    
        Parameters
        ----------
        impulses : list
            List of pairs (delta_time, delta_velocity)
        """
    
        self.impulses = args
        # HACK: Change API or validation code
        _dts, _dvs = zip(*args)
        self._dts, self._dvs = self._initialize(
            [(_dt * u.one).value for _dt in _dts] * (_dts[0] * u.one).unit,
>          [(_dv * u.one).value for _dv in _dvs] * (_dvs[0] * u.one).unit,
        )
      TypeError: can't multiply sequence by non-int of type 'CompositeUnit'
```

Here is the code for Hohmann I have coded:

```python
r_i = norm(orbit_i.r)                                                   
v_i = orbit_i.v                                                         
k = orbit_i.attractor.k                                                 
                                                                                 
a_trans = (r_i + r_f) / 2                                               
dv_a = (np.sqrt(2 * k / r_i - k / a_trans) - v_i).decompose()           
dv_b = (np.sqrt(k / r_f) - np.sqrt(2 * k / r_f - k / a_trans)).decompose()
                                                                               
t_trans = np.pi * np.sqrt(a_trans ** 3 / k)                             
                                                                                
return  cls((0 * u.s, dv_a), (t_trans, dv_b)) 
```

I have tried also the `astropy.units.decompose()` but raises the same error. Not sure how to deal with this :sweat_smile:

PD: I still have to add a checker to propagate orbit till pericenter or apocenter

jorgepiloto:
After a little debugging, I found that the increments for de velocity `dv_a` and `dv_b` have as first component `NaN`. I will try to rewrite equations in terms of the previous value `R`. Hope to finish this tomorrow. It overwhelms me when I have a PR open more than 3 days :laughing:

Juanlu001:
> It overwhelms me when I have a PR open more than 3 days :laughing: 

Happens to the best! :stuck_out_tongue: Good to see you figured out on your own though!

jorgepiloto:
I finally realized something. Hohmann formulations are always defined in books for "2D Orbits". However, our `Orbit` objects have three dimensions. This results in some weird behaviour such us `NaN` or `Inf` results when trying to apply the Hohmann algorithm, because:

v_i = [rx_i,  ry_i, rz_i] (generally in IJK)
v_f = [final_altitude] (this is always in PQW frame, that is the reason behind planar orbits in books)

One can not operate with theses two vectors directly because v_f is measured from PQW, not from body centred frame at attractor.

I became with the following idea:

* Rotate the position and velocity vectors from [IJK] to [PQW]
* Apply Hohmann transfer (it is like working in planar orbit)
* Undo rotation, this means from [PQW] to [IJK]

This was the only solution I became with for the moment :disappointed:  What do you think about it? May be good another kind of approach to this problem?

Juanlu001:
But I think we currently use the magnitude of final radius, which does not depend on the reference frame, right?

jorgepiloto:
That is true. When working with magnitudes the increments in the velocity are computed perfectly. However, the problem is that the constructor of `Maneuver` needs all velocities to have 3-components. Therefore that magnitude of the velocity should be converted into its components.

jorgepiloto:
I was finally able to write the Hohmann maneuver for a 3D general case, since in all books it is treated orbits with inclination equal to zero. However, and although the impulses,  cost and time are right for the test case I am not sure why the final orbit eccentricity is 0.998...

astrojuanlu:
I don't remember what happened here but probably we finished this in #599.

Issue #579, opened 2019-02-12
Stop using rtd_sphinx_theme fork


Juanlu001:
Basically undo these commits https://github.com/poliastro/poliastro/compare/6019013ee5d48e416dc72944c7d61309ceba0bae...1d0fa6b80f5716897f2fa248fdc07aec2f9fbd39 because https://github.com/rtfd/sphinx_rtd_theme/pull/710 was merged.

(We would keep the `require.js` though)

Juanlu001:
Closing this one, although I think the widgets do not show in the latest documentation... Again.

PR #593, opened 2019-03-01
Revert rtd fork

jorgepiloto:
This tries to solve #579. The `.js` were keep as said in this issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ jobs:
             pip install numpy --quiet
             pip install . --quiet
             pip install .[dev] --quiet
-            pip install https://github.com/Juanlu001/sphinx_rtd_theme/archive/js-head.zip
+            pip install https://github.com/rtfd/sphinx_rtd_theme/archive/a42c4d74fec660835393a4c9b1e7cb9e654ccc90.zip
 
       - run:
           name: run tests
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,6 @@ jobs:
             pip install numpy --quiet
             pip install . --quiet
             pip install .[dev] --quiet
-            pip install https://github.com/rtfd/sphinx_rtd_theme/archive/a42c4d74fec660835393a4c9b1e7cb9e654ccc90.zip
 
       - run:
           name: run tests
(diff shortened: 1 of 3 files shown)


Juanlu001:
I don't think that particular version will work:

https://github.com/rtfd/sphinx_rtd_theme/blob/a42c4d74fec660835393a4c9b1e7cb9e654ccc90/sphinx_rtd_theme/__init__.py#L8

Because we need this change:

https://github.com/rtfd/sphinx_rtd_theme/pull/710

So we would have to use a more recent commit.

jorgepiloto:
Oh, it is true! Therefore, since those changes were merged onto `Sphinx_rdt_theme master`, should I directly point towards its repository or to the dev branch?

* https://github.com/rtfd/sphinx_rtd_theme
* https://github.com/rtfd/sphinx_rtd_theme/tree/0.4.3

Juanlu001:
Oh wow, I didn't notice there is a new release! https://github.com/rtfd/sphinx_rtd_theme/releases/tag/0.4.3 We don't need any hacks then, just put `>=0.4.3` in `setup.py`.

.circleci/config.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             pip install numpy --quiet
             pip install . --quiet
             pip install .[dev] --quiet
-            pip install https://github.com/Juanlu001/sphinx_rtd_theme/archive/js-head.zip
+            pip install https://github.com/rtfd/sphinx_rtd_theme/archive/a42c4d74fec660835393a4c9b1e7cb9e654ccc90.zip
```
Juanlu001: Not needed anymore

requirements.txt
```diff
@@ -1,4 +1,4 @@
 # Used for beta.mybinder.org and Read the Docs
 https://github.com/poliastro/poliastro/archive/master.zip
 # Fix Plotly graphs, see https://github.com/rtfd/readthedocs.org/issues/4367
-https://github.com/Juanlu001/sphinx_rtd_theme/archive/js-head.zip
+https://github.com/rtfd/sphinx_rtd_theme/archive/a42c4d74fec660835393a4c9b1e7cb9e654ccc90.zip
```
Juanlu001: Not needed

setup.py
```diff
@@ -61,7 +61,7 @@
             "pytest-cov<2.6.0",
             "pytest>=3.2",
             "sphinx",
-            # "sphinx_rtd_theme",  # Use https://github.com/Juanlu001/sphinx_rtd_theme/archive/js-head.zip
+            # "sphinx_rtd_theme",  # Use https://github.com/rtfd/sphinx_rtd_theme/tree/a42c4d74
```
Juanlu001: Bring back the normal requirement

Juanlu001:
Thanks @jorgepiloto! Merging when the tests pass :)

PR #598, opened 2019-03-13
Heliosynchronous orbit calculation added

sashank27:
Addresses #514

### DIFF
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -581,6 +581,32 @@ class Orbit(object):
         return cls.circular(attractor, altitude)
 
     @classmethod
+    @u.quantity_input(a=u.m, ecc=u.one, w=u.rad / u.s, period=u.s)
+    def heliocentric(cls, attractor, a, ecc, w=None, period=None):
+        if w is None and period is None:
+            raise ValueError("At least one among w or period must be passed")
+
+        if w is None:
+            w = 2 * np.pi / period
+
+        if a is None:
+            raise ValueError("The value of semi major axis is required")
+        mean_motion = np.sqrt(attractor.k / np.cube(a))
+        h = a * (1 - ecc ** 2)
+        theta = (-2 / 3) * ((h / attractor.k) ** 2) * (w / (mean_motion * attractor.J2))
+        inc = (180 / np.pi) * np.arccos(theta)
+
+        raan = (0 * u.deg,)
+        argp = (0 * u.deg,)
+        arglat = (0 * u.deg,)
+        epoch = (J2000,)
+        plane = Planes.EARTH_EQUATOR
+
+        return cls.from_classical(
+            attractor, a, ecc, inc, raan, argp, arglat, epoch, plane
+        )
+
+    @classmethod
     @u.quantity_input(p=u.m, inc=u.rad, raan=u.rad, argp=u.rad, nu=u.rad)
     def parabolic(
         cls, attractor, p, inc, raan, argp, nu, epoch=J2000, plane=Planes.EARTH_EQUATOR
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
       - run:
           name: install dependencies
           command: |
-            python3 -m pip install tox --quiet
+            python3 -m pip install tox
 
       - run:
           name: check quality
@@ -25,6 +25,9 @@ jobs:
 
     working_directory: ~/repo
 
+    environment:
+      PIP_PREFER_BINARY: true
+
     steps:
       - checkout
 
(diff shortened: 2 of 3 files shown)


sashank27:
@Juanlu001 please let me know how to pass all the test cases

Juanlu001:
@sashank27 can you close and reopen? I think some builds failed for no reason

Juanlu001:
Sorry, it was because of lack of llvmlite and numba wheels in Python 3.5. Should be fixed soon.

Juanlu001:
@sashank27 can you rebase on current `master`? I fixed the CI issues.

Juanlu001:
This will need some tests :)

sashank27:
Could you tell me where to get the data so as to create the tests?

Juanlu001:
I would try to look for some textbook example in Curtis or Vallado (see https://docs.poliastro.space/en/stable/references.html#books-and-papers). If not, some paper, blog article, that contains this kind of orbit. And if not, we'll have to use some external software https://github.com/poliastro/poliastro/wiki/Validation

src/poliastro/twobody/orbit.py
```diff
@@ -580,6 +580,32 @@ def geostationary(
         altitude = geo_radius - attractor.R
         return cls.circular(attractor, altitude)
 
+    @classmethod
+    @u.quantity_input(a=u.m, ecc=u.one, w=u.rad / u.s, period=u.s)
+    def heliocentric(cls, attractor, a, ecc, w=None, period=None):
```
Juanlu001: Shouldn't this be called `heliosynchronous`? Heliocentric means a different thing.

src/poliastro/twobody/orbit.py
```diff
@@ -580,6 +580,32 @@ def geostationary(
...
+        if w is None and period is None:
+            raise ValueError("At least one among w or period must be passed")
+
+        if w is None:
+            w = 2 * np.pi / period
+
+        if a is None:
+            raise ValueError("The value of semi major axis is required")
+        mean_motion = np.sqrt(attractor.k / np.cube(a))
+        h = a * (1 - ecc ** 2)
```
Juanlu001: Could you please document what's `h` and where does the `theta` and `inc` values come from?

src/poliastro/twobody/orbit.py
```diff
@@ -580,6 +580,32 @@ def geostationary(
...
+        theta = (-2 / 3) * ((h / attractor.k) ** 2) * (w / (mean_motion * attractor.J2))
+        inc = (180 / np.pi) * np.arccos(theta)
+
+        raan = (0 * u.deg,)
```
Juanlu001: I'm not sure why we're using tuples here, these probably make the code crash somewhere.

Juanlu001: That means that, apart from tests validating the numerical aspect, we would need something that at least uses the function.

src/poliastro/twobody/orbit.py
```diff
@@ -580,6 +580,32 @@ def geostationary(
...
+        argp = (0 * u.deg,)
```
Juanlu001: Also, why are we setting these values to `0`?

Juanlu001:
Example 11-2 from Vallado 3rd edition has what we need:

photo5803431081498750838

photo5803431081498750839

src/poliastro/twobody/orbit.py
jorgepiloto: May I suggest that the variable `mean_motion` could be called `n`? Since mean_motion is one of the propagation methods that `poliastro`includes, it could lead to confusion or coding errors in the future.

Juanlu001:
@sashank27 are you still working on this?

sashank27:
Sorry for the delay. I was caught up in some college work. Will be continuing the work again.

src/poliastro/twobody/orbit.py
sashank27: Sorry, a typo

src/poliastro/twobody/orbit.py
sashank27: `h` is the semi-latus rectum. `inc` is the cosine of inclination for a given value of semi-major axis, eccentricity, and time period (in radians). `theta` is the original angle in degrees.
Also, I will replace `mean_motion` by `n`.

src/poliastro/twobody/orbit.py
sashank27: I am not sure what would be the value of right ascension of the ascending node, argument of the pericenter, and anomaly. So, I took them as 0.

jorgepiloto:
This can be closed now that #774 :rocket: :smile:

Juanlu001:
Thanks for the reminder @jorgepiloto and thanks @sashank27 for giving this a try! It was a difficult pull request, both for the implementation and because of some design decisions that we must revisit. Hope to review some other contribution of yours soon!

PR #599, opened 2019-03-13
3D maneuvers and docstrings

jorgepiloto:
This PR enables to compute the different maneuvers (Hohmann and Bielliptic) no matter the inclination of the orbit. Also the corresponding docstrings have been included for each function.

I reopened due to an strange error in previous PR.

### DIFF
--- a/src/poliastro/maneuver.py
+++ b/src/poliastro/maneuver.py
@@ -4,6 +4,8 @@
 import numpy as np
 from astropy import units as u
 
+from poliastro.core.elements import pqw2ijk, rv_pqw
+from poliastro.core.util import cross
 from poliastro.util import norm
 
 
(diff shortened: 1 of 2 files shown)


Juanlu001:
The reason why Python 3.5 tests are failing is because of https://github.com/numba/llvmlite/issues/471. Options:

* Wait, because the developers said half an hour ago they are producing llvmlite and numba wheels.
* Make numba optional, and don't depend on it for Python 3.5 ("in this version your code will be slow, please upgrade")
* Temporarily limit the version of numba/llvmlite until the "problem" is fixed.

Note that this is not actually a problem: it just makes the installation more cumbersome (ah, the good old times without wheels...).

Juanlu001:
My sorry attempt to fix this in a quick way failed. Will continue in another PR, sorry for the noise.

src/poliastro/maneuver.py
```diff
@@ -60,41 +58,159 @@ def __getitem__(self, key):
...
-        dv_a = ((np.sqrt(2 * Rs / (1 + Rs)) - 1) * v_i).decompose()
-        dv_b = (
-            -np.sqrt(2 / Rs) * (np.sqrt(1 / (1 + Rs / R)) - np.sqrt(1 / (1 + Rs))) * v_i
-        ).decompose()
-        dv_c = (-(np.sqrt(2 * Rs / (R + Rs)) - 1) / np.sqrt(R) * v_i).decompose()
-        t_trans1 = (np.pi * np.sqrt((r_i * (1 + Rs) / 2) ** 3 / k)).decompose()
-        t_trans2 = (np.pi * np.sqrt((r_i * (R + Rs) / 2) ** 3 / k)).decompose()
+        r_i = orbit_i.r
+        v_i = orbit_i.v
+        h_i = norm(cross(r_i.to(u.m), v_i.to(u.m / u.s)) * u.m ** 2 / u.s)
```
Juanlu001:
```suggestion
        h_i = norm(cross(r_i.to(u.m), v_i.to(u.m / u.s)))
```

jorgepiloto: I tried to supress those units as suggested but then the tests did not passed, since the following exception is raised:

```
AttributeError                            Traceback (most recent call last)
<ipython-input-15-1ae6b7d64845> in <module>
      1 orbit_i = Orbit.circular(Earth, alt=191.34411 * u.km, inc=0*u.deg)
----> 2 hohmann(orbit_i, r_f=35781.34857*u.km + Earth.R)

<ipython-input-14-a8a7f89cc4a4> in hohmann(orbit_i, r_f)
     29 
     30     #h_i = norm((cross(r_i.to(u.m), v_i.to(u.m / u.s)))) * u.m ** 2 / u.s
---> 31     h_i = norm(cross(r_i.to(u.m), v_i.to(u.m / u.s)))
     32     print(h_i)
     33     p = h_i ** 2 / k.to(u.m**3 / u.s**2)

~/Documentos/Github/poliastro/src/poliastro/util.py in norm(vec)
     79 
     80     """
---> 81     return norm_fast(vec.value) * vec.unit
     82 
     83 

AttributeError: 'numpy.ndarray' object has no attribute 'value'
```

Juanlu001: Oops, perhaps `* u.m ** 2 / u.s` has to be _outside_ of the parentheses?

jorgepiloto: There we go! The thing was with the `core` functions. Do you think it would be interesting to implement them `(rv_pqw, pqw2ijk, cross)` in the high level api, like for example the `util.norm` function?

Juanlu001: Maybe! Let's leave that for a future PR :)

Juanlu001:
Thanks a lot @Jorgepiloto! It took a lot of time but it was worth it 🎉

Juanlu001:
@jorgepiloto I notice we didn't add any new tests to this pull request. Can you please find an example of a Hohmann transfer that could not be done before it, that now works?

PR #601, opened 2019-03-15
Adds basic CZML extractor

Sedictious:
This is a draft CZML extractor. It currently can:

* Generate the orbital CZML data for Cesium's Sandcastle
* Edit the label, path and id details (text, fonts, colours)
* Allows for an arbitary number of orbits to be defined

**Update**: 
* The reference frames now default to inertial

Inertial


src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,181 @@
+import numpy as np
+import json
+
+from astropy.coordinates import CartesianRepresentation
+from astropy import units as u
+
+from poliastro.contrib.czml_extract_default_params import PIC_SATELLITE, DEFAULTS
+
+
+class ExtractorCZML:
```
Juanlu001: What about `CZMLExtractor` instead?

src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,181 @@
...
+        self.orbits = {1: [orbit, N, orbit.period, orbit.epoch]}
+
+        self.i = 1  # Current index id, used for insertion of new elements
+
+        self.fixed_rf = fixed_rf
+        self.start_epoch = ExtractorCZML.format_date(orbit.epoch.iso)
+        self.end_epoch = ExtractorCZML.format_date((orbit.epoch + orbit.period).iso)
+        self.init_czml()
+
+        return None
```
Juanlu001: All functions `return None` by default, you can remove this line

src/poliastro/contrib/extract_czml.py
Sedictious: It was initially ``CZMLExtractor``, but I though the camel case made it more readable that way. However, I have no problem to change it back if you think it'd be better

Sedictious:
@Juanlu001 Thanks for the review! If I understand correctly, Code climate fails due to "high cyclomatic complexity", which I take refers to the many unnested if statements (if I'm not mistaken, it can have 2^6 different outcomes). Now the solution would be to either "break" the function into two smaller ones or omit a variable altogether.

I'm a bit lost as to why circleci fails. Is it related to the project's dependencies?

Juanlu001:
Don't worry about codeclimate for the moment, we can think of ways of simplifying that method in the future (or just ignore it, which is what we do 80 % of the times).

Regarding circleci: we use [isort](https://github.com/timothycrosley/isort) to sort the `import` statements in a consistent way throughout the project, and it's telling you that the order is wrong:

```
isort --check-only --diff --recursive --project poliastro --section-default THIRDPARTY src setup.py
ERROR: /root/repo/src/poliastro/contrib/extract_czml.py Imports are incorrectly sorted.
--- /root/repo/src/poliastro/contrib/extract_czml.py:before	2019-03-15 16:16:25.732628
+++ /root/repo/src/poliastro/contrib/extract_czml.py:after	2019-03-15 16:16:37.236302
@@ -1,9 +1,9 @@
+import json
+
 import numpy as np
-import json
-
+from astropy.coordinates import CartesianRepresentation
 from astropy.time import Time
-from astropy.coordinates import CartesianRepresentation
-from astropy import units as u
 
 from poliastro.contrib.czml_extract_default_params import DEFAULTS
 
```

Instead of trying to follow that manually, you can apply `isort` yourself using `isort --recursive --project poliastro --section-default THIRDPARTY src setup.py` (or, if you have poliastro installed in your development environment, which is the usual thing to do, just `isort --recursive src`).

Sedictious:
@Juanlu001 Thanks I reformatted and now everything passes except for codeclimate and codecov. Regarding the code coverage, I'm planning on writing tests relatively soon. I'm still undecided about using nested dictionaries for JSON as it seems a bit inelegant, but it's the only efficient and comprehensible way I could come up with.

Juanlu001:
> Thanks I reformatted and now everything passes except for codeclimate and codecov.

Excellent!

> I'm still undecided about using nested dictionaries for JSON as it seems a bit inelegant, but it's the only efficient and comprehensible way I could come up with.

I don't have a strong opinion against it, but I would still need to review the code in depth. Give me one day to do it and I will write more meaningful comments and confirm the approach.

Juanlu001:
I am in process of reviewing this code, and I find it a bit difficult to use. I tried to create a class and then paste its `.czml` attribute in the Cesium Sandcastle, but it was somehow malformed and I had to slightly change its structure to make it work (namely, convert the result from a dictionary with keys `0` and `1` to a list of dictionaries). Then I found the `extract` method, which does the right thing. But in general, the class has too many public methods which I suppose are not normally used?

I propose making everything that won't be used by end users private, so perhaps leave only `extract`, `add_orbit` and `del_orbit` (rename to `remove_orbit`?)

Now I'm making some comments in the code.

src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,288 @@
...
+                ("interpolationAlgorithm", "LAGRANGE"),
+                ("interpolationDegree", 5),
+                ("referenceFrame", "FIXED"),
+                ("epoch", start_epoch),
+                ("cartesian", list()),
+            ],
+        )
+        self.init_orbit_packet_cords(i)
+
+    def init_orbit_packet_cords(self, i):
```
Juanlu001: Probably this whole function should leverage `Orbit.sample`

src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,288 @@
...
+                .represent_as(CartesianRepresentation)
+                .xyz.to(u.meter)
+                .value
+            )
+            cords = np.insert(cords, 0, h.value * k, axis=0)
+
+            self.czml[i]["position"]["cartesian"] += cords.tolist()
+            self.orbits[i][0] = self.orbits[i][0].propagate(h)
+
+    def init_czml(self):
```
Juanlu001: This should be private

src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,288 @@
...
+    def init_orbit_packet(self, i):
+        """
+        Sets the default values for a given orbit
+
+        Parameters
+        ----------
+        i : int
+            Index of referenced orbit
+        """
+        for t_key, t_val in DEFAULTS:
```
Juanlu001: Instead of storing the `DEFAULTS` as a list of tuples externally and then using `parse_dict_tuples` on it, wouldn't it be simpler to have some default `._czml` and then add properties to it?

src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,288 @@
...
+            Ending epoch
+        N: int
+            Number of sample points
+        fixed_rf: bool
+            Determines whether the reference frame is fixed
+            When set to false, it assumes the frame moving
+            with constant velocity.
+        """
+        self.czml = dict()
+        self.orbits = {1: [orbit, N, orbit.period, orbit.epoch]}
```
Juanlu001: Why do we need to store the period and epoch of the orbit? We have the `orbit` object itself, after all.

Juanlu001: Also, do we need `orbits` to be a dictionary? Or can it be a list?

src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,288 @@
...
+        ----------
+        date : str
+            date of the form "yyyy-mm-dd hh:mm:ss.ssss"
+
+        Returns
+        -------
+        formatted_date : str
+            date of the form "yyyy-mm-ddThh:mm:ssZ"
+        """
+        return date[:10] + "T" + date[11:-4] + "Z"
```
Juanlu001: Let's use `.strftime` instead: http://strftime.org/

src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,288 @@
...
+            orbit = orbit.propagate(Time(self.start_epoch) - orbit.epoch)
+        elif orbit.epoch > Time(self.end_epoch):
+            raise ValueError(
+                "The orbit's epoch cannot exceed the constructors ending epoch"
+            )
+
+        self.orbits[self.i] = [orbit, N, orbit.period, orbit.epoch]
+        self.init_orbit_packet(self.i)
+
+    def del_orbit(self, i):
```
Juanlu001: I propose that we remove this method. If someone wants to "delete an orbit", they should create another extractor class. Maybe this makes the rest of the code simpler.

src/poliastro/contrib/extract_czml.py
Sedictious: That's a really great idea and maybe far more intuitive. I initially had the idea of having some sort of choice between ``DEFAULTS`` but it didn't pan out.

src/poliastro/contrib/extract_czml.py
Sedictious: The problem with the current version of the code is that it has many leftover scribbles for things I'd like to implement next. 
I had this idea of adjusting the epoch of the timeline to the period of the orbit. As for storing the period, we don't need it but I think it would be a bit more readable than calling ``orbits[0].period`` then again... maybe not.

I used a dictionary for defining ``orbits`` as a way to quickly ID each orbit. The same thing could theoretically work with a list IF we decide on not deleting objects. Otherwise, using a list would be just too inefficient.

src/poliastro/contrib/extract_czml.py
Sedictious: I tend to agree, but I'm a bit reluctant to introduce this level of immutability, especially when the only alternative to defining one wrong orbit is starting all over again.

src/poliastro/contrib/extract_czml.py
Juanlu001: Yeah, let's change it to `CZMLExtractor`, the other name feels a bit backwards

src/poliastro/contrib/extract_czml.py
Juanlu001: It's true, but if you think about the workflow, most people will use editors like Spyder, Jupyter notebooks, Jupyter Lab... I don't think it's a big burden to change one or two lines of code, and from the implementation side it simplifies things a lot.

src/poliastro/contrib/extract_czml.py
Juanlu001: > it has many leftover scribbles for things I'd like to implement next.

KISS and YAGNI!

https://www.wikiwand.com/en/KISS_principle
https://www.wikiwand.com/en/You_aren%27t_gonna_need_it

If it's not needed now, let's dump it :) There's always time to add it later when we realize "oh, now I need this kind of functionality, and the current implementation is limiting me". Let's strive for a simple implementation now **with good test coverage**: that will allow us to refactor later and assure that we are not breaking present functionality.

I prefer `orbits[0].period` than troublesome redundancy ("does this period _really_ correspond to this orbit?"). And yeah, I think we can remove the `del_orbit` functionality.

Sedictious: Ah yes... I'm afraid I'm guilty of somewhat neglecting both practices... Anyway, thanks for the insightful comments. I'll start working on the changes as soon as I get back home!

Sedictious:
@Juanlu001 I fixed most of the things you mentioned (most notably, ``orbits`` is now a list, I removed the ability to delete orbits and made ``DEFAULTS`` into a dictionary which is simply copied when initializing a new orbit. 

What I didn't implement (yet):

- Changing the way path/text parameters are changed:

I feel like loading up the initializer with so many parameters would be a bit overwhelming. Maybe we could add the "basic" parameters (name, path colour) and keep the other methods as a way to add even more customization options.

- Using ``Orbit.sample()``:

Admittedly, I didn't look too much into the source, but I didn't find the way to set the time values

- Extract the data to a file

Would a simple ``f.write(path)`` suffice?

Juanlu001:
> I feel like loading up the initializer with so many parameters would be a bit overwhelming. Maybe we could add the "basic" parameters (name, path colour) and keep the other methods as a way to add even more customization options.

Perhaps, but take into account that there's *lots* of parameters in stuff like pandas plotting, or even matplotlib. The tradeoff is: more methods with fewer parameters, or fewer methods with more parameters. I lean towards the latter, to imitate what others do.

> Admittedly, I didn't look too much into the source, but I didn't find the way to set the time values

Oops, you're right! `Orbit.sample` takes a fixed number of points now. The way of doing it now would be:

```python
from poliastro.twobody.propagation import propagate

positions = propagate(
    iss,
    time.TimeDelta(...),
)
```

(see for example these notebooks [[1]](https://docs.poliastro.space/en/stable/examples/Propagation%20using%20Cowell's%20formulation.html), [[2]](https://docs.poliastro.space/en/stable/examples/Natural%20and%20artificial%20perturbations.html))

> Would a simple f.write(path) suffice?

Use this instead:

```
with open("results.json", "w") as fp:
    extractor.write_czml(fp)
```

(and, internally, `json.dump` on the input `fp`)

src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,279 @@
...
+        # We only want to pass a reference czml, then by modifying our reference, we'll be modifying our base dictionary
+        # which allows us to walk through a path of arbitrary length
+
+        curr = self.czml
+        for p in path:
+            curr = curr.setdefault(p, {})
+        for t in tups:
+            curr[t[0]] = t[1]
+
+    def __init_orbit_packet__(self, i):
```
Juanlu001:
```suggestion
    def _init_orbit_packet(self, i):
```

The double underscores are reserved for internal Python methods.

src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,279 @@
...
+        if color is not None:
+            self.parse_dict_tuples(
+                [i, "path", "material", "solidColor", "color"], [("rgba", color)]
+            )
+        if width is not None:
+            self.parse_dict_tuples([i, "path"], [("width", width)])
+        if show is not None:
+            self.parse_dict_tuples([i, "path", "show"], [("boolean", show)])
+
+    def change_label_params(
```
Juanlu001: I tried this function and seemed to change the labels of all orbits, by the way.

src/poliastro/contrib/extract_czml.py
```diff
@@ -0,0 +1,279 @@
...
+            [i, "path", "show"],
+            [("interval", start_epoch.value + "/" + self.end_epoch.value)],
+        )
+
+        self.parse_dict_tuples(
+            [i, "position"],
+            [
+                ("interpolationAlgorithm", "LAGRANGE"),
+                ("interpolationDegree", 5),
+                ("referenceFrame", "FIXED"),
```
Juanlu001: I wonder if this should be `"INERTIAL"` instead

src/poliastro/contrib/extract_czml.py
Sedictious: @Juanlu001 I just looked into it and you're right. It was my mistake for using ``copy()`` which creates a shallow copy of the ``DEFAULTS``. Anyway, I'll push the fix a bit later along with other changes you proposed.

Sedictious:
@Juanlu001 I've been thinking about a way to correctly implement the parameter customization. I tried to put everything in one method but it became way too big and hard to follow. Here's the way I'd approach this: Make the parameters_change methods private and call them from the orbit initializer. That way, the user won't get overwhelmed by an abundance of oftentimes redundant methods and the code will still be relatively concise.

Juanlu001:
Sounds like a good approach to me!

Sedictious:
@Juanlu001 So I moved all the additional parameters to ``add_orbit`` (now codeclimate has another thing to complain about, yay!) and I think you were right. It is now easier for someone to easily define the parameters they want instead of potentially calling three different methods.

I wanted to ask you: What would the best way of specifying reference frames be? Aren't poliastro's Orbits implicitly inertial unless specified otherwise? If so, is there a way to change said reference system?

Juanlu001:
Codeclimate is too picky :) I agree it looks easier now!

> Aren't poliastro's Orbits implicitly inertial unless specified otherwise?

The fact is that the orbits are internally converted to an inertial reference frame, even when created with the new `Orbit.from_coords` classmethod. This simplifies propagation and many other things. Therefore, I'd say we always assume it's an inertial system, at least for now.

Sedictious:
@Juanlu001 Oh, I see! I changed the rf to inertial and it now works as expected with your Sandcastle template

Juanlu001:
Merged! Congratulations @Sedictious for this contribution :clap: and hope it's only the beginning! :wink:

PR #602, opened 2019-03-16
Added tests for rot_elements_at_epoch function

amsuhane:
Added tests for rot_elements_at_epoch function


Juanlu001:
Hi @amsuhane, thanks for this pull request! I will need some more time to read the tests because they are a bit long, but in the meanwhile could you please fix the import order? The "quality" step is failing because of that reason. You can apply isort yourself using `isort --recursive --project poliastro --section-default THIRDPARTY src setup.py` (or, if you have poliastro installed in your development environment, which is the usual thing to do, just `isort --recursive src`).

setup.py
```diff
@@ -44,6 +44,7 @@
         "pandas",
         "plotly>=3.6,<4.*",
         "requests",
+        "spiceypy>=2.2.0",
```
Juanlu001: Can we add this to `extras_require["dev"]` instead?

src/poliastro/bodies.py
```diff
@@ -133,9 +133,9 @@ class _Mercury(_Body):
 
     @staticmethod
     def _rot_elements_at_epoch(T, d):
-        ra = (281.01 - 0.033 * T) * u.deg
-        dec = (61.45 - 0.005 * T) * u.deg
-        W = (329.548 + 6.1385025 * d) * u.deg
+        ra = (281.0097 - 0.0328 * T) * u.deg
```
Juanlu001: Can you please write a link at the top of the file to see where can we get these values from?

src/poliastro/tests/kernels/pck00010.tpc
```diff
@@ -0,0 +1,4061 @@
+KPL/PCK
```
Juanlu001: I wonder if there's a way we can download this kernel on-demand, or at least know the precise URL.

src/poliastro/bodies.py
amsuhane: Sure

amsuhane:
Orientation models to calculate the rotational elements (ICRF right ascension and declination and prime meridian location) are given in the kernel file "pck00010.tpc", provided by Nasa. These models have been used to calculate the rotational elements. 

The kernel file provides the coefficients of polynomials and trigonometric terms of the models. Hence, in the tests, coefficients are taken from the kernel file and plugged into the polynomial/trigonometric terms.

Url for kernel file is https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/pck00010.tpc

src/poliastro/tests/kernels/pck00010.tpc
amsuhane: The kernel can easily be downloaded on demand using requests. Once the tests are done should the kernel file be deleted?

setup.py
amsuhane: Ok

amsuhane:
@Juanlu001 I am not sure how to correct the error ```FileNotFoundError: [Errno 2] No such file or directory: '/home/circleci/repo/src/poliastro/tests/kernels/pck00010.tpc'```. The tests are running fine locally.

src/poliastro/tests/kernels/pck00010.tpc
Juanlu001: > Once the tests are done should the kernel file be deleted?

Nah, not a big deal.

src/poliastro/tests/test_bodies.py
```diff
@@ -412,3 +427,5 @@ def test_Moon(self, set_epoch):
 
         for i, j in zip(value_from_poliastro, value_from_spice):
             assert_quantity_allclose(i, j)
+
+    os.remove(kernel_name)
```
Juanlu001: This line should be in some teardown method or similar?

src/poliastro/tests/test_bodies.py
```diff
@@ -90,9 +91,23 @@ def test_from_relative():
 
 
 class TestRotElements:
-
+    """Orientation models to calculate the rotational elements (ICRF right ascension
```
Juanlu001: Perfect!

Juanlu001:
> The tests are running fine locally.

Perhaps because you already created the directory?

Juanlu001:
May I request that the only two SPICE experts that I know, @AndrewAnnex and @esaSPICEservice, take a look at this pull request when they have the time and will to do so? :blush: It would be great to confirm that we are following the right approach here.

Juanlu001:
May I request that the only two SPICE experts that I know, @AndrewAnnex and @marcsit, take a look at this pull request when they have the time and will to do so? :blush: It would be great to confirm that we are following the right approach here.

Juanlu001:
I just discovered https://github.com/esaSPICEservice/spiops by the way, in case it's any useful.

src/poliastro/tests/test_bodies.py
amsuhane: I removed the line, as removal of the downloaded kernel is not needed.

amsuhane:
I'll look into spiops. Thanks for the link!

src/poliastro/bodies.py
Juanlu001: I think this was not done yet? We should have this information in the code, not only in the tests. Just a line in the top docstring stating where do they come from, as it's done in other modules.

src/poliastro/tests/test_bodies.py
```diff
@@ -69,3 +88,348 @@ def test_from_relative():
...
+        body_code = spice.bodn2c("Jupiter")
+
+        BODY_POLE_RA = spice.gdpool("BODY" + str(body_code) + "_POLE_RA", 0, 3)
+        BODY_POLE_DEC = spice.gdpool("BODY" + str(body_code) + "_POLE_DEC", 0, 3)
+        BODY_PM = spice.gdpool("BODY" + str(body_code) + "_PM", 0, 3)
+        BODY5_NUT_PREC_ANGLES = spice.gdpool("BODY5_NUT_PREC_ANGLES", 0, 30)
+        BODY599_NUT_PREC_RA = spice.gdpool("BODY599_NUT_PREC_RA", 0, 15)
+        BODY599_NUT_PREC_DEC = spice.gdpool("BODY599_NUT_PREC_DEC", 0, 15)
+
+        Ja = BODY5_NUT_PREC_ANGLES[20] + BODY5_NUT_PREC_ANGLES[21] * T
```
Juanlu001: I was shocked when I saw the code the first time some days ago, but then I realized it's exactly the same logic we're applying in `bodies.py`. Perhaps it would be better to have functions that can be reused, so they just receive appropriate numerical values? That would make these tests shorter.

amsuhane: I'll start working on the functions.

amsuhane: I tried to write functions that would help in reducing the length of the tests. The problem is that the big tests are unique and there isn't a generalized way of calculating them. So, I'll just end up passing all the coefficients in the functions, which won't really make the tests shorter Instead, I was thinking I could just write them in a more compact like:
new_format
instead of
old

astrojuanlu:
Validating against SPICE (using SpiceyPy for example) is something we would also move to https://github.com/poliastro/validation/, so I am closing this. Thanks again @amsuhane for the massive effort in building this pull request.

PR #603, opened 2019-03-16
Added frozen orbit

Meuge:


src/poliastro/twobody/orbit.py
```diff
@@ -610,6 +610,71 @@ def parabolic(
         ss = ClassicalState(attractor, p, 1.0 * u.one, inc, raan, argp, nu)
         return cls(ss, epoch, plane)
 
+    @classmethod
+    @u.quantity_input(alt=u.m, inc=u.rad, argp=u.rad, raan=u.rad, arglat=u.rad)
+    def frozen(
+        cls,
+        alt,
+        inc=Earth.critical_inclinations[0],
```
Juanlu001: If I understand correctly, your logic is:

1. If the provided argument of periapsis is close to the critical one, use the provided inclination and freeze the eccentricity
2. If not, but the provided inclination is close to the critical one, use it
3. Else, raise an error

Now I realize this was not as straightforward as imagined! (also because I had forgotten what frozen orbits actually were).

I think it would be nice to document this logic, and also add some mathematical background. According to [J. Hoogland master thesis "Earth frozen orbits: design, injection and stability"](https://repository.tudelft.nl/islandora/object/uuid:fc0a57ce-33df-4cd2-a61b-66e702cc9cf8/datastream/OBJ/download), and probably also Vallado (latest editions not easily available on the Internet), there are two cases of frozen orbits (page 20):

1. (And now I wish GitHub had LaTeX support) (1 - 5/4 * sin^2 i) == 0. There are two critical inclinations and they don't depend on the attractor.
2. Argument of periapsis = 90 deg (for every attractor) and eccentricity given by J2, J3, the radius of the attractor, the semimajor axis and the _chosen_ inclination.

src/poliastro/twobody/orbit.py
```diff
@@ -610,6 +610,71 @@ def parabolic(
...
+        epoch: ~astropy.time.Time, optional
+            Epoch, default to J2000.
+        plane : ~poliastro.frames.Planes
+            Fundamental plane of the frame.
+        """
+        if alt <= 0:
+            raise ValueError(
+                "The semimajor axis may not be smaller that the earth's radius"
+            )
+        a = Earth.R + alt
```
Juanlu001: Frozen orbits can also appear in other planets with a non uniform gravitational field, therefore we should not make this method Earth specific. If the `attractor` doesn't have J2 and J3, we can just show the corresponding `AttributeError` to the user.

src/poliastro/twobody/orbit.py
```diff
@@ -610,6 +610,71 @@ def parabolic(
...
+
+    @classmethod
+    def _find_closest_value(cls, value, values):
+        critical_values = [critical_arg.value for critical_arg in values]
+        index = np.abs(np.asarray(critical_values) * u.rad - value).argmin()
+        return values[index]
+
+    @classmethod
+    def _frozen_critical_argp(cls, a, inc, argp, raan, arglat, epoch, plane):
+        ecc = -Earth.J3 * Earth.R * np.sin(inc) / 2 / Earth.J2 / a
```
Juanlu001: Let's move this line to `poliastro.util` and make it not Earth-specific. There's already `circular_velocity` and `hyp_nu_limit` there, so I think it makes sense.

src/poliastro/twobody/orbit.py
```diff
@@ -610,6 +610,71 @@ def parabolic(
...
+        return cls.from_classical(Earth, a, ecc, inc, raan, argp, arglat, epoch, plane)
+
+    @classmethod
+    def _frozen_critical_inclination(cls, a, inc, argp, raan, arglat, epoch, plane):
+        ecc = 5.5e-2 * u.one  # Same as the moon
```
Juanlu001: Let's have the formula to calculate this eccentricity in `poliastro.util` as well.

src/poliastro/bodies.py
```diff
@@ -168,6 +168,8 @@ class _Earth(_Body):
     J3 = constants.J3_earth
     H0 = constants.H0_earth
     rho0 = constants.rho0_earth
+    critical_inclinations = constants.critical_inclinations_earth
```
Juanlu001: All attractors have the same critical inclinations and arguments of periapsis, see my other comment.

src/poliastro/constants.py
```diff
@@ -337,3 +338,45 @@
...
 )
+
+critical_inclinations_earth = [
+    Constant(
+        "i1_earth",
+        "first critical inclination for earth's frozen orbit",
+        63.4349 * np.pi / 180,
+        "rad",
+        0,
+        "Orlov, A. A (1953)",
```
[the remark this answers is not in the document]
Juanlu001: Also: if we store these values here, they should not reference the Earth because these properties do not depend on the attractor.

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -205,6 +205,56 @@ def test_pqw_for_circular_equatorial_orbit():
     assert_allclose(w, expected_w)
 
 
+def test_frozen_orbit_argp():
```
Juanlu001: The only body that has J3 now is the Earth. Could we look for some other rocky body with a documented J3 value and add some tests that show that

1. One can create frozen orbits with attractors different than the Earth, and
2. If the given attractor does not have J3, the `AttributeError` is see from the outside?

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -205,6 +205,56 @@ def test_pqw_for_circular_equatorial_orbit():
...
+    assert orbit.inc in Earth.critical_inclinations
+
+
+def test_frozen_orbit_non_critical_inclination():
+    orbit = Orbit.frozen(1e3 * u.km, inc=0 * u.deg)  # Non-critical value
+    assert orbit.argp in Earth.critical_argps
+
+
+def test_frozen_orbit_invalid_alt():
+    with pytest.raises(ValueError):
```
Juanlu001: Here you can also check that the exception test is what you expect, take a look at other tests in poliastro for that. In this way, we avoid catching a `ValueError` that is raised for the wrong reasons.

src/poliastro/twobody/orbit.py
```diff
@@ -610,6 +615,155 @@ def parabolic(
...
+        raan=0 * u.deg,
+        arglat=0 * u.deg,
+        epoch=J2000,
+        plane=Planes.EARTH_EQUATOR,
+    ):
+        r"""Return frozen Orbit.
+
+        To achieve frozen orbit these two equations have to be set to zero.
+
+        .. math::
```
Juanlu001: Gorgeous!! :heart_eyes: https://2744-11178845-gh.circle-artifacts.com/0/docs-html/api/safe/twobody/orbit.html#poliastro.twobody.orbit.Orbit.frozen

src/poliastro/util.py
```diff
@@ -129,3 +129,27 @@ def hyp_nu_limit(ecc, r_max_ratio=np.inf):
...
+         Inclination, default to critical value.
+
+    """
+    ecc = -attractor.J3 * attractor.R * np.sin(inc) / 2 / attractor.J2 / a
+    return ecc
+
+
+def get_eccentricity_critical_inc():
+    """Calculates the eccentricity when a frozen orbit has critical inclination"""
+    ecc = 5.5e-2 * u.one  # Same as the moon
```
Juanlu001: Isn't this value free? If I understand correctly, for critical inclinations there's no restriction on the eccentricity (as long as it's small). If it's not free, where does it come from?

src/poliastro/twobody/orbit.py
```diff
@@ -610,6 +615,155 @@ def parabolic(
...
+                    attractor, a, inc, argp, raan, arglat, epoch, plane
+                )
+
+            raise ValueError("Can't create a frozen orbit from given arguments")
+
+        except Exception as exception:
+            cls._frozen_handle_exception(exception, attractor)
+
+    @classmethod
+    def _find_closest_value(cls, value, values):
```
Juanlu001: This function is generic enough that can go to `poliastro.util`, maybe? There's nothing specific to critical inclinations or even orbits here.

src/poliastro/twobody/orbit.py
```diff
@@ -610,6 +615,155 @@ def parabolic(
...
+        critical_values = [critical_arg.value for critical_arg in values]
+        index = np.abs(np.asarray(critical_values) * u.rad - value).argmin()
+        return values[index]
+
+    @classmethod
+    def _frozen_check_preconditions(cls, attractor, alt):
```
Juanlu001: I acknowledge that you introduced these classmethods to lower the cyclomatic complexity, but I think they make the code more difficult to read. I propose we get rid of all of them, let codeclimate complain, and we'll decide what to do at the end. I'm even considering disabling it.

src/poliastro/constants.py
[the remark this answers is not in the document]
Juanlu001: Too bad these old papers are so difficult to find...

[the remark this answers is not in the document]
Juanlu001: I still think these values should not be constants, but calculated with the formula.

src/poliastro/constants.py
```diff
@@ -337,3 +378,45 @@
...
+        "second critical inclination for earth's frozen orbit",
+        116.5651 * np.pi / 180,
+        "rad",
+        0,
+        "Orlov, A. A (1953)",
+        system="si",
+    ),
+]
+
+critical_argps = [
```
Juanlu001: These values are "unimportant enough" that we can hardcode them in the `frozen` code, I think.

src/poliastro/twobody/orbit.py
```diff
@@ -610,6 +615,155 @@ def parabolic(
...
+            \overline {e}=-\dfrac {J_{3}R_{E}}{2J_{2}\overline {a}\left( 1-\overline {e}^{2}\right) }\left( \dfrac {\sin ^{2}\overline {i}-\overline {e}\cos ^{2} \overline {i}}{\sin \overline {i}}\right)
+
+        Assuming that e is negligible compared to J2, it can be shown that:
+
+        .. math::
+            \overline {e}\approx -\dfrac {J_{3}R_{E}}{2J_{2}\overline {a}}\sin \overline {i}
+
+        The implementation is divided in the following cases:
+
+        1- When the user gives a negative altitude, the method will raise a ValueError
```
Juanlu001: This list was not rendered correctly though: https://2744-11178845-gh.circle-artifacts.com/0/docs-html/api/safe/twobody/orbit.html#poliastro.twobody.orbit.Orbit.frozen

src/poliastro/util.py
Juanlu001: And if it's arbitrary ("let's choose the Moon value because it's reasonable") let's not have it in a function and hardcoded inside `frozen`, I'd say.

Juanlu001:
This is in! Congratulations and thanks for this outstanding contribution!

Issue #401, opened 2018-07-06
Improve error message for Orbit.from_body_ephem(Pluto)


shreyasbapat:
Adds to #382  

🐞 **Problem**

We can't plot pluto using the ephimerids data.

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/.local/lib/python3.6/site-packages/astropy/coordinates/solar_system.py in _get_body_barycentric_posvel(body, time, ephemeris, get_velocity)
    233                 try:
--> 234                     body_index = PLAN94_BODY_NAME_TO_PLANET_INDEX[body]
    235                 except KeyError:

KeyError: 'pluto'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
<ipython-input-16-3b386b096513> in <module>()
      1 earth = Orbit.from_body_ephem(Earth)
----> 2 plu = Orbit.from_body_ephem(Pluto)

~/GitHub Clones/poliastro/src/poliastro/twobody/orbit.py in from_body_ephem(cls, body, epoch)
    163                  "{}. Use Time(..., scale='tdb') instead."
    164                  .format(epoch.tdb.value), TimeScaleWarning)
--> 165         r, v = get_body_barycentric_posvel(body.name, epoch)
    166         return cls.from_vectors(body.parent, r.xyz.to(u.km), v.xyz.to(u.km / u.day), epoch)
    167 

~/.local/lib/python3.6/site-packages/astropy/coordinates/solar_system.py in get_body_barycentric_posvel(body, time, ephemeris)
    328 
    329     """
--> 330     return _get_body_barycentric_posvel(body, time, ephemeris)
    331 
    332 

~/.local/lib/python3.6/site-packages/astropy/coordinates/solar_system.py in _get_body_barycentric_posvel(body, time, ephemeris, get_velocity)
    236                     raise KeyError("{0}'s position and velocity cannot be "
    237                                    "calculated with the '{1}' ephemeris."
--> 238                                    .format(body, ephemeris))
    239                 body_pv_helio = erfa.plan94(jd1, jd2, body_index)
    240                 body_pv_bary = body_pv_helio + sun_pv_bary``

KeyError: "pluto's position and velocity cannot be calculated with the 'builtin' ephemeris."
```

🖥 **Please paste the output of following commands**

`conda info -a`

`python -c "import poliastro.testing; poliastro.testing.test()"`

`pip freeze | grep astropy`

`pip freeze | grep poliastro`

[...]

PR #607, opened 2019-03-18
Improve error message for Orbit.from_body_ephem(Pluto)

Himanshu-Garg:
as per issue #401 
I have changed the error message which appear when Orbit.from_body_ephem(Object) runs with the name  of the Object which is not there in the selected Ephemeris

such as pluto is not there in the default ephemeris,
so now it shows the following error message
 -----------------------------------------------------------------------------------------------

KeyError - Wrong ephemeris selected, cannot load Pluto data, as Pluto is not a part of selected ephemeris. Do run this command before calling from_body_ephem ------> solar_system_ephemeris.set ('XXXXX') ------> where XXXXX contains the ephemeris having Pluto as one of its objects"

-----------------------------------------------------------------------------------------------------
 pls review #401

### DIFF
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -369,54 +369,50 @@ class Orbit(object):
     @classmethod
     def from_body_ephem(cls, body, epoch=None):
         """Return osculating `Orbit` of a body at a given time."""
+
         # TODO: https://github.com/poliastro/poliastro/issues/445
+
         if body.name == "Pluto":
             raise NotImplementedError(
                 """Default Ephemeris selected. To change it, please do
+                >>> solar_system_ephemeris.set('de432s')"""
+            )
 
-            >>> solar_system_ephemeris.set('de432s')
+        if not epoch:
+            epoch = time.Time.now().tdb
 
-            """
+        elif epoch.scale != "tdb":
+            epoch = epoch.tdb
+            warn(
+                "Input time was converted to scale='tdb' with value "
+                "{}. Use Time(..., scale='tdb') instead.".format(epoch.tdb.value),
+                TimeScaleWarning,
+            )
+
+        r, v = get_body_barycentric_posvel(body.name, epoch)
+
+        if body == Moon:
+            # TODO: The attractor is in fact the Earth-Moon Barycenter
+            icrs_cart = r.with_differentials(v.represent_as(CartesianDifferential))
+            gcrs_cart = (
+                ICRS(icrs_cart)
+                .transform_to(GCRS(obstime=epoch))
+                .represent_as(CartesianRepresentation)
+            )
+            ss = cls.from_vectors(
+                Earth,
+                gcrs_cart.xyz.to(u.km),
+                gcrs_cart.differentials["s"].d_xyz.to(u.km / u.day),
+                epoch,
             )
 
         else:
-            if not epoch:
-                epoch = time.Time.now().tdb
+            # TODO: The attractor is not really the Sun, but the Solar System
+            # Barycenter
+            ss = cls.from_vectors(Sun, r.xyz.to(u.km), v.xyz.to(u.km / u.day), epoch)
+            ss._frame = ICRS()  # Hack!
 
-            elif epoch.scale != "tdb":
-                epoch = epoch.tdb
-                warn(
-                    "Input time was converted to scale='tdb' with value "
-                    "{}. Use Time(..., scale='tdb') instead.".format(epoch.tdb.value),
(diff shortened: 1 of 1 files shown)


src/poliastro/twobody/orbit.py
```diff
@@ -372,40 +373,47 @@ def from_body_ephem(cls, body, epoch=None):
...
-        elif epoch.scale != "tdb":
-            epoch = epoch.tdb
-            warn(
-                "Input time was converted to scale='tdb' with value "
-                "{}. Use Time(..., scale='tdb') instead.".format(epoch.tdb.value),
-                TimeScaleWarning,
-            )
-
-        r, v = get_body_barycentric_posvel(body.name, epoch)
+        if body.name.lower() not in solar_system_ephemeris.bodies:
```
shreyasbapat: I guess this was not needed. You just had to apply a check for Pluto and Moon in that case!

src/poliastro/twobody/orbit.py
```diff
@@ -4,6 +4,7 @@
 from astropy import time, units as u
 from astropy.coordinates import (
     GCRS,
+    solar_system_ephemeris,
```
shreyasbapat: This import was not needed

src/poliastro/twobody/orbit.py
```diff
@@ -372,40 +373,47 @@ def from_body_ephem(cls, body, epoch=None):
...
-            warn(
-                "Input time was converted to scale='tdb' with value "
-                "{}. Use Time(..., scale='tdb') instead.".format(epoch.tdb.value),
-                TimeScaleWarning,
-            )
-
-        r, v = get_body_barycentric_posvel(body.name, epoch)
+        if body.name.lower() not in solar_system_ephemeris.bodies:
+            error_message="Wrong ephemeris selected, cannot load " + str(body.name) + " data, as " + str(body.name) + " is not a part of selected ephemeris. "  
+            error_message+="Do run this command before calling from_body_ephem ------> solar_system_ephemeris.set('XXXXX') ------> where XXXXX contains the ephemeris having " + str(body.name) + " as one of its objects"
```
shreyasbapat: Have a look at some errors raised by poliastro previously. It will be better if you work on that line. 
See this: https://github.com/poliastro/poliastro/blob/master/src/poliastro/twobody/orbit.py#L563
Or for giving code examples, see this: https://github.com/poliastro/poliastro/blob/master/src/poliastro/plotting/static.py#L18-L24

Himanshu-Garg: thanks a lot
now will show the error in same format

src/poliastro/twobody/orbit.py
Himanshu-Garg: i think, we need this as there are several ephemeris available and some contain pluto and some may not
here is the list of ephemeris i found on astropy
- 'builtin': polynomial approximations to the orbital elements.
- 'de430' or 'de432s': short-cuts for recent JPL dynamical models.
- 'jpl': Alias for the default JPL ephemeris (currently, 'de430').
- URL: (str) The url to a SPK ephemeris in SPICE binary (.bsp) format.
- `None`: Ensure an Exception is raised without an explicit ephemeris

pls review

Himanshu-Garg: here, using url we can add any ephemeris

shreyasbapat: I believe we didn't intend to achieve this from the issue! @Juanlu001  I will seek your help.

src/poliastro/twobody/orbit.py
```diff
@@ -368,12 +369,23 @@ def from_equinoctial(
...
+        """Return osculating `Orbit` of a body at a given time."""
+
         # TODO: https://github.com/poliastro/poliastro/issues/445
+
+        if (
+            body.name == "Pluto"
+            and body.name.lower() not in solar_system_ephemeris.bodies
+        ):
+            raise KeyError(
+                """Default Ephemeris selected. To change it, please do
```
shreyasbapat: I would rather do: 

```suggestion
                """These bodies can not be found in the "builtin" ephemeris. Please change the ephemeris.
                >>> from astropy.coordinates import solar_system_ephemeris
                >>> solar_system_ephemeris.set('jpl')"""

```

shreyasbapat: And I am not sure if raising `KeyError` is the correct thing to do!

Himanshu-Garg: i was thinking to set it as Not-Implemented Error, but have to change test cases accordingt to it.

should I do that?
and change test cases also?

Issue #609, opened 2019-03-18
Raise error in Orbit.from_sbdb when more than one orbit is returned


shreyasbapat:
🐞 **Problem**

When `Orbit.from_sbdb` is called with names which are too common, like "Halley" or "*alley", it returns a keyerror. Raise a suitable warning in those cases.

Code to reproduce error: 
```
Orbit.from_sbdb("Halley")
```

🖥 **Please paste the output of following commands**

* `conda info -a` (only if you have conda)
* `conda list` (only if you have conda)
* `pip freeze`

<details>
  
```bash
# Paste your output here:

```
</details>

* `python -c "import poliastro.testing; poliastro.testing.test()"`

<details>

```bash
# Paste your output here:

```
</details>

🎯 **Goal**

Related to #534 
To make `from_sbdb` more user friendly.

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
Rather than a warning, I think we should raise an error, in the same way that we did with the old API:

![Screenshot_2019-03-18 Using NEOS package — poliastro 0 12 0 documentation](https://user-images.githubusercontent.com/316517/54541408-2b320380-499a-11e9-8898-b3adbfb79278.png)

Ivan9528:
What about raising an error a bit more elaborated/understandable for the user about the different bodies found and then giving the user the chance to chose one of them by entering the complete name or the code number associated with it?

Juanlu001:
@Ivan9528 thanks for chiming in! The problem with that is that it requires interactivity, and in a script that runs unattended it can block the program. #611 is unfinished, so perhaps you want to take care of this one?

michiboo:
Hi!  is this issue still unsolved? if not I will try tackle it!

Ivan9528:
Hi @michiboo!! It is already solved actually, I just have to make the pull request but the coding is done, I will do it tonight if time allows me to :D Thanks!!

Ivan9528:
@michiboo I am sorry, I am not working on this issue but on #611. Anyway, @Juanlu001, please confirm that this issue is unsolved because I have run the code I have got a value error instead of a key one which I presume it was the purpose of the issue...

shreyasbapat:
[...]

PR #611, opened 2019-03-18
changed from key error to value error in from_sbdb

Himanshu-Garg:
as per issue #609 
now when no. of objects are more than 1, instead of key error, its showing Value error as discussed in comments section of #609 

here is the new output ...
Screenshot (156)

pls review!!!

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -731,9 +731,9 @@ def test_from_sbdb():
 
         assert ss_classical == SBDB_DATA[target_name]
 
+
 def test_from_sbdb_raise_valueerror():
     with pytest.raises(ValueError) as e:
         Orbit.from_sbdb(name="Halley")
-        
-    assert str(e.value)[0]=='2'
 
+    assert str(e.value)[0] == "2"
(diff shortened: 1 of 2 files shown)


src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -730,3 +730,10 @@ def test_from_sbdb():
         ss_classical = ss_target.classical()
 
         assert ss_classical == SBDB_DATA[target_name]
+
+
+def test_from_sbdb_raise_valueerror():
+    with pytest.raises(ValueError) as e:
+        Orbit.from_sbdb(name="Halley")
+
+    assert str(e.value)[0] == "2"
```
Juanlu001: Can we test the full message?

src/poliastro/twobody/orbit.py
```diff
@@ -469,31 +469,46 @@ def from_sbdb(cls, name, **kargs):
...
-            a,
-            ecc,
-            inc,
-            raan,
-            argp,
-            nu,
-            epoch=epoch.tdb,
-            plane=Planes.EARTH_ECLIPTIC,
-        )
+        except KeyError:
```
Juanlu001: I don't like this way of using an error to control the flow. Perhaps could do instead:

```
if "count" in obj:
    # Handle the error
    raise ValueError(...)

# No need of else block, just do the rest of the stuff
a = obj["orbit"][...]
...
```

Juanlu001:
@Himanshu-Garg are you still working on this?

src/poliastro/twobody/orbit.py
Ivan9528: I will take care of these changes asap! The idea is pretty much the same as in issue #609, right? Raise a proper message (not a key) when an error is produced.

Juanlu001: Correct!

PR #612, opened 2019-03-18
Improve error message for Orbit.from_body_ephem(Pluto)

Himanshu-Garg:
as per issue #401
I have changed the error message which appear when Orbit.from_body_ephem(Object) runs with the name of the Object which is not there in the selected Ephemeris

such as pluto is not there in the default ephemeris,
so now it shows the following error message
KeyError - Wrong ephemeris selected, cannot load Pluto data, as Pluto is not a part of selected ephemeris. Do run this command before calling from_body_ephem ------> solar_system_ephemeris.set ('XXXXX') ------> where XXXXX contains the ephemeris having Pluto as one of its objects"

pls review #401

### DIFF
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -9,6 +9,7 @@ from astropy.coordinates import (
     CartesianDifferential,
     CartesianRepresentation,
     get_body_barycentric_posvel,
+    solar_system_ephemeris,
 )
 from astroquery.jplhorizons import Horizons
 from astroquery.jplsbdb import SBDB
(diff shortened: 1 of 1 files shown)


Himanshu-Garg:
its same as pull request #607 
again attaching because by mistake, deleted the whole repo from my ID

src/poliastro/twobody/orbit.py
```diff
@@ -368,12 +369,23 @@ def from_equinoctial(
...
     def from_body_ephem(cls, body, epoch=None):
-        """Return osculating `Orbit` of a body at a given time.
 
-        """
+        """Return osculating `Orbit` of a body at a given time."""
+
         # TODO: https://github.com/poliastro/poliastro/issues/445
+
+        if (
+            body.name == "Pluto"
```
jorgepiloto: Thanks for the PR @Himanshu-Garg. As said in #401, for the case of the Moon same happens and therefore some message should be raised. The problem is that 'Moon' is included in both `astropy` ephemerides: `builtin` and `de432s`. So therefore, here is my code suggestion:

```suggestion
            body.name == "Moon"
            or body.name.lower() not in solar_system_ephemeris.bodies
```

Juanlu001:
@Himanshu-Garg are you still working on this?

Juanlu001:
Superseded by #657.

Issue #597, opened 2019-03-10
Migrate to setup.cfg


Juanlu001:
TIL: https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

(Thanks https://gitlab.com/librespacefoundation/satnogs/satnogs-client/!)

Juanlu001:
Closed in #613

PR #613, opened 2019-03-19
Move setup.py to setup.cfg

jorgepiloto:
This solves #597 :snake:


Juanlu001:
Thanks for this PR! I think the errors are because our `src` layout: https://setuptools.readthedocs.io/en/latest/setuptools.html#using-a-src-layout

Juanlu001:
Awesome! :muscle:

Issue #486, opened 2018-11-03
Leverage new numba objmode context manager


astrojuanlu:
As said in https://github.com/numba/numba/issues/2788#issuecomment-435343673:

> the latest release of Numba now has the capacity to allow jumping into the interpreter from nopython mode

https://numba.pydata.org/numba-doc/latest/user/withobjmode.html#the-objmode-context-manager

This would be really interesting to use, since there are places where we read external data and we cannot `njit` those functions. I'm thinking specifically about the 3rd body perturbations.

astrojuanlu:
For 3rd body perturbations, let's better create an interpolator or similar, instead of using `with.objmode`.

PR #631, opened 2019-03-26
Extended Gauss' propagator

jorgepiloto:
I took directly the algorithm from Battin 1999, section 5.6 and tested the algorithm with the proposed example 5-30.

This algorithm still needs to be finished, so CI is going to fail. But I decided to make the PR just for receive code reviews and keep working on this. Last part of Battin's explanation about the algorithm is a little bit confusing.

This may solve when finished #475


src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -55,6 +55,21 @@ def test_propagation(method):
     assert_quantity_allclose(v, expected_v, rtol=1e-4)
 
 
+def test_gauss():
+
```
Juanlu001: Can you write right here where does this data come from?

src/poliastro/twobody/propagation.py
```diff
@@ -8,13 +8,16 @@
...
+from scipy.optimize import root
 
 from poliastro.core.propagation import (
     func_twobody,
     kepler as kepler_fast,
     mean_motion as mean_motion_fast,
 )
+
 from poliastro.integrators import DOP835
+from poliastro.util import norm
```
Juanlu001: You might want to use `poliastro.core.util.norm` instead

src/poliastro/twobody/propagation.py
```diff
@@ -189,6 +192,134 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+            kappa = alpha_prime * Tm_prime /
+            """
+            n += 1
+
+    return KeyError("SOLUTION NOT FOUND")
+
+
+def gauss(k, r, v, time_of_flight, rtol=1e-8, **kwargs):
+
+    results = [_gauss(k, r, v, tof, rtol, **kwargs) for tof in time_of_flight]
```
Juanlu001: My recommendation is that you strip the units from `r` and `v` before calling `_gauss`, so you can use lower level functions and, if you're lucky, `njit` all the thing to make it blazing fast :sunglasses:

src/poliastro/twobody/propagation.py
```diff
@@ -189,6 +192,134 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+
+        epsilon = 1 + eta * phi0_prime
+        b = np.abs(epsilon + eta * (phi0_2prime + chi0_prime * psi))
+
+        def f(x, epsilon, b):
+            """ This is eq. 5.83 in Battin. """
+            return x ** 3 - 3 * epsilon * x - 2 * b
+
+        # Solving for the roots
+        x = root(f, b, args=(epsilon, b)).x[0]
```
Juanlu001: Is there a chance we have the derivative of the function? I think the answer is yes, so I would use `newton` instead. Even if we don't provide the derivative, I would not use `root`, because it's too general purpose and impossible to accelerate with numba (unless we explore https://github.com/poliastro/poliastro/issues/486)

src/poliastro/twobody/propagation.py
```diff
@@ -189,6 +192,134 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+            xi = 1 - x
+
+        theta = psi * eta / xi ** 2
+
+        A = gamma0_prime * theta
+
+        if np.abs(A) <= A_m:
+            """ If value A is less than Amax. """
+            if np.abs(A - A_old) <= rtol:
+                # Since 'lambda' is a reserved function, 'lamda' is used
```
Juanlu001: In that case, I would use `lambda_` instead

src/poliastro/twobody/propagation.py
```diff
@@ -189,6 +192,134 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+            xi = 1 + x
+        else:
+            xi = 1 - x
+
+        theta = psi * eta / xi ** 2
+
+        A = gamma0_prime * theta
+
+        if np.abs(A) <= A_m:
+            """ If value A is less than Amax. """
```
Juanlu001: Use `# ` instead, this comment style should be used only in docstrings

jorgepiloto:
I've decided to work again on this since I have some free time today. However, even if the tests pass some things still need to be done:

* [ ] When the algorithm checks if the value of A is higher than A_m and gets inside the conditional, the returned orbit is propagated wrong. I have spent hours checking and debugging and not sure if this part is well implemented in the code. Still working on this 

* [ ] Furthermore, some docstings are needed

What a pity Riot-Matrix is down today :cry:

Juanlu001:
Pending:

* [ ] Add _failing_ tests
* [ ] Add link to paper that contains full flow chart

jorgepiloto:
Link to paper on Gauss Algorithm: https://dspace.mit.edu/handle/1721.1/16383

jorgepiloto:
This is going to be my priority during the following days. I will read in detail the paper because even that if Battin is the advisor,  I am not sure if it is the same algorithm that the one explained on his book. 
And as it is said in the introduction to the paper: 

`The final
algorithm is a general form valid for any orbit of any eccentricity and
requires no knowledge of the nature of tKo orbit for application.`

So this is supposed to be what I call the MOAP (Mother Of All Propagators) :laughing:

jorgepiloto:
Current status of this algorithm:

*  There are two implemented tests: one is from Battin (passing) and another one that tests all the propagators in poliastro (failing in gauss).

*  Algorithm works well if not entering in the `elif` section, that means that `A` never exceeds `A_max = 0.3`. Otherwise algorithm does not converge well. I have done line by line debugging but everything looks good when executing different parts of the code. Problem may be located at this section but I am not able to find the bug.

* Notice that on the failing tests, results just diverge from a few km.

jorgepiloto:
Current status and updates on this pull request:

- After #908 and #918 this algorithm interest has become just "archeological". As other Kepler routines, this might be interesting for comparison purposes.

- :warning: Just found the original paper for the algorithm with DOI: https://doi.org/10.2514/3.55860
  This might lead to the final implementation of this solver.

PR #634, opened 2019-03-27
Added some tests for CZML Extractor

Sedictious:
I added a couple of tests to the CZML Extractor (and also renamed the ``contrib`` dir to ``czml``). 

I know the unit tests are supposed to test a very specific functionality to narrow down the list of possible culprits in case something goes wrong, but I felt it was more natural to test the whole thing. So the main test function checks nothing is wrong with the more bug-prone parts of the program (no overriding parameters, correct coordinates)

Please let me know if it'd be better to simply create more and smaller unit tests (e.g.``test_coordinates`` and ``test_parameters``)


src/poliastro/czml/extract_czml.py
```diff
@@ -6,7 +6,7 @@
 from astropy.coordinates import CartesianRepresentation
 from astropy.time import Time, TimeDelta
 
-from poliastro.contrib.czml_extract_default_params import DEFAULTS
+from poliastro.czml.czml_extract_default_params import DEFAULTS
```
Juanlu001: I think the `ImportError` on Windows comes from the fact that `poliastro/czml` is missing an `__init__.py`, so Python understands it's a package.

src/poliastro/tests/test_czml.py
```diff
@@ -0,0 +1,59 @@
...
+
+from poliastro.czml.extract_czml import CZMLExtractor
+from poliastro.examples import iss, molniya
+
+
+def test_czml_add_orbit():
+    start_epoch = iss.epoch
+    end_epoch = iss.epoch + molniya.period
+
+    Extractor = CZMLExtractor(start_epoch, end_epoch, 10)
```
Juanlu001:
```suggestion
    extractor = CZMLExtractor(start_epoch, end_epoch, 10)
```

Let's use lowercase here, it was a bit confusing at the beginning (I'm so used to PEP8 naming that at first I thought it was a class)

src/poliastro/tests/test_czml.py
```diff
@@ -0,0 +1,59 @@
...
+    assert Extractor.czml[0]["label"]["fillColor"]["rgba"] == [125, 80, 120, 255]
+
+    assert Extractor.czml[1]["path"]["show"]["boolean"] is False
+
+
+def test_czml_invalid_orbit_epoch_error():
+    start_epoch = molniya.epoch
+    end_epoch = molniya.epoch + molniya.period
+
+    Extractor = CZMLExtractor(start_epoch, end_epoch, 10)
```
Juanlu001:
```suggestion
    extractor = CZMLExtractor(start_epoch, end_epoch, 10)
```

Same as above

src/poliastro/tests/test_czml.py
```diff
@@ -0,0 +1,59 @@
...
+    end_epoch = iss.epoch + molniya.period
+
+    Extractor = CZMLExtractor(start_epoch, end_epoch, 10)
+
+    Extractor.add_orbit(
+        molniya, label_text="Molniya", label_fill_color=[125, 80, 120, 255]
+    )
+    Extractor.add_orbit(iss, label_text="ISS", path_show=False)
+
+    cords_molniya = Extractor.czml[0]["position"]["cartesian"][4 * 5 : 4 * 6]
```
Juanlu001: Instead of testing a small subset, I would choose one of two options:

1. Checking that the full coordinates list is equal to the appropriate `propagate(...)`
2. Don't check the values at all, and only that `["position"]["cartesian"]` has the expected number of elements

I don't think checking the actual values is that important in this case. In general, let's keep the test simple to follow and understand.

src/poliastro/czml/extract_czml.py
Sedictious: Ah you're completely right. I was really confused as to why it was failing, since everything seemed to work locally.

src/poliastro/czml/extract_czml.py
```diff
@@ -29,8 +29,8 @@ def __init__(self, start_epoch, end_epoch, N):
             of sampled data points will be N when calling
             add_orbit()
         """
-        self.czml = dict()
-        self.orbits = []
+        self.czml = dict()  # type : dict
```
Juanlu001:
```suggestion
        self.czml = dict()  # type: dict
```

Otherwise, mypy doesn't understand it :sweat_smile:

src/poliastro/czml/extract_czml.py
```diff
@@ -29,8 +29,8 @@ def __init__(self, start_epoch, end_epoch, N):
             of sampled data points will be N when calling
             add_orbit()
         """
-        self.czml = dict()
-        self.orbits = []
+        self.czml = dict()  # type: Dict[int, Any]
```
Juanlu001: `from typing import Dict, List, Any`

Sedictious: Thank you! I was really confused as to why it was throwing this error...

Juanlu001:
Thanks! :tada:

PR #635, opened 2019-03-29
Enable porkchops between Orbit objects

jorgepiloto:
With this modifications porkchops are able to plot figures between `Orbit` instances. Therefore, it is possible to create porkchops for asteroids by using the `from_sbdb`. I tried this option for `Apophis`:

apophis_porkchop

Here you can compare with the porkchop figure for same launch and arrival spans: http://www.ssdl.gatech.edu/sites/default/files/papers/conferencePapers/AAS-2008-251.pdf

paper_porkchop

However, I would like to find a way of testing porkchops rather than visual one...


src/poliastro/plotting/porkchop.py
```diff
@@ -17,12 +18,23 @@ def _targetting(departure_body, target_body, t_launch, t_arrival):
     """
 
     # Compute departure and arrival positions
-    rr_dpt_body, vv_dpt_body = coord.get_body_barycentric_posvel(
-        departure_body.name, t_launch
-    )
-    rr_arr_body, vv_arr_body = coord.get_body_barycentric_posvel(
-        target_body.name, t_arrival
-    )
+    if isinstance(departure_body, Orbit):
```
Juanlu001: Rather than all these `isinstance` checks, which are extremely slow and complicate the code, I would create a new function for the new interface. And if there's lots of duplicate code between the new and the current one, then there's an opportunity to refactor and put the common code in a private function, for example.

Juanlu001:
Regarding validation, I agree just looking at plots is not enought, but a porkchop plot in the end is "just" finding Lambert solutions in a double loop.

Perhaps we can draw inspiration from matplotlib itself:

https://matplotlib.org/devel/testing.html#writing-an-image-comparison-test

src/poliastro/plotting/porkchop.py
jorgepiloto: Sorry for the delay :sweat_smile:  I finally wrote a private function called `_get_state` that returns the state vectors (rr, vv) of a body depending if it belongs to `Orbit` or `Body` instances. This function does not make use of `isinstance` but requires importing all poliastro bodies instead.

jorgepiloto:
I took a look to the image comparasion example from matplotlib. However a warning about deprecation and further elimination of that module is given when the test is executed...

After some research I found that some people just compare the returned vector from the function with a previously computed solution. This means trusting that the figure will be the same as the expected :confused:

Juanlu001:
> I took a look to the image comparasion example from matplotlib. However a warning about deprecation and further elimination of that module is given when the test is executed...

Can you paste the deprecation warning you are seeing? I didn't find anything relevant here:

https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/testing/decorators.py

jorgepiloto:
The warning I am getting is the following one: 

```bash
MatplotlibDeprecationWarning: The ImageComparisonTest class was deprecated in Matplotlib 3.0 and will be removed in 3.2.
-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

Juanlu001:
I see where it was deprecated 
https://github.com/matplotlib/matplotlib/pull/10906 but on the other 
hand it's not mentioned in this documentation 
https://matplotlib.org/devel/testing.html#writing-an-image-comparison-test 
so I understand there's a simpler alternative now using those 
decorators?

Juanlu001:
The codecov check is failing because, even though the total coverage is increasing (which is not difficult, because this part was not tested before) it also checks what percentage of *the lines being added* is tested. For me it is not that strict, so I will make a review and, if everything looks good, merge it.

setup.cfg
```diff
@@ -88,6 +88,7 @@ markers =
     slow
     remote_data
     filterwarnings
+    mpl_image_compare
```
Juanlu001: Don't we need to add `pytest-mlp` to `options.extras_require`?

src/poliastro/plotting/porkchop.py
Juanlu001: Fair enough! If this piece of code is going away, please remove it (don't leave it as a multistring comment)

src/poliastro/tests/tests_plotting/test_porkchop.py
```diff
@@ -0,0 +1,21 @@
...
+import pytest
+
+from poliastro.bodies import Earth, Mars
+from poliastro.plotting.porkchop import porkchop
+from poliastro.util import time_range
+
+
+@pytest.mark.mpl_image_compare
+def test_porkchop_plotting():
+
```
Juanlu001: Remove this extra line?

src/poliastro/tests/tests_plotting/test_porkchop.py
```diff
@@ -0,0 +1,21 @@
...
+    fig = plt.figure()
```
Juanlu001: Let's use `plt.subplots` instead

src/poliastro/tests/tests_plotting/test_porkchop.py
```diff
@@ -0,0 +1,21 @@
+import matplotlib.pyplot as plt
+import pytest
+
+from poliastro.bodies import Earth, Mars
+from poliastro.plotting.porkchop import porkchop
+from poliastro.util import time_range
+
+
+@pytest.mark.mpl_image_compare
```
Juanlu001: Excellent!!

src/poliastro/tests/tests_plotting/test_porkchop.py
jorgepiloto: Oh, so sorry! Forgot to clean this before uploading it :sweat_smile:

Juanlu001:
Merged! Thanks a lot @jorgepiloto :tada:

Issue #638, opened 2019-03-31
Add contributing section in docs


AunSiro:
The current [documentation in Read the Docs](https://docs.poliastro.space/en/stable/index.html) doesn't contain the information about contributing present in [CONTRIBUTING.rst](https://github.com/poliastro/poliastro/blob/master/CONTRIBUTING.rst)

In order to solve this, knowledge about how to write documentation in Read the Docs is required.

michiboo:
Hi can I take this?

Juanlu001:
@michiboo, this was closed in #640!

Issue #608, opened 2019-03-18
Add tox environment to reformat code


Juanlu001:
Although `tox -e check` tells you what's wrong with the code, it's not obvious for newcomers how to fix those issues. We should have a `tox -e format` or similar that actually reformats the code.

azlkiniue:
I am interested to try this.
From this [comment](https://github.com/poliastro/poliastro/pull/640#discussion_r27183904) I know that we need to use `black` and `isort`. Is it enough?

Juanlu001:
First `isort`, then `black` :) Correct!

Juanlu001:
@azlkiniue still thinking about working on this?

azlkiniue:
> @azlkiniue still thinking about working on this?

Oh, I'm sorry.. almost forget this one..
Are these commands correct?
```
isort --recursive --section-default THIRDPARTY src setup.py
black src setup.py
```

Juanlu001:
The isort one would be `isort --recursive --project poliastro --section-default THIRDPARTY src setup.py`. If you have poliastro installed in development mode it would be just `isort --recursive src setup.py`, but I prefer the safe version.

Juanlu001:
Closed in #655

PR #640, opened 2019-04-03
Add contributing section and fix some typos in docs

azlkiniue:
This PR is for  #638 
I also add VSCode folder to .gitignore

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -70,9 +70,10 @@ You will also need to understand how git works. git is a decentralized
 version control system that preserves the history of the software, helps
 tracking changes and allows for multiple versions of the code to exist
 at the same time. If you are new to git and version control, I recommend
-following `the Try Git tutorial`_.
+following `the Try Git tutorial`_ and `Git Workflow tutorial`_.
 
 .. _`the Try Git tutorial`: https://try.github.io/
+.. _`Git Workflow tutorial`: https://www.asmeurer.com/git-workflow/
 
 If you already know how all this works and would like to contribute new
 features then that's awesome! Before rushing out though please make sure it
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -1,0 +1,121 @@
+Contributing
+============
+
+poliastro is a community project, hence all contributions are more than
+welcome!
+
+Bug reporting
+-------------
+
+Not only things break all the time, but also different people have different
+use cases for the project. If you find anything that doesn't work as expected
+or have suggestions, please refer to the `issue tracker`_ on GitHub.
+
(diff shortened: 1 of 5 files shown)


docs/source/contributing.rst
```diff
@@ -0,0 +1,121 @@
+Contributing
```
shreyasbapat: This is a great addition, can you also include the quality checks that we perform in this? For example, black and isort?

docs/source/contributing.rst
```diff
@@ -0,0 +1,121 @@
...
+.. _`the Try Git tutorial`: https://try.github.io/
+
+If you already know how all this works and would like to contribute new
+features then that's awesome! Before rushing out though please make sure it
+is within the scope of the library so you don't waste your time -
+`the mailing list`_ or `the chat`_ are good places to ask.
+
+.. _`the mailing list`: https://groups.io/g/poliastro-dev
+.. _`the chat`: https://riot.im/app/#/room/#poliastro:matrix.org
+
```
shreyasbapat: I would prefer the code quality block here! @Juanlu001  will tell more if something is needed. As a lot of contributors get confused in running `black` and `isort`, it will be better to have it documented here. 

Specifically, we would want these two commands: 

```
black changed_filename.py
isort changed_filename.py
```
here.

Sedictious: @azlkiniue Actually they're required to pass the tests (had to learn this the hard way). Maybe consider adding a section for configuring those tools on some of the most popular IDEs or link to the corresponding wikis of each.

Juanlu001: Related: https://github.com/poliastro/poliastro/issues/608 I think we should address this one first!

docs/source/contributing.rst
```diff
@@ -0,0 +1,121 @@
...
+steep learning curve). Other people use Spyder, emacs, gedit, Notepad++,
+Sublime, Atom...
+
+You will also need to understand how git works. git is a decentralized
+version control system that preserves the history of the software, helps
+tracking changes and allows for multiple versions of the code to exist
+at the same time. If you are new to git and version control, I recommend
+following `the Try Git tutorial`_.
+
+.. _`the Try Git tutorial`: https://try.github.io/
```
Juanlu001: I think this tutorial is also very good and worth adding:

https://www.asmeurer.com/git-workflow/

azlkiniue: Added!

docs/source/contributing.rst
shreyasbapat: Yes :/

Juanlu001:
Sorry for the long delay! I think it's time to merge this :) Thanks a lot @azlkiniue!

PR #646, opened 2019-04-09
Allow extractor to accept different attractor radii

Sedictious:
I've been thinking a bit about how to represent non-geocentric orbits and I think simply scaling the trajectory is the cleanest and most straight-forward way.

 Obviously deleting or hiding the main body and defining a new ellipse is not a great idea. There are ways to resize the main globe in Cesium, but they require extra javascript code and I think it's better to do most of the job here, rather than hiding half the code behind the js application -especially for something this simple-.

For now, you can't really change the imagery layers, but this is the next thing I'm planning to work on.

### DIFF
--- a/src/poliastro/czml/extract_czml.py
+++ b/src/poliastro/czml/extract_czml.py
@@ -7,6 +7,7 @@ from astropy import units as u
 from astropy.coordinates import CartesianRepresentation
 from astropy.time import Time, TimeDelta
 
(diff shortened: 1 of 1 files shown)


Juanlu001:
For the record, after discussing a bit, we decided not to "hack" the planet radius and use a more consistent approach instead, even if it's more complicated.

PR #648, opened 2019-04-16
Tox for testing: docs, py3X and coverage

jorgepiloto:
This file runs tests on docs, different python environments and coverage. However, there are some warnings I was not able to solve by my own:

* As far as I read py35, pypy and pypy3 are raising `InterpreterNotFound` errors since tox is not able to find this environments in Anaconda.
* There is also an error coming from graphviz from the docs: `dot command 'dot' cannot be run (needed for graphviz output), check the graphviz_dot setting`

I also found that by applying `isort -rc --atomic .` it is possible to format all the code when running `tox` :relaxed:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,40 +1,45 @@
-# The only current way of using OS X virtual machines in Travis CI
-# is setting up a generic environment
-# This is not a problem because Python will be installed through conda
-
-language: generic
-os: osx
-
-branches:
-  only:
-    - master
-    - 0.3.x
-    - 0.4.x
-    - 0.5.x
-    - 0.6.x
-    - 0.7.x
-    - 0.8.x
-    - 0.9.x
-    - 0.10.x
-    - 0.11.x
-    - 0.12.x
-
+language: python
+dist: xenial
 env:
   global:
-    - MPLBACKEND="Agg"
-
+    - LD_PRELOAD=/lib/x86_64-linux-gnu/libSegFault.so
+    - SEGFAULT_SIGNALS=all
+matrix:
+  include:
+    - python: '3.6'
+      env:
+        - TOXENV=check
+    - python: '3.6'
+      env:
+        - TOXENV=docs
+    - python: '3.5'
+      env:
+        - TOXENV=py35,report,codecov
+    - python: '3.6'
+      env:
+        - TOXENV=py36,report,codecov
+    - python: '3.7'
+      env:
+        - TOXENV=py37,report,codecov
+    - python: 'pypy3.5-6.0'
+      env:
+        - TOXENV=pypy3,report,codecov
 before_install:
-  - HOMEBREW_NO_AUTO_UPDATE=1 brew install python3
-  - python3 -m venv venv
-  - source venv/bin/activate
-  - python3 -m pip install --upgrade pip
-
+  - python --version
+  - uname -a
+  - lsb_release -a
 install:
-  - pip install numpy  # Required
-  - pip install .  # Test installation correctness
-  - pip install .[dev]  # Test development dependencies
-
+  - pip install tox
+  - virtualenv --version
+  - easy_install --version
+  - pip --version
+  - tox --version
 script:
-  - pytest -vv  # Test against installed code
-
-sudo: false
+  - tox -v
+after_failure:
+  - more .tox/log/* | cat
+  - more .tox/*/log/* | cat
+notifications:
+  email:
+    on_success: always
+    on_failure: always
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
-language: python
-dist: xenial
-env:
-  global:
-    - LD_PRELOAD=/lib/x86_64-linux-gnu/libSegFault.so
-    - SEGFAULT_SIGNALS=all
+# The only current way of using OS X virtual machines in Travis CI
+# is setting up a generic environment
+# This is not a problem because Python will be installed through conda
+
+language: generic
+ox: osx
 matrix:
   include:
     - python: '3.6'
@@ -14,16 +14,16 @@ matrix:
         - TOXENV=docs
     - python: '3.5'
       env:
-        - TOXENV=py35,report,codecov
+        - TOXENV=py35
     - python: '3.6'
       env:
-        - TOXENV=py36,report,codecov
+        - TOXENV=py36
     - python: '3.7'
       env:
-        - TOXENV=py37,report,codecov
+        - TOXENV=py37
     - python: 'pypy3.5-6.0'
       env:
-        - TOXENV=pypy3,report,codecov
+        - TOXENV=pypy3
 before_install:
   - python --version
   - uname -a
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,40 +1,45 @@
...
+  - uname -a
+  - lsb_release -a
 install:
-  - pip install numpy  # Required
-  - pip install .  # Test installation correctness
-  - pip install .[dev]  # Test development dependencies
-
+  - pip install tox
+  - virtualenv --version
+  - easy_install --version
```
Juanlu001: `easy_install` must burn in hell :) please remove this line so I don't have nightmares at night!

.travis.yml
```diff
@@ -1,40 +1,45 @@
...
+  - pip --version
+  - tox --version
 script:
-  - pytest -vv  # Test against installed code
-
-sudo: false
+  - tox -v
+after_failure:
+  - more .tox/log/* | cat
```
Juanlu001: What's the difference with `cat .tox/log/*`?

.travis.yml
```diff
@@ -1,40 +1,45 @@
...
+  - more .tox/*/log/* | cat
+notifications:
+  email:
+    on_success: always
```
Juanlu001: Please no :) My inbox suffers too much already :joy:

tox.ini
```diff
@@ -0,0 +1,94 @@
...
+    {py36,docs,spell}: {env:TOXPYTHON:python3.6}
+    py37: {env:TOXPYTHON:python3.7}
+    {clean,check,report,codecov}: {env:TOXPYTHON:python3}
+setenv =
+    PYTHONPATH={toxinidir}/tests
+    PYTHONUNBUFFERED=yes
+passenv =
+    *
+usedevelop = false
+deps =
```
Juanlu001: Isn't it enough with installing poliastro and let `pip` + `install_requires` in `setup.cfg` do the rest?

tox.ini
```diff
@@ -0,0 +1,94 @@
...
+    numba>=0.39
+    numpy
+    pandas
+    plotly>=3.6,<4.*
+    scipy>=1.0
+commands =
+    {posargs:pytest --cov --cov-report=term-missing -vv src/poliastro/tests}
+
+[testenv:check]
+deps =
```
Juanlu001: Same comment as above, but for `extras_require` in `setup.cfg`

tox.ini
```diff
@@ -0,0 +1,94 @@
...
+usedevelop = false
+commands =
+    python setup.py check --strict --metadata --restructuredtext
+    flake8 src setup.py
+    isort --check-only --diff --recursive --project poliastro --section-default THIRDPARTY src setup.py
+    black --check src setup.py
+    mypy --ignore-missing-imports --check-untyped-defs --no-strict-optional src
+
+[testenv:docs]
+description = invoke sphinx-build to build the HTML docs
```
Juanlu001: Didn't know we could write descriptions! Let's have more of those :smile:

tox.ini
```diff
@@ -0,0 +1,94 @@
...
+
+[testenv:codecov]
+deps =
+    codecov
+skip_install = true
+commands =
+    coverage xml --ignore-errors
+    codecov []
+
+[flake8]
```
Juanlu001: It's a bit strange to me to put the `flake8` config in `tox.ini`. Does it work if it's invoked _without_ tox as well? Maybe we can leave it at `setup.cfg`?

tox.ini
```diff
@@ -0,0 +1,94 @@
...
+ignore = E203, E266, E501, W503
+max-line-length = 80
+max-complexity = 18
+select = B,C,E,F,W,T4,B9
+
+[testenv:flake8]
```
Juanlu001: Do we need this environment, having the `check` one that does everything?

.travis.yml
```diff
@@ -1,40 +1,45 @@
-# The only current way of using OS X virtual machines in Travis CI
-# is setting up a generic environment
-# This is not a problem because Python will be installed through conda
-
-language: generic
-os: osx
```
Juanlu001: Beware! We were using Travis CI to test things on OS X. For Linux, it's Circle CI, and perhaps you'll need to touch `appveyor.yml` as well.

tox.ini
jorgepiloto: That is what I thought when coding this. However, and as far as I read tox.ini looks for the requirements.txt file in order to install all the dependencies. It is kind of redundant to have a requirements.txt and a setup.py or setup.cfg with the same install_requires.

I will look for more information on the topic to try solving this.

tox.ini
jorgepiloto: Sure! :+1:

tox.ini
jorgepiloto: Tox i s now intalling the dependencies from setup.cfg, so everything is right. Probably was because I let the `deps = `. So this is solved.

tox.ini
jorgepiloto: Sure! Moving to setup.cfg :+1:

jorgepiloto:
With this tox.ini configuration, previous requested changes are solved (now I feel more confortable working with this kind of files). By running `tox`, all the environments are checked: 

- clean
- check
- docs
- py35, py36, py37, pypy, pypy3

If user/developer wants to test an specific environment, just run `tox -e name_environment`, for example to check the code style just run `tox -e check`.

However I still need to work in `.travis.yml` and `appveyor.yml` files.

appveyor.yml
```diff
@@ -18,23 +18,52 @@ branches:
...
-  # Install dependencies
-  - "choco install pandoc"
+init:
+  - ps: echo $env:TOXENV
+  - ps: ls C:\Python*
 
-  - "%PYTHON_BIN% -m pip install --upgrade pip"
-  - "%PIP_BIN% install .[dev] --prefer-binary"  # Test installation correctness
-
-build: off
```
Juanlu001: You have to retain this one to remove the error :)

appveyor.yml
```diff
@@ -18,23 +18,52 @@ branches:
...
-matrix:
-  fast_finish: true
+environment:
+  matrix:
+    - TOXENV: check
+      TOXPYTHON: C:\Python36\python.exe
+      PYTHON_HOME: C:\Python36
+      PYTHON_VERSION: '3.6'
+      PYTHON_ARCH: '32'
+    - TOXENV: 'py35'
```
Juanlu001: Windows users tend to have more up-to-date Python installations, see for example https://github.com/ContinuumIO/anaconda-package-data:

Screenshot_2019-04-21 anaconda_package_data_example(1)

Therefore, I propose we remove Python 3.5 altogether, and even the 32 bit builds. The will probably slow us down for very little gain.

Juanlu001:
AppVeyor failed because of timeouts, but I think the configuration is good now.

Juanlu001:
I am merging even if I should rebase to be 100 % sure. Everything will be fine :) Thanks @jorgepiloto for your patience and dedication!

PR #651, opened 2019-04-20
Raise proper error sbdb, fixes #611

Ivan9528:
Fixes issue #611 by changing the way of dealing with the error. The error message keeps the same.

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -887,9 +887,9 @@ def test_from_sbdb():
 
         assert ss_classical == SBDB_DATA[target_name]
 
+
 def test_from_sbdb_raise_valueerror():
     with pytest.raises(ValueError) as e:
         Orbit.from_sbdb(name="Halley")
-        
-    assert str(e.value)[0]=='2'
 
+    assert str(e.value)[0] == "2"
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -892,4 +892,7 @@ def test_from_sbdb_raise_valueerror():
     with pytest.raises(ValueError) as e:
         Orbit.from_sbdb(name="Halley")
 
-    assert str(e.value)[0] == "2"
+    assert (
+        str(e.value)
+        == "2 different objects found: \n2688 Halley (1982 HG1)\n1P/Halley\n"
+    )
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -886,3 +886,10 @@ def test_from_sbdb():
         ss_classical = ss_target.classical()
 
         assert ss_classical == SBDB_DATA[target_name]
+
+def test_from_sbdb_raise_valueerror():
+    with pytest.raises(ValueError) as e:
+        Orbit.from_sbdb(name="Halley")
+        
+    assert str(e.value)[0]=='2'
+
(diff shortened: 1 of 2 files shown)


jorgepiloto:
Thanks for the pull request @Ivan9528. I saw you fetched @Himanshu-Garg pull request in order to keep his commits and contributions. However since #611 was opened, different changes into main code have been done. You rebased this branch onto master (that is perfect) but somehow you deleted useful lines of code. Also you added others that were not used.

I would recommend you to revert your branch to commit 00209294df89de874029551b869515023d88368e so you just can apply the changes requested in #611 in order to merge.

Before pushing again, you can run `tox -e check` to see if the format and tests are successful :+1:

Juanlu001:
Thanks @jorgepiloto, just wanted to say that this is probably half my fault 😅 I was instructing @Ivan9528 how to do the git stuff properly, and didn't realize some stuff was removed in the process. No big deal though :) We can bring it back!

src/poliastro/twobody/orbit.py
```diff
@@ -612,157 +593,17 @@ def parabolic(
...
-
         """
         ss = ClassicalState(attractor, p, 1.0 * u.one, inc, raan, argp, nu)
         return cls(ss, epoch, plane)
 
-    @classmethod
-    @u.quantity_input(
-        alt=u.m, inc=u.rad, argp=u.rad, raan=u.rad, arglat=u.rad, ecc=u.one
-    )
-    def frozen(
```
Juanlu001: This is the method that should be brought back :)

jorgepiloto:
Oh! No problem then :smile:

Ivan9528:
Thank you for everything! I am trying to add the stuff which is left and correct the quality issues, I will push as soon as I make it! :)

jorgepiloto:
Alright, do not worry this is going in the right direction :+1: 

There is another method that was partially removed, in this case `def plot(self, label=None, use_3d=False, static=False)`. Notice that the argument `static=False` has been removed. This is why the tests are not passing. Also, other lines related to `plot()` method documentation have disappeared. But since you already solved the other method, this will be easy now :smile: 

 I will leave you here the same requests here as @Juanlu001 asked for in #611

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -886,3 +886,10 @@ def test_from_sbdb():
         ss_classical = ss_target.classical()
 
         assert ss_classical == SBDB_DATA[target_name]
+
+
+def test_from_sbdb_raise_valueerror():
```
jorgepiloto: In this test you should check the full message

src/poliastro/twobody/orbit.py
```diff
@@ -476,6 +455,19 @@ def from_sbdb(cls, name, **kargs):
 
         obj = SBDB.query(name, full_precision=True, **kargs)
 
+        if "count" in obj:
```
jorgepiloto: As previously requested, it should be as easy as:

```python
if "count" in obj:
# Raise Error
else:
# Keep working with the algorithm
```

src/poliastro/twobody/orbit.py
```diff
@@ -1085,25 +1032,16 @@ def apply_maneuver(self, maneuver, intermediate=False):
             res = orbit_new
         return res
 
-    def plot(self, label=None, use_3d=False, static=False):
+    def plot(self, label=None, use_3d=False):
```
jorgepiloto: Check that no lines are removed in this function when pushing again. Once you have solved all requested changes, apply to your code `tox -e check`. This will format your code and you will be almost finished :tada:

Juanlu001:
It took some effort, but here we are :) Thanks a lot @Ivan9528 for this small contribution and especially for your patience, which is arguably one of the most important traits when working in open source. Merging!

PR #653, opened 2019-04-29
Changed CONTRIBUTING.rst

saurav-malani:
As per setup.cfg, installation requires python>=3.5. So, 'pip3' should be used instead of 'pip'. At times using pip might throw an error. 
Also, during my installation, I tried using relative path but it throws an error and when I tried with the absolute path everything worked fine. I don't know, if I made some blunder in relative path case, due to which I was getting the error. But, in similar scenarios, it is always a good practice to ask people to use an absolute path in the documentation. As this will be less prone to human error. So, I made changes corresponding to this also.

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -109,7 +109,7 @@ These are some succint steps to set up a development environment:
 3. `Fork poliastro <https://help.github.com/articles/fork-a-repo/>`_.
 4. `Clone your fork <https://help.github.com/articles/cloning-a-repository/>`_.
 5. Install it in development mode using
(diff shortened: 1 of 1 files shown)


saurav-malani:
@Juanlu001 can you please review the changes?

Juanlu001:
> Also, during my installation, I tried using relative path but it throws an error and when I tried with the absolute path everything worked fine. I don't know, if I made some blunder in relative path case, due to which I was getting the error.

Can you please describe which error? I normally do `pip install -e .[dev]` which works without problems.

Juanlu001:
Closing, but feel free to continue discussing here or in our chat.

PR #658, opened 2019-05-05
Do not fail propagation if there is no frame

Juanlu001:
Fix #654.

PR #660, opened 2019-05-05
Validate dimensions

Juanlu001:
Fix a subtle error introduced in #658.


Juanlu001:
cc @fatiherturk61 I think we got it now!

fatiherturk61:
It works great, thanks!

PR #661, opened 2019-05-07
Create custom CZML packets

Sedictious:
Add the ability to create custom CZML packets to work with the Sandcastle [template](https://github.com/poliastro/cesium-app/blob/master/poliastro-demo.js)

It currently accepts two parameters:

``ellipsoid`` : which defines the shape of the attractor

``pr_map``: which is the path to the ellipsoid's UV map

Note that resizing objects needs a bit more work, but this refers to the Cesium app and shouldn't have an impact on the extractor's code


Juanlu001:
Sorry for the long delay, merging this!

PR #669, opened 2019-05-14
rotation matrix for coe2rv

wselwood:
Speed up coe2rv by using a rotation matrix and applying it to both vectors rather than doing six rotations.

Master:
wselwood@SA1LT156:/git/poliastro$ python -m timeit "from poliastro.core.elements import coe2rv; coe2rv(132712442099.00002, 6, 0.6, 0.8, 1.4, 6.4, 4)"
10 loops, best of 3: 5.68 usec per loop

Test:
(venv) wselwood@SA1LT156:/git/poliastro$ python -m timeit "from poliastro.core.elements import coe2rv; coe2rv(132712442099.00002, 6, 0.6, 0.8, 1.4, 6.4, 4)"
10 loops, best of 3: 4.67 usec per loop

PR #672, opened 2019-05-15
remove pqw2ijk and update maneuver to use rotation matrix

wselwood:
The same trick used in #669 can be used in maneuver and then pqw2ijk can be removed.

This should give a nice speed boost too


Juanlu001:
:tada:

Issue #676, opened 2019-05-19
Provide a way to change orbit attractor


Juanlu001:
See for example https://gist.github.com/Juanlu001/0a730a29c6aadf69638f909b2895a006

Juanlu001:
As I wrote in #679:

> The original intent of the function was to change the attractor for the Galilean moons from the Sun to Jupiter. This makes sense, because the real attractor of these moons is Jupiter. In fact, if one takes the data of .from_horizons and just .propagate, the results are completely wrong, because the propagation happens as if Jupiter does not even exist.

> I think having the possibility to change the attractor of any orbit in this way is quite dangerous. I should have thought about the problem when writing the issue.

However, there is one place where we would benefit from a way to change the attractor:

https://docs.poliastro.space/en/stable/examples/Catch%20that%20asteroid!.html

We changed from an heliocentric orbit to a geocentric orbit **to study the osculating orbit with respect to the Earth**. The important thing to note here is that, while the object of study is within the sphere of influence of the Earth, this makes total sense, and is in fact the basis of the patched conics method.

Therefore, this is a tricky issue because _sometimes_ it makes sense to change the attractor, and sometimes it doesn't. We need to discuss in detail what cases are there before providing an error-prone API.

Juanlu001:
Let's imagine the Sun and any planet, and some body in the Solar System.

* If the body is in a **closed** orbit around the planet, it **always** make sense to have the planet as the attractor.
  - Propagating around the planet ignores the influence of any other body of the Solar System unless explicitly accounted through perturbations, and therefore gives almost perfect results.
  - Propagating around the Sun ignores the influence of the planet and therefore transforms a spiral heliocentric orbit into a elliptic orbit.
* If the body is in an **open** orbit around the planet, it **does** make sense to have the planet as the attractor _up to a certain point_, and it **does** make sense to have the Sun as the attractor.
  - Propagating around the planet is correct until the body escapes the planet sphere of influence, and therefore gives correct results which start degrading from that point.
  - Propagating around the Sun assumes the body already escaped the planet sphere of influence, which gives correct results within the patched conics assumption.
* If the body is in an orbit around the Sun (regardless of its nature), it **almost always** makes sense to have the Sun as the attractor.
  - Propagating around the Sun correctly assumes that the main gravitational influence is the Sun, **but** it ignores any flybys the object might make around nearby planets, and therefore it gives correct results within the patched conics assumption.

Juanlu001:
[...]

Issue #445, opened 2018-09-05
Offer an alternative to Orbit.propagate for the Moon and planets


astrojuanlu:
🐞 **Problem**

Currently `Orbit.from_body_ephem` is a convenient way to get the osculating orbit of a planet or the Moon for plotting, to retrieve some classical element, or other things. However, there are some inconsistencies:

* It gives the false impression that these objects can be accurately propagated or sampled, while the correct way to retrieve planetary trajectories would be directly querying the ephemerides or using [the undocumented `build_ephem_interpolant`](https://github.com/poliastro/poliastro/blob/a9a5a235e47b569542920df64b93879fab4f99b5/src/poliastro/ephem.py#L13).
* For planets, it sets the attractor to the Sun, while in all seriousness it should be the Solar System Barycenter, to match the original data.

The places where `from_body_ephem` is used are:

1. Get _approximate_ orbital elements from the planets to compute spheres of influence in `poliastro.threebody.soi`
2. Plot _approximate_ (osculating) orbits of planets
3. Get _precise_ position vectors of planets to compute Lambert arcs

🎯 **Goal**

We should provide an alternative that works equally well for all use cases but that prevents the user from shooting themself in the foot.

💡 **Possible solutions**

We can have a special "attractor", the SSB, that does not have a `k` value and therefore prevents the orbit from being propagated and sampled. However, this still does not fix the Moon case, unless we have another object which is EMB (Earth-Moon Barycenter).

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
Idea:

We have `Orbit` objects, which represent an osculating orbit. Of course, they can be propagated with any propagator, including one that includes perturbations.

Perhaps we should have another type of object that follows a similar API than `Orbit`, but whose `propagate` method does a totally different thing and, for example, looks for external ephemerides.

Sedictious:
[...]

PR #679, opened 2019-05-26
Add method change_attractor

jorgepiloto:
Tries to solve #676. Just changed syntax from gist so it can act as an Orbit method. I also included a test based on iss from the examples module.

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -46,7 +46,12 @@ from poliastro.frames import (
     VenusICRS,
     get_frame,
 )
-from poliastro.twobody.orbit import Orbit, OrbitSamplingWarning, TimeScaleWarning
+from poliastro.twobody.orbit import (
+    Orbit,
+    OrbitSamplingWarning,
+    PatchedConicsWarning,
+    TimeScaleWarning,
+)
 
 
 @pytest.fixture()
@@ -876,10 +881,8 @@ def test_from_coord_if_coord_is_not_of_shape_zero():
 
 
 def test_from_sbdb():
-
     # Dictionary with structure: 'Object': [a, e, i, raan, argp, nu, epoch]
     # Notice JPL provides Mean anomaly, a conversion is needed to obtain nu
-
     SBDB_DATA = {
         "Ceres": (
             2.76916515450648 * u.AU,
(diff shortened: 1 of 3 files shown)


Juanlu001:
Thanks for the PR @jorgepiloto! I see one danger with this:

The original intent of the function was to change the attractor for the Galilean moons from the Sun to Jupiter. This makes sense, because the real attractor of these moons is Jupiter. In fact, if one takes the data of `.from_horizons` and just `.propagate`, the results are completely wrong, because the propagation happens as if Jupiter does not even exist.

I think having the possibility to change the attractor of any orbit in this way is quite dangerous. I should have thought about the problem when writing the issue.

What do you think about making this method a private function in `poliastro.twobody.orbit` and looking for a way to fix the mismatch of `from_horizons`?

jorgepiloto:
Sure! Therefore, should we add something like:

```python
```

jorgepiloto:
Sure! What do you think about something like this?

```python
ss = Orbit.from_horizons(name, attractor=Sun, **kwargs)
```

Juanlu001:
Looking at https://astroquery.readthedocs.io/en/latest/api/astroquery.jplhorizons.HorizonsClass.html, I don't think `from_horizons` can _know_ which attractor should it use automatically. So yes, there should be a way for the user to specify that.

After thinking about your proposal a bit more (see thoughts in #676), I think `Orbit.from_horizons(name).change_attractor(Jupiter)` is a good idea. However, perhaps we could check wether the object falls inside the Sphere of Influence of the new attractor (not sure whether Laplace or Hill) and, if not, display a warning or even raise an error. What do you think @jorgepiloto?

jorgepiloto:
It would be very easy to check by using the `laplace_radius` in `poliastro.threebody.soi` if the assumption of patched conics is suitable.

The problem comes when the orbit is under this sphere of influence but its nature is parabolic or hyperbolic (flyby case). The algorithm will change the attractor, but if the user tries to propagate, there would be some point at which computations are no longer correct and attractor should be changed again, as you said in  [your comment](https://github.com/poliastro/poliastro/issues/676#issuecomment-497959261).

Knowing the point at which the previous situation occurs is easy but implement that inside the propagators methods I think may be crazy...

We could enable this method to work if two conditions are fulfilled:
* We are under the sphere of influence
* Orbit of the body is elliptical or circular (closed orbit) 

Otherwise, we raise an error for the moment. Could this be a good approach?

Juanlu001:
> Knowing the point at which the previous situation occurs is easy but implement that inside the propagators methods I think may be crazy...

Actually, now that [SciPy propagators have support for events](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html), it's not that difficult :wink: but let's leave that for a future issue.

I agree with raising an error if the orbit is outside the sphere of influence of the target attractor with respect to the Sun. However, let's not consider the eccentricity here, to allow open orbits as well.

Juanlu001:
Wow, investigating these failures.

Juanlu001:
Oops, we introduced a circular import:

```
  File "/home/circleci/repo/venv/lib/python3.7/site-packages/poliastro/twobody/__init__.py", line 1, in <module>
    from .orbit import Orbit
  File "/home/circleci/repo/venv/lib/python3.7/site-packages/poliastro/twobody/orbit.py", line 22, in <module>
    from poliastro.threebody.soi import laplace_radius
  File "/home/circleci/repo/venv/lib/python3.7/site-packages/poliastro/threebody/soi.py", line 32, in <module>
```

Juanlu001:
The `Orbit` object is used in the SOI module to... Compute the classical elements of the planets!

https://github.com/jorgepiloto/poliastro/blob/dbbf332386e135eb64d0bf62d114350deb5c072d/src/poliastro/threebody/soi.py#L54-L55

This leads me to two things:

* My dreaded `from_body_ephem` strikes back https://github.com/poliastro/poliastro/issues/445 (see some comments at the end about sbpy)
* We had a long discussion with @AunSiro about adding some "mean" elements to the `Body` classes that we could use for this kind of thing https://github.com/poliastro/poliastro/pull/153

So my solution to this circular import is to make the `Orbit` import in `poliastro.threebody.soi` "delayed" while we figure out a better way.

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -915,3 +916,9 @@ def test_orbit_wrong_dimensions_fails():
     with pytest.raises(ValueError) as excinfo:
         Orbit.from_vectors(Earth, bad_r, bad_v)
     assert "ValueError: Vectors must have dimension 1, got 2 and 3" in excinfo.exconly()
+
+
+def test_orbit_change_attractor_error():
+    with pytest.raises(ValueError) as excinfo:
+        iss.change_attractor(Sun)
```
Juanlu001: It turns out that computing the radius of the sphere of influence of the Sun makes no sense :)

> In the patched conic approximation, once an object leaves the planet's SOI, the primary/only gravitational influence is the Sun (until the object enters another body's SOI).

https://www.wikiwand.com/en/Sphere_of_influence_(astrodynamics)

I think we need two kinds of errors:

* If the attractor is set to the `Sun` and we do `set_attractor(new_attractor)`, we check the SOI of `new_attractor` and we raise the error you wrote here (`"ValueError: Body is out of new attractor's SOI."`)
* If the attractor is set to `attractor` and we do `set_attractor(attractor.parent)`, we check the SOI of `attractor` and, _if we are still inside the SOI of `attractor`_:
  - If the orbit is closed, raise `"ValueError: Body will never leave the SOI of its current attractor"`
  - If the orbit is open, just do it! (I think emitting a warning here might be too noisy, but we could as well say `"PatchedConicsWarning: Leaving the SOI of the current attractor"`

What do you think @jorgepiloto?

jorgepiloto: I think raising these 2 errors and the warning, depending on the case, it is very useful. Hope to have this in a couple of minutes :+1:

Juanlu001:
I made several changes to your code:

* Made it more general: instead of checking for the `Sun`, I'm checking for `attractor.parent` (in case people want to define their own planetary systems)
* Removed the capability to change the plane to simplify the code (probably we should have another method for that)
* Implement a short-circuit in case the target and current attractor are the same
* Allow only changes from attractor to parent and vice-versa

Juanlu001:
Aaaaaand merging! Thanks @jorgepiloto to get the ball rolling!

Issue #114, opened 2016-01-27
Lambert transfers should be Maneuvers


Juanlu001:
After writing http://nbviewer.jupyter.org/github/poliastro/poliastro/blob/6d5b231/examples/Exploring%20the%20New%20Horizons%20launch.ipynb#From-Earth-to-Jupiter I realized that the API for the Lambert problem is not consistent with the rest of maneuvers. I already considered this in https://github.com/poliastro/poliastro/wiki/API-Refactor:-Maneuver-objects.

Juanlu001:
With the arrival of multi revolution solutions this is a bit unclear. I'm postponing the changes to next version.

Juanlu001:
See https://github.com/poliastro/poliastro/wiki/API-Refactor:-Multi-revolution-Lambert-problem for details.

Juanlu001:
It seems that @nikita-astronaut and I are doing some work on Maneuvers in the coming months, so I'm moving this to the next milestone.

Juanlu001:
Out of time to make this work for this release, but we will be in a good position for the next one, also because we will be able to combine it with the flybys that we just introduced.

Juanlu001:
Tons of work on reference frames and documentation for the next sprint. Postponing.

Juanlu001:
Addressed in #680! It would be great to update [the User Guide](https://docs.poliastro.space/en/latest/user_guide.html#traveling-through-space-solving-the-lambert-problem) accordingly

jorgepiloto:
Should we also update the notebooks [Going to Mars with Python using poliastro](https://docs.poliastro.space/en/latest/examples/Going%20to%20Mars%20with%20Python%20using%20poliastro.html) and [Going to Jupiter with Python using Jupyter and poliastro](https://docs.poliastro.space/en/latest/examples/Going%20to%20Jupiter%20with%20Python%20using%20Jupyter%20and%20poliastro.html)?

Juanlu001:
Definitely!

Juanlu001:
More than three years later, this is in :)

Issue #495, opened 2018-11-15
Fix minor issues in Lambert


Juanlu001:
_Follow-up from #492_

Several things to do:

* [x] Fix chord calculation, see https://github.com/poliastro/poliastro/issues/492#issuecomment-438808398
* [x] Improve error message when the vectors are parallel, see https://github.com/poliastro/poliastro/issues/492#issuecomment-439071294
* [ ] Remove misleading example from Lambert notebook `(_, v_l), (_, v_r) = izzo.lambert(k, r0, r, tof, M=1)` (do not remember what the intention was here), see https://space.stackexchange.com/questions/31967/multiple-revolution-lambert-problem/32020#comment97323_32020
* [ ] Clarify multi-revolution API (isn't the number of branches always 1 or 2? is it necessary to use a generator?), no need to wait for #114

Juanlu001:
The first two have been addressed in #505.

Juanlu001:
@jorgepiloto now that we understand the algorithm a bit better, perhaps we can keep the "low level Lambert API" in this notebook https://docs.poliastro.space/en/latest/examples/Revisiting%20Lamberts%20problem%20in%20Python.html#Multiple-revolutions but decide whether we still want to use a generator, or at least change the `(_, v_l), (_, v_r) = izzo.lambert(k, r0, r, tof, M=1)`?

Juanlu001:
Done!

Issue #348, opened 2018-04-20
Examine solutions in Lambert notebook


Juanlu001:
🐞 **Problem**

With the current master version, I'm getting solutions in strange locations of the plot:
![png image 1200 x 900 pixels](https://user-images.githubusercontent.com/316517/39049730-00060cda-44a3-11e8-88bf-54aeeb2dcfb5.png)

I am not sure this was always the case, so it's worth it to rerun the notebook and assess whether these results are correct.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
This has been like that at least for two years https://docs.poliastro.space/en/v0.7.0/examples/Revisiting%20Lambert's%20problem%20in%20Python.html perhaps it was never quite right?

Juanlu001:
The plot is taken from the Izzo paper:

![Screenshot_2019-06-06 1403 2705 pdf](https://user-images.githubusercontent.com/316517/59024962-5dddf080-8853-11e9-91ec-df2409a32501.png)

https://arxiv.org/abs/1403.2705

jorgepiloto:
I also saw that the labels `Elliptic` and `Hyperbolic` are switched, I was uploading this image but you were faster than me :smile:

jorgepiloto:
The current state of this:

1. After comparing the poliastro Izzo's algorithm against [pykep](https://esa.github.io/pykep/) it got exactly the same results, so algorithm implementation seems to be right.

2. I focused my attention on the `_compute_T_min` private function and realized that wrong results were obtained only for those values of `ll = -1`. If we increase the `y-axis` limit to also see the solutions `M=3` we can notice that red crosses corresponding to `ll = -1` are the problematic points:

![problematic_lambert](https://user-images.githubusercontent.com/28702884/59948217-b4f8dd80-946f-11e9-9ff7-023ba7473352.png)

3. This value of lambda is a problematic one as [this comment](https://github.com/poliastro/poliastro/blob/c9afdc9b5fcbd666f8da3d0c231842a13f72d3bb/src/poliastro/core/iod.py#L368) states. Therefore, I will try to further inspect this code section :+1:

jorgepiloto:
[...]

PR #680, opened 2019-05-27
Lambert becomes Maneuver

jorgepiloto:
This tries to solve for #114. I added the option to pass as an argument a lambert solver. The idea was taken from the method `propagate` in the Orbit class. Just a simple change in porkchops and they keep working without problems.

Some things still need to be done:

* Correct notebooks and documentation
* When adding new tests, should we move the ones from `iod` module to the maneuver one? Or do we implement new ones with the same data? 

But before correcting them I just wanted to be sure if you think this is the way on solving this :rocket:

### DIFF
--- a/src/poliastro/plotting/porkchop.py
+++ b/src/poliastro/plotting/porkchop.py
@@ -19,7 +19,8 @@ from poliastro.bodies import (
     Uranus,
     Venus,
 )
-from poliastro.iod import lambert
+from poliastro.maneuver import Maneuver
+from poliastro.twobody.orbit import Orbit
 from poliastro.util import norm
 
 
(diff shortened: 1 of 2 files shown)


jorgepiloto:
> This looks very good @jorgepiloto!
> 
> While we are at this, do you think we can take care of #495?

Sure! I also saw #348. I will start working on all of them this afternoon :+1:

src/poliastro/plotting/porkchop.py
```diff
@@ -48,29 +49,36 @@ def _get_state(body, time):
...
+    # Define time of flight
+    tof = ss_arr.epoch - ss_dpt.epoch
 
     if tof <= 0:
         return None, None, None, None, None
 
     try:
-        (v_dpt, v_arr), = lambert(Sun.k, rr_dpt_body.xyz, rr_arr_body.xyz, tof)
+        # Lambert is now a Maneuver object
+        man_lambert = Maneuver.lambert(ss_dpt, ss_arr)
```
Juanlu001: Much nicer!

PR #688, opened 2019-06-05
Add trail plotting option

jorgepiloto:
While working on the different Lambert's issues I thought it may be useful to have a faded trail plotting option. Just pass the keyword  `trail=True` and the path of the orbit will be shown. As an example:

```python
plotter = StaticOrbitPlotter()                                     
plotter.plot(iss, trail=True, color="red")                                  
plt.show()
```
plot


Juanlu001:
This looks very nice! It reminds me of this library we mentioned in the chat a while ago: https://github.com/hannorein/rebound/blob/master/ipython_examples/OrbitPlot.ipynb

I have two requests:

* From the implementation point of view, perhaps we could make it simpler? https://matplotlib.org/3.1.0/gallery/lines_bars_and_markers/multicolored_line.html
* From the visual point of view, could we make the gradient in a way the complete orbit can be seen? (as in rebound)

jorgepiloto:
Regarding the implementation of the algorithm I tried the suggested `multicolored lines`. The problem is that it does not work with `colors` but instead with `cmap`. Also, we can not make use of the method `get_color` in `LineCollection` instances... Both things increase the difficulty of working with `multicolored lines`.

But plotting the whole orbit is not a huge problem. In fact, it looks nicer than just plotting some part of it:

earth

Juanlu001:
> Also, we can not make use of the method get_color in LineCollection instances...

What do you mean? In principle it's listed there https://matplotlib.org/3.1.0/api/collections_api.html#matplotlib.collections.LineCollection.get_color and also https://matplotlib.org/3.1.0/api/collections_api.html#matplotlib.collections.LineCollection.get_colors

> The problem is that it does not work with colors but instead with cmap

True, but maybe we can do a colormap between the target color and the same color with `alpha=0`?

Juanlu001:
What I'm concerned is that we are plotting the points one by one, and that makes the appearance of the line less smooth and perhaps difficults the rendering

src/poliastro/tests/tests_plotting/test_static.py
```diff
@@ -116,3 +117,12 @@ def test_redraw_keeps_trajectories():
     op.set_frame(*mars.pqw())
 
     assert len(op.trajectories) == 2
+
+
+@pytest.mark.mpl_image_compare
+def test_trail_plotting():
```
Juanlu001: I think this test is not testing anything. Tried adding a red thick line in the image and even removing it, and still passed.

Juanlu001:
Oh, we need to use `pytest --mpl` for that!

Juanlu001:
This will need a rebase after https://github.com/poliastro/poliastro/pull/693 is merged.

jorgepiloto:
Ups! Sorry for the error uploading the test. Now it is working properly :+1:

Juanlu001:
Merging, thanks a lot @jorgepiloto!

PR #695, opened 2019-06-10
Perifocal frame and plotter set_frame

astrojuanlu:
This is a proof of concept of how a `*Plotter.set_frame` could work. I am not very happy with the implementation, but to be honest working with Astropy frames is a little bit hacky at times :) Before merging this, I would like to:

* [ ] Test the Geocentric Solar Ecliptic (GSE) frame that @shreyasbapat implemented a while ago to reproduce [these beautiful plots](https://github.com/poliastro/poliastro/issues/335#issue-302625991)
* [ ] Add more image tests (now that we know how they work) to make this robust in the face of redraws
* [ ] Clean the implementation
* [ ] Evaluate a better name? Perhaps we can use `get_pqw` or `get_pqw_frame` instead of `get_perifocal_frame`. Thoughts?
* [ ] Translate functionality to interactive plotters for consistency
* [ ] Deprecate `orbit.pqw`

### DIFF
--- a/src/poliastro/plotting/misc.py
+++ b/src/poliastro/plotting/misc.py
@@ -27,15 +27,16 @@ def _plot_bodies(orbit_plotter, outer=True, epoch=None):
 
 
 def _plot_solar_system_2d(outer=True, epoch=None, interactive=False):
-    pqw = Orbit.from_body_ephem(Earth, epoch).pqw()
     if interactive:
         orbit_plotter = (
             OrbitPlotter2D()
         )  # type: Union[OrbitPlotter2D, StaticOrbitPlotter]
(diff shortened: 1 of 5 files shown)


astrojuanlu:
Oh wow, I can't even imagine how I started implementing this without doing #876 first.

astrojuanlu:
* The `poliastro.plotting` code is horrible, hopefully #876 addresses that a bit
* This is still coupled with `Orbit.from_body_ephem`, which we intend to remove
* The `get_perifocal_frame` code is horrible because... well, Astropy magic

I'm closing this, and hope I can address it... by the end of this year.

PR #696, opened 2019-06-10
Add ipywidgets to tox -e images

jorgepiloto:
Just adding `ipywidgets` to `images` environment in `tox` :+1:


Juanlu001:
Oh, now I remember why these tests were passing - because Circle CI does not use tox (yet?) and we didn't add image checking to AppVeyor. Anyway, this is good to go!

Issue #665, opened 2019-05-09
One failing test in Astropy 3.2


Juanlu001:
One test from the new `GeocentricSolarEcliptic` frame is failing, probably "just" because we hardcoded the values?

```
=============================================================== FAILURES ===============================================================
______________________________________________ test_GeocentricSolarEcliptic_against_data _______________________________________________

    def test_GeocentricSolarEcliptic_against_data():
        gcrs = GCRS(ra="02h31m49.09s", dec="+89d15m50.8s", distance=200 * u.km)
        gse = gcrs.transform_to(GeocentricSolarEcliptic(obstime=Time("J2000")))
        lon = 233.11663895663975
        lat = 48.64652559835358
>       assert_quantity_allclose(gse.lat.value, lat, atol=1e-7)

src/poliastro/tests/test_frames.py:129: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

actual = 48.64606410986667, desired = 48.64652559835358, rtol = 1e-07, atol = 1e-07, kwargs = {}
np = <module 'numpy' from '/home/juanlu/.miniconda36/envs/poliastro37/lib/python3.7/site-packages/numpy/__init__.py'>
_unquantify_allclose_arguments = <function _unquantify_allclose_arguments at 0x7fd6ca3e0400>

    def assert_quantity_allclose(actual, desired, rtol=1.e-7, atol=None,
                                 **kwargs):
        """
        Raise an assertion if two objects are not equal up to desired tolerance.
    
        This is a :class:`~astropy.units.Quantity`-aware version of
        :func:`numpy.testing.assert_allclose`.
        """
        import numpy as np
        from astropy.units.quantity import _unquantify_allclose_arguments
        np.testing.assert_allclose(*_unquantify_allclose_arguments(
>           actual, desired, rtol, atol), **kwargs)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=1e-07
E       
E       (mismatch 100.0%)
E        x: array(48.646064)
E        y: array(48.646526)

../../../.miniconda36/envs/poliastro37/lib/python3.7/site-packages/astropy/tests/helper.py:478: AssertionError
```

[...]

PR #699, opened 2019-06-11
Update user's guide to new Lambert maneuvers

jorgepiloto:
This tries to update the user's guide and also the different notebooks that make use of the Lambert algorithm, so they can be consistent with the new nature of `Maneuver.lambert`.

I still need to update the [Going to Mars](https://docs.poliastro.space/en/latest/examples/Going%20to%20Mars%20with%20Python%20using%20poliastro.html) example. 

Just decided to start the PR for the reviewing process :+1:


Juanlu001:
Tests failing because of https://github.com/poliastro/poliastro/issues/665

jorgepiloto:
I simplified Mars' notebook by making use of `from_body_epehem` constructor and the `sample` method. However, you better tell me if it looks good or something needs to be changed :+1:

Juanlu001:
Thanks a lot @jorgepiloto!

PR #700, opened 2019-06-12
Update tests to astropy 3.2

jorgepiloto:
Just a small update on some tests due to astropy version 3.2 :rocket: 

Solving #665

### DIFF
--- a/src/poliastro/tests/test_frames.py
+++ b/src/poliastro/tests/test_frames.py
@@ -123,9 +123,9 @@ def test_round_trip_from_GeocentricSolarEcliptic_gives_same_results():
 
 def test_GeocentricSolarEcliptic_against_data():
     gcrs = GCRS(ra="02h31m49.09s", dec="+89d15m50.8s", distance=200 * u.km)
-    gse = gcrs.transform_to(GeocentricSolarEcliptic(obstime=Time("J2000")))
-    lon = 233.11663895663975
-    lat = 48.64652559835358
+    gse = gcrs.transform_to(GeocentricSolarEcliptic(obstime=J2000))
+    lon = 233.11691362602866
+    lat = 48.64606410986667
     assert_quantity_allclose(gse.lat.value, lat, atol=1e-7)
     assert_quantity_allclose(gse.lon.value, lon, atol=1e-7)
 
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -45,7 +45,6 @@ from poliastro.frames import (
     VenusICRS,
     get_frame,
 )
-from poliastro.twobody.angles import M_to_nu
 from poliastro.twobody.orbit import Orbit, OrbitSamplingWarning, TimeScaleWarning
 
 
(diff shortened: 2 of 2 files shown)


jorgepiloto:
This push should be one. Sorry for the delay solving this, not sure why the tests were running properly on my local machine :confused:

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -591,12 +590,12 @@ def test_orbit_from_horizons_has_expected_elements():
     # Taken from https://ssd.jpl.nasa.gov/horizons.cgi
     ss = Orbit.from_classical(
         Sun,
-        2.767107592216510 * u.au,
-        7.554803091400027e-2 * u.one,
-        2.718502494739172e1 * u.deg,
-        2.336913218336299e1 * u.deg,
-        1.322919809219236e2 * u.deg,
-        2.128957916690369e1 * u.deg,
+        2.767107584017257 * u.au,
```
Juanlu001: These tests are going to fail :) See: https://matrix.to/#/!DdazxkLDTPnBLekbbL:openastronomy.org/$1560168960782926ZNgkh:matrix.org?via=openastronomy.org&via=matrix.org

src/poliastro/tests/test_frames.py
```diff
@@ -124,8 +124,8 @@ def test_round_trip_from_GeocentricSolarEcliptic_gives_same_results():
 def test_GeocentricSolarEcliptic_against_data():
     gcrs = GCRS(ra="02h31m49.09s", dec="+89d15m50.8s", distance=200 * u.km)
     gse = gcrs.transform_to(GeocentricSolarEcliptic(obstime=Time("J2000")))
-    lon = 233.11663895663975
-    lat = 48.64652559835358
+    lon = 233.1169136260286
```
Juanlu001: Do we know _why_ the numbers are different here?

jorgepiloto: They were failing with my previous local caché but with the updated one this is not showing as a failing test anymore.

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -591,12 +590,12 @@ def test_orbit_from_horizons_has_expected_elements():
     # Taken from https://ssd.jpl.nasa.gov/horizons.cgi
     ss = Orbit.from_classical(
         Sun,
-        2.767107592216510 * u.au,
-        7.554803091400027e-2 * u.one,
-        2.718502494739172e1 * u.deg,
-        2.336913218336299e1 * u.deg,
-        1.322919809219236e2 * u.deg,
-        2.128957916690369e1 * u.deg,
+        2.76710759221651 * u.au,
```
Juanlu001: I was squinting at this and then realized that you just "moved the comma" :)

jorgepiloto: Oh! Yeah, I thought it was more natural removing the scientific notation.

Juanlu001:
Now the tests passed, but the question is: where in this code did the change happen? https://github.com/poliastro/poliastro/blob/f66c46c3925f97842b3033bc5219042b01a53c36/src/poliastro/frames.py#L271-L297

Juanlu001:
Perhaps @shreyasbapat has more information?

jorgepiloto:
This may have the key:

[What's new in Astropy 3.2?](https://docs.astropy.org/en/v3.2/whatsnew/3.2.html)

Juanlu001:
Spot the differences :wink: 

Screenshot from 2019-06-13 14-56-14

src/poliastro/tests/test_frames.py
```diff
@@ -124,8 +124,8 @@ def test_round_trip_from_GeocentricSolarEcliptic_gives_same_results():
 def test_GeocentricSolarEcliptic_against_data():
     gcrs = GCRS(ra="02h31m49.09s", dec="+89d15m50.8s", distance=200 * u.km)
     gse = gcrs.transform_to(GeocentricSolarEcliptic(obstime=Time("J2000")))
```
Juanlu001: Here is the problem: we should have used `poliastro.constants.J2000`, which explicitly specifies the scale. My fault for not spotting it when I reviewed the code.

src/poliastro/tests/test_frames.py
Juanlu001: See my comment below.

Juanlu001:
Source: https://docs.astropy.org/en/v3.2/changelog.html#id21

src/poliastro/tests/test_frames.py
jorgepiloto: Great! Thank you @Juanlu001, I saw in the changelog that the default scale was changed but thought it was because of the `float64`. Solving immediately :+1:

jorgepiloto:
Since the expected values were in `utc` scale, we finally decided to rewrite them in the `tt` one, according to `J2000`.

PR #701, opened 2019-06-12
Use sphinx-notfound-page extension to build a nice 404 page

humitos:
When Read the Docs finds a 404.html page in the project it will serve it instead the default one.

The extension installed, takes care about all the links being generated properly (absolute) so the page can be served at any URL where the page is not found.

Screenshot_2019-06-12_13-00-57

> NOTE: there are some extra configurations that can be tweaked in case you want to change the default message: https://sphinx-notfound-page.readthedocs.io/en/latest/configuration.html

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -44,6 +44,7 @@ extensions = [
     'IPython.sphinxext.ipython_console_highlighting',
     'sphinx.ext.mathjax',  # Maths visualization
     'sphinx.ext.graphviz',  # Dependency diagrams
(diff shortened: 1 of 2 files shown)


Juanlu001:
Awesome! I know I told you I would do it, so thanks for doing it yourself :)

The tests failed because of https://github.com/poliastro/poliastro/issues/665, so I am merging this right away and let's see how it goes.

humitos:
It seems it works: https://docs.poliastro.space/en/latest/not/found/page.html

PR #708, opened 2019-06-22
Use canonical units in core functions

### LINKED ISSUE #707: Evaluate use of canonical units
I have long been thinking about this, but this bit in the [pykep documentation](https://esa.github.io/pykep/documentation/core.html#pykep.lambert_problem.__init__) triggered the idea again:

> Example (non-dimensional units used)::
> ```
> l = lambert_problem([1,0,0],[0,1,0],5 * pi / 2. )
> ```

In our two-layered architecture, we are already normalizing to kilometers in some places, see for example

https://github.com/poliastro/poliastro/blob/c9afdc9b5fcbd666f8da3d0c231842a13f72d3bb/src/poliastro/iod/izzo.py#L37-L42

we would have to just go a step further, normalize everything with respect to the gravitational parameter, and have one parameter less in the `core` functions.

Possible benefits:

* Fewer operations (only noticeable if they happen in a loop, otherwise we would get a micro-optimization at least)
* More predictable behavior for corner cases? See #475

astrojuanlu:
First attempt to fix #707.

* Introduced new Halley orbit example (this can be splitted off this PR)
* All tests passing except for long propagation with Kepler
* Pending moving Hohmann and bielliptic calculations to fast functions
* Pending more uniform unit conversion

### DIFF
--- a/src/poliastro/coordinates.py
+++ b/src/poliastro/coordinates.py
@@ -9,6 +9,7 @@ import astropy.units as u
 import numpy as np
 
 from poliastro.core.elements import rv2coe
+from poliastro.util import norm
 
 
 def inertial_body_centered_to_pqw(r, v, source_body):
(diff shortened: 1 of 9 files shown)


astrojuanlu:
Bringing this back to life would need making a massive rebase. If I ever do it, I will start from scratch. Closing!

Issue #168, opened 2017-05-03
Split and optimize the libraries to make Poliastro lighter


meristian:
In function of make Poliastro lighter I open this issue.

Optimize and rewrite the code of the library itself and their libraries (numpy , scipy , etc.) will be the goal of this Issue. We can learn much more about the libraries we use , know wich parameters and variables are wrongly used.

Watching how will coders apply this library , Poliastro , make it lighter to pave the way for mobile devices and , why not , videogames. Poliastro is useful to make plotters of spacial travels but also is the easyest way to make computation of pyshic parameters.

Speed up all the apps that will be done with this library is the main purpouse.

I will updating this issue with another ways to optimize our code.

Juanlu001:
Thanks @Thextroyer for the idea! We can leave this as issue as a TODO item, but further information should go to the wiki as the proposal might be complex. If you intent to work on this for SOCIS, I will send the information to the mailing list very soon :)

Juanlu001:
I'm removing the SOCIS label, but leaving the issue open in case anyone wants to work on it :)

Juanlu001:
I am not sure whether the optimization should happen in poliastro itself, an alternative "simpler" implementation should be made or a server-client architecture using JavaScript/TypeScript should be explored. I'm marking this as a question for the future.

Juanlu001:
@ZachDischner said in https://github.com/poliastro/poliastro/issues/156#issuecomment-413333078 that he was having problems to zip all poliastro dependencies to AWS Lambda:

> Lambda wants you to upload a zip of the whole environment to use during runtime. Poliastro dependencies make the zip too big for their size limits (and it is a pain in the butt to build libs for Amazon Linux). 

I think that's a pity because probably the biggest contributor to dependencies weight is SciPy, and we only use a couple of packages in it. Also, for a setting like this you probably could get away without Jupyter and choosing one plotting backend (either Plotly or matplotlib). I would like to put some effort here to make poliastro lighter. Perhaps something like the Dask devs do:

https://github.com/dask/dask/blob/e7a71c092def9d8d16580bfa3456e6b85334656b/setup.py#L10-L17

Juanlu001:
I read that the limit for code uploaded to AWS Lambda is 250 MB https://hackernoon.com/exploring-the-aws-lambda-deployment-limits-9a8384b0bec3 @ZachDischner did all the poliastro dependencies weigh more than this, uncompressed?

Juanlu001:
[...]

PR #711, opened 2019-06-24
Use czml3 for the czml extractor

Sedictious:
This pretty much completely removes the dictionary-based czml representation of the extractor's previous implementation and uses poliastro's czml3 library instead.

(Note that there is a minor bug in the main application where the optional map needs to be defined, hence why the UV map doesn't default to none. I'll try to resolve it asap)


Juanlu001:
Please rebase on current master and fix the quality issues at your earliest convenience :) Remember you can do `tox -e reformat` and they will be fixed automatically.

Juanlu001:
Sorry, I spoke too fast. Leaving inline comments about mypy.

src/poliastro/czml/extract_czml.py
```diff
@@ -35,10 +44,8 @@ def __init__(self, start_epoch, end_epoch, N, ellipsoid=None, pr_map=None):
             of sampled data points will be N when calling
             add_orbit()
         """
-        self.czml = dict()  # type: Dict[int, Any]
-        self.cust_czml = dict()  # type: Dict[int, Any]
+        self.packets = []
```
Juanlu001:
```suggestion
        self.packets = []  # type: List[Packet]
```

src/poliastro/czml/extract_czml.py
```diff
@@ -129,7 +67,13 @@ def _init_orbit_packet_cords_(self, i):
         ----------
         i: int
             Index of referenced orbit
+
+        Returns
+        -------
+        coordinate list
         """
+        cart_cords = []
```
Juanlu001:
```suggestion
        cart_cords = []  # type: List[?]
```

(not entirely sure what line 85 does)

src/poliastro/czml/extract_czml.py
```diff
@@ -138,34 +82,28 @@ def _init_orbit_packet_cords_(self, i):
             cords = position.represent_as(CartesianRepresentation).xyz.to(u.meter).value
             cords = np.insert(cords, 0, h.value * k, axis=0)
 
-            self.czml[i]["position"]["cartesian"] += list(
-                map(lambda x: x[0], cords.tolist())
-            )
+            cart_cords += list(map(lambda x: x[0], cords.tolist()))
```
Juanlu001: Is there a more straightforward way to write this? It's a bit cryptic :sweat:

src/poliastro/czml/extract_czml.py
```diff
@@ -403,29 +187,43 @@ def add_ground_station(
...
-            id_description=id_description,
-            label_fill_color=label_fill_color,
-            label_font=label_font,
-            label_outline_color=label_outline_color,
-            label_text=label_text,
-            label_show=label_show,
+        pckt = Packet(
+            id="GS" + str(self.gs_n),
+            description=id_description,
+            availability=self.start_epoch.value + "/" + self.end_epoch.value,
```
Juanlu001: Can't we use a `TimeInterval` from `czml3` here?

src/poliastro/czml/extract_czml.py
```diff
@@ -403,29 +187,43 @@ def add_ground_station(
...
+            position=Position(cartesian=pos),
+            label=Label(
+                show=label_show,
+                text=label_text,
+                font=label_font if label_font is not None else "11pt Lucida Console",
```
Juanlu001: Labels have a default font of `30px sans-serif` https://github.com/AnalyticalGraphicsInc/czml-writer/wiki/Label#properties do we need to set it here?

src/poliastro/czml/extract_czml.py
```diff
@@ -403,29 +187,43 @@ def add_ground_station(
...
+                fillColor=Color(rgba=label_fill_color)
+                if label_fill_color is not None
+                else None,
+                outlineColor=Color(rgba=label_outline_color)
+                if label_outline_color is not None
+                else None,
+            ),
+            billboard=Billboard(
+                image=PIC_GROUNDSTATION,
+                horizontalOrigin="CENTER",
```
Juanlu001: I think we can replace these by `czml3.enums.HorizontalOrigins.CENTER`

src/poliastro/czml/extract_czml.py
```diff
@@ -403,29 +187,43 @@ def add_ground_station(
...
+                verticalOrigin="CENTER",
```
Juanlu001: Same for `czml3.enums.VerticalOrigins.CENTER`

Juanlu001: However, again these are the defauls and probably don't need to be specified https://github.com/AnalyticalGraphicsInc/czml-writer/wiki/Billboard#properties

src/poliastro/czml/extract_czml.py
```diff
@@ -496,28 +285,58 @@ def add_orbit(
...
-            outline_color=label_outline_color,
-            font=label_font,
-            text=label_text,
-            show=label_show,
+
+        pckt = Packet(
+            id=self.i,
+            name=id_name,
+            description=id_description,
+            availability=start_epoch.value + "/" + self.end_epoch.value,
```
Juanlu001: Same as above

Sedictious:
@Juanlu001 I run tox and was so ready it was going to pass circleci. Turns out a failing test is one missing type annotation away 😅

src/poliastro/czml/extract_czml.py
```diff
@@ -496,28 +285,58 @@ def add_orbit(
...
+                else Material(
+                    solidColor=SolidColorMaterial(color=Color(rgba=[255, 255, 0, 255]))
+                ),
+                resolution=120,
+            ),
+            label=Label(
+                text=label_text,
+                font=label_font,
+                show=label_show,
+                fillColor=Color(rgba=label_fill_color)
```
Juanlu001: The default is already `white` https://github.com/AnalyticalGraphicsInc/czml-writer/wiki/Label#properties so probably we don't need to specify it either?

src/poliastro/czml/extract_czml.py
```diff
@@ -496,28 +285,58 @@ def add_orbit(
...
+                if label_fill_color is not None
+                else Color(rgba=[255, 255, 0, 255]),
+                outlineColor=Color(rgba=label_outline_color)
+                if label_outline_color is not None
+                else Color(rgba=[255, 255, 0, 255]),
+            ),
+            billboard=Billboard(
+                horizontalOrigin="CENTER",
```
Juanlu001: Same as above

src/poliastro/czml/extract_czml.py
```diff
@@ -496,28 +285,58 @@ def add_orbit(
...
+                verticalOrigin="CENTER",
+                image=PIC_SATELLITE,
+                scale=1,
```
Juanlu001: Scale is `1.0` by default https://github.com/AnalyticalGraphicsInc/czml-writer/wiki/Billboard#properties

setup.cfg
```diff
@@ -65,6 +65,7 @@ jupyter = notebook; ipywidgets>=7.0
 dev =
     black ; python_version>='3.6'
     coverage
+    czml3
```
Juanlu001: Hmmm let's add it under `install_requires` instead until we sort https://github.com/poliastro/poliastro/issues/168#issuecomment-421963430 out

Juanlu001:
> As for adding czml3 as a dependency, is there anything more I'll need to do?

Good question, just left a comment for it. There should be a dedicated section for `czml3` (instead of `dev`) but let's not worry much about that now.

src/poliastro/czml/extract_czml.py
Sedictious: I figured that it would be better for the default to be yellow because it's what the original demo uses and in my opinion it looks more "clear". Same thing I changed the font default: 30pt is way too big for a label and it took up half the view-port

src/poliastro/czml/extract_czml.py
Sedictious: This simply flattens the list (since ``cords.tolist()`` returns a list of lists). I guess we could also use something like ``itertools.chain``?

Juanlu001: Oh, I see! I think `itertools.chain` is better recognized as "flattening the list" but, whether you change it or not, please add a comment explaining

src/poliastro/czml/extract_czml.py
```diff
@@ -190,7 +191,9 @@ def add_ground_station(
         pckt = Packet(
             id="GS" + str(self.gs_n),
             description=id_description,
-            availability=self.start_epoch.value + "/" + self.end_epoch.value,
+            availability=TimeInterval(
+                start=self.start_epoch.datetime, end=self.end_epoch.value
```
Juanlu001: Why `.datetime` and `.value`?

src/poliastro/czml/extract_czml.py
```diff
@@ -287,15 +284,15 @@ def add_orbit(
...
-        )
+        start_epoch = Time(min(self.orbits[self.i][2], self.start_epoch), format="isot")
 
         pckt = Packet(
             id=self.i,
             name=id_name,
             description=id_description,
-            availability=start_epoch.value + "/" + self.end_epoch.value,
+            availability=TimeInterval(
+                start=self.start_epoch.datetime, end=self.end_epoch.value
```
Juanlu001: Same as above

src/poliastro/czml/extract_czml.py
Juanlu001: Perfect then!

src/poliastro/czml/extract_czml.py
```diff
@@ -129,7 +67,13 @@ def _init_orbit_packet_cords_(self, i):
         ----------
         i: int
             Index of referenced orbit
+
+        Returns
+        -------
+        coordinate list
         """
+        cart_cords = []  # type: List[float]
```
Juanlu001: :+1:

Juanlu001:
Some tests are failing because of... timezones :smile: 

```
E         -     "availability": "2013-03-18T12:00:00Z/2013-03-18T23:59:35.108",
E         ?                                  ^
E         +     "availability": "2013-03-18T10:00:00Z/2013-03-18T23:59:35.108",
E         ?  
```

Juanlu001:
Interestingly, there are also some small precision issues. This will need closer inspection.

Juanlu001:
The `test_GeocentricSolarEcliptic_against_data` is failing because this branch still needs a rebase. @Sedictious when you have some spare 20 minutes ping me privately :)

Juanlu001:
This PR has been in the works for far too long already, so I decided to merge it straight away and we can fix any remaining issues afterwards :) Thanks @Sedictious for the hard work!

PR #713, opened 2019-06-25
Trigger CI

Juanlu001:
Investigating test failures seen in https://github.com/poliastro/poliastro/pull/711.


Juanlu001:
Nothing to see here, #711 needs a rebase.

PR #717, opened 2019-07-01
Additional orbit properties

Sedictious:
Adds 3 orbit properties:

* Angular momentum

* Mean anomaly

* Time elapsed from last perifocal passage

* Tests taken from Curtis, H. D. (2005). Orbital mechanics for engineering students, p115, Example 3.1

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -721,12 +721,12 @@ def test_expected_mean_anomaly():
     orbit = Orbit.from_classical(attractor, a, ecc, _a, _a, _a, nu)
     orbit_M = orbit.M
 
-    assert_allclose(orbit_M.value, expected_mean_anomaly.value, rtol=1e-2)
+    assert_quantity_allclose(orbit_M.value, expected_mean_anomaly.value, rtol=1e-2)
 
 
 def test_expected_angular_momentum():
     # Example from Curtis
-    expected_ang_mom = 72472 * u.km ** 2
+    expected_ang_mag = 72472 * u.km ** 2
 
     attractor = Earth
 
@@ -736,9 +736,9 @@ def test_expected_angular_momentum():
     nu = 120 * u.deg
 
     orbit = Orbit.from_classical(attractor, a, ecc, _a, _a, _a, nu)
-    orbit_h_mom = orbit.h_mom
+    orbit_h_mag = orbit.h_mag
 
-    assert_allclose(orbit_h_mom.value, expected_ang_mom.value, rtol=1e-2)
+    assert_quantity_allclose(orbit_h_mag.value, expected_ang_mag.value, rtol=1e-2)
 
 
 def test_expected_last_perifocal_passage():
@@ -755,7 +755,7 @@ def test_expected_last_perifocal_passage():
     orbit = Orbit.from_classical(attractor, a, ecc, _a, _a, _a, nu)
     orbit_t_p = orbit.t_p
 
-    assert_allclose(orbit_t_p.value, expected_t_p.value, rtol=1e-2)
+    assert_quantity_allclose(orbit_t_p.value, expected_t_p.value, rtol=1e-2)
 
 
 def test_convert_from_rv_to_coe():
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -707,6 +707,57 @@ def test_perigee_and_apogee():
     assert_allclose(ss.r_p.to(u.km).value, expected_r_p.to(u.km).value)
 
 
+def test_expected_mean_anomaly():
+    # Example from Curtis
+    expected_mean_anomaly = 77.93 * u.deg
+
+    attractor = Earth
+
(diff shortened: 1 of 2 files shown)


src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -707,6 +707,57 @@ def test_perigee_and_apogee():
...
+
+    _a = 1.0 * u.deg  # Unused angle
+    a = 15300 * u.km
+    ecc = 0.37255 * u.one
+    nu = 120 * u.deg
+
+    orbit = Orbit.from_classical(attractor, a, ecc, _a, _a, _a, nu)
+    orbit_M = orbit.M
+
+    assert_allclose(orbit_M.value, expected_mean_anomaly.value, rtol=1e-2)
```
Juanlu001: There is an `assert_quantity_allclose` in Astropy that you can use https://astropy.readthedocs.io/en/stable/api/astropy.tests.helper.assert_quantity_allclose.html

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -707,6 +707,57 @@ def test_perigee_and_apogee():
...
+
+    _a = 1.0 * u.deg  # Unused angle
+    a = 15300 * u.km
+    ecc = 0.37255 * u.one
+    nu = 120 * u.deg
+
+    orbit = Orbit.from_classical(attractor, a, ecc, _a, _a, _a, nu)
+    orbit_h_mom = orbit.h_mom
+
+    assert_allclose(orbit_h_mom.value, expected_ang_mom.value, rtol=1e-2)
```
Juanlu001: Same as above

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -707,6 +707,57 @@ def test_perigee_and_apogee():
...
+
+    attractor = Earth
+
+    _a = 1.0 * u.deg  # Unused angle
+    a = 15300 * u.km
+    ecc = 0.37255 * u.one
+    nu = 120 * u.deg
+
+    orbit = Orbit.from_classical(attractor, a, ecc, _a, _a, _a, nu)
+    orbit_h_mom = orbit.h_mom
```
Juanlu001: Not a fan of `h_mom`, what about `h_mag` (of "magnitude")?

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -707,6 +707,57 @@ def test_perigee_and_apogee():
...
+
+    _a = 1.0 * u.deg  # Unused angle
+    a = 15300 * u.km
+    ecc = 0.37255 * u.one
+    nu = 120 * u.deg
+
+    orbit = Orbit.from_classical(attractor, a, ecc, _a, _a, _a, nu)
+    orbit_t_p = orbit.t_p
+
+    assert_allclose(orbit_t_p.value, expected_t_p.value, rtol=1e-2)
```
Juanlu001: Same as above

src/poliastro/twobody/orbit.py
```diff
@@ -216,12 +216,29 @@ def h_vec(self):
         h_vec = np.cross(self.r.to(u.km).value, self.v.to(u.km / u.s)) * u.km ** 2 / u.s
         return h_vec
 
+    @property
+    def h_mom(self):
+        """Angular momentum. """
+        h_mom = np.linalg.norm(self.h_vec) * u.km ** 2
```
Juanlu001: Instead of using `np.linalg.norm`, you can use `poliastro.util.norm`, which respects the units

src/poliastro/twobody/orbit.py
```diff
@@ -216,12 +216,29 @@ def h_vec(self):
         h_vec = np.cross(self.r.to(u.km).value, self.v.to(u.km / u.s)) * u.km ** 2 / u.s
         return h_vec
 
+    @property
+    def h_mom(self):
+        """Angular momentum. """
```
Juanlu001: Notice that this is the _specific_ angular momentum. In mechanics, they call "specific" to the magnitudes without mass, like this case, see https://www.wikiwand.com/en/Specific_relative_angular_momentum

src/poliastro/twobody/orbit.py
```diff
@@ -216,12 +216,29 @@ def h_vec(self):
...
 
+    @property
+    def M(self):
+        """Mean anomaly. """
+        return nu_to_M(self.nu, self.ecc)
+
+    @property
+    def t_p(self):
+        """Elapsed time since latest perifocal passage. """
+        t_p = self.period * self.M / (360 * u.deg)
```
Juanlu001: Interesting! I expected some subtraction from `self.epoch` but I guess this also works.

Also, perhaps instead of `/ (360 * u.deg)` we can have `/ (1 * u.cycle)`?

Sedictious: I started of like you said by subtracting. But I then saw that Curtis (as well as some various sources) expressed it as elapsed time rather than epoch which makes sense (and the result can then easily be subtracted).

Strangely, ``1 * u.cycle`` doesn't yield correct results

Juanlu001:
Merging, thanks!

Issue #714, opened 2019-06-25
Add new two-body propagators


Juanlu001:
(Not to be confused with #285)

It's already well known that some of our propagators have issues (see for example #475). Tricky stuff includes:

* Near parabolic orbits
* Very long propagation times
* Very low pericenters
* Others?

It would be nice to have more propagators (and also to benchmark them, see #253).

Juanlu001:
@giordano points us to this interesting method: http://dx.doi.org/10.1007/BF00691917

> Kepler's Equation is solved over the entire range of elliptic motion by a fifth-order refinement of the solution of a cubic equation. This method is not iterative, and requires only four transcendental function evaluations: a square root, a cube root, and two trigonometric functions. The maximum relative error of the algorithm is less than one part in 10^18, exceeding the capability of double-precision computer arithmetic. Roundoff errors in double-precision implementation of the algorithm are addressed, and procedures to avoid them are developed.

Definitely looks very promising! @jorgepiloto is surely looking into it :wink:

jorgepiloto:
I have implemented the previous Kepler equation solver in [this gist](https://gist.github.com/jorgepiloto/8d572ea4bb50c9ffed93c56dfaa81160).

However, when trying to solve for the vectors in #474 it seems to get stucked at the function `nu_to_M`. This may be a clue to what is going one with those vectors :rocket:

Juanlu001:
If it gets stuck, run the code again after exporting `NUMBA_DISABLE_JIT=1`. In my experience, numba sometimes "get stuck" in infinite loops because exceptions are not properly handled.

Juanlu001:
Relevant: https://github.com/poliastro/poliastro/pull/394/

jorgepiloto:
Here are some interesting papers that I have collected during the last months in which different methods for solving Kepler's equation are shown:

* [Robust resolution of Kepler’s equation in all eccentricity regimes](https://link.springer.com/article/10.1007/s10569-013-9476-9)

* [Accurate Kepler equation solver without transcendental function evaluations](https://www.researchgate.net/publication/288133262_Accurate_Kepler_equation_solver_without_transcendental_function_evaluations)

* [A Fast Procedure Solving Gauss' Form of Kepler's Equation](https://www.researchgate.net/publication/226072650_A_Fast_Procedure_Solving_Gauss%27_Form_of_Kepler%27s_Equation)

[...]

Issue #383, opened 2018-06-06
Document classical elements singularities


Juanlu001:
🐞 **Problem**

Some unexpected results might arise when creating equatorial and/or circular orbits:

```
ecc = 0.01 * u.one
a = 6678 * u.km
raan = 10.0 * u.deg
inc = 0.0 * u.deg
argp = 10.0 * u.deg
nu = 0.0 * u.rad
epoch = Time(j_date, format='jd', scale='tdb')
initial = Orbit.from_classical(Earth, a, ecc, inc, raan, argp, nu, epoch=epoch)
initial.state.to_vectors.raan  # 0 rad
```

🎯 **Goal**

To make it easier and less surprising for experts to create any kind of orbit, without worrying about singularities. When doing experiments, this is very common.

💡 **Possible solutions**

1. The singularities should be properly documented, and what poliastro does about them
2. Proper warnings should be raised when creating, e.g., equatorial orbits with a nonzero `raan`
3. Perhaps more constructors should be added, in the same way that we already have `Orbit.circular`?
4. More special elements should be implemented, in particular the argument of latitude and longitude of periapsis.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

jorgepiloto:
Could be interesting: https://doi.org/10.2514/1.14760

astrojuanlu:
After we reorganized our documentation, this would be a nice "Background" section.

PR #718, opened 2019-07-01
New powerful twobody propagators

jorgepiloto:
The following PR tries to solve for #475 and #714. It includes the following new propagators for solving the Kepler equation:

* `mikkola`:  https://doi.org/10.1007/BF01235850
* `markley`: https://doi.org/10.1007/BF00691917 (this is the previously called `kepler_improved`)
* `pimienta`: [link to the paper](https://www.researchgate.net/publication/288133262_Accurate_Kepler_equation_solver_without_transcendental_function_evaluations)
* `gooding`: https://doi.org/10.1007/BF01235540
* `danby`: https://doi.org/10.1007/BF00691917

New features and bug solutions:
1. These algorithms work really well and do not require evaluation of transcendental functions either several iterations to converge to an accurate solution. 

2. It was found that orbit from #474, related with #475, is nearly equatorial and therefore singularities appear and produce the error.

3. The test marked with the `skipped` command is now super fast if solved with some of the new propagators and not making so many iterations with the `mean_motion`.

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_propagation.py
+++ b/src/poliastro/tests/tests_twobody/test_propagation.py
@@ -13,9 +13,13 @@ from poliastro.examples import iss
 from poliastro.frames import Planes
 from poliastro.twobody import Orbit
 from poliastro.twobody.propagation import (
+    ALL_PROPAGATORS,
+    ELLIPTIC_PROPAGATORS,
+    HYPERBOLIC_PROPAGATORS,
+    PARABOLIC_PROPAGATORS,
     cowell,
     danby,
-    goodingI,
+    gooding,
     kepler,
     markley,
     mean_motion,
@@ -24,14 +28,14 @@ from poliastro.twobody.propagation import (
 )
 from poliastro.util import norm
 
-propagators_set = [mean_motion, kepler, mikkola, markley, pimienta, goodingI, danby]
-accurate_propagators_elliptic = [markley, pimienta, goodingI, danby]
-accurate_propagators_hyperbolic = [kepler, danby]
 
-
-@pytest.mark.parametrize("ecc", [0.99, 0.99, 0.999, 0.9999])
-@pytest.mark.parametrize("propagator", accurate_propagators_elliptic)
+@pytest.mark.parametrize("ecc", [0.9, 0.99, 0.999, 0.9999, 0.99999])
+@pytest.mark.parametrize("propagator", ELLIPTIC_PROPAGATORS)
 def test_elliptic_near_parabolic(ecc, propagator):
+    # 'kepler fails if really close to parabolic'. Refer to issue #714.
+    if propagator in [kepler] and ecc > 0.99:
+        pytest.xfail()
+
     _a = 0.0 * u.rad
     tof = 1.0 * u.min
     ss0 = Orbit.from_classical(
@@ -46,8 +50,12 @@ def test_elliptic_near_parabolic(ecc, propagator):
 
 
 @pytest.mark.parametrize("ecc", [1.0001, 1.001, 1.01, 1.1])
-@pytest.mark.parametrize("propagator", accurate_propagators_hyperbolic)
+@pytest.mark.parametrize("propagator", HYPERBOLIC_PROPAGATORS)
 def test_hyperbolic_near_parabolic(ecc, propagator):
+    # Still not implemented. Refer to issue #714.
+    if propagator in [pimienta, gooding]:
+        pytest.skip()
+
     _a = 0.0 * u.rad
     tof = 1.0 * u.min
     ss0 = Orbit.from_classical(
@@ -61,7 +69,7 @@ def test_hyperbolic_near_parabolic(ecc, propagator):
     assert_quantity_allclose(ss_propagator.v, ss_cowell.v)
 
 
-@pytest.mark.parametrize("propagator", [markley, pimienta])
+@pytest.mark.parametrize("propagator", [markley])
 def test_near_equatorial(propagator):
     r = [8.0e3, 1.0e3, 0.0] * u.km
     v = [-0.5, -0.5, 0.0001] * u.km / u.s
@@ -75,7 +83,7 @@ def test_near_equatorial(propagator):
     assert_quantity_allclose(ss_propagator.v, ss_cowell.v, rtol=1e-4)
 
 
-@pytest.mark.parametrize("propagator", propagators_set)
+@pytest.mark.parametrize("propagator", ALL_PROPAGATORS)
 def test_propagation(propagator):
     # Data from Vallado, example 2.4
     r0 = [1131.340, -2282.343, 6672.423] * u.km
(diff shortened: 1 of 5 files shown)


Juanlu001:
There is a bug with the `markley` propagator:

```
(Pdb) p ss0
15 x 8088 km x 180.0 deg (GCRS) orbit around Earth (♁) at epoch J2000.000 (TT)
(Pdb) p ss0.r.to(u.km)
<Quantity [8000., 1000.,    0.] km>
(Pdb) p ss0.propagate(0 * u.s, method=cowell).r.to(u.km)
<Quantity [8000., 1000.,    0.] km>
(Pdb) p ss0.propagate(1 * u.s, method=cowell).r.to(u.km)
<Quantity [7999.49695737,  999.49961973,    0.        ] km>
(Pdb) p ss0.propagate(0 * u.s, method=kepler_improved).r.to(u.km)
<Quantity [ 8.06225775e+03, -1.79061317e-11,  2.19235174e-27] km>
(Pdb) p ss0.propagate(1 * u.s, method=kepler_improved).r.to(u.km)
<Quantity [ 8.06169653e+03, -4.34121516e-01,  5.31645525e-17] km>
```

Juanlu001:
We checked and it's because the orbit is equatorial. With non-equatorial orbits it works like a charm!

Juanlu001:
Possibly relevant: https://github.com/poliastro/poliastro/issues/383

Juanlu001:
By the way:

So ~ many ~ propagators :heart_eyes:

src/poliastro/tests/tests_twobody/test_propagation.py
```diff
@@ -250,24 +288,25 @@ def test_propagate_to_date_has_proper_epoch():
...
-        mean_motion,
-        pytest.param(kepler, marks=pytest.mark.xfail),
-        pytest.param(
-            cowell,
-            marks=pytest.mark.skip(reason="for now propagation takes too long time"),
-        ),
-    ],
-)
-def test_propagate_long_times_keeps_geometry(method):
+# @pytest.mark.filterwarnings("ignore:ERFA")
```
Juanlu001: If we don't need this code anymore, let's remove it!

src/poliastro/twobody/propagation.py
```diff
@@ -188,6 +195,868 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+    Note
+    ----
+    Original paper: https://doi.org/10.1007/BF01235850
+    """
+
+    # Solving for the classical elements
+    p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
+    if ecc >= 1.0:
+        raise ValueError("Mikkola only works with elliptical orbits.")
+    M0 = nu_to_M(nu, ecc)
```
Juanlu001: I have a question though: a few of the new propagators use `nu_to_M`. Isn't it easier to keep `delta` and do `nu_to_M(nu, ecc, delta=0)` in these cases? That way, perhaps we can retain the behavior of `mean_motion` and making all the rest do the same. What do you think?

src/poliastro/twobody/propagation.py
```diff
@@ -188,6 +195,868 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+    vv : ~astropy.units.Quantity
+
+    Note
+    ----
+    Original paper: https://doi.org/10.1007/BF01235850
+    """
+
+    # Solving for the classical elements
+    p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
+    if ecc >= 1.0:
```
Juanlu001: According to the paper:

> The method can be used for all orbit types, including hyperbolic

Even with all the time you devoted to these propagators, I think it would be better to, at least, raise a `NotImplementedError`, to communicate that "now it doesn't work, but in the future maybe yes".

src/poliastro/twobody/propagation.py
```diff
@@ -188,6 +195,868 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+    tofs = tofs.to(u.s).value
+
+    results = [_mikkola(k, r0, v0, tof) for tof in tofs]
+    return (
+        [result[0] for result in results] * u.m,
+        [result[1] for result in results] * u.m / u.s,
+    )
+
+
+def _mikkola(k, r0, v0, tof, rtol=None):
```
Juanlu001: Is there a chance we apply `jit` to all of these propagators, in the same way it's done to `_kepler` and others?

src/poliastro/twobody/propagation.py
```diff
@@ -188,6 +195,868 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+
+    Note
+    ----
+    The following algorithm was taken from http://dx.doi.org/10.1007/BF00691917.
+    """
+
+    # Solve first for eccentricity and mean anomaly
+    p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
+    if ecc >= 1.0:
+        raise ValueError("Markley only works with elliptical orbits.")
```
Juanlu001: In this case it seems true that this propagator only claims to work for eccentric orbits

Juanlu001: However,

> but the eccentricity can be arbitrarily close to this limit

So perhaps we should check `if ecc > 1.0` instead?

src/poliastro/twobody/propagation.py
```diff
@@ -188,6 +195,868 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+    tof : float
+        Time of flight.
+
+    Returns
+    -------
+    rf: array
+        Final position vector
+    vf: array
+        Final velocity vector
+    """
```
Juanlu001: Can we have a link with the DOI also here?

src/poliastro/twobody/propagation.py
```diff
@@ -188,6 +195,868 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+        - 560 * w ** 3
+        + 15 * w
+    )
+
+    nu = E_to_nu(E, ecc)
+
+    return coe2rv(k, p, ecc, inc, raan, argp, nu)
+
+
+def goodingI(k, r, v, tofs, numiter=150, rtol=1e-8):
```
Juanlu001: Is there a `goodingII`? :)

src/poliastro/twobody/propagation.py
```diff
@@ -188,6 +195,868 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+     vv : 1x3 vector
+
+    Note
+    ----
+    Original paper for the algorithm: https://doi.org/10.1007/BF01238923
+    """
+
+    # Solve first for eccentricity and mean anomaly
+    p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
+    if ecc >= 1.0:
```
Juanlu001: Does it work for the case `ecc == 1.0`?

src/poliastro/twobody/propagation.py
```diff
@@ -188,6 +195,868 @@ def _kepler(k, r0, v0, tof, *, numiter):
...
+    n = np.sqrt(k / np.abs(semi_axis_a) ** 3)
+    M = M0 + n * tof
+
+    # Start the computation
+    n = 0
+    c = ecc * np.cos(M)
+    s = ecc * np.sin(M)
+    psi = s / np.sqrt(1 - 2 * c + ecc ** 2)
+    f = 1.0
+    while f ** 2 >= rtol and n <= numiter:
```
Juanlu001: What happens if the maximum number of iterations is reached?

src/poliastro/twobody/propagation.py
jorgepiloto: Sure! I have checked and by keeping delta, `mean_motion` works without problem for any `ecc` regime :+1:

src/poliastro/twobody/propagation.py
```diff
@@ -222,3 +478,36 @@ def propagate(orbit, time_of_flight, *, method=mean_motion, rtol=1e-10, **kwargs
     )
 
     return cartesian
+
+
+PROPAGATORS_SET = [
```
Juanlu001: Instead of building `PROPAGATORS_SET` manually, why don't we intersect 

```
ALL_PROPAGATORS = list(set(ELLIPTIC_PROPAGATORS) & set(...) & set(...))
```

?

Juanlu001:
Only failure:

```
docstring of poliastro.twobody.propagation.pimienta:24:Unexpected indentation.
```

This is almost good to go!

jorgepiloto:
I spent the whole afternoon trying to fix some of those `NotImplementedError` but could not solve them. Since this pull request is getting bigger (sorry for that and thanks for the reviews) I think is enough for the moment.

More information on propagators current status is available at #714 :earth_africa:

Juanlu001:
Awesome work @jorgepiloto, this is in!

Issue #703, opened 2019-06-17
Include more IAU-WGCCRE constants


astrojuanlu:
Follow-up from https://github.com/poliastro/poliastro/pull/674#discussion_r290166801

The constants from The IAU Working Group on Cartographic Coordinates and Rotational Elements to Astropy. They contain things like ellipsoidal and mean radiuses, reference meridians, rotational velocities for planets and some minor bodies... The IAU-WGCCRE publishes reports that contain updated values of these constants:

https://astrogeology.usgs.gov/groups/IAU-WGCCRE

We are already using the radiuses of the planets and also the rotational elements, and we could include the ellipsoid parameters as well.

We decided it would be better to have them outside Astropy: https://groups.google.com/d/msg/astropy-dev/Bpmxlfp9z30/S6VVrEBWBQAJ

Related: #109

Juanlu001:
@Sedictious already updated the radii values, and also added some new ones. Next step: reference meridians and rotational velocities.

astrojuanlu:
The radiuses are already in https://github.com/poliastro/poliastro/blob/master/src/poliastro/constants/general.py and we already have the rotational periods in https://github.com/poliastro/poliastro/blob/master/src/poliastro/constants/rotational_elements.py.

Regarding the reference meridians, as it can be seen in https://github.com/poliastro/poliastro/blob/master/src/poliastro/frames/fixed.py and the WGCCRE report itself, some of them are dynamic, so it's not practical to stablish them as constants. Therefore, I'm closing this issue as finished.

PR #719, opened 2019-07-03
[WIP] Implement pass rise and set times predictor

Sedictious:
Implement the algorithm described on Escobal, P.R.. (1963). Rise and set time of a satellite about an oblate planet. AIAA Journal. 1. 10.2514/3.2057.

### DIFF
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -869,7 +869,7 @@ class Orbit(object):
         t = time.Time.now()
         gmst06 = erfa.gmst06(t.ut1.jd1, t.ut1.jd2, t.tt.jd1, t.tt.jd2)
 
-        th0 = long + gmst06 * u.rad # sidereal station time (rad)
+        th0 = long + gmst06 * u.rad  # sidereal station time (rad)
         r_s = 1  # TODO: sidereal rate of change
 
         def Z(E):
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -856,7 +856,7 @@ class Orbit(object):
 
     def calculate_pass(self, lat, long, H):
         a = self.attractor.R  # semimajor axis
-        b = self.attractor.R_polar #semiminor axis
+        b = self.attractor.R_polar  # semiminor axis
         e = self.ecc  # eccentricity
         i = self.inc  # equatorial inclination
         O = self.long_asc  # longitude of ascending node
@@ -901,6 +901,11 @@ class Orbit(object):
         )
 
         Z_av = 0.5 * (Z(2 * np.pi) + Z(0))
+
+        E = a - np.arccos(
+            (G + a * e * np.dot(P, Z_av))
+            / (a * np.sqrt(np.dot(P, Z_av) ** 2 + np.dot(Q, Z_av) ** 2 * (1 - e ** 2)))
+        )
 
     def represent_as(self, representation):
         """Converts the orbit to a specific representation.
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -187,6 +187,19 @@ class Orbit(object):
         return self.raan + self.argp + self.nu
 
     @property
+    def long_asc(self):
+        """Longitude of ascending node. """
+        h = self.h_vec
+        k = np.array([0, 0, 1])
+        n = np.cross(k, h)
+        long_asc = (
+            np.arccos(n[0] / np.linalg.norm(n))
+            if n[1] >= 0
+            else 2 * np.pi - np.arccos(n[0] / np.linalg.norm(n))
+        )
+        return long_asc
+
+    @property
     def period(self):
         """Period of the orbit. """
         return 2 * np.pi * u.rad / self.n
@@ -841,6 +854,54 @@ class Orbit(object):
             raise ValueError(
                 f"The semimajor axis may not be smaller that {attractor.name}'s radius"
             )
+
+    def calculate_pass(self, lat, long, H):
+        a = self.a  # semimajor axis
+        e = self.ecc  # eccentricity
+        i = self.inc  # equatorial inclination
+        O = self.long_asc  # longitude of ascending node
+        o = self.argp  # argument of perigee
+        T = self.period  # time of perifocal passage
+
+        n = self.n  # mean motion
+        ar = self.attractor.r  # equatorial radius
+        f = 0  # TODO: flattening
+
+        t0 = 1  # TODO: epoch universal time (rad)
+        th0 = 1  # TODO: sidereal station time (rad)
+        r_s = 1  # TODO: sidereal rate of change
+
+        def Z(E):
+            return np.array(
+                [
+                    np.cos(lat) * np.cos(th0 + r_s * (E - e * np.sin(E) / n + T - t0)),
+                    np.cos(lat) * np.sin(th0 + r_s * (E - e * np.sin(E) / n + T - t0)),
+                    np.sin(lat),
+                ]
+            )
+
+        G1 = np.array(ar / np.sqrt(1 - (2 * f - f * f) * np.sin(lat) ** 2) + H)
+        G2 = np.array(
+            (1 - f) ** 2 * ar / np.sqrt(1 - (2 * f - f ** 2) * np.sin(lat) ** 2) + H
+        )
+        G = G1 * np.cos(lat) ** 2 + G2 * np.sin(lat) ** 2
+
+        P = np.array(
+            [
+                np.cos(o) * np.cos(O) - np.sin(o) * np.sin(O) * np.cos(i),
+                np.cos(o) * np.sin(O) + np.sin(o) * np.cos(O) * np.cos(i),
+                np.sin(o) * np.sin(i),
+            ]
+        )
+        Q = np.array(
+            [
+                -np.sin(o) * np.cos(O) - np.cos(o) * np.sin(O) * np.cos(i),
+                -np.sin(o) * np.sin(O) + np.cos(o) * np.cos(O) * np.cos(i),
+                np.cos(o) * np.sin(i),
+            ]
+        )
+
+        Z_av = 0.5 * (Z(2 * np.pi) + Z(0))
 
     def represent_as(self, representation):
         """Converts the orbit to a specific representation.
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -1,7 +1,7 @@
 from warnings import warn
 
 import numpy as np
-from astropy import time, units as u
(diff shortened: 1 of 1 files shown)


src/poliastro/twobody/orbit.py
```diff
@@ -839,6 +852,53 @@ def frozen(
...
+        o = self.argp  # argument of perigee
+        T = self.period  # time of perifocal passage
+
+        n = self.n  # mean motion
+        ar = self.attractor.r  # equatorial radius
+        f = 0  # TODO: flattening
+
+        t0 = 1  # TODO: epoch universal time (rad)
+        th0 = 1  # TODO: sidereal station time (rad)
+        r_s = 1  # TODO: sidereal rate of change
```
Juanlu001: This should be the attractor "mean angular rotation in degrees per second". For the case of the Earth, it's not included in the 2015 WGCRE report:

> "Previous reports also included approximate expressions for the Earth. Their accuracy was poor, and the expressions failed near the fundamental epoch (J2000.0), yet they were sometimes used as a recommended model. Users should refer to the International Earth Rotation and Reference Systems Service (IERS, https://www.iers.org) for appropriate models of the Earth’s rotation."

However, we are indeed interest in a mean value, which is included here:

https://github.com/poliastro/poliastro/blob/fa131ae470b14ff1e239d6c923c4e541a8ebb0a2/src/poliastro/bodies.py#L211

And comes from the 2009 report. This is somehow coupled with https://github.com/poliastro/poliastro/issues/703.

src/poliastro/twobody/orbit.py
```diff
@@ -839,6 +852,53 @@ def frozen(
...
+        i = self.inc  # equatorial inclination
+        O = self.long_asc  # longitude of ascending node
+        o = self.argp  # argument of perigee
+        T = self.period  # time of perifocal passage
+
+        n = self.n  # mean motion
+        ar = self.attractor.r  # equatorial radius
+        f = 0  # TODO: flattening
+
+        t0 = 1  # TODO: epoch universal time (rad)
```
Juanlu001: I propose we set this reference time to the corresponding "time of last perifocal passage" of the of the orbit we're studying. In that case, in (18) the `T` cancels out (we don't even need it here in the code) and we're left with calculating the Greenwich Mean Sidereal Time at that particular instant.

src/poliastro/twobody/orbit.py
```diff
@@ -839,6 +852,53 @@ def frozen(
...
+        th0 = 1  # TODO: sidereal station time (rad)
+        r_s = 1  # TODO: sidereal rate of change
+
+        def Z(E):
+            return np.array(
+                [
+                    np.cos(lat) * np.cos(th0 + r_s * ((E - e * np.sin(E)) / n + T - t0)),
```
Juanlu001: As per my comment above, `T = t0` and therefore we could remove it.

src/poliastro/twobody/orbit.py
```diff
@@ -839,6 +852,53 @@ def frozen(
...
+        O = self.long_asc  # longitude of ascending node
+        o = self.argp  # argument of perigee
+        T = self.period  # time of perifocal passage
+
+        n = self.n  # mean motion
+        ar = self.attractor.r  # equatorial radius
+        f = 0  # TODO: flattening
+
+        t0 = 1  # TODO: epoch universal time (rad)
+        th0 = 1  # TODO: sidereal station time (rad)
```
Juanlu001: Having selected the reference time, according to the Vallado equation (3-43), this will be equal the longitude plus `erfa.gmst06`:

```
In [20]: t = Time.now()                                                                                                                 

In [21]: t.ut1.jd1                                                                                                                      
Out[21]: 2458701.0

In [22]: t.ut1.jd2                                                                                                                      
Out[22]: 0.432900555821122

In [23]: t.tt.jd1                                                                                                                       
Out[23]: 2458701.0

In [24]: t.tt.jd2                                                                                                                       
Out[24]: 0.43370315784722224

In [25]: erfa.gmst06(t.ut1.jd1, t.ut1.jd2, t.tt.jd1, t.tt.jd2)                                                                          
Out[25]: 5.061874805765537
```

Juanlu001: A comment about the applicability of this for other planets. From `gmst06` docs:

> 2) Both UT1 and TT are required, UT1 to predict the Earth rotation
   and TT to predict the effects of precession.  If UT1 is used for
   both purposes, errors of order 100 microarcseconds result.

Source code:

https://github.com/liberfa/erfa/blob/95bdbb87dc0ac444c35da95cbaa1b46b9500c6bb/src/gmst06.c#L68-L79

After implementing the Earth case, we should carefully reflect on how to apply the formulas of WGCRE here, and probably remove the use of ERFA completely. But let's leave that for a future PR.

src/poliastro/twobody/orbit.py
Sedictious: Wasn't this partly addressed in #735 ?

src/poliastro/twobody/orbit.py
Sedictious: Is ``erfa`` an astropy module? I can't seem to find it.

Juanlu001:
Oops! Yes, it's `from astropy import _erfa as erfa`

src/poliastro/twobody/orbit.py
Juanlu001: It's unrelated - in #735 lots of values were updated, but not the ones from the Earth. In any case, what we want is to either (a) use those values, if they are suitable for a _mean_ rotational velocity, or (b) find a better value to use.

src/poliastro/twobody/orbit.py
Juanlu001: Oops! Yes, it's `from astropy import _erfa as erfa`

src/poliastro/twobody/orbit.py
Juanlu001: It's unrelated - in #735 lots of values were updated, but not the ones from the Earth. In any case, what we want is to either (a) use those values, if they are suitable for a _mean_ rotational velocity, or (b) find a better value to use.

Juanlu001: It's unrelated - in #735 lots of values were updated, but not the ones from the Earth. In any case, what we want is to either (a) use those values, if they are suitable for a _mean_ rotational velocity, or (b) find a better value to use.

Sedictious: If I recall, the only values were updated were the ones that were changed so the Earth values are up to date with the IAU 15 report

src/poliastro/twobody/orbit.py
Sedictious: @Juanlu001 Just another question regarding your comment: 

> this will be equal to the longitude plus erfa.gmst06

the output of ``erfa.gmst06`` is a float while longitude is calculated in rads. Do we need to take the value of the longitude or can we also convert ``erfa.gmst06`` to rads (presumably, theta must be in rads anyway)

Juanlu001: According to the docs:

https://github.com/liberfa/erfa/blob/7009153a0fdbc2783e9a3442ddde88b9d9e8b5ef/src/gmst06.c#L15-L16

`erfa.gmst06` is already in radians, so at most we would need to attach a unit.

src/poliastro/twobody/orbit.py
Juanlu001: The thing is that the Earth does not appear in IAU 15 :) but it's an unimportant detail, we can use the value from IAU 09

Juanlu001:
Completely testing this will be a bit tricky, but feasible. For example, I tested the pass predictor of `orbit-predictor` against STK:

https://github.com/satellogic/orbit-predictor/blob/7a75f6d675fe84a84e2df0708b9f929ac47ddc06/tests/test_accurate_predictor.py#L158-L172

The difference is that orbit-predictor propagates the orbits with SGP4, whereas we are implicitly considering Keplerian orbits for this algorithm. I am not quite sure whether the results will be _slightly_ different or _very different_, unfortunately.

In our case, we could use GMAT instead:

https://sourceforge.net/projects/gmat/

Which supports Keplerian propagation and, if I'm not mistaken, pass prediction.

Juanlu001:
@Sedictious there has been several changes and fixes in `master`, could you please rebase this?

Sedictious:
@Juanlu001 Done 😄

astrojuanlu:
Will probably try to get this rolling again soon. We got stuck trying to generalize this for all bodies, so now we'll try to be less ambitious.

astrojuanlu:
Well, I have good news @Sedictious ! Today I need this code at work 😂  So I will devote some time to understand it, probably copy parts of it to https://github.com/satellogic/orbit-predictor, and if I am lucky in understanding the whole thing, making a decision about this pull request. I would like to apologize for the _slighly long delay_ and thank you again for the wonderful work you did last year ❤️

astrojuanlu:
Well, now I have some bitter news - I can't make sense of the paper! I gave this a shot at https://github.com/satellogic/orbit-predictor/pull/114 and left some comments in https://space.stackexchange.com/q/49136/10716.

In short, I have several problems:

- Escobal "preliminary estimate" doesn't really work. Sure, it does work for polar satellites in Low-Earth Orbit, but I tried a mid-inclination satellite and the amplitude is not constant anymore.
- Therefore, quoting the paper, "it will be necessary to hunt for the zeros of Eq. (15) directly". But the paper doesn't give any hint about how to hunt for these zeros! The Newton method only works if the initial guess is somewhat close to the solution.

I am about to try a final approach using `find_peaks` and see if I can make it work.

Sedictious:
The [last paper](https://arxiv.org/ftp/arxiv/papers/1611/1611.02402.pdf) linked in the Stack Exchange thread gives a numerical self-adaptive curve-fitting method which looks interesting. Did you have any luck with that?

astrojuanlu:
I did not try implementing the other papers just yet, but this one and a few others linked there look promising!

astrojuanlu:
Coming back to this pull request would involve doing a massive rebase, for a method that we were not able to fully validate, nor get a satisfactory response. 

If we ever want to reboot this pull request, I would take a different approach:

1. Translate the equations from the paper in the most direct way possible, probably in a totally different module
2. Try to reproduce the figures of the paper
3. Add it to the `poliastro.core` sub-package
4. Open a new pull request that brings that to the `Orbit` object.

I am sure that with enough focus and dedication we could do it, but unfortunately this currently falls under the category of "too difficult" things. To remove some clutter from the repository, I think it would be better to close this pull request.

It is also my fault, because the package has grown somewhat organically and I admit that I am still not 100 % sold on how we are integrating the Earth-related functionalities. This makes me hesitant to merge code that's Earth-specific, and it's still bothering me almost two years later.

We should take a step back and reconsider some parts of the design very carefully, again with enough time and focus.

Thanks a lot @Sedictious for the work you did back then ❤️

PR #720, opened 2019-07-04
Add SpheroidLocation class

Sedictious:
Adds ``SpheroidLocation``:

* Conversion to and from Cartesian coordinates

* Find tangential vectors (and by extension the tangential plane)

* Determine if a point is visible to the ground station in a given epoch.

* Add accompanying tests

### DIFF
--- a/src/poliastro/ground_station.py
+++ b/src/poliastro/ground_station.py
@@ -1,0 +1,38 @@
+import numpy as np
+
+from astropy import units as u
+
+
+class GroundStation(object):
+    """"""
+
+    def __init__(self, lon, lat, h, a, b):
+        """
+
+        Parameters
+        ----------
+        lon: ~astropy.units.core.Unit
+            geodetic longitude
+
+        lat: ~astropy.units.core.Unit
+            geodetic latitude
+
+        h: ~astropy.units.core.Unit
+            geodetic height
+        """
+        self._lon = lon
+        self._lat = lat
+        self._h = h
+        self._a = a
+        self._b = b
+
+    @property
+    def cartesian_cords(self):
+        """Convert to the Cartesian Coordinate system. """
+        e2 = 1 - (self._b / self._a) ** 2
+        N = self._a / np.sqrt(1 - e2 * np.sin(self._lon) ** 2)
+
+        x = (N + self._h) * np.cos(self._lon) * np.cos(self._lat)
+        y = (N + self._h) * np.cos(self._lon) * np.sin(self._lat)
+        z = ((1 - e2) * N + self._h) * np.sin(self._lon)
+        return x, y, z
--- a/src/poliastro/ground_station.py
+++ b/src/poliastro/ground_station.py
@@ -1,24 +1,29 @@
 import numpy as np
 
-from astropy import units as u
-
 
 class GroundStation(object):
-    """"""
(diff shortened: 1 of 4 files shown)


src/poliastro/ground_station.py
```diff
@@ -0,0 +1,145 @@
...
+        """Returns orthonormal vectors tangential to the ellipsoid at the point of the ground station.
+        """
+        N = self.N
+        u = np.array([1.0, 0, 0])
+        u -= u.dot(N) * N
+        u /= np.linalg.norm(u)
+        v = np.cross(N, u)
+        return u, v
+
+    def propagate(self, t):
```
Juanlu001: I am not exacly sure what `gs.propagate` is supposed to do, but it doesn't seem quite right to me. What do you think will be the use case?

Juanlu001: To clarify: to me, a `GroundStation` is something static, attached to the surface of something. Surely you can propagate that something, but propagating the ground station still sounds weird.

src/poliastro/ground_station.py
```diff
@@ -0,0 +1,145 @@
...
+        c: ~astropy.units.quantity.Quantity
+            Third axis defining the ellipsoid.
+            If not set, it defaults to the value of a, making the ellipsoid
+            a spheroid.
+        """
+        self._lon = lon
+        self._lat = lat
+        self._h = h
+        self._epoch = epoch
+        self._angular_velocity = angular_velocity
```
Juanlu001: Instead of passing angular velocity and parameters of the ellipsoid, what about passing a `poliastro.bodies.Body`?

Juanlu001: And then these properties should be in the bodies themselves

src/poliastro/ground_station.py
```diff
@@ -0,0 +1,145 @@
...
+
+        c: ~astropy.units.quantity.Quantity
+            Third axis defining the ellipsoid.
+            If not set, it defaults to the value of a, making the ellipsoid
+            a spheroid.
+        """
+        self._lon = lon
+        self._lat = lat
+        self._h = h
+        self._epoch = epoch
```
Juanlu001: I guess the `epoch` is only needed for the propagation? As per my remarks about `.propagate`, probably we won't need to specify the *time* of a ground station, only its position in planet-fixed coordinates.

src/poliastro/ground_station.py
Sedictious: I agree and I actually spend a lot of time thinking about it, but couldn't come up with anything better. Now we may want to account for the rotation of the body, so there should be a way to specify said rotation in a given epoch or after a certain interval. That being said, calling ``gs.rotate(time)`` sounds even less reasonable.

But, there may be a way to deal with this if we allow for a ``Body`` as you mentioned bellow, since we could then use the body itself and not the ground station (this could probably require some work on the ``Body`` class itself). Alternatively, we could use something simple like ``specify_epoch(time)`` which makes more sense.

Sedictious:
@I thought more about this and started researching what could be added, I think it would make more sense if we renamed ``GroundStation`` in something like ``SpheroidLocation`` (kind like a generalized astropy's ``EarthLocation``).

Juanlu001:
Let's merge this, and we will take care about the `Body` objects later! Thanks a lot @Sedictious :clap:

PR #721, opened 2019-07-09
Add 2D Scene support

Sedictious:
This is a small commit to compliment the latest change to the Cesium application which allows users to define the Viewer Scene from the extractor.


Juanlu001:
Merging this, thanks @Sedictious and sorry for the delay!

PR #723, opened 2019-07-14
Drop Python 3.5, test on 3.8 beta

Juanlu001:


Juanlu001:
Temporarily sticking to Python 3.7.3 because of https://github.com/numba/numba/issues/4323.

Juanlu001:
On the other hand, numba and Python 3.8 might not be trivial:

https://github.com/numba/numba/issues/4071

Just in case Python 3.8 gets released before poliastro, I will limit it.

PR #729, opened 2019-07-15
Add 'change_attractor' force argument

jorgepiloto:
Small pull request related to #679. It adds the `force` argument in the `change_attractor` method and warns users that this option is available under their responsibility.

The `Catch that asteroid` notebook still needs to be updated...


Juanlu001:
Thanks a lot @jorgepiloto!

PR #737, opened 2019-07-19
SpheroidLocation accepts poliastro bodies instead of spheroid parameters

Sedictious:


src/poliastro/spheroid_location.py
```diff
@@ -19,18 +19,17 @@ def __init__(self, lon, lat, h, a, c):
...
-        c: ~astropy.units.quantity.Quantity
-            equatorial semi-minor axis
         """
         self._lon = lon
         self._lat = lat
         self._h = h
-        self._a = a
-        self._b = a
-        self._c = c
+        self._body = body
```
Juanlu001: Do we need to store the particular `Body`?

Sedictious: You're right, no need for adding it now. I was thinking about adding relevant methods later, but we can add it back again if and when it's needed

Issue #702, opened 2019-06-13
Add better atmosphere model for the Earth


astrojuanlu:
(Coming from #694)

See for example:

* https://github.com/DeepHorizons/Python-NRLMSISE-00
* https://github.com/space-physics/msise00

(and probably others)

Juanlu001:
This was (partially) addressed in #738. @jorgepiloto would you please leave a comment here stating what needs to be done, or alternatively close the issue?

jorgepiloto:
Sure! I will make a brief resume on the current status of this:

After #738 was merged two new atmospheric models were added in `poliastro`:
* COESA62: up to 700km
* COESA76: up to 1000km

However, some difficulties appeared when implementing pressure and density in previous atmospheres above some altitudes. Different `#TODO` comments were left in the code to know what is still required to be done.

On the he other hand, more models could be added such as:
* Jacchia77
* NRLMSISE-00

This last model has been implemented in Python packages as stated at the beginning of this issue. However, I have tried to install [this package](https://github.com/space-physics/msise00) and it requires to have a `Fortran` compiler installed in user's machine.

Before closing this I would like to solve the following tasks:
- [ ] Full implemented COESA62
- [ ] Full implemented COESA76
- [ ] Full implemented Jacchia77 (could be included as dependence)
- [ ] Full implemented NRLMSISE-00 (could be included as dependence)

jorgepiloto:
Just found this [amazing atmospheric models website](http://www.dem.inpe.br/~val/atmod/default.html) which holds source codes written in Fortran and C/C++ for different Jacchia's models and the famous MSIS-86. I will take a closer look in a few days :rocket:

GorgiAstro:
If NRLMSISE-00 gets implemented, it would be nice to feed it with the daily space weather data combined by CSSI/AGI: ftp://ftp.agi.com/pub/DynamicEarthData/SpaceWeather-All-v1.2.txt and http://celestrak.com/SpaceData/SpaceWx-format.php

It contains three-hourly Ap geomagnetic index and daily F10.7 flux required by NRLMSISE-00.

The CSSI/AGI file is for instance also packaged into GMAT (in the folder `GMAT/R2019aBeta1/data/atmosphere/earth`). If we choose to package this file in the poliastro repo, I don't know what's the best way to keep this file up to date without needing to manually push new versions. This file is updated several times a day. Although if an extreme accuracy is not required, the file does not need to be updated very often as it offers predictions until 25 years in the future.

[...]

Issue #694, opened 2019-06-10
Issue: Air Drag possibly inconsistent and tolerance problems


WolfsSky:
WILL EDIT ISSUE LATER
Comparison of poliastro Cowell, Python sgp4 and TLE from Tiangong 1. 

What can be seen is that for a torelance of 1e-5 the orbit decays, but for 1e-7 the orbit does not. Of course one needs to take into account the less accurate SGP4 than Cowell, however, based on the TLE's Cowell seems to be inaccurate.

Relative Tolerance = 1e-5
![Test1Drag1e5](https://user-images.githubusercontent.com/6907283/59194378-22ac2c00-8b89-11e9-98a9-680d20e7fa0c.png)

Relative Tolerance = 1e-7
![Test1Drag1e7](https://user-images.githubusercontent.com/6907283/59194379-22ac2c00-8b89-11e9-97ea-fa2b44b83e32.png)

🐞 **Problem**

The relative Cowell tolerance has an influence on the outcome of the propagation. SGP4 is known to be inconsistent for long propagation times, however Cowell/Dopri853 should be rather accurate/similar to TLE data.

🖥 **Please paste the output of following commands**

* `conda info -a` (only if you have conda)
* `conda list` (only if you have conda)
* `pip freeze`

<details>
  
```bash
# Paste your output here:

```
</details>

* `python -c "import poliastro.testing; poliastro.testing.test()"`

<details>

```bash
# Paste your output here:

```
</details>

🎯 **Goal**

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
Hmmm interesting. Please keep us posted when you're done with the analysis, this is important.

WolfsSky:
**Finding 1**
What I find the most peculiar is the too long **Linear decay**, compared to TLE and SGP4. The atmospheric model uses a scale height of 8 km with a rho_0=1.3 kg/m^3. This is true for an atmosphere till 100 km, but above that it is enormous erroneous compared to for example the data found in Space Mission Analysis and Design by Wertz for higher altitudes:
[...]

PR #738, opened 2019-07-19
Add atmospheric models

jorgepiloto:
:warning: DRAFT PULL REQUEST :warning: 

[ + ] Last update on 20th July 2019

This pull request adds different atmospheric models to improve results on drag perturbations, as requested in #702. It may be the solution to #694. The following atmospheric models are planned to be implemented:

* [x] ISA-ICAO: accurate up to 80 km.
* [x] COESA62: accurate up to 86 km.
* [x] COESA76: accurate up to 86 km.
* [ ] Jacchia77: from 125km up to 700km.
* [ ] NRLMSISE-00: up to 1000km (very used for drag perturbation)

Main features:

* The differeten atmospheric models implemented till now are instances of the `Atmosphere()` class. It only needs the so-called *geopotential layer properties* to compute all the different properties.

```python
from poliastro.atmosphere.models import coesa62, coesa76
T, p, rho = coesa62.properties(50 * u.km)
```

* Some constants on the papers such us `R_earth` or `g0` are not updated. Others that are required are not present in `poliastro.contants` such us `M0_air` or `R_air`.

* The `test_perturbations[drag]` should be updated in order to check if atmospheric models work properly.

* Tests still need to be implemented.

By opening this, you can review better the code and review it as I continue adding more models :+1:


jorgepiloto:
Sorry, I forgot about the option `Draft Pull Request` :sweat:

Juanlu001:
Gave a quick look, no major issues found :) Just left one comment.

On the other hand, even though I agree with you that it's better to not use classes when not necessary, I think we will encounter the need to pass "a generic atmosphere" around, or make it a configurable thing. In that case, it's easier to pass an object than a module, I think. Either of those is a form of namespacing, but objects are easier to inspect.

src/poliastro/atmosphere/isa.py
```diff
@@ -0,0 +1,176 @@
...
+            # We are at layer. Return those values
+            Tb = Tb_layers[i]
+            pb = pb_layers[i]
+            if gradient:
+                beta = beta_set[i]
+            else:
+                beta = 0 * u.K / u.km
+
+            return H, Hb, Tb, pb, beta
+        else:
```
Juanlu001: This `else: pass` is not needed, I think

jorgepiloto:
Moving to classes after not only reading the previous comment, but also seeing that most atmospheric models are defined by layers and methods are shared among them.  :sweat_smile:

src/poliastro/atmosphere/isa.py
jorgepiloto: With the class implementation, this has been simplified. It's not necessary anymore :+1:

Juanlu001:
In case serves you as inspiration:

* https://github.com/DeepHorizons/Python-NRLMSISE-00 (MIT)
* https://github.com/space-physics/msise00 (MIT)

We can either get inspiration from them, or include them as dependencies, or ignore them (probably a bad take!)

jorgepiloto:
The `COESA76` works up to 86km for any variable and only up to 1000km for the temperature. Regarding `COESA62` I still need to copy the layer parameters from the tables but the mathematics are exactly the same ones behind coesa76 and even easier.

Juanlu001:
Loved the notebook :smile: Could you please rebase this on top of current `master`?

Also, from our private conversation:

> From 86 kilometers upwards, we have to apply a long formula to obtain `n_i` [Equation 33c]. However one of the variables in the formula depends on that same parameter, it's very strange.

From the paper:

> Neither n_i, the number densities of individual species, nor sum n_i, the sum of the individual number densities, is known directly. Consequently, pressures above 86 km cannot be computed without first determining n_i for each of the significant gas species.

And then section 1.3.2 spans for more than two pages explaining how to do such a thing. Honestly I didn't bother reading it entirely, and it's probably a little bit ambitious for this PR. Therefore, can we throw a `NotImplementedError` for these cases, and leave it for later?

Juanlu001:
@jorgepiloto please rebase!

jorgepiloto:
Sure! Hope to add all requested changes in a couple of hours :rocket:

jorgepiloto:
I am adding some tests, hope to be ready soon :+1:

Juanlu001:
```
sphinx.errors.SphinxWarning: autodoc: failed to import module 'modules' from module 'poliastro.atmosphere'; the following exception was raised:
No module named 'poliastro.atmosphere.modules'
```

src/poliastro/atmosphere/models.py
```diff
@@ -0,0 +1,661 @@
...
+        H: ~astropy.units.Quantity
+            Geopotential altitude.
+
+        Returns
+        -------
+        i: int
+            Index position in table.
+        """
+
+        for i, Hb in enumerate(self._Hb_table):
```
Juanlu001: Isn't this duplicated in `util.get_base_index`?

src/poliastro/atmosphere/models.py
```diff
@@ -0,0 +1,661 @@
...
+            1.5381e-3,
+            7.1042e-5,
+            2.5382e-5,
+            3.0236e-9,
+            7.5138e-11,
+        ] * u.mbar
+
+        # Generate table
+        table_coesa76 = [Zb_table, Hb_table, Tb_table, Lb_table, pb_table]
+        COESA.__init__(self, table_coesa76)
```
Juanlu001: I would use `super().__init__` instead

src/poliastro/atmosphere/models.py
```diff
@@ -0,0 +1,661 @@
...
+    +--------+---------+---------+-----------+---------------+---------------+
+    | 1000.0 | 864.071 |   1000  | 7.5138e-5 |   3.561e-15   |  exponential  |
+    +--------+---------+---------+-----------+---------------+---------------+
+
+    """
+
+    def __init__(self):
+
+        # Lower layers (Z < 86km)
+        Hb_lower = [0.0, 11.0, 20.0, 32.0, 47.0, 51.0, 71.0] * u.km
```
Juanlu001: Any chance these valuable constants can be stored outside of the class?

src/poliastro/atmosphere/models.py
```diff
@@ -0,0 +1,661 @@
...
+        -------
+        T: ~astropy.units.Quantity
+            Temperature at given height.
+        """
+        # Check if valid range and convert to geopotential
+        Z, H = self._get_heights(alt, geometric=geometric)
+        Zb, Hb, Tb, Lb, _ = self._get_base_parameters(H)
+
+        # Different altitudes imply different equations
+        # TODO: Include mean air molecular weight corrections
```
Juanlu001: Does this `TODO` affect all the layers, or only the last one?

src/poliastro/atmosphere/util.py
```diff
@@ -0,0 +1,93 @@
...
+    -------
+    i: int
+        Index position in table.
+    """
+
+    for i, Hb in enumerate(table):
+        if i < len(table) and H > Hb:
+            pass
+        elif H == Hb:
+            return [i]
```
Juanlu001: Why do we need to return lists with one element in both cases?

src/poliastro/atmosphere/models.py
jorgepiloto: Ups! You are right, I'll get rid of this function in the `util` script :+1:

src/poliastro/atmosphere/util.py
jorgepiloto: No idea why I coded the output as a list but is not necessary at all. Only an integer value should be returned :confused:

src/poliastro/atmosphere/models.py
jorgepiloto: No problem! For the moment I will just move them outside the class but in the future, it may be interesting to have them stored in some kind of `*.dat` files as discussed privately

src/poliastro/atmosphere/models.py
jorgepiloto: No, it only affects the layers between 80km and 86km. Below 80km the mean molecular weight of the air can be assumed to be constant but above this value, till the 86km layer, a small correction may be applied.

At 80km the ratio M/M0 is known to be 1.0, while at 86km this ratio becomes 0.999579, being M the real molecular weight and M0 the one at sea.

I will rewrite this #TODO so it becomes clear what is required :+1:

PR #739, opened 2019-07-20
[WIP] Add ground track plotting 

Sedictious:
This is the initial commit for the ground track plotting feature. In theory this should work just fine but after testing, it seems like Cesium can't render exceptionally big paths (in particular, it seems that the problem is with defining a ground track arc wider than 180°)

💡 Possible solutions:
* Use polylines instead of paths (that would mean we would loose the ability to animate the ground track)

* Break down the path into different packets and load them dynamically.


Juanlu001:
> it seems like Cesium can't render exceptionally big paths (in particular, it seems that the problem is with defining a ground track arc wider than 180°)

Hmm interesting, perhaps it has to do with this?

https://macwright.org/2016/09/26/the-180th-meridian.html

Can you craft an example so we can reproduce the problem?

Sedictious:
@Juanlu001 Interesting read, I didn't even know it had a name :P 
I'll upload an example to make testing easier

Sedictious:
@Juanlu001 Huh... I tested it again and it seems to be working just fine (I uploaded the JSON in the [app examples](https://github.com/poliastro/cesium-app/tree/master/examples/ex2-groundtrack)). I'll see if I can find any relevant information but it may've been a bug that was resolved!

src/poliastro/czml/extract_czml.py
```diff
@@ -100,6 +100,52 @@ def _init_orbit_packet_cords_(self, i, rtol):
...
+        # Get rounding factor given the relative tolerance
+        rf = 0
+        while rtol < 1:
+            rtol *= 10
+            rf += 1
+
+        ellipsoid = self.cust_prop[0]
+
+        for k in range(self.orbits[i][1] + 2):
+            position = propagate(self.orbits[i][0], TimeDelta(k * h), rtol=rtol)
```
Juanlu001: If I understand correctly, perhaps we could replace this `for` loop by `orbit.sample`, or perhaps propagating along the whole time interval?

src/poliastro/czml/extract_czml.py
```diff
@@ -100,6 +100,52 @@ def _init_orbit_packet_cords_(self, i, rtol):
...
+
+            # flatten list
+            cords = list(map(lambda x: round(x[0], rf), cords.tolist()))
+            t, p = cords[0], cords[1:]
+            pr_p = project_point_on_ellipsoid(
+                p[0], p[1], p[2], ellipsoid[0], ellipsoid[1], ellipsoid[2]
+            )
+            #  This is a hack to ensure that our point lies above the surface of the
+            # ellipsoid. An iterative method could be used instead but the error margin
+            # is too small to be worth it.
```
Juanlu001: Good to have a docstring here - could you comment a bit on how the hack works though? Not sure what `+ 0.1` does (to clarify: could you expand the code comment?)

src/poliastro/czml/extract_czml.py
```diff
@@ -129,10 +175,13 @@ def _change_custom_params(self, ellipsoid, pr_map, scene3D):
 
         if ellipsoid is None:
             ellipsoid = [6378137.0, 6378137.0, 6356752.3142451793]
```
Juanlu001: Now that we have these values in constants, before making this general for any ellipsoid, can we at least take the values from there?

src/poliastro/czml/extract_czml.py
```diff
@@ -192,10 +241,14 @@ def add_ground_station(
...
+
             if self.cust_prop[0]:
-                a, b = self.cust_prop[0][:1]  # get semi-major and semi-minor axises
+                a, b = (
+                    self.cust_prop[0][0],
+                    self.cust_prop[0][2],
+                )  # get semi-major and semi-minor axises
             else:
-                a, b = 6378137.0, 6378137.0
+                a, b = 6378137.0, 6356752.3142451793
```
Juanlu001: Same as above

src/poliastro/czml/utils.py
```diff
@@ -30,3 +30,94 @@ def ellipsoidal_to_cartesian(a, b, lat, lon):
...
+
+    Parameters
+    ----------
+    x, y, z: Cartesian coordinates of point
+
+    a, b, c: semi-axises of the ellipsoid
+    """
+    p1, p2 = intersection_ellipsoid_line(x, y, z, x, y, z, a, b, c)
+
+    if np.linalg.norm([p1[0] - x, p1[1] - y, p1[2] - z]) <= np.linalg.norm(
```
Juanlu001: If I understand correctly, this blocks selects the closest one of the solutions, right?

src/poliastro/tests/test_czml.py
```diff
@@ -1,19 +1,11 @@
...
 
-try:
-    from poliastro.czml.extract_czml import CZMLExtractor  # noqa: E402
 
-    import czml3  # noqa: F401
-except ImportError:
-    pass
-
-
-@pytest.mark.skipif("czml3" not in sys.modules, reason="requires czml3")
```
Juanlu001: Let's leave this like it was before, now that we are properly installing it in the CI

src/poliastro/czml/extract_czml.py
Sedictious: Sure but I'm not sure whether I want to replace them with Earth's constants since those are based on Cesium's default Ellipsoid ``WGS84`` values (perhaps we could also introduce it as a constant)

src/poliastro/czml/extract_czml.py
Sedictious: Ok, I'll try to make it more clear. The added constant simply ensures that the point lies above the surface. The reason we have to do this, is because the resulting point may be slightly bellow the surface because of numerical errors (it should be rounded up anyway, but better be sure that no graphical glitch will result out of this)

src/poliastro/czml/utils.py
Sedictious: Yes that's it!

src/poliastro/czml/extract_czml.py
Sedictious: We could use ``orbit.sample`` but there is not way moving around the ``for`` loop unless we use some kind of mapping (which would make the code more hard to understand). Is there any specific advantage to using ``orbit.sample``?

src/poliastro/czml/extract_czml.py
Juanlu001: It's a valid concern. The way I see this is:

a) The discrepancies in radius is an Earth-specific problem
b) CZML is a visualization system, so if I have to choose I prefer the propagation to have correct results

So, even though what we are doing is not exactly accurate, it is up to ~30 meters.

src/poliastro/czml/extract_czml.py
Juanlu001: If it's going to make the code harder to understand, then no :)

PR #743, opened 2019-07-25
Allow groundtrack lead/trail time

Sedictious:
Just a quick PR to add those two parameters to the ground track plotter, allowing you to animate the ground track


Juanlu001:
All good, merging!

Issue #740, opened 2019-07-20
Incorrect Hohmann maneuver when not at pericenter?


astrojuanlu:
🐞 **Problem**

I think there is a potential problem here:

https://github.com/poliastro/poliastro/blob/bc4c00887a6263601143366007edb6b7a711be61/src/poliastro/maneuver.py#L100-L101
https://github.com/poliastro/poliastro/blob/bc4c00887a6263601143366007edb6b7a711be61/src/poliastro/maneuver.py#L144

in that the time to get to the pericenter is not being taken into account for the maneuver, and therefore perhaps doing an `apply` on the initial orbit won't have the intended effect.

💡 **Possible solutions**

_Note I didn't check whether this is an actual bug or not yet_

Do not hardcode the initial maneuver to `0 * u.s` and use the time to get to the pericenter.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

jorgepiloto:
Interesting... The same happens in `Bielliptic` ones. Some possible solutions that came to my mind:

* We could raise `ValueError("Not in pericenter...")` so the user is forced to propagate before solving the maneuver.
* In case-solving internally time to pericenter, the set of impulses should be like:

```python
{propagation: till periapsis, second: first hohmann impulse, third: second hohmann impulse}
```

That propagation impulse always has `deltaV = 0 * u.km / u.s` and time may vary between [0 and T)

Juanlu001:
I would rather do the latter instead of raising the error.

Juanlu001:
Fixed in #744.

astrojuanlu:
I think I forgot to close this. Doing it now.

Issue #731, opened 2019-07-17
Error when using different units with bielliptic maneuver


Sedictious:
🐞 **Problem**

There seems to be a bug when using different units in bielliptic maneuver.

🖥 **How to reproduce this bug**

* In ``constants.py`` replace the definition of ``Earth`` with: 

``` python
R_earth = Constant(
    "R_earth",
    "Earth equatorial radius",
    6.3781366e3,
    "km",
    0.0001,
    "IAU Working Group on Cartographic Coordinates and Rotational Elements: 2015",
    system="si",
)
```

After running the tests in ``test_maneuver.py``, ``test_bielliptic_maneuver()`` should fail:
```
actual = <Quantity 653697.65245305 s>, desired = <Quantity 593.919803 h>
rtol = 1e-06, atol = None, kwargs = {}
np = <module 'numpy' from 'C:\\Users\\P\\Desktop\\poliastro\\venv\\lib\\site-packages\\numpy\\__init__.py'>
_unquantify_allclose_arguments = <function _unquantify_allclose_arguments at 0x0000027155408F28>

    def assert_quantity_allclose(actual, desired, rtol=1.e-7, atol=None,
                                 **kwargs):
        """
        Raise an assertion if two objects are not equal up to desired tolerance.
    
        This is a :class:`~astropy.units.Quantity`-aware version of
        :func:`numpy.testing.assert_allclose`.
        """
        import numpy as np
        from astropy.units.quantity import _unquantify_allclose_arguments
        np.testing.assert_allclose(*_unquantify_allclose_arguments(
>           actual, desired, rtol, atol), **kwargs)
E       AssertionError: 
E       Not equal to tolerance rtol=1e-06, atol=0
E       
E       Mismatch: 100%
E       Max absolute difference: 1484413.63834695
E       Max relative difference: 0.69426397
E        x: array(653697.652453)
E        y: array(2138111.2908)
```

(Tested with Python 3.6)

💡 **Possible solutions**

If this bug is reproduce-able, it may be the case that somewhere the astropy units are converted into raw values, so fixing this may fix the error.

jorgepiloto:
Investigating this... Could it be the fact that you wrote the number in `km` and variable `system='si'` is set to the international system?

Sedictious:
@jorgepiloto Yes it seems to work when set to meters (this also seems to be the only test that fails).

Juanlu001:
[...]

PR #744, opened 2019-07-26
Hohmann time to pericenter

jorgepiloto:
Tries to solve #740 and #731 . This pull request includes:

* New orbit class method: `time_to_anomaly`
* Propagates `hohmann` and `bielliptic` to pericenter and adds that `time of flight` in the `Maneuver` instance.
* Makes use of `Quantity.decompose()` to avoid weird behavior with units inside `Maneuver` instances.

### DIFF
--- a/src/poliastro/maneuver.py
+++ b/src/poliastro/maneuver.py
@@ -97,8 +97,12 @@ class Maneuver(object):
             Final altitude of the orbit
         """
 
-        if orbit_i.nu is not 0 * u.deg:
+        # Propagate till periapsis
+        if orbit_i.nu != 0 * u.deg:
+            t_pericenter = orbit_i.time_to_anomaly(0 * u.deg)
             orbit_i = orbit_i.propagate_to_anomaly(0 * u.deg)
+        else:
+            t_pericenter = 0 * u.s
 
         # Initial orbit data
         k = orbit_i.attractor.k
@@ -141,7 +145,9 @@ class Maneuver(object):
 
         t_trans = np.pi * np.sqrt(a_trans ** 3 / k)
 
-        return cls((0 * u.s, dv_a), (t_trans, dv_b))
+        return cls(
+            (t_pericenter.decompose(), dv_a.decompose()), (t_trans.decompose(), dv_b.decompose())
+        )
 
     @classmethod
     def bielliptic(cls, orbit_i, r_b, r_f):
@@ -184,8 +190,13 @@ class Maneuver(object):
         r_f: astropy.unit.Quantity
             Final altitude of the orbit
         """
-        if orbit_i.nu is not 0 * u.deg:
+
+        # Propagate till periapsis
+        if orbit_i.nu != 0 * u.deg:
+            t_pericenter = orbit_i.time_to_anomaly(0 * u.deg)
             orbit_i = orbit_i.propagate_to_anomaly(0 * u.deg)
+        else:
+            t_pericenter = 0 * u.s
 
         # Initial orbit data
         k = orbit_i.attractor.k
@@ -234,7 +245,11 @@ class Maneuver(object):
         t_trans1 = np.pi * np.sqrt(a_trans1 ** 3 / k)
         t_trans2 = np.pi * np.sqrt(a_trans2 ** 3 / k)
 
-        return cls((0 * u.s, dv_a), (t_trans1, dv_b), (t_trans2, dv_c))
+        return cls(
+            (t_pericenter.decompose(), dv_a.decompose()),
+            (t_trans1.decompose(), dv_b.decompose()),
+            (t_trans2.decompose(), dv_c.decompose()),
+        )
 
     @classmethod
     def lambert(cls, orbit_i, orbit_f, method=lambert_izzo, short=True, **kwargs):
@@ -269,7 +284,10 @@ class Maneuver(object):
         else:
             dv_a, dv_b = sols[-1]
 
-        return cls((0 * u.s, dv_a - orbit_i.v), (tof.to(u.s), orbit_f.v - dv_b))
+        return cls(
+            (0 * u.s, (dv_a - orbit_i.v).decompose()),
+            (tof.to(u.s), (orbit_f.v - dv_b).decompose()),
+        )
 
     def get_total_time(self):
         """Returns total time of the maneuver.
--- a/src/poliastro/maneuver.py
+++ b/src/poliastro/maneuver.py
@@ -146,7 +146,8 @@ class Maneuver(object):
         t_trans = np.pi * np.sqrt(a_trans ** 3 / k)
 
         return cls(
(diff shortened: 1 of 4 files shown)


src/poliastro/twobody/orbit.py
```diff
@@ -1024,6 +1024,34 @@ def propagate(self, value, method=mean_motion, rtol=1e-10, **kwargs):
...
+        value : ~astropy.units.Quantity
+
+        Returns
+        -------
+        tof: ~astropy.units.Quantity
+            Time of flight required.
+        """
+
+        # TODO: Avoid repeating logic with mean_motion?
+        p, ecc, inc, raan, argp, _ = rv2coe(
```
Juanlu001: Are we using these values anywhere?

src/poliastro/maneuver.py
```diff
@@ -97,8 +97,12 @@ def hohmann(cls, orbit_i, r_f):
             Final altitude of the orbit
         """
 
-        if orbit_i.nu is not 0 * u.deg:
+        # Propagate till periapsis
+        if orbit_i.nu != 0 * u.deg:
+            tof = orbit_i.time_to_anomaly(0 * u.deg)
```
Juanlu001: Can we add a test for this case?

src/poliastro/maneuver.py
```diff
@@ -184,8 +188,13 @@ def bielliptic(cls, orbit_i, r_b, r_f):
         r_f: astropy.unit.Quantity
             Final altitude of the orbit
         """
-        if orbit_i.nu is not 0 * u.deg:
+
+        # Propagate till periapsis
+        if orbit_i.nu != 0 * u.deg:
+            tof = orbit_i.time_to_anomaly(0 * u.deg)
```
Juanlu001: Same as above

jorgepiloto:
As commented in the official Riot channel, while working on this, a strange behavior was found in the `get_total_time()` method in the `Maneuver` class. This was caused somehow by the `astropy.units.Quantity`. By making use of the `decompose()` method, this problem was avoided.

This problem was causing #731, so it is also solved :+1:

src/poliastro/twobody/orbit.py
jorgepiloto: Sure! Sorry, did not notice this :sweat:

src/poliastro/twobody/orbit.py
```diff
@@ -1024,6 +1024,27 @@ def propagate(self, value, method=mean_motion, rtol=1e-10, **kwargs):
...
+        Returns
+        -------
+        tof: ~astropy.units.Quantity
+            Time of flight required.
+        """
+
+        # Compute time of flight for correct epoch
+        M = nu_to_M(self.nu, self.ecc)
+        new_M = nu_to_M(value, self.ecc)
+        tof = Angle(new_M - M).wrap_at(360 * u.deg) / self.n
```
Juanlu001: Can we reuse this logic in `propagate_to_anomaly`? We could do

```
time_of_flight = self.time_to_anomaly(value)
```

and drop lines 1062 to 1071.

src/poliastro/tests/test_maneuver.py
```diff
@@ -90,3 +137,7 @@ def test_apply_maneuver_correct_dimensions():
 
     assert new_orb.r.ndim == 1
     assert new_orb.v.ndim == 1
+
+
+if __name__ == "__main__":
```
Juanlu001: Leftover from testing, I suppose?

src/poliastro/constants.py
```diff
@@ -286,8 +286,8 @@
 R_earth = Constant(
     "R_earth",
     "Earth equatorial radius",
-    6.3781366e6,
-    "m",
+    6.3781366e3,
```
Juanlu001: Can we remove this unrelated change?

Juanlu001: Also, remember we still have https://github.com/poliastro/poliastro/issues/731

src/poliastro/maneuver.py
```diff
@@ -141,7 +145,9 @@ def hohmann(cls, orbit_i, r_f):
 
         t_trans = np.pi * np.sqrt(a_trans ** 3 / k)
 
-        return cls((0 * u.s, dv_a), (t_trans, dv_b))
+        return cls(
+            (tof.decompose(), dv_a.decompose()), (t_trans.decompose(), dv_b.decompose())
```
Juanlu001: Shouldn't this be `t_trans + tof`? (Did not think about it in depth, please confirm)

src/poliastro/maneuver.py
```diff
@@ -234,7 +245,11 @@ def bielliptic(cls, orbit_i, r_b, r_f):
         t_trans1 = np.pi * np.sqrt(a_trans1 ** 3 / k)
         t_trans2 = np.pi * np.sqrt(a_trans2 ** 3 / k)
 
-        return cls((0 * u.s, dv_a), (t_trans1, dv_b), (t_trans2, dv_c))
+        return cls(
+            (tof.decompose(), dv_a.decompose()),
+            (t_trans1.decompose(), dv_b.decompose()),
```
Juanlu001: Same as above, please confirm we don't have to add `tof` (which shouldn't in fact be called `tof`, but `t_pericenter` or something like that? to avoid confusion) to the other times.

src/poliastro/tests/test_maneuver.py
```diff
@@ -61,6 +61,31 @@ def test_hohmann_maneuver():
     )
 
 
+def test_hohmann_maneuver_not_at_pericenter():
+
```
Juanlu001: Instead of creating a new test, I propose parametrizing `test_hohmann_maneuver`, which is already using the data from example 6-1, and instead of starting from a circular orbit start from an elliptical one, whose pericenter matches the expected altitude. Therefore, when the anomaly is 0, we are in the case where the time to pericenter will be zero, and if we move to a different starting anomaly, we will have a different time.

src/poliastro/tests/test_maneuver.py
```diff
@@ -77,6 +102,28 @@ def test_bielliptic_maneuver():
     assert_quantity_allclose(man.get_total_time(), expected_t_trans, rtol=1e-6)
 
 
+def test_bielliptic_maneuver_not_at_pericenter():
```
Juanlu001: Same comment as above

src/poliastro/tests/test_maneuver.py
jorgepiloto: Ups, sorry! :sweat_smile:

src/poliastro/twobody/orbit.py
jorgepiloto: I simplified the `propagate_to_anomaly` by making use of this new method and also using `self.property` instead of calling the `rv2coe` function, which lead to dropping those lines :+1:

src/poliastro/constants.py
jorgepiloto: Sure! Will open a pull request for this new change :smile:

PR #746, opened 2019-08-01
Add angular speed and rotational period

Sedictious:


src/poliastro/constants/rotational_elements.py
```diff
@@ -0,0 +1,209 @@
+from astropy.constants import Constant
+
+T_earth = Constant(
```
Juanlu001: Could you please point me out to where exactly are these constants defined in Vallado? In the dictionary of symbols, `T` is not used for rotational period.

src/poliastro/constants/rotational_elements.py
```diff
@@ -0,0 +1,209 @@
...
+    "T_earth",
+    "Earth rotational period",
+    0.9972698,
+    "d",
```
Juanlu001: Shouldn't the unit be angles per unit time, like deg/second or similar?

src/poliastro/constants/rotational_elements.py
```diff
@@ -0,0 +1,209 @@
...
+    "angular_speed_earth",
+    "Earth angular speed",
+    7.29211502e-05,
+    "rad / s",
+    0,
+    "Vallado, D. and McClain, W. Fundamentals of astrodynamics and applications.",
+    system="si",
+)
+
+angular_speed_sun = Constant(
```
Juanlu001: I propose we use `omega_Sun`, `omega_Earth`, etc

Juanlu001: Also, aren't these definitions redundant with the rotational periods? I think we should only keep the angular velocities

Juanlu001: Also, if you could also point out where are these values defined?

Juanlu001: What we discussed in private is that this collisions with how the WGCRE defines the rotational velocities. It's nice to have mean values though, so we will probably want to keep both...?

src/poliastro/constants/rotational_elements.py
Sedictious: It would probably make more sense but this is how it was referenced so I kept it like this for consistency

src/poliastro/constants/rotational_elements.py
Sedictious: It's not used so I should've probably replace it with ``rotational_period_`` (that is if we decide to keep this constant at all)

src/poliastro/constants/rotational_elements.py
Sedictious: @Juanlu001 This values weren't explicitly defined in Vallado and they were derived from the rotational period reference (this is partly why I included both instead of only the rotational period). I can remove the rotational period and add a docstring explaining this if it'd be better

src/poliastro/constants/rotational_elements.py
Juanlu001: Well, if it's a period then it makes total sense that it's in [time] units only, so let's leave it like that.

src/poliastro/constants/rotational_elements.py
Juanlu001: Yes, if these are derived values let's not include them in constants, and only the rotational periods. Then we could have a property like `body.angular_speed` that would do the required operation. This way we keep `poliastro.constants` only with stuff we can check in a book, a paper or similar.

Juanlu001:
Copying a comment I left in an outdated review (oh well... I sometimes mess up with this)

> Yes, if these are derived values let's not include them in constants, and only the rotational periods. Then we could have a property like `body.angular_speed` that would do the required operation. This way we keep `poliastro.constants` only with stuff we can check in a book, a paper or similar.

On the other hand, this will need a rebase.

Juanlu001:
Ready to merge when the tests pass!

Juanlu001:
Merged, thanks a lot!

PR #756, opened 2019-08-14
Reformat on atmospheres

jorgepiloto:
This pull request reformats the actual `poliastro.atmosphere` module by including the following features:

* Parameters for each atmosphere have been stored in `atmosphere/data/*.dat` by making use of Astropy capabilities, which translates into a cleaner implementation in each `*.py` file.

* Pure atmospheric models have been partially or fully implemented by making use of outdated official paper constants.

* Values are easier to modify and read since each line of the data files has the same format as official tables.

I am opening this as a draft since I still need to:

- [x] :warning: Modify `setup.cfg` and to install `*.dat` files. :warning: 
- [x] More tests, so coverage is not decreased.
- [x] More documentation on atmospheres.
- [ ] Test if this solves #694


Juanlu001:
:clap: Merged!

Issue #380, opened 2018-05-29
Return velocity in Orbit.sample


Juanlu001:
🐞 **Problem**

At the moment, `Orbit.sample` does return only an array of times and positions. This was done on purpose to deliver the feature more quickly and because it was enough for plotting:

https://github.com/poliastro/poliastro/issues/237#issuecomment-335436315

However, the velocity is not returned. Sometimes the user just wants that.

On the other hand, although there are [orbit determination algorithms](https://github.com/aerospaceresearch/orbitdeterminator) that can recover an orbit from observations data, `Orbit.sample` should give more complete information.

🎯 **Goal**

To have the velocity apart from the already available position in `Orbit.sample`.

💡 **Possible solutions**

As anticipated in https://github.com/poliastro/poliastro/issues/237, at some point it would be useful to sample any interesting quantity of a propagation of a trajectory. However, as a starting point we can focus on the velocity.

The combination of `CartesianRepresentation` and `CartesianDifferential` is still considered as an experimental API (http://docs.astropy.org/en/stable/coordinates/representations.html#attaching-differential-s-to-representation-s), so we should use it with care. Perhaps we could start by just returning a multidimensional `Quantity`, but it feels kind of wrong. Also, let's not forget about #257, which perhaps should come first.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
I don't like returning different things depending on a parameter - at the moment we're doing:

https://github.com/poliastro/poliastro/blob/b659a947d9960e13ca844e7d049206e680fb0b8a/src/poliastro/plotting.py#L181

So perhaps we should do

```
positions = orbit.sample(self.num_points, method)[1]
```

instead, for the moment.

nikita-astronaut:
OK. Do people really need the feature with velocity? If yes, why not doing

Juanlu001:
I changed the title of the issue because I think we should focus on sampling arbitrary properties, to avoid code like this:

https://github.com/poliastro/poliastro/blob/335a37cf619b8822e837a2353edeb5b3002bad5a/src/poliastro/tests/tests_twobody/test_perturbations.py#L260-L269

If people are interested in getting the velocity, I'm sure they will want to know the time history of the right ascension or the true anomaly as well.

[...]

PR #757, opened 2019-08-16
Sampler for any Orbit attribute

jorgepiloto:
This is just a possible implementation by making use of `np.vectorize` function. The desired attribute should be passed as string. This works in the following way:

```python
from poliastro.util import time_range
t_span = time_range(ss.epoch, end=ss.epoch + 24 * u.h)
nu_span = ss.sample_attr("nu", t_span)
``` 

This was developed to solve #380 :+1:


astrojuanlu:
In light of my latest comment at #380, I think we should close this old PR @jorgepiloto, what do you think?

jorgepiloto:
Sure! No problem with that :rocket:


20/04/26 01:17, Juan Luis Cano Rodríguez wrote:

PR #763, opened 2019-08-18
Refactor coordinates, frames, bodies, everything!!1!

Juanlu001:
Since we introduced the bodies with static methods I always found them ugly. Finally I took the opportunity to move all the rotation information and the ugliness to `poliastro.frames`. There is some ugliness we can't escape because astropy frames require some magic to initialize (I have some ideas about how to improve that, but I need more time).

I am not 100 % happy with this pull request though, because I dropped some features along the way, I broke one test, and I didn't check the notebooks. A bit more work is needed, but I'm glad I finally laid down my idea for `poliastro.bodies`.


src/poliastro/bodies.py
```diff
@@ -19,97 +19,50 @@
...
+    def angular_velocity(self):
+        return (2 * math.pi * u.rad) / (self.rotational_period).to(u.s)
+
     def __str__(self):
         return f"{self.name} ({self.symbol})"
 
     def __repr__(self):
         return self.__str__()
 
-    def rot_elements_at_epoch(self, epoch):
```
Juanlu001: These are the functions that moved

Juanlu001:
Missing:

* [ ] Write down all the features and functions I dropped
* [ ] Check `atmosphere` - that code uses some custom constants attached to `Earth` (I think) and will probably break.
* [ ] Think of alternative for bodies with custom properties
  - Adding a dictionary breaks the frame mapping, because we are using the bodies as keys
* [ ] Add at least some smoke tests for the fixed frames
  - Even if the https://github.com/poliastro/validation/ machinery is not ready, we could do a test with GMAT and copy-paste the values
  - Alternatively, we could at least write a test copy-pasting the result from our current implementation, to notice when we change anything

jorgepiloto:
Old constant values were used in `atmosphere` for each one of the atmospheric models, so no dependency on `bodies` has been stablished. This is the reason why `tests_atmosphere` were passing.

However the `plotting.porkchop` failed, so I will investigate why :confused:

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -459,7 +462,7 @@ def normalize_to_Curtis(t0, sun_r):
     return 149600000 * r / norm(r)
 
 
-@pytest.mark.slow
+# @pytest.mark.slow
```
Juanlu001: Leftover?

src/poliastro/tests/tests_twobody/test_perturbations.py
```diff
@@ -192,11 +193,13 @@ def test_atmospheric_drag():
...
 
     # parameters of the atmosphere
-    rho0 = Earth.rho0.to(u.kg / u.km ** 3).value  # kg/km^3
-    H0 = Earth.H0.to(u.km).value
+    rho0 = rho0_earth.value  # kg/km^3
+    H0 = H0_earth.value
     tof = 100000  # s
 
-    dr_expected = -B * rho0 * np.exp(-(norm(r0) - R) / H0) * np.sqrt(k * norm(r0)) * tof
+    dr_expected = (
```
Juanlu001: Not sure why we were using the raw value before, doesn't this work using quantities?

Juanlu001:
One doctest failing:

```
_______________ src/poliastro/tests/tests_plotting/test_core.py ________________
142: error: "None" has no attribute "template"
```

src/poliastro/tests/tests_twobody/test_perturbations.py
jorgepiloto: Ups! My fault :hand:

src/poliastro/tests/tests_twobody/test_perturbations.py
jorgepiloto: Finally solved this, sorry for being so late. I think units were avoid because when used, complexity inside `cowell`  makes it return a wrong result. Something similar happened some time ago in the `poliastro.maneuver` module

jorgepiloto:
I opened a pull request to @Juanlu001 branch to solve those minor errors in `test_perturbations` :+1:

Juanlu001:
We will have to apply changes similar to https://github.com/poliastro/poliastro/pull/763/commits/d72bdabf9ef55b6d42f7f6e8bbcbccfd66aa038e to the notebooks as well.

Juanlu001:
Should rebase after https://github.com/poliastro/poliastro/pull/785 is merged.

Juanlu001:
Well well well, the coverage check is failing but this is long overdue. The sooner we land this refactor and start testing it ourselves and feeling comfortable with the new code, the better. We can adjust later.

Merging!

PR #758, opened 2019-08-16
Mark tests that use remote data

olebole:
In Debian, we run the build time and CI tests without access to the internet. This makes some of the poliastro tests fail.

To handle this, there is the "pytest-remotedata" package (and the "pytest-doctestplus" package for doctests) that add a "remotedata" option to pytest. Both packages are already pulled with the "astropy-helpers" package. This patch adds the options on the necessary places.

Would you consider applying this to make my life easier here? Thanks!

PR #764, opened 2019-08-18
EXP: Why are you using astroquery conftest.py

pllim:
Just trying random things in response to Slack post from @Juanlu001 regarding failure of #758 .

Ref: https://docs.pytest.org/en/latest/pythonpath.html

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,1 @@
+# Configure pytest behavior here.
--- a/conftest.py
+++ b/conftest.py
@@ -1,1 +1,1 @@
-# Configure pytest behavior here.
+# Configure pytest behavior here
--- a/src/poliastro/tests/__init__.py
+++ b/src/poliastro/tests/__init__.py
@@ -1,0 +1,1 @@
+# Tests for poliastro


pllim:
Interesting... Adding 2 empty files caused this failure: https://circleci.com/gh/poliastro/poliastro/5015

Juanlu001:
Perplexing. Thanks @pllim, will give it a look!

pllim:
Unfortunately, I am not familiar enough with this package to proceed. So, feel free to close this or push to my branch here using your maintainer power.

Juanlu001:
Well, not that strange. This means we have a couple of mypy errors in the tests, which were not being detected before because the tests directory was not a Python package. Should be unrelated to the astroquery thing, I guess (I will fix them anyway).

Juanlu001:
@pllim fixed mypy errors, now the tests are running. Let's see what happens...

Juanlu001:
Adding these two empty files no longer has any effect.

Juanlu001:
As described in https://github.com/poliastro/poliastro/pull/758#issuecomment-522350802, the issue is probably in the collection strategy of pytest-doctestplus. Closing this PR.

PR #767, opened 2019-08-18
Remove executable bit from some Python sources

olebole:
Just a small glitch that I discovered during packaging 0.13. ;-)


Juanlu001:
Woops! Nice catch :) Merging!

PR #785, opened 2019-10-20
Fix typing errors

Juanlu001:


Juanlu001:
Reason: https://mypy-lang.blogspot.com/2019/10/mypy-0740-released.html?spref=tw

PR #793, opened 2019-11-02
Fix #781 : Fix the last broken link

### LINKED ISSUE #781: binder link: embedded hyperlinks broken
🐞 **Problem**

Clicking on the binder link in the README leads to a page with broken hyperlinks. Cannot navigate to examples. 

The cached version that shows up during loading does work, though.

shreyasbapat:
Fix #781 finally.


Juanlu001:
Oops, good catch @shreyasbapat! Merging!

PR #771, opened 2019-09-05
Remove poliastro wrappers of NumPy functions currently supported in numba

Juanlu001:
I realized today that recent numba versions have support for `numpy.linalg.norm`. Four tests fail locally though, and I did a hack that will probably crash the quality step.

PR #794, opened 2019-11-04
Finally remove norm

Juanlu001:
Second attempt after https://github.com/poliastro/poliastro/pull/771. I applied @stuartarchibald suggestion from https://github.com/numba/numba/issues/4701#issuecomment-541630071.


Juanlu001:
Removing code is always such a pleasure!

PR #797, opened 2019-11-10
Prepare refactor of Bodies and orbits

Juanlu001:
I am in the middle of a refactor of `Body` classes, so I decided to put some preliminary work first, including some changes and deprecations.

The most important one is that `Orbit.sample` does not return a frame anymore, but a representation. I think it's equally useful and less burden in the implementation side.


Juanlu001:
These are my notes: https://github.com/poliastro/poliastro/wiki/Body-orbits

Juanlu001:
Note that the `docs` step will likely fail, I didn't touch the notebooks yet.

Juanlu001:
Thanks a lot for the fixes @jorgepiloto! Waiting for the tests to pass

Juanlu001:
Merged :tada:

PR #802, opened 2019-11-17
Migrate to flit

Juanlu001:
Some pending items noted:

* [x] No short description?
* [x] Keywords cannot contain spaces?
* [x] PEP 508 URL requirements not supported in extras?

cc @takluyver


Juanlu001:
I tried `flit install` and `flit build` locally and it seems to work! :tada:

Juanlu001:
Being hit by https://github.com/takluyver/flit/issues/75#issuecomment-290139035, supposedly fixed in https://github.com/tox-dev/tox/pull/954.

Juanlu001:
* [ ] `flit install --symlink` does not install non-extra requirements? See https://circleci.com/gh/poliastro/poliastro/5423

takluyver:
Thanks! The short description is actually taken from the first line of the module's docstring - slightly strange, but it saves a bit of repetition.

I'd forgotten that there even was a keywords field in there. Assuming the metadata spec & PyPI allow spaces, it should be easy to allow a list for that.

I'll try to get to looking at the points about requirements.

Juanlu001:
Thanks a lot @takluyver!

takluyver:
The keywords field is a space-separated list [in the standard metadata format](https://packaging.python.org/specifications/core-metadata/#keywords), and no quoting rules are defined, so keywords with spaces aren't feasible.

It may be nice to allow specifying them as a toml list rather than a single string, anyway. I've opened an issue for that: https://github.com/takluyver/flit/issues/293

takluyver:
Looking at your released files, setuptools appears to create a comma separated list of keywords, not space separated. I'll bring it up on the distutils-sig mailing list - it might be the spec that gets changed.

takluyver:
Here's the thread I've started: https://mail.python.org/archives/list/distutils-sig@python.org/thread/LFTLGMIIB5PRG7NUUIGGGX4COGX53DMR/

Juanlu001:
> Looking at your released files, setuptools appears to create a comma separated list of keywords, not space separated.

Hah, I've been wrong all this time :) Thanks for the leg work

takluyver:
distutils has been wrong all this time. Or possibly the spec has been 'wrong' :wink:

takluyver:
The issues with requirements are hopefully fixed in RC3, if you want to give it another try.

We seem to have agreed that the spec was wrong, and keywords should be comma separated. I think you can put them directly in like that.

pyproject.toml
```diff
@@ -10,11 +10,9 @@ build-backend = "flit_core.buildapi"
 module = "poliastro"
 author = "Juan Luis Cano Rodríguez"
 author-email = "hello@juanlu.space"
-# TODO: No short description?
 description-file = "README.rst"
 home-page = "https://github.com/poliastro/poliastro/"
-# TODO: Keywords cannot contain spaces?
-keywords = "aero aerospace engineering astrodynamics orbits kepler orbital mechanics"
+keywords = "aero, aerospace, engineering, astrodynamics, orbits, kepler, orbital mechanics"
```
takluyver: It looks like distutils & setuptools do a simple `.split(',')`, so I suspect you want no spaces after the commas. I'll try to sand the rough edges off this soon.

Juanlu001: Thanks @takluyver! That should be good enough, will change it now.

Juanlu001:
Awesome, everything is working now! Thanks a lot @takluyver, merging this when the tests pass :heart:

takluyver:
Thanks :-)

PR #804, opened 2019-11-19
frames: Fix conversion between inertial & fixed frames

### LINKED ISSUE #801: Fixed planetary frame transformation error
When creating a coordinate in a planetary fixed frame a transformation is not possible due to 
an error in `frames/fixed.py`

🐞 **Problem**

```
testloc = frames.fixed.MarsFixed(10 * u.deg, 20 * u.deg, distance=bodies.Mars.R, obstime=start_time)
testloc.transform_to(frames.ICRS)
```

```
~/git/misc/poliastro/src/poliastro/frames/fixed.py in to_equatorial(fixed_coo, equatorial_frame)
     73         assert fixed_coo.body == equatorial_frame.body
     74 
---> 75         r = fixed_coo.data.xyz
     76         v = fixed_coo.data.differentials["s"].d_xyz
     77 

AttributeError: 'SphericalRepresentation' object has no attribute 'xyz'
```

💡 **Possible solutions**

Converting to CartesianRepresentation and filling in the CartesianDifferentials. Possibly similar to how it is handled for `EarthLocation` in `earth.py` in astropy.

noc0lour:
For conversion from fixed frames to inertial frames coordinates
in the fixed coordinate system need to be extracted as cartesian and
thus the internal data structure needs to be converted to a
CartesianRepresentation first.

Closes #801

I've had no time to add a test yet and I consider this code rather a hack than a real solution, suggestions very welcome.  I took code for `EarthLocation` in `earth.py` in astropy as inspiration.

### DIFF
--- a/src/poliastro/frames/fixed.py
+++ b/src/poliastro/frames/fixed.py
@@ -1,5 +1,4 @@
-import math
-
+import numpy as np
 from astropy import units as u
 from astropy.coordinates import (
     HCRS,
@@ -10,7 +9,6 @@ from astropy.coordinates import (
     FunctionTransform,
     TimeAttribute,
     frame_transform_graph,
-    get_body_barycentric_posvel,
 )
 from astropy.coordinates.builtin_frames.utils import DEFAULT_OBSTIME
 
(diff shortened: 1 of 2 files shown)


src/poliastro/frames/fixed.py
```diff
@@ -72,8 +73,15 @@ def __new__(cls, *args, **kwargs):
     def to_equatorial(fixed_coo, equatorial_frame):
         assert fixed_coo.body == equatorial_frame.body
 
-        r = fixed_coo.data.xyz
-        v = fixed_coo.data.differentials["s"].d_xyz
+        fixed_coo_cart = fixed_coo.data.to_cartesian()
```
Juanlu001: I guess this line is the one fixing the issue?

Juanlu001:
Thanks for the PR @noc0lour! I see that you also fixed another problem: the lack of differentials.

My advice is that you write tests for both cases (they should be simplified cases of the code that made you discover the failure). I anticipate that testing the first problem will be easy, but the second will not :) Also, I don't agree with setting the derivatives to zero, and would rather prefer to not return the velocity if the differentials are not present.

noc0lour:
@Juanlu001 the differentials are quite important for usage with `transform_to` since even though you have no differentials in your fixed frame (if you are not moving w.r.t the body) a transformation to an inertial frame will give you differentials due to the orbit of the body & rotation of the body. Adding zero velocities is more or less what I took from `astropy` (which is of course only applicable for non-moving objects as it is the case for EarthLocation). Thus I assumed if one does not give differentials the velocity is 0?

noc0lour:
Ah, that was a misconception on my side. I'll update my PR to provide a clean solution.

noc0lour:
This PR has now a list of incremental commits each fixing a isolated issue with the fixed frames.

Juanlu001:
Good! I just checked that the tests fail in `master` :) Merging, thanks a lot @noc0lour!

PR #808, opened 2019-11-25
Update docs/contributing.rst

kerel-fs:
Three commits updating the contributing guide as discussed via matrix.

some small note:
The new installation method using `flit install --symlink` might break on Windows, where `flit install --pth-file` might be a better option (see [docs: flit install --pth-file option](https://flit.readthedocs.io/en/latest/cmdline.html#cmdoption-flit-install-pth-file)).


Juanlu001:
Thanks a lot @kerel-fs and congratulations for your first pull request! :tada:

PR #810, opened 2019-11-29
Fixed frame bugfixes

noc0lour:
This PR contains mostly bugfixes w.r.t how the obstime is handled in the `transform_to` call and allowing vectorized coordinate inputs by using astropy routines. This also speeds up the conversion significately since the ERFA routines are quite fast.


noc0lour:
I've added a separate commit to this PR removing the `rotate` and `transform` function from util since the rotation functions in astropy are much faster & provide vectorized input support.
There are no users within poliastro left for these functions, but this is a change in the API which is not backwards compatible so you might want to cherry-pick this commit later for a major release.

Juanlu001:
Amazing that the astropy functions are faster! Good that we also gained vectorization in the transition :)

We don't have to worry about that breaking change. Could you please rebase on current `master` so we can merge straight away?

noc0lour:
Rebased & pushed.

Juanlu001:
Merging, thanks a lot!

PR #813, opened 2019-12-04
Fix typo in CZML user guide

kerel-fs:
Note: This typo is already fixed in the [examples/CZML Tutorial.ipynb](https://github.com/poliastro/poliastro/blob/master/docs/source/examples/CZML%20Tutorial.ipynb).

(OT: The tutorial duplicates the user guide but is more complete (ideally it should be the other way around imo), although it's still missing some steps on how to conveniently write `extractor.packets` to json files). I'll try to create a proper MR when I figured out how to ues poliastro+czml in a convenient way (planning to use it for matching navarea warnings to last orbits / plotting).


Juanlu001:
Thanks @kerel-fs! Merging when the tests pass :muscle:

Issue #780, opened 2019-10-04
Documentation error: units for rho0 in atmospheric_drag perturbation


mc-allen:
🐞 In the atmospheric drag function, the rho0 units are indicated as kg/m^3 (https://github.com/poliastro/poliastro/blob/v0.13.0/src/poliastro/core/perturbations.py#L112).

However, I think that the function actually wants kg/km^3, which is consistent with the example in https://docs.poliastro.space/en/stable/examples/Natural%20and%20artificial%20perturbations.html#Atmospheric-drag

Juanlu001:
Thanks @mc-allen, good catch! @jorgepiloto looks like @mc-allen is correct, can you please confirm? And then we can accept pull requests from new contributors for this issue.

Juanlu001:
Closed in #816

PR #816, opened 2019-12-08
Minimal error in docstring

jorgepiloto:
Solves #780 

The function requires that user introduced the air's density in [kg/km3]. However the docstring claimed to be input in [kg/m3]. This pull request solves this minimal issue :relaxed:


Juanlu001:
Thanks @jorgepiloto ! A small step towards correctness :)

PR #825, opened 2019-12-21
Add mean elements to bodies, remove from_body_ephem

### LINKED ISSUE #790: Reconsider having reference elements for planets
💡 **Feature request**

To have representative/reference/mean orbital parameters for major bodies.

🎯 **Goal**

The issue of having mean elements for planets instead of relying on the ephemerides has appeared many times in the past.

One of such times was [this pull request by @AunSiro](https://github.com/poliastro/poliastro/pull/153). Even though it made the computation of the radius of the Sphere of Influence trivial, I was not convinced:

> I'm not sure if the semimajor axis and the eccentricity belong here. While properties like the mass and the planet mean radius are clearly property of the planets themselves, the parameters of the orbit are a consequence of that body being in orbit around a certain attractor. For instance, you can retrieve the osculating orbit using ephem.planet_ephem, as described here http://poliastro.readthedocs.io/en/latest/user_guide.html#where-are-the-planets-computing-ephemerides. One could argue that having the mean properties of the orbit attached to the bodies might be practical, but still I'm not convinced from a code point of view.

This appeared again [some months after that](https://github.com/poliastro/poliastro/issues/150#issuecomment-326518564) and I said:

> I still maintain that we should not add orbital parameters to bodies.

As a result, this is the approach we are following for the SOI radius:

https://github.com/poliastro/poliastro/blob/68dc1de3170dc7e28ecbc2c11d65182427a9c7ff/src/poliastro/threebody/soi.py#L56-L64

One could argue that, after all, this is equivalent to hardcoding some parameters! But instead of having a representative value, we are using the instantaneous one at J2000.

And finally, in https://github.com/poliastro/poliastro/pull/774 we came back to the same problem: we need to know the mean motion of the planets around the Sun to model the Sun-synchronous condition.

JPL provides [lower accuracy formulas for planetary positions](https://ssd.jpl.nasa.gov/?planet_pos) and a nice justification:

> Lower accuracy formulae for planetary positions have a number of important applications when one doesn't need the full accuracy of an integrated ephemeris. They are often used in observation scheduling, telescope pointing, and prediction of certain phenomena as well as in the planning and design of spacecraft missions.

Therefore, I think the time has come to include them in poliastro.

astrojuanlu:
Fix #790, address #445.

Notice that the title of #445 is "Offer an alternative to Orbit.propagate for the Moon and planets", but rather than offering an alternative, here we're saying "use something else". A user could still do `Earth.get_mean_orbit().propagate`, but it's more evident than before that the result of the propagation is approximate.

Besides, I'm still not 100 % happy with two things:

* `poliastro.ephem.get_mean_orbit` ended up being somewhat complex. However, I could not think of a simpler solution: if we just have `body.a`, `body.ecc`, `body.inc`... it's still not clear _with respect to what_ those elements are measured, and we are back to having implicit reference frames, which I don't like. By having the elements and the reference frame we ended up having an `Orbit`. This opens the door for having more low-precision ephemerides, but I chose the Standish JPL one because it's simple (linear regressions) and should be representative over a wide period of time.
* `body.get_mean_orbit` still introduces a circular dependency because of planetary frames.

There are a couple of tests that need fixing.

### DIFF
--- a/src/poliastro/ephem.py
+++ b/src/poliastro/ephem.py
@@ -9,6 +9,128 @@ from astropy.coordinates import (
 from astropy.time import Time
 from scipy.interpolate import interp1d
 
+from .bodies import Earth, Jupiter, Mars, Mercury, Neptune, Pluto, Saturn, Uranus, Venus
+from .constants import J2000
+from .frames import Planes
+from .twobody.angles import M_to_nu
+from .twobody.orbit import Orbit
+
+# Source: https://ssd.jpl.nasa.gov/?planet_pos
+MEAN_ELEMENTS_1800AD_2050AD = {
+    Mercury: {
+        "a": (0.38709927, 0.00000037),
+        "ecc": (0.20563593, 0.00001906),
+        "inc": (7.00497902, -0.00594749),
+        "meanlon": (252.25032350, 149472.67411175),
+        "lonper": (77.45779628, 0.16047689),
+        "node": (48.33076593, -0.12534081),
+    },
+    Venus: {
+        "a": (0.72333566, 3.9e-06),
+        "ecc": (0.00677672, -4.107e-05),
+        "inc": (3.39467605, -0.0007889),
+        "meanlon": (181.9790995, 58517.81538729),
+        "lonper": (131.60246718, 0.00268329),
+        "node": (76.67984255, -0.27769418),
+    },
+    Earth: {
+        "a": (1.00000261, 5.62e-06),
+        "ecc": (0.01671123, -4.392e-05),
+        "inc": (-1.531e-05, -0.01294668),
+        "meanlon": (100.46457166, 35999.37244981),
+        "lonper": (102.93768193, 0.32327364),
+        "node": (0.0, 0.0),
+    },
+    Mars: {
+        "a": (1.52371034, 1.847e-05),
+        "ecc": (0.0933941, 7.882e-05),
+        "inc": (1.84969142, -0.00813131),
+        "meanlon": (-4.55343205, 19140.30268499),
+        "lonper": (-23.94362959, 0.44441088),
+        "node": (49.55953891, -0.29257343),
+    },
+    Jupiter: {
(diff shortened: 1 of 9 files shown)


src/poliastro/ephem.py
```diff
@@ -9,6 +9,128 @@
...
+
+
+def _get_elements(body, epoch=J2000):
+    elements = MEAN_ELEMENTS_1800AD_2050AD[body]
+    return {
+        name: _get_element(elem[0], elem[1], epoch) for name, elem in elements.items()
+    }
+
+
+def get_mean_orbit(body, epoch=J2000):
```
Juanlu001: Perhaps we should move this to `Orbit.from_body_mean_elements` and make `get_elements` public?

Juanlu001:
In the end, what we gained here is:

* A convenience method `Body.get_mean_orbit`. This was already possible before, placing an `import` inside the method the same way I did in this PR. Also, @shreyasbapat argues that it's awkward to have several places from which to create orbits, and he has a point.
* More representative mean orbital elements for planets by using a very low precision ephemerides. This was already possible before by using the `builtin` ephemerides in Astropy, based in Simon et al. (1994) [and implemented in ERFA](https://github.com/liberfa/erfa/blob/v1.7.0/src/plan94.c). 

Therefore, after dedicating the whole afternoon to this, I'm not sure it's worth the breakage.

Juanlu001:
[From our chat](https://matrix.to/#/!sENfoZnXsOePhaLRJE:matrix.org/$dp_BiQIMpHDoT_2j1dbXKS-GAqXzY6F5qX797FFvS3s?via=matrix.org&via=openastronomy.org):

> Perhaps it's a good opportunity to separate the orbit creation in two conceptual fields: orbits of spacecraft and orbits of natural bodies. For the former we want propagation, for the latter we need ephemerides

astrojuanlu:
Closing in favor of #864.

Issue #501, opened 2018-11-29
Use sphinx-gallery for the examples


astrojuanlu:
sphinx-gallery has some interesting features:

* It turns scripts into notebooks, reducing the burden in version control and helping maintainability
* It arranges thumbnails nicely
* It has some Binder support

We should look into migrating the current examples to using that.

shreyasbapat:
Does sphinx gallery does not support plotly?
It's not clear in their documentation. 

> Sphinx-Gallery will not manage its dependencies when installing, thus you are required to install them manually. Our minimal dependencies are:
Sphinx
Matplotlib
Pillow

> Sphinx-Gallery has also support for packages like:
Seaborn
Mayavi

Juanlu001:
> Does sphinx-gallery support plotly?

I don't know :)

https://github.com/sphinx-gallery/sphinx-gallery/issues?utf8=%E2%9C%93&q=plotly

If you are interested, maybe you can try.

> If it doesn't, is it like we move some of our examples to sphinx gallery and we don't move the others?

I prefer not to, to be honest. Either we move all of them, we add plotly support to sphinx-gallery, or we stay like this.

Juanlu001:
On behalf of @shreyasbapat: https://github.com/sphinx-gallery/sphinx-gallery/issues/441

Juanlu001:
Another option to consider is jupytext https://towardsdatascience.com/jupytext-1-0-highlights-a49dca9baa7b

cc @jorgepiloto

Juanlu001:
Support for Plotly was added to sphinx-gallery, so this is again open for consideration.

astrojuanlu:
Now nbsphinx will support thumbnails :) https://github.com/spatialaudio/nbsphinx/pull/392

jorgepiloto:
We might consider closing this after #973 was merged :+1:

astrojuanlu:
Yay! Closing!

PR #828, opened 2019-12-26
Sphinx-gallery and documentation update

jorgepiloto:
This pull-request tries to solve for the following issues:

* Sphinx-gallery for documentation examples: #501

The procedure was the following: all the Jupiter Notebooks were converted to Python files with proper comments layout readed by Sphinx-Gallery. The different configuration files were also modified to work with this new doc extension and sphinx-gallery was added as a one of the developers dependencies.


Juanlu001:
Wow @jorgepiloto , thanks for this massive pull request! This will be a huge improvement in many fronts :clap: 

Apparently the only failing tests are related to DASTCOM5:

```
==================================== ERRORS ====================================
____ ERROR collecting docs/source/auto_examples/plot_using_neos_package.py _____
docs/source/auto_examples/plot_using_neos_package.py:135: in <module>
    atira = dastcom5.orbit_from_name("atira")[0]  # NEO
.tox/images/lib/python3.6/site-packages/poliastro/neos/dastcom5.py:338: in orbit_from_name
    records = record_from_name(name)
.tox/images/lib/python3.6/site-packages/poliastro/neos/dastcom5.py:393: in record_from_name
    lines = string_record_from_name(name)
.tox/images/lib/python3.6/site-packages/poliastro/neos/dastcom5.py:418: in string_record_from_name
    with open(idx_path, "r") as inF:
E   FileNotFoundError: [Errno 2] No such file or directory: '/root/.poliastro/dastcom5/dat/dastcom.idx'
------------------------------- Captured stdout --------------------------------
There exist several objects with that name
Eros epoch:  2020-05-31 00:01:09.185
Eros November epoch:  2017-09-03 12:00:00.000
_______ ERROR collecting docs/source/examples/plot_using_neos_package.py _______
docs/source/examples/plot_using_neos_package.py:135: in <module>
    atira = dastcom5.orbit_from_name("atira")[0]  # NEO
.tox/images/lib/python3.6/site-packages/poliastro/neos/dastcom5.py:338: in orbit_from_name
    records = record_from_name(name)
.tox/images/lib/python3.6/site-packages/poliastro/neos/dastcom5.py:393: in record_from_name
    lines = string_record_from_name(name)
.tox/images/lib/python3.6/site-packages/poliastro/neos/dastcom5.py:418: in string_record_from_name
    with open(idx_path, "r") as inF:
E   FileNotFoundError: [Errno 2] No such file or directory: '/root/.poliastro/dastcom5/dat/dastcom.idx'
```

Do you have an idea why this might be the case?

Juanlu001:
After giving this a quick look, I have a lot of questions coming from my ignorance of how sphinx-gallery works. They boil down to two:

* What are the `.zip`, autogenerated `.html`, and `.pickle` files?
* I see that every example was turned into a `.ipynb` + `.py` + `.rst` (+ perhaps something else). Do we need to keep all these files in version control?

@jorgepiloto you clearly have spent a lot of time with sphinx-gallery, so a few interesting links to understand the former would help me doing the review.

Regarding the latter, I guess there is some tool to keep all these versions in sync. If there is, then I think we should not have all of them in `git`. There are some pretty big files in this PR, even though the net contribution is a decrease. If there is not such an automated way, then having to sync all these versions is probably going to be a nightmare.

Finally, I wonder if we could make a separate PR for the changes that have nothing to do with sphinx-gallery.

Thanks again for putting so much work into this :heart:

jorgepiloto:
## How sphinx-gallery works

This tool generates a set of `*rst` files from original Python ones with a specific layout. There exist online tools such us [this one](https://gist.github.com/chsasank/7218ca16f8d022e02a9c0deb94a310fe) to convert from Jupyter Notebook to Python-Sphinx-Gallery format. In addition, `*.py` and `*.ipynb` are also generated so user can download each notebook in both formats.

Regarding other "strange" output files and refering to [official docs](https://sphinx-gallery.github.io/stable/getting_started.html#configure-and-use-sphinx-gallery):

* `.md5`: a md5 hash of the .py file, used to determine if changes have been made to the file and thus if new output files need to be generated.
* `.pickle`: used to identify function names and to which module they belong (more details in Identifying function names in a script)

**Notice that only the python files under the `examples/` directory are required in order to build up sphinx general documentation, so we can get rid of all the other inside our repository: they will be automatically generated once user runs `make html`**

I do not remember if **Read the docs** automatically builds from config.py, but in principle there should not be any problem if we get rid of those non-python files and just keep the Python ones under the examples' folder.

jorgepiloto:
I will just focus on solving issue #501 and let the others for future pull-requests. Furthermore, let me better organize commit history :rocket:

Juanlu001:
Thanks for the explanations @jorgepiloto! Happy to split this PR into smaller ones.

jorgepiloto:
Closing this is favor of #973

Issue #791, opened 2019-10-31
Different namespace for Earth-specific functionality


astrojuanlu:
💡 **Feature request**

Add a new module `poliastro.earth` or `poliastro.twobody.earth` for Earth-specific functionality, possibly including an `EarthOrbit` object.

🎯 **Goal**

We traditionally wanted poliastro to be as generic as possible and don't include any Earth-specific functionality. As stated in the README:

https://github.com/poliastro/poliastro/blob/5d710592f33a41065eb22f26aa2ede0b77e1fa0a/README.rst#L222-L223

However, we already added at least two: [export to CZML/Cesium](https://github.com/poliastro/poliastro/pulls?q=is%3Apr+author%3Asedictious+is%3Aclosed) (mostly for the Earth until we figure out a way to represent the other planets), and [Earth atmospheric models](https://github.com/poliastro/poliastro/pull/738). In fact, we proposed and accepted a Google Summer of Code project titled ["Improving poliastro Earth capabilities"](https://summerofcode.withgoogle.com/projects/#6467241724346368)!

Advantages of making poliastro more useful for Earthlings :alien::

* Some algorithms are better documented for the Earth, making them easier to implement (and later some of them could be generalized)
* Most commercial applications are Earth-bound, so adding some functionality could make poliastro more attractive for these users (for example, being able to propagate TLEs with SGP4!)

Even if we decided not to add more functionality, we should probably find a better place for the code we already merged, and we will expand it anyway.

noc0lour:
Given poliastro is in a unique position to give non-aerospace people an easy entry into calculating propagation in Python this would come in handy for a lot of applications. Addition of Earth-specifics shouldn't come at cost of losing functionality for other planets though.

shreyasbapat:
What functionalities will be in Earth Namespace? 
A rough list? So that I don't miss them.

astrojuanlu:
Let's tackle this after https://github.com/poliastro/poliastro/issues/445 is solved, see https://github.com/poliastro/poliastro/projects/3.

astrojuanlu:
Quick proposal (before I start writing a page in the wiki, which might have to wait a few months still):

```python
from poliastro.earth import EarthOrbit  # Separate child class of Orbit with extra methods
from poliastro.earth.atmosphere import COESA76
from poliastro.earth.gravity import EGM84  # Does not exist... yet

EarthOrbit.heliosynchronous(...)  # Already present in Orbit
[...]

PR #833, opened 2020-01-14
Create a new namespace for earth related properties

shreyasbapat:
Addresses #791 

This is still WIP.

But I would like to initiate a discussion and a little bit clarity on how we want this to happen.

@Juanlu001


astrojuanlu:
Hi @shreyasbapat, sorry I didn't reply to this! Before starting with the change we need to discuss it, if anything because the massive `git mv` will be a bit disruptive and very difficult to rebase, therefore it has to be done in one strike. I would also love to have @jorgepiloto input but he has exams soon. Let's get back to #791 in a few weeks.

shreyasbapat:
True. As you wish. I am on a quest to close pending PRs for some days now. Whenever you feel, we can have a discussion.

astrojuanlu:
Thanks again for this PR @shreyasbapat ! However it was a bit "ahead of its time" 😅 I'm not even 100 % of the current way we do things, but in any case, this work has been superseded. Closing 🙏🏽

Issue #832, opened 2020-01-10
Ship our setup.cfg


Juanlu001:
🐞 **Problem**

Our pytest configuration is not being picked up by users that want to run the test suite because we are not shipping our `setup.cfg`.

Also, other packages mistakenly shipping `setup.cfg` to the wrong location can interfere in poliastro tests, see #750 and #784.

💡 **Possible solutions**

Ship our own `setup.cfg`. Using setuptools, this was possible doing

```
package_data={
    "poliastro": "setup.cfg"
}
```

(see [this statsmodels discussion](https://github.com/statsmodels/statsmodels/issues/6114))

Not sure if flit supports this.

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

priyanshurohilla:
I want to work on this.

susanka068:
I'd like to work on this issue

Juanlu001:
@susanka068, @priyanshurohilla is already working on it at #836

noc0lour:
This can be closed I assume.

Juanlu001:
Thanks for the reminder @noc0lour. I wanted to test this before closing... and I saw we're not there yet.

We added `setup.cfg` under `sdist`, but it's not being included in the wheel. I'll see what can be done.

Juanlu001:
Well, I don't think it can be easily done for a file that is out of the `src` tree. This is probably a symptom that we shouldn't be shipping our tests, even though other packages in the scientific Python ecosystem do it.

Out of curiosity, I checked SciPy wheels and, even though they have `include tox.ini pytest.ini` in their `MANIFEST.in`, their pytest configuration doesn't make it to the wheels.

It's OK that we managed to include `setup.cfg` in our source distribution, but perhaps we should think about removing the tests from the wheels. I won't even bother about opening an issue for it.

In the meanwhile, I close this.

astrojuanlu:
> I won't even bother about opening an issue for it.

Well, I did: #863

PR #836, opened 2020-01-31
setup.cfg

priyanshurohilla:
Addresses #832 
i have tried to resolve the issue please review it

### DIFF
--- a/src/poliastro/constants/general.py
+++ b/src/poliastro/constants/general.py
@@ -489,7 +489,7 @@ R_polar_mars = Constant(
 R_polar_jupiter = Constant(
     "R_polar_jupiter",
     "Jupiter polar radius",
-    6.6854,
+    6.6854e7,
     "m",
     10000,
     "IAU Working Group on Cartographic Coordinates and Rotational Elements: 2009",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,6 +76,8 @@ dev = [
 Source = "https://github.com/poliastro/poliastro"
 Tracker = "https://github.com/poliastro/poliastro/issues"
 
+
+
 [tool.flit.entrypoints."console_scripts"]
 poliastro = "poliastro.cli:main"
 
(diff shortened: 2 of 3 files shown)


Juanlu001:
Hi @priyanshurohilla, thanks for your pull request! It looks to me though that you're using setuptools config (in setup.cfg) which is not used in poliastro, and therefore these changes would have no effect. Please take a look at how `flit` does this.

priyanshurohilla:
@Juanlu001 i looked up how flit does that and i found that the flit was made so that u don't have to ship any package. see this
https://github.com/takluyver/flit/issues/115#issuecomment-310930758

 It automatically detects and ships all the packages. So maybe what am i thinking is that we need to change the location of the `setup.cfg` file which is currently in poliastro folder to the folder where `flit` is located. Therefore, it will import 'setup.cfg' . This we have to do as there is currently no operation to import specific file from another location in `flit`.

see this
https://github.com/takluyver/flit/issues/115#issuecomment-310931356

jorgepiloto:
I think [this](https://flit.readthedocs.io/en/latest/pyproject_toml.html#sdist-section) the part in which we should include something like:

```bash
[tool.flit.sdist]
include = [
    ...
    "src/poliastro/atmosphere/data/coesa76_rho.dat",
    "setup.cfg"
]
```

Not sure though :thinking:

priyanshurohilla:
@jorgepiloto Actually the `sdist` command itself produces a minimal default set into the source distribution:

- all Python source files ` py_modules` and `packages`
- all C source files in the `ext_modules` or `libraries` 
- all test scripts: `test/test*.py`
- README.txt, setup.py and setup.cfg 

so basically it automatically imports that also, but in this case the file is present in another folder so i think explicitly including it will solve the issue.

Juanlu001:
I think this should suffice! @priyanshurohilla can you please rebase or squash these commits into only one and `git push --force` so we can simplify the history?

src/poliastro/constants/general.py
```diff
@@ -489,7 +489,7 @@
 R_polar_jupiter = Constant(
     "R_polar_jupiter",
     "Jupiter polar radius",
-    6.6854,
+    6.6854e7,
```
Juanlu001: This is because the branch is not up to date with current `master`. But let's not be picky with that now.

PR #842, opened 2020-02-14
twobody: correct raan_from_ltan calculation by 180 degrees

### LINKED ISSUE #841: RAAN from LTAN calculation off-by 180 degrees
The pull request which added RAAN from ltan function to `twobody.angles` module 
shifts the angle by 180 degrees and it ends up having LTAN and LTDN switched around. 

🐞 **Problem**

Verification with actual orbits obtained from TLEs shows this discrepancy

Example Sentinel-1: 
```
In [1]: import astropy.time

In [2]: import astropy.units as u

In [3]: from poliastro import twobody

In [4]: s1_epoch = astropy.time.Time(astropy.time.Time('2020-01-01 00:00').to_value('mjd') + 45.05318108 - 1, format='mjd')

In [5]: twobody.angles.raan_from_ltan(s1_epoch, 18.00 * u.hourangle)
Out[5]: <Angle 593.23886941 deg>

In [6]: twobody.angles.raan_from_ltan(s1_epoch, 18.00 * u.hourangle).wrap_at(360 * u.deg)
Out[6]: <Angle 233.23886941 deg>

In [7]: twobody.angles.raan_from_ltan(s1_epoch, 18.00 * u.hourangle).wrap_at(360 * u.deg) - 180 * u.deg
Out[7]: <Angle 53.23886941 deg>
```

Value from Celestrak TLE for this: 53.8031

💡 **Possible solutions**

Removing `-12 * u.hourangle` from `alpha`
```
    alpha = (coordinates.Angle(ltan).wrap_at(24 * u.hourangle) - 12 * u.hourangle).to(
        u.rad
    )
```

I crosschecked with Sentinel-2, METOP-B and if I change this it checks out (+- 0.5 degree).
Calculation of the epoch coming from the TLE is a bit special since 2020-01-01 00:00 is expressed as 20001.0000 and thus I need to subtract 1.0 before calculating the appropriate `MJD`.

I've created this issue to discuss if this is actually wrong or if I'm just confusing something...

noc0lour:
Closes #841

Should I also add a unit test checking if the function behaves correctly for a
few orbits which I can extract from public TLEs?

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_angles.py
+++ b/src/poliastro/tests/tests_twobody/test_angles.py
@@ -1,3 +1,4 @@
+import astropy.time
 import numpy as np
 import pytest
 from astropy import units as u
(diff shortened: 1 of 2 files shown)


Juanlu001:
hi @noc0lour, thanks for this PR! this needs a `tox -e reformat`.

> Should I also add a unit test checking if the function behaves correctly for a
few orbits which I can extract from public TLEs?

Absolutely!

noc0lour:
Should be fine now, I've added a test with a MetOp-B & Sentinel-5P TLE for their respective LTAN and RAAN matches for up to 0.3 degrees, I think that's fine as the S-5P LTAN is not as precise as the published MetOp-B LTAN.

src/poliastro/tests/tests_twobody/test_angles.py
```diff
@@ -240,3 +241,34 @@ def test_convert_coe_and_rv_circular_equatorial(circular_equatorial):
     k, expected_res = circular_equatorial
     res = rv2coe(k, *coe2rv(k, *expected_res))
     assert_allclose(res, expected_res, atol=1e-8)
+
+
+def test_raan_from_ltan_metopb():
+    # MetOp-B: LTAN: 21:31:45
```
Juanlu001: Where does the LTAN come from?

src/poliastro/tests/tests_twobody/test_angles.py
```diff
@@ -240,3 +241,34 @@ def test_convert_coe_and_rv_circular_equatorial(circular_equatorial):
...
+        astropy.time.Time("2020-01-01 00:00").to_value("mjd") + 49.95408566 - 1,
+        format="mjd",
+    )
+    expected_raan = 110.9899 * u.deg
+    raan = angles.raan_from_ltan(epoch, ltan)
+    assert_allclose(raan.wrap_at(360 * u.deg).to(u.deg), expected_raan, atol=0.3)
+
+
+def test_raan_from_ltan_sentinel5p():
+    # SENTINEL-5P
```
Juanlu001: I guess we could have a similar comment like

```
# LTAN: 13:30:00
```

with the source included?

Juanlu001:
@noc0lour looking good! For this to be perfect let's add a comment explicitly stating where does the expected LTAN come from, and I'll merge :muscle:

src/poliastro/tests/tests_twobody/test_angles.py
noc0lour: Added comment stating LTAN & source

src/poliastro/tests/tests_twobody/test_angles.py
noc0lour: Added a comment stating source of LTAN

noc0lour:
Thanks for pointing out the missing sources, I meant to include them but just overlooked.

Juanlu001:
Thanks to you! Merged 💪

PR #844, opened 2020-02-18
CZML extractor: timezone issue #817

### LINKED ISSUE #817: CZML extractor: timezone issues (clock.Interval and currentTime not tz-aware)
🐞 **Problem**

The ISO8601 timestamps in the `clock.interval` and `currentTime` fields by the czml extractor are missing the 'Z' character indicating UTC. As a result, when the users locale is UTC+1 the CZMLWidget will start one hour too early (so that the satellites will only be visible after one hour passed in the simulation.

[Example notebook](https://gist.github.com/kerel-fs/6cdcfe33b933329f807382b0c5bbf1a9) [Binder](https://mybinder.org/v2/gist/kerel-fs/6cdcfe33b933329f807382b0c5bbf1a9/master)

🖥 **Envorionment**
* `pip freeze`
  - `poliastro==0.13.post0.dev0`

💡 **Possible solutions**

- option A: Extend `czml3.types.TimeInterval` to have also a `_value` like `IntervalValue` has. Then use `czml3.types.TimeInterval` instead of `czml3.types.IntervalValue` in [extract_czml.py#L174-L181](https://github.com/Juanlu001/poliastro/blob/1ddf2d5a8082b7e7d191b0c3ae8de2839b3f51a8/src/poliastro/czml/extract_czml.py#L174-L181):
  ```python
  pckt = Preamble(
      name="document_packet",
      clock=IntervalValue(
          start=self.start_epoch.value,
          end=self.end_epoch.value,
          value=Clock(currentTime=self.start_epoch.value, multiplier=60),
      ),
  )
  ```
- option B: Copy datetime formatting code from `czml3.types.TimeInterval` into the relevant section in `extract_czml.py`

A similar issue was fixed in #783.

sundeshgupta:
Fixes #817. 

I have also added a test to check that `clock.interval` and `clock.currentTime` fields by the czml extractor have the 'Z' character indicating UTC.

### DIFF
--- a/src/poliastro/tests/test_czml.py
+++ b/src/poliastro/tests/test_czml.py
@@ -70,8 +70,8 @@ def test_czml_add_orbit():
     "version": "1.0",
     "name": "document_packet",
     "clock": {
-        "interval": "2013-03-18T12:00:00.000/2013-03-18T23:59:35.108",
-        "currentTime": "2013-03-18T12:00:00.000",
+        "interval": "2013-03-18T12:00:00Z/2013-03-18T23:59:35Z",
+        "currentTime": "2013-03-18T12:00:00Z",
         "multiplier": 60,
         "range": "LOOP_STOP",
         "step": "SYSTEM_CLOCK_MULTIPLIER"
@@ -98,7 +98,7 @@ def test_czml_add_orbit():
     "id": 0,
     "availability": "2013-03-18T12:00:00Z/2013-03-18T23:59:35Z",
     "position": {
-        "epoch": "2013-03-18T12:00:00.000",
+        "epoch": "2013-03-18T12:00:00Z",
         "interpolationAlgorithm": "LAGRANGE",
         "interpolationDegree": 5,
         "referenceFrame": "INERTIAL",
@@ -198,7 +198,7 @@ def test_czml_add_orbit():
     "id": 1,
     "availability": "2013-03-18T12:00:00Z/2013-03-18T23:59:35Z",
     "position": {
-        "epoch": "2013-03-18T12:00:00.000",
+        "epoch": "2013-03-18T12:00:00Z",
         "interpolationAlgorithm": "LAGRANGE",
         "interpolationDegree": 5,
         "referenceFrame": "INERTIAL",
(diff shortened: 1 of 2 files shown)


Sedictious:
Looks good but I have to test once I get back home!

sundeshgupta:
Sure! No problem :)

sundeshgupta:
I have updated the existing tests. Kindly review this PR.

src/poliastro/tests/test_czml.py
```diff
@@ -70,8 +70,8 @@ def test_czml_add_orbit():
     "version": "1.0",
     "name": "document_packet",
     "clock": {
-        "interval": "2013-03-18T12:00:00.000/2013-03-18T23:59:35.108",
-        "currentTime": "2013-03-18T12:00:00.000",
+        "interval": "2013-03-18T12:00:00Z/2013-03-18T23:59:35Z",
+        "currentTime": "2013-03-18T12:00:00Z",
```
Sedictious: I think there isn't any reason not to convert the rest of the properties (like ``epoch``) to the ISO8601 format.

src/poliastro/tests/test_czml.py
```diff
@@ -645,3 +645,45 @@ def test_czml_invalid_orbit_epoch_error():
         "ValueError: The orbit's epoch cannot exceed the constructor's ending epoch"
         in excinfo.exconly()
     )
+
+
+@pytest.mark.skipif("czml3" not in sys.modules, reason="requires czml3")
+def test_czml_timezone():
```
Sedictious: Isn't this test a bit redundant since it's guaranteed that the timezone format will be tested in every other test anyway?

src/poliastro/tests/test_czml.py
sundeshgupta: I'll update the commit to convert the rest of the properties to ISO8601 format.

src/poliastro/tests/test_czml.py
sundeshgupta: Yes. I also think that the test is redundant. Should I delete this test?

sundeshgupta:
I have converted `epoch` to ISO8601 format. Please tell if should delete `test_czml_timezone()`.

src/poliastro/tests/test_czml.py
Sedictious: I think it would be better if you did since it doesn't test for any specific edge-case or other functionality

sundeshgupta: I have removed the test. Please tell if there is anything else to update.

Sedictious: Looks good! Could you perhaps squash and rebase?

sundeshgupta: I have tried to squash the commits and rebase to master. Please check if I did it correctly.

Juanlu001:
Thanks a lot @sedictious for reviewing this, and @sundeshgupta for the pull request! Merging :muscle:

Issue #840, opened 2020-02-07
vallado.lambert fails for long way transfers


paul-witsberger:
When trying to set up poliastro.iod.vallado.lambert, I receive the following error when I set short=False (long path transfer):
```
Traceback (most recent call last):
  File "C:/Users/pawit/.PyCharmCE2019.3/config/scratches/scratch_4.py", line 11, in <module>
    v = lambert_v(k, r0, rf, tof, short=short, numiter=numiter, rtol=rtol).__next__()
  File "C:\Users\pawit\Documents\research_env\lib\site-packages\poliastro\iod\vallado.py", line 48, in lambert
    v0, v = vallado_fast(k_, r0_, r_, tof_, short, numiter, rtol)
  File "C:\Users\pawit\Documents\research_env\lib\site-packages\poliastro\core\iod.py", line 151, in vallado
    raise RuntimeError("Maximum number of iterations reached")
RuntimeError: Maximum number of iterations reached
```

Here is my minimal (non)working example:
```
from astropy import units as u, constants as c
from poliastro.iod.vallado import lambert as lambert_v

k = c.GM_earth.to(u.km ** 3/ u.s ** 2)
r0 = [10000., 0, 0] * u.km
rf = [8000., -5000, 0] * u.km
tof = (2 * u.hour).to(u.s)
short = False
numiter = 1000
rtol = 1e-6
v = lambert_v(k, r0, rf, tof, short=short, numiter=numiter, rtol=rtol).__next__()

print('v1 = ')
print(v[0])
print('v2 = ')
print(v[1])
```

It has worked for all of the tests I gave it when short=True, and failed every time short=False.

NanuSai:
I'm a newbie, but may I take this issue?.

jorgepiloto:
Hi @paul-witsberger, I have been able to reproduce your issue.  Let me try to investigate a little more :smile: By the way, @NanuSai feel free to take this issue :+1:

NanuSai:
After much digging the internet and talks with @Juanlu001 , I've drawn the following conclusions from the problem.

1. I tried to compare izzo and vallado(short = True) and apparently they both yield different answers, 
```
from poliastro.iod.vallado import lambert as lambert_v
from poliastro.iod.izzo import lambert as lambert_i 
from astropy import units as u, constants as c


if __name__ == "__main__":
    k = (c.GM_earth.to(u.km ** 3/ u.s ** 2))
    r0 = ([10000., 0, 0] * u.km)
[...]

PR #845, opened 2020-02-24
Issue 840

NanuSai:
- It highlights after which iterations tof stops changing
- It returns the velocity vectors of the boundary value tof when tof_input exceeds the limit. So if the user give tof > tof_max, then v_tof_max would be returned, similarly for tof < tof_min

[Link to discussion](https://github.com/poliastro/poliastro/issues/840)

### DIFF
--- a/src/poliastro/tests/test_iod.py
+++ b/src/poliastro/tests/test_iod.py
@@ -1,5 +1,5 @@
 import pytest
-from astropy import units as u
+from astropy import constants as c, units as u
 from astropy.tests.helper import assert_quantity_allclose
 
 from poliastro.bodies import Earth
@@ -43,11 +43,12 @@ def test_curtis53(lambert):
     r0 = [273378.0, 0.0, 0.0] * u.km
     r = [145820.0, 12758.0, 0.0] * u.km
     tof = 13.5 * u.h
+    numiter = 100
 
     # ERRATA: j component is positive
     expected_va = [-2.4356, 0.26741, 0.0] * u.km / u.s
 
-    va, vb = next(lambert(k, r0, r, tof))
+    va, vb = next(lambert(k, r0, r, tof, numiter=numiter))
     assert_quantity_allclose(va, expected_va, rtol=1e-4)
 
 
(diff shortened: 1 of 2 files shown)


src/poliastro/tests/test_iod.py
```diff
@@ -1,13 +1,16 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
NanuSai: My bad :sweat_smile: . Yeah I've reformatted, works like charm

Juanlu001:
So the problem was... That the number of iterations was too short? :flushed: @jorgepiloto do you want to take a look? If I understood correctly @NanuSai PR, the solution to #840 is trivial.

NanuSai:
Currently the PR highlights the loose ends of the algorithm as shown by test case I added. The end goal, I believe, is to account for all kinds of tof , and if the tof is not suitable it should say so. 

And about the number of iterations, I tried 10**9 iterations, but it doesn't work. As I highlighted tof_new stops changing after itr 55, so the algorithm is not "proceeding" after that iteration.

Juanlu001:
> And about the number of iterations, I tried 10**9 iterations, but it doesn't work. As I highlighted tof_new stops changing after itr 55, so the algorithm is not "proceeding" after that iteration.

When I read the test, and I saw it passed, I wrongly understood that the issue was fixed. If it's not the case, then please rewrite the test in a way it demonstrates the buggy behavior (i.e. it fails).

I admit I am not devoting 100 % of my concentration to this issue yet, but I will do within a few days.

src/poliastro/tests/test_iod.py
```diff
@@ -17,12 +20,21 @@ def test_vallado75(lambert):
...
     assert_quantity_allclose(vb, expected_vb, rtol=1e-4)
 
+#Added by Riot: @JonPsy Github : @NanuSai
+@pytest.mark.parametrize('lambert',[vallado.lambert])
+def test_issue840(lambert):
+    k = (c.GM_earth.to(u.km ** 3/ u.s ** 2))
+    r0 = [10000., 0, 0] * u.km
+    rf = [8000., -5000, 0] * u.km
+    tof = 2 * u.hour
+    assert(next(lambert(k,r0,rf,tof,short=False)) != None)
```
NanuSai: Running this test with the master branch would through Runtime error "Maximum error reached" and hence would fail the test.  Running it on my PR would return a generator object and hence would pass the test 😊

src/poliastro/tests/test_iod.py
```diff
@@ -17,11 +17,21 @@ def test_vallado75(lambert):
...
+# Added by Riot: @JonPsy Github : @NanuSai
+@pytest.mark.parametrize("lambert", [vallado.lambert])
+def test_issue840(lambert):
+    k = c.GM_earth.to(u.km ** 3 / u.s ** 2)
+    r0 = [10000.0, 0, 0] * u.km
+    rf = [8000.0, -5000, 0] * u.km
+    tof = 2 * u.hour
+    assert next(lambert(k, r0, rf, tof, short=False)) is not None
+
+
```
NanuSai: Running this test in the main branch would throw **Runtime Error "Maximum iterations reached"** thereby failing the test  , running this on my PR would return velocity vectors(which may be trivial as you've said) and hence pass the test 😊.

I think this may lead to something big so it's okay to take necessary time to solve this issue 😁.

jorgepiloto:
Dear @NanuSai, I am so sorry for this late answer and promise to review in depth the code this night. 

It is kind of surprise that the number of iterations cause the problem because user is able to pass a maximum value for them, but let me check. Thank you for working on this issue! Hope to review it in a few hours!

Juanlu001:
Notice that the docstring of `poliastro.core.iod.vallado` is now failing!

jorgepiloto:
Looks good! As Juanlu said, you will have to fix the example on the docstring. Furthermore, when referring to the original algorithm I noticed it also supports multi-revolution solutions: it might be interesting to implement that feature, but for the moment let us solve and close this when ready :+1:

PR #864, opened 2020-03-15
Body mean elements, take 2

### LINKED ISSUE #790: Reconsider having reference elements for planets

astrojuanlu:
Supersedes #825. Fix #790.

Still needs a second pass to create an `Ephem` object to be used in plotting instead of `Orbit.from_body_ephem` as outlined in https://github.com/poliastro/poliastro/projects/3.


astrojuanlu:
This will also fix #790.

astrojuanlu:
Super weird error in docs... Rerunning.

astrojuanlu:
Rebased on top of #865.

Issue #861, opened 2020-03-14
from_sbdb  unit tests fails with default installation


iMurfyD:
🐞 **Problem**

🖥 **Please paste the output of following commands**

* `conda info -a` (only if you have conda)
```
(poliastropreinstalled) C:\Users\IMDes>conda info -a

     active environment : poliastropreinstalled
    active env location : C:\Users\IMDes\Anaconda3\envs\poliastropreinstalled
            shell level : 1
       user config file : C:\Users\IMDes\.condarc
 populated config files : C:\Users\IMDes\.condarc
          conda version : 4.8.2
    conda-build version : 3.18.9
         python version : 3.7.4.final.0
       virtual packages : __cuda=10.2
       base environment : C:\Users\IMDes\Anaconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/win-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/win-64
                          https://repo.anaconda.com/pkgs/r/noarch
                          https://repo.anaconda.com/pkgs/msys2/win-64
                          https://repo.anaconda.com/pkgs/msys2/noarch
          package cache : C:\Users\IMDes\Anaconda3\pkgs
                          C:\Users\IMDes\.conda\pkgs
                          C:\Users\IMDes\AppData\Local\conda\conda\pkgs
       envs directories : C:\Users\IMDes\Anaconda3\envs
                          C:\Users\IMDes\.conda\envs
                          C:\Users\IMDes\AppData\Local\conda\conda\envs
               platform : win-64
             user-agent : conda/4.8.2 requests/2.22.0 CPython/3.7.4 Windows/10 Windows/10.0.18362
          administrator : False
             netrc file : None
           offline mode : False

# conda environments:
#
base                     C:\Users\IMDes\Anaconda3
calc.py                  C:\Users\IMDes\Anaconda3\envs\calc.py
lowthrustsim             C:\Users\IMDes\Anaconda3\envs\lowthrustsim
plasmapydev              C:\Users\IMDes\Anaconda3\envs\plasmapydev
poliastropreinstalled  *  C:\Users\IMDes\Anaconda3\envs\poliastropreinstalled

sys.version: 3.7.4 (default, Aug  9 2019, 18:34:13) [...
sys.prefix: C:\Users\IMDes\Anaconda3
sys.executable: C:\Users\IMDes\Anaconda3\python.exe
[...]

PR #867, opened 2020-03-16
Changed unit test to compare databases

iMurfyD:
Mostly explained in issue #861 

Changes unit test to cross reference JPL databases instead of use hard-coded value, which will change over time.

### DIFF
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -983,8 +983,7 @@ def test_from_sbdb():
 
         # Two DB are not exactly the same because orbits perturb slowly and
         # measurment taken at different epochs in Horizons and SBDB
-        max_err = [0.0005 * y for y in ss_classical] # Maximum error of 0.05%
-                                                     # Chosen arbitrarily 
+        max_err = [0.0005 * y for y in ss_classical] # Maximum error of 0.05% (chosen arbitarily)
 
         assert diff_bw_orbits < max_err
 
--- a/src/poliastro/tests/tests_twobody/test_orbit.py
+++ b/src/poliastro/tests/tests_twobody/test_orbit.py
@@ -975,15 +975,21 @@ def test_from_sbdb():
         ss_target = Orbit.from_sbdb(target_name)
         ss_classical = ss_target.classical()
 
-        ss_ref = Orbit.from_horizons(name=target_name, attractor=Sun, plane=Planes.EARTH_ECLIPTIC)
-        ss_ref = ss_ref.propagate_to_anomaly(ss_classical[5])  # Catch reference orbit to same epoch
+        ss_ref = Orbit.from_horizons(
+            name=target_name, attractor=Sun, plane=Planes.EARTH_ECLIPTIC
+        )
+        ss_ref = ss_ref.propagate_to_anomaly(
+            ss_classical[5]
+        )  # Catch reference orbit to same epoch
         ss_ref_class = ss_ref.classical()
 
         diff_bw_orbits = [abs(x[1] - x[0]) for x in zip(ss_classical, ss_ref_class)]
 
         # Two DB are not exactly the same because orbits perturb slowly and
         # measurment taken at different epochs in Horizons and SBDB
-        max_err = [0.0005 * y for y in ss_classical]  # Maximum error of 0.05% (chosen arbitarily)
+        max_err = [
+            0.0005 * y for y in ss_classical
+        ]  # Maximum error of 0.05% (chosen arbitarily)
 
         assert diff_bw_orbits < max_err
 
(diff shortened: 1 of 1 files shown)


astrojuanlu:
Thanks for the pull request! Best way to solve quality issues is to do `tox -e reformat`

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -968,25 +968,30 @@ def test_from_coord_if_coord_is_not_of_shape_zero():
...
-        "Ceres": (
-            2.76916515450648 * u.AU,
-            0.07600902910070946 * u.one,
-            10.59406704424526 * u.deg,
-            80.30553156826473 * u.deg,
-            73.597694115971 * u.deg,
-            86.01851107780747 * u.deg,
-        )
-    }
+    targets = ["Ceres", "Vesta", "Eros"]  # Objects in both JPL SBDB and JPL Horizons
```
astrojuanlu: Let's replace this with `@pytest.mark.parametrize`, as done in other tests

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -968,25 +968,30 @@ def test_from_coord_if_coord_is_not_of_shape_zero():
 
 @pytest.mark.remote_data
 def test_from_sbdb():
```
astrojuanlu: Good moment to rename this test? For example

```suggestion
def test_from_sbdb_and_from_horizons_give_similar_results():
```

src/poliastro/tests/tests_twobody/test_orbit.py
```diff
@@ -968,25 +968,30 @@ def test_from_coord_if_coord_is_not_of_shape_zero():
...
-        assert ss_classical == SBDB_DATA[target_name]
+        ss_ref = Orbit.from_horizons(
+            name=target_name, attractor=Sun, plane=Planes.EARTH_ECLIPTIC
+        )
+        ss_ref = ss_ref.propagate_to_anomaly(
+            ss_classical[5]
+        )  # Catch reference orbit to same epoch
+        ss_ref_class = ss_ref.classical()
+
+        diff_bw_orbits = [abs(x[1] - x[0]) for x in zip(ss_classical, ss_ref_class)]
```
astrojuanlu: What about something like

```python
for element, element_ref in zip(ss_classical, ss_ref_class):
    assert_quantity_allclose(element, element_ref, rtol=...)
```

?

astrojuanlu:
We changed our CI entirely, this will need a rebase.

src/poliastro/tests/tests_twobody/test_orbit.py
iMurfyD: Changed name as suggested

src/poliastro/tests/tests_twobody/test_orbit.py
iMurfyD: Good idea - added parametrization. I left the remote decorator in as well. I think python will apply both decorators from bottom to top, so this should be ok.

src/poliastro/tests/tests_twobody/test_orbit.py
iMurfyD: Oh boy - I didn't even know we could do that. Good on the astropy developers for adding that feature in.

astrojuanlu:
@iMurfyD the tests are failing: https://dev.azure.com/poliastro/poliastro/_build/results?buildId=4&view=logs&j=538a0c56-c006-575f-81ef-6a7057413c77&t=b32c51f5-9aff-5da6-8e1a-744748933fbc&l=1393

astrojuanlu:
This requires another `tox -e reformat`

astrojuanlu:
Sorry, wrong button.

astrojuanlu:
> This requires another tox -e reformat

No it doesn't. Why are the checks failing then? Investigating.

astrojuanlu:
Oh, yes it does, I ran the checks against the wrong branch locally.

astrojuanlu:
Don't take offence @iMurfyD, I rebased `git push --force` your PR! I did this to accelerate things a bit, we're in an unstable phase with the Continuous Integration now so I want to make it work ASAP. Your authorship is retained, as you can see :pray:

astrojuanlu:
Passing all tests on Linux, merging :tada: Thanks again @iMurfyD!

iMurfyD:
Great- thank you for speeding this through! The CI pipeline is a bit of a learning curve for me. In the future, are there any local things I can do to test the CI pipeline? My process was run a test script which ran just the method in question and then rerun all the unit tests (although that took a while on my machine), but sometimes that didn't hit everything either.

astrojuanlu:
Hi @iMurfyD, sorry for the delay! Yeah, we're all learning about CI stuff... all the time :sweat_smile: 

> In the future, are there any local things I can do to test the CI pipeline?

Yes: the best thing to do is to run `tox -e py38` (or `tox -e py37` ir you're in Python 3.7). That will run all the tests in exactly the same way the CI does.

If you want to run only the "fast" tests (not all the perturbation ones) you can do `tox -e py38 -- -m "not slow"` (where `-m "not slow"` will be passed to `pytest`).

PR #878, opened 2020-03-23
Stop shipping tests

### LINKED ISSUE #863: Stop shipping tests in wheels
Well, I don't think it can be easily done for a file that is out of the `src` tree. This is probably a symptom that we shouldn't be shipping our tests, even though other packages in the scientific Python ecosystem do it.

Out of curiosity, I checked SciPy wheels and, even though they have `include tox.ini pytest.ini` in their `MANIFEST.in`, their pytest configuration doesn't make it to the wheels.

It's OK that we managed to include `setup.cfg` in our source distribution, but perhaps we should think about removing the tests from the wheels.

_Originally posted by @astrojuanlu in https://github.com/poliastro/poliastro/issues/832#issuecomment-590885043_

astrojuanlu:
A sign of maturity, I guess!

Fix #863.


astrojuanlu:
_So long, and thanks for all the fish_ :wave:

PR #882, opened 2020-03-27
Ephem objects

### LINKED ISSUE #445: Offer an alternative to Orbit.propagate for the Moon and planets

astrojuanlu:
See https://github.com/poliastro/poliastro/wiki/Ephem-objects for details. Closes #445.

To do:

* [x] Addition of more `@classmethod`s coming from `Orbit`
* [x] Tests
* [x] Plotting
* [x] Internal usage
* [x] Documentation
* [x] Add deprecations


astrojuanlu:
Ideas for plotting:

* Have a `plotter.plot_body_orbit(SolarSystemBody)` which takes care of everything and hides the gory details
* Have a `plotter2d.set_frame_from_orbit(Orbit)`, and deprecate both `plotter2d.set_frame` and `Orbit.pqw`, adding a `Orbit._pqw` while we figure out a better way

astrojuanlu:
Plotting ideas above implemented in #891, I will rebase this one afterwards.

astrojuanlu:
Notes to self:

* [ ] Several WIP comments, write tests or complete
* [ ] After the refactor, `plot_body_orbit` makes attractor (Sun) appear too small

astrojuanlu:
@jorgepiloto you're of course invited to review the whole pull request, but if you are on a tight schedule, could you take a look at the notebooks at least? They're greatly simplified in some places :)

There are two I couldn't update:

* The multi revolution Lambert notebook because of https://github.com/poliastro/poliastro/issues/901
* The NEOS notebook because of https://github.com/poliastro/poliastro/issues/902

@meuge is already having a look at the former, so I'll leave that for a future PR.

This PR is ready for review, just lacking some tests that I want to add before merging.

astrojuanlu:
Documentation available at https://poliastro--882.org.readthedocs.build/en/882/api/safe/ephem.html

astrojuanlu:
After a couple of days, I am merging this as declared. Welcome to the future! :rocket:

Issue #887, opened 2020-03-31
Investigate sudden drop in coverage


astrojuanlu:
Before #846 we were at ~89 % coverage, and after #884 we are at ~71 % coverage. How did this happen? What are we measuring differently?

https://codecov.io/github/poliastro/poliastro?branch=master

Perhaps the tests were included in the source coverage? If 71 % is the real number, then we have *a lot* of work to do.

astrojuanlu:
* Before: https://codecov.io/gh/poliastro/poliastro/tree/1563d4671f529519150dcc048066511883294b3e/src/poliastro
* After: https://codecov.io/gh/poliastro/poliastro/tree/e7be2537e0a00f9cd7327bddc228e729d82005a7/src/poliastro

astrojuanlu:
I suspect we're running coverage checks without disabling Numba.

astrojuanlu:
No we aren't! Or at least, we're trying.

https://github.com/poliastro/poliastro/blob/93042f5a3c2b43ab5c55ba7ba3eacb1e6dbb0799/azure-pipelines.yml#L74

PR #872, opened 2020-03-19
Czml tests improvements

tomaszmrugalski:
Here's my first PR, hoping to address issue #862 

Here the list of changes:

- I think I fixed the original problem with the code assuming local timezone rather than UTC. I don't have much experience with python yet and almost zero understanding of poliastro code, so if the fix is not correct, please point me in the right direction and I'll be happy to update this fix;

- improved czml_add_to_orbit check. It now really checks the exported CZML (there was assert missing in line 306/321).

- added new tests for preamble-only CZML packet. It was useful for me for debugging, but I thought it may stay. Hey, there's not such thing as too many tests;

- reordered image and show fields in billboard in expected CZML values, so the alphabetic order is accepted;

- added comments to existing tests;

- addressed almost all problems reported by pylint;

Right now all tests, except test_czml_add_orbit, which differs very slightly from the expected values (I think by millimeters). More info about this in the first comment. There's also a minor problem reported by pylint (note in second comment).

The code has been run through pylint, tox -e reformat, tox -e check.

I've tested this code on Ubuntu 19.10 and MacOS 10.15.3, both with python 3.7.5 with all dependencies installed using pip in virtualenv.

PR #888, opened 2020-03-31
Adjust tox

astrojuanlu:
Adjustments done while investigating #887. When all tests pass I will try to fix that issue as well.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,53 +26,52 @@ jobs:
       - linux: check
         name : quality
         env:
-          - FLIT_ROOT_INSTALL: 1 
+          - FLIT_ROOT_INSTALL: 1
           - MPLBACKEND: Agg
       
       - linux: py36
         name: py36_test
         env:
-          - FLIT_ROOT_INSTALL: 1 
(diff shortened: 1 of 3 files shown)


astrojuanlu:
This is _very_ interesting, the test failure we fixed in #872 is back to having the "other" results. Perhaps it has to do with how dependencies were being installed? I also saw it failing in my system.

@tomaszmrugalski I will try to revert 4e831695d6614dba0782285cb0b67c4f5b5c0f53 and 4ea9f3163841ef8520d0b007fd22ee7dbd236318.

astrojuanlu:
I get it now @tomaszmrugalski: the test results change slightly with numba.

After I reverted your commits, `tox -e py38 -- -k "czml_add_orbit"` fails but `NUMBA_DISABLE_JIT=1 tox -e py38 -- -k "czml_add_orbit"` passes. I will keep what we had in your PR and disable the test if numba is not running.

astrojuanlu:
The online tests failed because we've been hammering HORIZONS all day :sweat_smile: Code coverage went to normal, and now we have a `tox.ini` we can understand. Merging!

PR #890, opened 2020-03-31
Move installation of LaTeX dependencies to CI config

astrojuanlu:


jains8844:
@astrojuanlu I tried doing the same before adding this to the ```tox.ini``` file. Perhaps we have to specify the libraries like this : https://github.com/sunpy/sunpy/blob/master/azure-pipelines.yml#L47 and then mention them in the particular test like this : https://github.com/sunpy/sunpy/blob/master/azure-pipelines.yml#L68.

astrojuanlu:
@jains8844 that's similar to what [they told me in Astropy ci-issues](https://matrix.to/#/!vImbmhgZYeXojXObXc:openastronomy.org/$1585670232416uftzH:openastronomy.org?via=cadair.com&via=openastronomy.org&via=matrix.org), let's give it a try! I'll push the changes now

astrojuanlu:
Done. I will merge when the tests pass.

Issue #858, opened 2020-03-07
Add multiple revolution case to vallado solver for Lambert


NanuSai:
🐞 **Problem**

After a brief chat with @jorgepiloto. I think we need to update our core vallado solver for Lamberts problem to account for multiple revolution case.  The izzo algorithm already solves for it, fixing this issue would also solve issue #726. As he suggested the given [site](https://celestrak.com/software/vallado-sw.php) has the updated  code implemented in well known languages like MATLAB,C++,C#..

📋  **Steps to solve the problem**

- Check the site and understand the code 
- Implement the code in python 
- Try to fit it inside our codebase

I'd personally like to solve this issue because I've already worked on #840.

astrojuanlu:
Thanks @NanuSai!

To add some more information: when I implemented the Vallado algorithm, I had the 3rd edition of the book in my hands and couldn't find the 4th online. If the algorithm has been updated, I think it would be great to have it in poliastro :)

PR #895, opened 2020-04-04
[WIP] Multi rev vallado

jonpsy:
Updating to add multi revolution vallado case. Using [this](https://celestrak.com/software/vallado-sw.php) see matlab code, lambertu for reference. In response to issue #858 

Help's appreciated !

### DIFF
--- a/src/poliastro/core/iod.py
+++ b/src/poliastro/core/iod.py
@@ -1,3 +1,5 @@
+from math import gamma
+
 import numpy as np
 from numpy import cross, pi
 from numpy.linalg import norm
@@ -9,7 +11,7 @@ from ._jit import jit
 
 
 @jit
-def vallado(k, r0, r, tof, short, numiter, rtol):
+def vallado(k, r0, r, tof, num_rev, short, numiter, rtol):
     r"""Solves the Lambert's problem.
 
     The algorithm returns the initial velocity vector and the final one, these are
@@ -61,10 +63,13 @@ def vallado(k, r0, r, tof, short, numiter, rtol):
         Final position vector
     tof: ~float
         Time of flight
+    num_rev: int
+        Number of revolutions
     numiter: int
         Number of iterations to
-    rtol: int
-        Number of revolutions
+    rtol: float
+        Error tolerance
+
 
     Returns
     -------
@@ -114,39 +119,73 @@ def vallado(k, r0, r, tof, short, numiter, rtol):
     if A == 0.0:
         raise RuntimeError("Cannot compute orbit, phase angle is 180 degrees")
 
-    psi = 0.0
-    psi_low = -4 * np.pi ** 2
-    psi_up = 4 * np.pi ** 2
+    psi = 0
+
+    if num_rev == 0:  # Usual num_rev == 0 case
+        psi_low = -4.0 * np.pi ** 2
+        psi_up = 4.0 * np.pi ** 2
+
+    else:
+        psi_low = 4.0 * num_rev ** 2 * np.pi ** 2
+        psi_up = 4.0 * (num_rev + 1.0) ** 2 * np.pi ** 2
+
+    psi = _initial_guess_vallado(psi, psi_low, psi_up, num_rev, tof, short)
 
     count = 0
+    tof_new = -10
 
     while count < numiter:
-        y = norm_r0_plus_norm_r + A * (psi * c3(psi) - 1) / c2(psi) ** 0.5
-        if A > 0.0:
+
+        if np.abs(c2(psi) > rtol):
+            y = norm_r0_plus_norm_r + A * (1.0 - psi * c3(psi)) / c2(psi) ** 0.5
+        else:
+            y = norm_r0_plus_norm_r
+
+        if short:
             # Readjust xi_low until y > 0.0
             # Translated directly from Vallado
-            while y < 0.0:
-                psi_low = psi
+            neg_y_max_iter = 1
+
+            while y < 0.0 and neg_y_max_iter < 10:
                 psi = (
                     0.8
                     * (1.0 / c3(psi))
                     * (1.0 - norm_r0_times_norm_r * np.sqrt(c2(psi)) / A)
                 )
-                y = norm_r0_plus_norm_r + A * (psi * c3(psi) - 1) / c2(psi) ** 0.5
+                psi_low = psi
+                if np.abs(c2(psi)) > rtol:
+                    y = norm_r0_plus_norm_r + A * (
+                        (1.0 - psi * c3(psi)) / c2(psi) ** 0.5
+                    )
+                else:
+                    y = norm_r0_plus_norm_r
+                neg_y_max_iter = neg_y_max_iter + 1
 
-        xi = np.sqrt(y / c2(psi))
-        tof_new = (xi ** 3 * c3(psi) + A * np.sqrt(y)) / np.sqrt(k)
+            break  # Break out of count < numiter loop
 
-        # Convergence check
-        if np.abs((tof_new - tof) / tof) < rtol:
-            break
-        count += 1
-        # Bisection check
-        condition = tof_new <= tof
-        psi_low = psi_low + (psi - psi_low) * condition
-        psi_up = psi_up + (psi - psi_up) * (not condition)
+        else:
+            if np.abs(c2(psi)) > rtol:
+                xi = np.sqrt(y / c2(psi))
+            else:
+                xi = 0.0
 
-        psi = (psi_up + psi_low) / 2
+            tof_new = (xi ** 3 * c3(psi) + A * np.sqrt(y)) / np.sqrt(k)
+            psi_new = _newton_rhapson(psi, A, xi, y, k, tof, tof_new)
+            psi_new, psi_low, psi_up = _check_bounds(
+                psi, psi_new, psi_up, psi_low, tof_new, tof
+            )
+            psi = psi_new
+            count += 1
+
+            converged = np.abs((tof_new - tof) / tof) < rtol
+
+            # Convergence check
+            if converged:
+                # make sure the first guess isn't too close
+                if count == 1:
+                    tof_new = tof - 1.0
+                else:
+                    break  # Break out of count < numiter loop
     else:
         raise RuntimeError("Maximum number of iterations reached")
 
@@ -162,8 +201,8 @@ def vallado(k, r0, r, tof, short, numiter, rtol):
 
 
 @jit
-def izzo(k, r1, r2, tof, M, numiter, rtol):
-    """ Aplies izzo algorithm to solve Lambert's problem.
+def izzo(k, r1, r2, tof, num_rev, numiter, rtol):
+    """ Applies izzo algorithm to solve Lambert's problem.
 
     Parameters
     ----------
@@ -175,7 +214,7 @@ def izzo(k, r1, r2, tof, M, numiter, rtol):
         Final position vector
     tof: float
         Time of flight between both positions
-    M: int
+    num_rev: int
         Number of revolutions
     numiter: int
         Numbert of iterations
@@ -225,7 +264,7 @@ def izzo(k, r1, r2, tof, M, numiter, rtol):
     T = np.sqrt(2 * k / s ** 3) * tof
 
     # Find solutions
-    xy = _find_xy(ll, T, M, numiter, rtol)
+    xy = _find_xy(ll, T, num_rev, numiter, rtol)
 
     # Reconstruct
     gamma = np.sqrt(k * s / 2)
@@ -254,7 +293,7 @@ def _reconstruct(x, y, r1, r2, ll, gamma, rho, sigma):
 
 
 @jit
-def _find_xy(ll, T, M, numiter, rtol):
+def _find_xy(ll, T, num_rev, numiter, rtol):
     """Computes all x, y for given number of revolutions.
 
     """
@@ -262,24 +301,24 @@ def _find_xy(ll, T, M, numiter, rtol):
     assert abs(ll) < 1
     assert T > 0  # Mistake on original paper
 
-    M_max = np.floor(T / pi)
+    num_rev_max = np.floor(T / pi)
     T_00 = np.arccos(ll) + ll * np.sqrt(1 - ll ** 2)  # T_xM
 
     # Refine maximum number of revolutions if necessary
-    if T < T_00 + M_max * pi and M_max > 0:
-        _, T_min = _compute_T_min(ll, M_max, numiter, rtol)
+    if T < T_00 + num_rev_max * pi and num_rev_max > 0:
+        _, T_min = _compute_T_min(ll, num_rev_max, numiter, rtol)
         if T < T_min:
-            M_max -= 1
+            num_rev_max -= 1
 
     # Check if a feasible solution exist for the given number of revolutions
     # This departs from the original paper in that we do not compute all solutions
-    if M > M_max:
-        raise ValueError("No feasible solution, try lower M")
+    if num_rev > num_rev_max:
+        raise ValueError("No feasible solution, try lower num_rev")
 
     # Initial guess
-    for x_0 in _initial_guess(T, ll, M):
+    for x_0 in _initial_guess(T, ll, num_rev):
         # Start Householder iterations from x_0 and find x, y
-        x = _householder(x_0, T, ll, M, rtol, numiter)
+        x = _householder(x_0, T, ll, num_rev, rtol, numiter)
         y = _compute_y(x, ll)
 
         yield x, y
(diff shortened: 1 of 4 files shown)


src/poliastro/iod/vallado.py
```diff
@@ -8,7 +8,7 @@
 kms = u.km / u.s
 
 
-def lambert(k, r0, r, tof, short=True, numiter=35, rtol=1e-8):
+def lambert(k, r0, r, tof, M=0, short=True, numiter=35, rtol=1e-8, direct=True):
```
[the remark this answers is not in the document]
NanuSai: I suspect that very much! 
But according to our [poliastro docs](https://docs.poliastro.space/en/stable/examples/Multirevolutions%20solution%20in%20Lamberts%20problem.html) 

Short refers to the orbit as a whole
But direct refers to prograde motion ``` Retrograde motion in astronomy is, in general, orbital or rotational motion of an object in the direction opposite the rotation of its primary, that is, the central object.```

But the way these parameters are handled are really confusing in the lambertu.m file, so it'll help if you have a look for sanity check ;D

src/poliastro/iod/vallado.py
```diff
@@ -8,7 +8,7 @@
...
     """Solves the Lambert problem.
```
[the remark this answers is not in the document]
NanuSai: oopsie daisy !. I'll fix that

src/poliastro/core/iod.py
```diff
@@ -163,7 +196,7 @@ def vallado(k, r0, r, tof, short, numiter, rtol):
 
 @jit
 def izzo(k, r1, r2, tof, M, numiter, rtol):
-    """ Aplies izzo algorithm to solve Lambert's problem.
+    """ Applies izzo algorithm to solve Lambert's problem.
```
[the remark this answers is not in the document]
NanuSai: No problamae ! ;D

src/poliastro/core/iod.py
```diff
@@ -9,7 +9,7 @@
 
 
 @jit
-def vallado(k, r0, r, tof, short, numiter, rtol):
+def vallado(k, r0, r, tof, M, direct, short, numiter, rtol):
```
astrojuanlu, line 12: Can we _not_ call this `M`? At first I thought this was supposed to be the true anomaly. I know, I know: this is the name we used for `izzo`, but in retrospect I now think it was bad :joy: No need to fix it there though, just use `num_rev` here.

astrojuanlu: Regarding the confusion between `direct` and `short` that @jorgepiloto already mentioned, the MATLAB code makes a distinction between:

* "Direction of motion": long period vs short period
* "Direction of flight": direct vs retrograde

However, I don't see that distinction mentioned in the book:

Screenshot from 2020-04-24 13-18-53

Look at these disturbing comments in the MATLAB code:

```matlab
if ( df == 'r' )  %dm == 'l'
...
if (df == 'd')  % dm == 's'
```

Does that mean that they are the same?

To simplify things, let's consider only **prograde orbits**, and maintain the `short` parameter.

src/poliastro/core/iod.py
```diff
@@ -111,42 +111,75 @@ def vallado(k, r0, r, tof, short, numiter, rtol):
 
     A = t_m * (norm_r * norm_r0 * (1 + cos_dnu)) ** 0.5
 
-    if A == 0.0:
+    if abs(A) <= 0.2:  # Not exactly zero
```
astrojuanlu, line 114: Can we keep it like before? If any test fails, we'll see why and how to fix it.

src/poliastro/core/iod.py
```diff
@@ -111,42 +111,75 @@ def vallado(k, r0, r, tof, short, numiter, rtol):
...
+            y = norm_r0_plus_norm_r
 
-    while count < numiter:
-        y = norm_r0_plus_norm_r + A * (psi * c3(psi) - 1) / c2(psi) ** 0.5
         if A > 0.0:
             # Readjust xi_low until y > 0.0
             # Translated directly from Vallado
-            while y < 0.0:
-                psi_low = psi
+            ynegktr = 1
```
astrojuanlu, line 143: Can we rename this lovely variable `ynegktr` to `neg_y_max_iter`?

src/poliastro/core/iod.py
```diff
@@ -111,42 +111,75 @@ def vallado(k, r0, r, tof, short, numiter, rtol):
...
+            psi_new = _newton_rhapson(psi, A, xi, y, k, tof, tof_new)
+            psi_new, psi_low, psi_up = _check_bounds(
+                psi, psi_new, psi_up, psi_low, tof_new, tof
+            )
+            psi = psi_new
+            count += 1
+
+            # make sure the first guess isn't too close
+            if abs(tof_new - tof) < rtol and count == 1:
+                tof_new = tof - 1.0
```
astrojuanlu, line 181: Am I missing an `else: break` if `ynegktr < 10` is False? To reach the `RuntimeError` below?

src/poliastro/core/iod.py
```diff
@@ -293,6 +326,79 @@ def _compute_y(x, ll):
...
+
+
+@jit
+def _newton_rhapson(psi, A, xi, y, k, tof, tof_new):
+
+    """Newton-Rhapson iteration to update psi
+
+    """
+    # Newton rhapson iteration
+    if abs(psi) > 1e-5:
```
astrojuanlu, line 358: Can we separate the objective functions from the Newton method in a similar way as done in `poliastro.core.angles:newton`?

src/poliastro/core/iod.py
```diff
@@ -293,6 +326,79 @@ def _compute_y(x, ll):
...
+        3.0 * c3 * np.sqrt(y) / c2(psi) + A / xi
+    ) * 1 / np.sqrt(k)
+
+    psi_new = psi - (tof_new - tof) / dtdpsi
+    return psi_new
+
+
+@jit
+def _initial_guess_vallado(psi, psi_low, psi_up, M, tof, direct):
+    """Returns Initial guess for psi for vallado solver
```
astrojuanlu, line 386: Remove this docstring, the method name is already good enough :)

src/poliastro/core/iod.py
NanuSai: Now that I think of it, once you've decided it's short of long there's **only one way** to go.  Like in Fig 7 you showed,  once you've chosen short that's it, no more "direct/retrograde". I think they're synonymous don't you agree?

astrojuanlu: Yeah, let's consider them synonymous!

src/poliastro/core/iod.py
NanuSai: What I'm surprised is, there is already
 ```python
    while count < numiter and ynegktr <= 10:
```
so, why again ```ynegktr < 10```. I'm thinking it's probably because of the first guess not being too close.

File: src/poliastro/core/iod.py [hunk 10]
```diff
@@ -111,42 +111,75 @@ def vallado(k, r0, r, tof, short, numiter, rtol):
...
+    while count < numiter and ynegktr <= 10:
+
+        if abs(c2(psi) > rtol):
+            y = norm_r0_plus_norm_r + A * (1.0 - psi * c3(psi)) / c2(psi) ** 0.5
+        else:
+            y = norm_r0_plus_norm_r
 
-    while count < numiter:
-        y = norm_r0_plus_norm_r + A * (psi * c3(psi) - 1) / c2(psi) ** 0.5
         if A > 0.0:
```
NanuSai, line 140: ```A > 0.0``` only when short is True, we can just replace this by ``` if short: ```

src/poliastro/core/iod.py
NanuSai: **FIXED:**
```python
            # Convergence check
            if converged:
                # make sure the first guess isn't too close
                if count == 1:
                    tof_new = tof - 1.0 
                else:
                    break
```
where the ```break``` statement would land us to the RuntimeError :D

src/poliastro/core/iod.py
NanuSai: Too long :( . Anything better ?

astrojuanlu: This is not FORTRAN, we can afford `neg_y_max_iter` 😝

src/poliastro/core/iod.py
NanuSai: Question, in the if else statement the conditions are
1. if np.abs(psi) > 1e -5 :
2. parabolic
What does the first function correspond to? :/

src/poliastro/core/iod.py
NanuSai: Haha, okay!

src/poliastro/core/iod.py
astrojuanlu: Non-parabolic? :) (Elliptic or hyperbolic)

NanuSai:
Slight problem, our test_iod.py have parameterized inputs. So lambert is an alias for both izzo and vallado.One adopts M and the other num_rev, there is a clash

src/poliastro/iod/vallado.py
```diff
@@ -23,6 +23,10 @@ def lambert(k, r0, r, tof, short=True, numiter=35, rtol=1e-8):
         Final position (km).
     tof : ~astropy.units.Quantity
         Time of flight (s).
+    M : int, optional
```
astrojuanlu, line 26:
```suggestion
    num_rev : int, optional
```

astrojuanlu:
> Slight problem, our test_iod.py have parameterized inputs. So lambert is an alias for both izzo and vallado.One adopts M and the other num_rev, there is a clash

Oh, true... Let's update it in the other one then.

src/poliastro/iod/vallado.py
NanuSai: Thanks!, I'm saving docstring tests for last. I'll fix in next push

src/poliastro/core/iod.py
```diff
@@ -114,39 +114,69 @@ def vallado(k, r0, r, tof, short, numiter, rtol):
     if A == 0.0:
         raise RuntimeError("Cannot compute orbit, phase angle is 180 degrees")
 
-    psi = 0.0
-    psi_low = -4 * np.pi ** 2
-    psi_up = 4 * np.pi ** 2
+    psi = 0
+
+    if num_rev == 0:  # Usual num_rev == 0 case
```
astrojuanlu, line 119: Also, `num_rev` missing in the docstring of this function

NanuSai:
It "looks" great, the tests are failing. I'll make the next push when our test_iod.py passes.

src/poliastro/core/iod.py
```diff
@@ -293,6 +330,97 @@ def _compute_y(x, ll):
...
+
+@jit
+def _c3dot(psi, c2, c3):
+    return 0.5 / psi * (c2(psi) - 3.0 * c3(psi))
+
+
+@jit
+def _c2dot_para(psi):
+    return (
+        -1.0 / np.math.factorial(4) + 2.0 * psi / np.math.factorial(6) - 3.0 * psi
```
astrojuanlu, line 368: `Unknown attribute 'factorial' of type Module(<module 'math' `

https://dev.azure.com/poliastro/poliastro/_build/results?buildId=259&view=logs&j=77e370ac-c6ce-58ed-b689-4eaf2ed67de9&t=ffe2475a-255d-544e-4141-d84b3eb0a61b&l=1457

You must use `math.gamma`, see https://github.com/poliastro/poliastro/pull/31/files#diff-bfecda9ed598400464e2ce964edbb901 and https://github.com/numba/numba/pull/987#issuecomment-84980245.

astrojuanlu:
```
>       v0, v = vallado_fast(k_, r0_, r_, tof_, num_rev, short, numiter, rtol)
E       TypeError: expected UniTuple(float64 x 3), got None
```

https://dev.azure.com/poliastro/poliastro/_build/results?buildId=261&view=logs&j=77e370ac-c6ce-58ed-b689-4eaf2ed67de9&t=ffe2475a-255d-544e-4141-d84b3eb0a61b&l=1303

astrojuanlu:
Running `NUMBA_DISABLE_JIT=1 pytest -k issue840 -vv` (or `NUMBA_DISABLE_JIT=1 tox -e py38 -- -k issue840 -vv` gives a clue about what's going on:

```

(poliastro38) juanlu@ephyra:~/Personal/poliastro/poliastro$ NUMBA_DISABLE_JIT=1 pytest -k issue840
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.8.0, pytest-5.3.2, py-1.8.1, pluggy-0.13.1
Matplotlib: 3.2.1
Freetype: 2.6.1
rootdir: /home/juanlu/Personal/poliastro/poliastro, inifile: setup.cfg
plugins: cov-2.5.1, mpl-0.11, mypy-0.4.2, doctestplus-0.5.0, remotedata-0.3.2, hypothesis-5.10.1
collected 915 items / 914 deselected / 1 selected

tests/test_iod.py F                                                                                                                                                                                                                    [100%]

================================================================================================================== FAILURES ==================================================================================================================
_______________________________________________________________________________________________________ test_issue840[lambert-lambert] _______________________________________________________________________________________________________

lambert_vallado = <function lambert at 0x7f8784901ee0>, lambert_izzo = <function lambert at 0x7f8784cd9ca0>

    @pytest.mark.parametrize(
        "lambert_vallado,lambert_izzo", [(vallado.lambert, izzo.lambert)]
    )
    def test_issue840(lambert_vallado, lambert_izzo):
        k = c.GM_earth.to(u.km ** 3 / u.s ** 2)
        r0 = [10000.0, 0, 0] * u.km
        rf = [8000.0, -5000, 0] * u.km
        tof = 2 * u.hour

        expected_va = [0.36591277, 5.8228806, 0.0] * u.km / u.s
        expected_vb = [3.99397599, 4.78236576, 0.0] * u.km / u.s

>       va_v, vb_v = next(lambert_vallado(k, r0, rf, tof, short=False))

tests/test_iod.py:141:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
src/poliastro/iod/vallado.py:52: in lambert
    v0, v = vallado_fast(k_, r0_, r_, tof_, num_rev, short, numiter, rtol)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

k = 398600.4, r0 = array([10000.,     0.,     0.]), r = array([ 8000., -5000.,     0.]), tof = 7200.0, num_rev = 0, short = False, numiter = 35, rtol = 1e-08

    @jit
    def vallado(k, r0, r, tof, num_rev, short, numiter, rtol):
        r"""Solves the Lambert's problem.

        The algorithm returns the initial velocity vector and the final one, these are
        computed by the following expresions:

        .. math::

            \vec{v_{o}} &= \frac{1}{g}(\vec{r} - f\vec{r_{0}}) \\
            \vec{v} &= \frac{1}{g}(\dot{g}\vec{r} - \vec{r_{0}})

        Therefore, the lagrange coefficients need to be computed. For the case of
        Lamber's problem, they can be expressed by terms of the initial and final vector:

        .. math::

            \begin{align}
                f = 1 -\frac{y}{r_{o}} \\
                g = A\sqrt{\frac{y}{\mu}} \\
                \dot{g} = 1 - \frac{y}{r} \\
            \end{align}

        Where y(z) is a function that depends on the :py:mod:`poliastro.core.stumpff` coefficients:

        .. math::

            y = r_{o} + r + A\frac{zS(z)-1}{\sqrt{C(z)}} \\
            A = \sin{(\Delta \nu)}\sqrt{\frac{rr_{o}}{1 - \cos{(\Delta \nu)}}}

        The value of z to evaluate the stump functions is solved by applying a Numerical method to
        the following equation:

        .. math::

            z_{i+1} = z_{i} - \frac{F(z_{i})}{F{}'(z_{i})}

        Function F(z)  to the expression:

        .. math::

            F(z) = \left [\frac{y(z)}{C(z)}  \right ]^{\frac{3}{2}}S(z) + A\sqrt{y(z)} - \sqrt{\mu}\Delta t

        Parameters
        ----------
        k: float
            Gravitational Parameter
        r0: ~np.array
            Initial position vector
        r: ~np.array
            Final position vector
        tof: ~float
            Time of flight
        num_rev: int
            Number of revolutions
        numiter: int
            Number of iterations to
        rtol: float
            Error tolerance


        Returns
        -------
        v0: ~np.array
            Initial velocity vector
        v: ~np.array
            Final velocity vector

        Examples
        --------

        >>> from poliastro.core.iod import vallado
        >>> from astropy import units as u
        >>> import numpy as np
        >>> from poliastro.bodies import Earth
        >>> k = Earth.k.to(u.km**3 / u.s**2)
        >>> r1 = np.array([5000, 10000, 2100])*u.km #Initial position vector
        >>> r2 = np.array([-14600, 2500, 7000])*u.km #Final position vector
        >>> tof = 3600*u.s #Time of fligh
        >>> v1, v2 = vallado(k.value, r1.value, r2.value, tof.value, short=True, numiter=35, rtol=1e-8)
        >>> v1 = v1*u.km / u.s
        >>> v2 = v2*u.km / u.s
        >>> print(v1, v2)
        [-5.99249503  1.92536671  3.24563805] km / s [-3.31245851 -4.19661901 -0.38528906] km / s

        Note
        ----
        This procedure can be found in section 5.3 of Curtis, with all the
        theoretical description of the problem. Analytical example can be found
        in the same book under name Example 5.2.

        """
        if short:
            t_m = +1
        else:
            t_m = -1

        norm_r0 = np.dot(r0, r0) ** 0.5
        norm_r = np.dot(r, r) ** 0.5
        norm_r0_times_norm_r = norm_r0 * norm_r
        norm_r0_plus_norm_r = norm_r0 + norm_r

        cos_dnu = np.dot(r0, r) / norm_r0_times_norm_r

        A = t_m * (norm_r * norm_r0 * (1 + cos_dnu)) ** 0.5

        if A == 0.0:
            raise RuntimeError("Cannot compute orbit, phase angle is 180 degrees")

        psi = 0

        if num_rev == 0:  # Usual num_rev == 0 case
            psi_low = -4.0 * np.pi ** 2
            psi_up = 4.0 * np.pi ** 2

        else:
            psi_low = 4.0 * num_rev ** 2 * np.pi ** 2
            psi_up = 4.0 * (num_rev + 1.0) ** 2 * np.pi ** 2

        psi = _initial_guess_vallado(psi, psi_low, psi_up, num_rev, tof, short)

        count = 0
        tof_new = -10

        while count < numiter:

            if np.abs(c2(psi) > rtol):
                y = norm_r0_plus_norm_r + A * (1.0 - psi * c3(psi)) / c2(psi) ** 0.5
            else:
                y = norm_r0_plus_norm_r

            if short:
                # Readjust xi_low until y > 0.0
                # Translated directly from Vallado
                neg_y_max_iter = 1

                while y < 0.0 and neg_y_max_iter < 10:
                    psi = (
                        0.8
                        * (1.0 / c3(psi))
                        * (1.0 - norm_r0_times_norm_r * np.sqrt(c2(psi)) / A)
                    )
                    psi_low = psi
                    if np.abs(c2(psi)) > rtol:
                        y = norm_r0_plus_norm_r + A * (
                            (1.0 - psi * c3(psi)) / c2(psi) ** 0.5
                        )
                    else:
                        y = norm_r0_plus_norm_r
                    neg_y_max_iter = neg_y_max_iter + 1

                break  # Break out of count < numiter loop

            else:
                if np.abs(c2(psi)) > rtol:
                    xi = np.sqrt(y / c2(psi))
                else:
                    xi = 0.0

                tof_new = (xi ** 3 * c3(psi) + A * np.sqrt(y)) / np.sqrt(k)
                psi_new = _newton_rhapson(psi, A, xi, y, k, tof, tof_new)
>               psi_new, psi_low, psi_up = _check_bounds(
                    psi, psi_new, psi_up, psi_low, tof_new, tof
                )
E               TypeError: cannot unpack non-iterable NoneType object

src/poliastro/core/iod.py:174: TypeError

```

src/poliastro/core/iod.py
```diff
@@ -293,6 +332,99 @@ def _compute_y(x, ll):
...
+        steep_condition_1 = tof_new < tof and psi > psi_low
+        psi_low = psi_low + (psi - psi_low) * steep_condition_1
+
+        # upper_bound
+        steep_condition_2 = tof_new > tof and psi < psi_up
+        psi_up = psi_up + (psi - psi_up) * steep_condition_2
+
+        psi_new = (psi_up + psi_low) / 2
+
+    return psi_new, psi_low, psi_up
```
NanuSai, line 354: The only difference previous code and this one is, the return statement. This time we return regardless of ```if(condition)```.

NanuSai:
```python
Expected:
y=10967.478908770492
f=0.31218281273585313
g = 3514.9491491164854
gdot = 0.31218265289197367
v0 = array([2.05891255, 2.91596461, 0.        ])
v = array([-3.4515665 ,  0.91031357,  0.        ])

Got:
y=1923.244317587858
f= 0.879385179771152
g=1471.91547564592
gdot = 0.8793851517411156
v0=array([-1.22782981,  6.96335318,  0.        ])
v = array([-3.53539075,  6.12346939,  0.        ])
```
Here's a brief expectation vs reality. Fixing the assert error now

jonpsy:
Re-open when there's some activity on this.

Issue #815, opened 2019-12-05
Unify propagate and Orbit.propagate API


noc0lour:
Currently to get intermediate sampling positions & the orbit at the end of simulation `propagate()` is essentially called twice and the whole propagation is performed twice. 

```
            results = propagate(
                sc.orbit,
                sampling_times,
                method=cowell,
                ad=J2_perturbation,
                J2=bodies.Earth.J2,
                R=bodies.Earth.R.to(u.km).value,
                rtol=1e-13,
            )
            sc.orbit = sc.orbit.propagate(
                sc.orbit.epoch + ticks * self._sampling,
                method=cowell,
                ad=J2_perturbation,
                J2=bodies.Earth.J2,
                R=bodies.Earth.R.to(u.km).value,
            )
```

💡 **Feature request**

Having either the `propagate()` function from `twobody` optionally return the new propagated orbit.
Or the `Orbit.propagate()` function to return a time-series of coordinates + a new orbit upon input of a time-series.

🎯 **Goal**

Speeding up simulations 2x by only `propagate()`ing once since orbit.propagate() calls propagate() internally.

priyanshurohilla:
i want to work on it.

astrojuanlu:
First things first: sorry @noc0lour for the looooooong delay, I know we talked in the chat but it doesn't usually take me that much time to respond to an issue in poliastro. I'm in a better position now :)

Simpler case:

```python
In [8]: times = time_range(Time("2020-04-04 12:00:00", scale="tt"), periods=3, end=Time("2020-04-04 14:00:00", scale="tt"))                                                                                                                   

In [9]: times                                                                                                                                                                                                                                 
Out[9]: 
<Time object: scale='tt' format='iso' value=['2020-04-04 12:00:00.000' '2020-04-04 13:00:00.000'
 '2020-04-04 14:00:00.000']>

In [10]: propagate(iss, times)  # Representation in all times, no Orbit                                                                                                                                                                                                              
Out[10]: 
<CartesianRepresentation (x, y, z) in km
    [( 5128.4602846 ,  3729.87885198, -2425.77381503),
     (  237.13966575, -4293.07533267,  5234.77691922),
     (-5384.63450019,  1436.20865414, -3861.90127496)]
 (has differentials w.r.t.: 's')>

In [11]: iss.propagate(times[-1])  # Orbit in chosen time, no intermediate positions                                                                                                                                                                                                           
[...]

PR #896, opened 2020-04-04
Optimized orbit.propagate()

priyanshurohilla:
Addresses #815 
Please review the changes

### DIFF
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -1138,6 +1138,7 @@ class Orbit:
         -------
         Orbit
             New orbit after propagation.
+            Propagate an orbit some time and return the cartesian.
 
         """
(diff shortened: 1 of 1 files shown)


priyanshurohilla:
Why are the tests failing?

astrojuanlu:
> Why are the tests failing?

I didn't look at the tests, but you changed the return value of a function that is used everywhere :)

priyanshurohilla:
> > Why are the tests failing?
> 
> I didn't look at the tests, but you changed the return value of a function that is used everywhere
Oh godd this didn't click to me!! I need to see what else I can do

astrojuanlu:
I expressed my dislike of changing the return value according to a parameter and also the original use case is covered with existing functions reusing the Astropy objects, so let's close this. Thanks @priyanshurohilla for the effort nonetheless!

PR #898, opened 2020-04-08
SolarSystemBody.plot()

astrojuanlu:
Convenience method to avoid using `Orbit.from_body_ephem`, comes from #882.


astrojuanlu:
Is the doctest failure on Windows just random? I'll go ahead and merge this pull request, and we'll see if it fails again or not.

PR #900, opened 2020-04-10
Final plotting refactors & fixes

astrojuanlu:
I was still not satisfied with the refactor of the orbit plotters, plus too many plotting changes were still accumulating in #882. I prefer to make them all now, and rebase later to minimize the plotting changes.

Also, this features the new body palette! 🎨

test_plot_inner_solar_system_static


astrojuanlu:
@humitos do you have any idea why is Azure Pipelines complaining that "hoverxref extension is not fully configured"? I suspect it's not reading the `conf.py` somehow, but it's weird because nothing changed in the past days, and [the RTD build looks normal](https://poliastro--900.org.readthedocs.build/en/900/) after I adapted the configuration for `0.2b1`.

astrojuanlu:
For the record, I can reproduce it locally with tox.

astrojuanlu:
Oh, I see:

```
setenv =
  READTHEDOCS_PROJECT = poliastro
  READTHEDOCS_VERSION = latest
```

not sure why it was working before, but this did the trick.

humitos:
Sorry for the issues.

I moved that warning to an upper level. Now, if the extension is installed and not well configured it warns. 

Before the last release it warns only if it was used and not well configured. Although, I don't fully understand why it wasn't warning before 🤷‍♂️

Adding those variables solves the issue, but it makes me think that more people will hit this issue and maybe I should remove that warning, but then you will be surprised that the tooltip doesn't work. There is a tradeoff...

PR #908, opened 2020-04-13
Fix near parabolic regimes

### LINKED ISSUE #475: Propagator mean_motion hangs for some r, v vectors around Earth

### LINKED ISSUE #907: Orbit.propagate_to_anomaly freezes
(TBC)

```python
In [86]: r_a                                                                                                          
Out[86]: <Quantity 2200. m>

In [87]: r_p = 10 * u.m                                                                                               

In [88]: a = (r_a + r_p) / 2; a                                                                                       
Out[88]: <Quantity 1105. m>

In [89]: e = r_a / a - 1                                                                                              

In [90]: e                                                                                                            
Out[90]: <Quantity 0.99095023>

In [91]: o_n = Orbit.from_classical(ast, a, e, 0 * u.deg, 0 * u.deg, 0 * u.deg, 180 * u.deg)                          

In [92]: o_n.r                                                                                                        
Out[92]: <Quantity [-2.20000000e+00,  2.69422296e-16,  0.00000000e+00] km>

In [93]: o_n.r_a                                                                                                      
Out[93]: <Quantity 2200. m>

In [94]: o_n.r_p                                                                                                      
Out[94]: <Quantity 10. m>

In [95]: o_n.nu                                                                                                       
Out[95]: <Quantity 180. deg>

In [96]: o_n.propagate_to_anomaly(191.35 * u.deg)
```

astrojuanlu:
Regions of `(nu, ecc)` from Farnocchia at al., 2013:

Screenshot from 2020-04-13 04-40-44

Previously our logic was wrong: we were incorrectly using the near parabolic solution where we should have been using the strong elliptic solution, and therefore it failed "in the vicinity of `nu = pi`, `0.985 < ecc < 1.0` (for `delta=1e-2`)" [as I correctly identified two years ago](https://github.com/poliastro/poliastro/issues/475#issuecomment-439002103).

* [x] Fix `nu_to_M` ([which causes `orbit.M` to hang](https://github.com/poliastro/poliastro/issues/907#issuecomment-612694540))
* [x] Fix `M_to_nu` ([which causes the `mean_motion` propagator to hang](https://github.com/poliastro/poliastro/issues/475))

Fix #475 (finally!), fix #907.

### DIFF
--- a/src/poliastro/core/angles.py
+++ b/src/poliastro/core/angles.py
@@ -282,6 +282,7 @@ def M_to_E(M, ecc):
     This uses a Newton iteration on the Kepler equation.
 
     """
+    assert -np.pi <= M <= np.pi
     if ecc < 0.8:
         E0 = M
     else:
--- a/src/poliastro/core/angles.py
+++ b/src/poliastro/core/angles.py
@@ -285,7 +285,7 @@ def M_to_E(M, ecc):
     if ecc < 0.8:
         E0 = M
     else:
-        E0 = np.pi
+        E0 = np.pi * np.sign(M)
     E = newton("elliptic", E0, args=(M, ecc))
     return E
 
--- a/src/poliastro/core/angles.py
+++ b/src/poliastro/core/angles.py
@@ -40,7 +40,7 @@ def newton(regime, x0, args=(), tol=1.48e-08, maxiter=50):
             return p
         p0 = p
 
-    return 1.0
+    return np.nan
 
 
 @jit
@@ -282,7 +282,10 @@ def M_to_E(M, ecc):
     This uses a Newton iteration on the Kepler equation.
 
     """
-    E0 = M
+    if ecc < 0.8:
+        E0 = M
+    else:
+        E0 = np.pi
     E = newton("elliptic", E0, args=(M, ecc))
     return E
 
--- a/src/poliastro/core/propagation/farnocchia.py
+++ b/src/poliastro/core/propagation/farnocchia.py
@@ -2,6 +2,7 @@ import numpy as np
 
 from .._jit import jit
 from ..angles import (
+    D_to_M,
     D_to_nu,
     E_to_M,
     E_to_nu,
@@ -19,7 +20,7 @@ from ..elements import coe2rv, rv2coe
 
 @jit
 def _kepler_equation_near_parabolic(D, M, ecc):
-    return D_to_M_near_parabolic(ecc, D) - M
+    return D_to_M_near_parabolic(D, ecc) - M
 
 
 @jit
@@ -79,7 +80,7 @@ def d2S_x_alt(ecc, x, atol=1e-12):
 
 
 @jit
-def D_to_M_near_parabolic(ecc, D):
+def D_to_M_near_parabolic(D, ecc):
     x = (ecc - 1.0) / (ecc + 1.0) * (D ** 2)
     assert abs(x) < 1
     S = S_x(ecc, x)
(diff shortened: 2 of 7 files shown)


astrojuanlu:
Asked [one question about notation in Math.StackExchange](https://math.stackexchange.com/questions/3623757/what-set-is-mathbbt1).

astrojuanlu:
Well, this was more difficult than expected. Getting errors in some old tests that compare the propagator with Cowell's method, still not sure what's happening. Will review tomorrow.

astrojuanlu:
Well, this turned out to be more difficult than I thought.

Several findings while working on this:

* Fixing the near parabolic region computation in `nu_to_M` and `M_to_nu` fixes some cases, but breaks others. The functions does not seem to be smooth at the boundaries, so small numerical errors can have a big effect and produce totally different results.
* We have a *huge* mess in the `Orbit.sample()` method and it's urgent that we refactor it to avoid several useless conversions between classical and cartesian elements.
* There is a very tight coupling between how the Farnocchia method separates the "near parabolic" region, and all the code in `poliastro.core.angles`. Which either means:
  - There shouldn't be such coupling, and all the near-parabolic code should be used only by `poliastro.core.propagation.farnocchia`.
  - All the propagators need such distinctions to work properly for near-parabolic orbits, and therefore the near-parabolic boundary will trascend the `farnocchia` propagator.

I still don't understand why `nu_to_M` and `M_to_nu` are not smooth. There's excellent mathematical background about them in the paper, but it requires deeper inspection.

We have a path to fix this, but it will take a long time.

astrojuanlu:
| Paper                                                                                                             | Numerical simulation                                                                                                       |
|-------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------|
| farnocchia | direct_107_12_19_53 |

Looking great so far! Now on to the next step :muscle:

astrojuanlu:
**All** the propagators in `poliastro.core.propagation` use some sort of angle conversion at some point, and in some algorithms branch depending on the eccentricity.

A very obvious case can be seen in Mikkola's method:

https://github.com/poliastro/poliastro/blob/643356d5cf858816bc3e70326d3f441926483264/src/poliastro/core/propagation.py#L327-L335

I feel that if we mix the treatment that Farnocchia does of the near-parabolic region with the other propagators we can get mixed results. Doing so probably requires a careful review of all the algorithms and implementations. The last two years we have devoted a lot of time to improve our propagators, and I have to say both @jorgepiloto and @nikita-astronaut did an amazing job. 

Therefore, I will try not to affect how other propagators work while fixing the issues with the Farnocchia one.

And to finally justify all this work, let's take a look at this table again:

https://github.com/poliastro/poliastro/blob/643356d5cf858816bc3e70326d3f441926483264/src/poliastro/twobody/propagation.py#L4-L22

Leaving Cowell's method aside, Farnocchia's and Vallado's are the only propagators we have at the moment that are supposed to work with all orbits. Vallado's method, being based on the universal variables approach, shares its pros and cons. From Farnocchia et al.:

Screenshot from 2020-04-18 11-56-56

Even though @nikita-astronaut fixed the convergence problems of Vallado's method in #362, it still requires a large number of Newton iterations (in line with the comment above) and the relative tolerance is bounded by numerical errors.

In conclusion:

* It's worth it to try to fix this propagator, and I will do so in this PR.
* The angles functions should be work as before because they are used in other propagators.
* We will keep Farnocchia's method as the one preferred for sampling, and we will do some refactoring to avoid unnecessary element conversions and accelerate it considerably.

astrojuanlu:
While working on this PR I came to an important realization:

Even if most books and papers talk about the mean anomaly as a quantity that can be defined for elliptic, parabolic and hyperbolic orbits, more rigorous texts like Battin are very careful not to abuse the term for eccentricities equal or larger than one.

For example, the "mean anomaly" for hyperbolic motion is referred to as "the quantity N":

Screenshot from 2020-04-18 17-23-39

The same can be said about the "parabolic anomaly" and the "hyperbolic anomaly", which are never called as such: the former is mentioned tangentially suggested when saying that "tan f/2 is the key variable in Barker's equation", and the latter is called "the quantity H", or the "hyperbolic analog to the eccentric anomaly".

My theory is that this is done because these quantities do not represent angles directly. However, for practical reasons this fact is often overlooked.

astrojuanlu:
My previous comment has important implications for poliastro. In the same way that `Orbit` objects do not have an `E` property (for the eccentric anomaly), **they shouldn't have an `M` property for the mean anomaly either**, and it should be treated at a lower level as a useful mathematical artifact, but nothing more.

astrojuanlu:
Current status:

inverse_110_21_36_50

Getting there.

astrojuanlu:
The reason for the divergence is in the Newton iteration process. After the first step, the anomaly gets out of the near parabolic region and diverges.

astrojuanlu:
Now the implementation seems to be working!

inverse_111_11_40_32

And therefore, I'll direct my attention to... the tests.

astrojuanlu:
Most of the failures seem legitimate. One could question the comparisons with Cowell, but it is strange that they were passing before.

One of the possible interpretations is that now the propagator is _actually_ using the near parabolic region, which was not the case before, and we are losing accuracy as a result. From the paper:

> δ = 10 −2 seems to strike a good compromise between efficiency and robustness, but other choices are possible.

Nothing is said about the accuracy of the result.

However, it's likely that there is still some problem in the code, otherwise `test_elliptic_near_parabolic[farnocchia-0.9]` would not fail, because that's well outside the near parabolic region for all sensible values of the parameter delta.

astrojuanlu:
Raising the value to `delta=1e-3` *did* improve the accuracy of some of the comparisons, but still there are lots of failures that indicate some bug somewhere.

astrojuanlu:
The difficult part of this pull request is done :tada:

astrojuanlu:
All tests passing on Linux! Problematic tests:

* `[doctest] poliastro.core.elements.rv2coe` (Windows)
* `test_propagation_near_parabolic_orbits_zero_seconds_gives_same_anomaly` (Windows and macOS)
* `test_long_propagations_vallado_agrees_farnocchia` (macOS)
* `test_long_propagation_preserves_orbit_elements` (macOS)

Can't reproduce locally :thinking:

astrojuanlu:
This is working perfectly on Linux now :sparkles: 

Regarding the problematic tests [I listed above](https://github.com/poliastro/poliastro/pull/908#issuecomment-619036367), I could put a lot of `print` statements, disable all the other tests, and try to debug this on the CI... but it's a lot of pain.

Therefore, if you have a Windows or macOS operative system, I would ask you:

1. Check out my branch locally
2. Run `tox -e py38 -- -k "propagation_near_parabolic_orbits_zero_seconds_gives_same_anomaly" -vv`
3. Confirm (or not) that the test fails in your computer
4. Repeat after `export NUMBA_DISABLE_JIT=1` (or the Windows equivalent)
5. Do the same for the other problematic tests
6. And if you can even debug the root cause, that would be awesome!

If we are not able to diagnose what's happening, I will mark these tests as `xfail` and tackle them again after fixing https://github.com/poliastro/poliastro/issues/921.

Syed-Osama-Hussain:
I encounter the same error while running the test for ```propagation_near_parabolic_orbits_zero_seconds_gives_same_anomaly``` both with and without the environment variable ```NUMBA_DISABLE_JIT``` set. Tested on python 3.6.
```
E           Not equal to tolerance rtol=1e-07, atol=0
E
E           x and y nan location mismatch:
E            x: array(nan)
E            y: array(-3.136673) ```

**Comment by Syed-Osama-Hussain** [comment]
To me the root problem looks like that `nu` is being calculated for `strong elliptic` instead of `near parabolic` with `delta_t=-1191.532` for `nu = -3.136673`.

**Comment by astrojuanlu** [comment]
I don't quite understand your comment: if I step inside `propagation_near_parabolic_orbits_zero_seconds_gives_same_anomaly`, I reach this point:

```
(Pdb) l
316  	    q = p / (1 + ecc)
317  	
318  	    delta_t0 = delta_t_from_nu(nu0, ecc, k, q)
319  	    delta_t = delta_t0 + tof
320  	
321  ->	    nu = nu_from_delta_t(delta_t, ecc, k, q)
322  	
323  	    return coe2rv(k, p, ecc, inc, raan, argp, nu)
[EOF]
(Pdb) p delta_t
-1191.5328125489923
(Pdb) p ecc
0.9962001546703796
(Pdb) p k
398600.44180000003
(Pdb) p q
15.395515243651763
(Pdb) p nu_from_delta_t(delta_t, ecc, k, q)
-3.136672923790313
```

Therefore, `nu_from_delta_t(-1191.5328125489923, 0.9962001546703796, 398600.44180000003, 15.395515243651763)` gives `-3.136672923790313` (which happens to be exactly `nu0`).

Could you please give more detail of which function call is returning `NaN`? (We're getting there!)

**Comment by helgee** [comment]
I am seeing the same thing as @Syed-Osama-Hussain. The branching in `nu_from_delta_t` seems to be incorrect. I can also produce a `nan` on Linux by doing `nu_from_delta_t(-1191.5328125489923, 0.98)`.

**Comment by astrojuanlu** [comment]
Thanks @helgee, I can reproduce that locally, both with and without numba. Now I have a starting point to debug.

**Comment by helgee** [comment]
It seems that `newton` chokes on the negative mean anomaly.

**Comment by astrojuanlu** [comment]
Probably a bad initial guess. Should be easily fixable then, let me see

**Comment by astrojuanlu** [comment]
After the initial guess change, the following tests are still failing on CI:

* `test_long_propagation_preserves_orbit_elements[farnocchia]` (macOS) 
* `test_long_propagations_vallado_agrees_farnocchia` (macOS)

The Windows build passed though!

**Comment by helgee** [comment]
Those pass locally with `NUMBA_DISABLE_JIT=1` on my machine.

**Comment by astrojuanlu** [comment]
Can't reproduce these locally, either with numba 0.48 or 0.49, with or without jitting.

**Comment by noc0lour** [comment]
> Those pass locally with `NUMBA_DISABLE_JIT=1` on my machine. But fail while jitted.

It would be interesting to know the processor type/supported SIMD instruction sets, if numba is optimizing w.r.t. those that might be a difference...

**Comment by helgee** [comment]
> It would be interesting to know the processor type/supported SIMD instruction sets, if numba is optimizing w.r.t. those that might be a difference...

`Intel(R) Core(TM) m7-6Y75 CPU @ 1.20GHz`

https://ark.intel.com/content/www/us/en/ark/products/88199/intel-core-m7-6y75-processor-4m-cache-up-to-3-10-ghz.html

**Comment by astrojuanlu** [comment]
Found another falsifying case locally: `nu_from_delta_t(3393554832.0050254, 0.9676000001129921, 132712442099.00002, 87574593.50447243)`

**Comment by astrojuanlu** [comment]
*All tests passed* :heart_eyes:

Thanks everyone who chimed in! I am now merging this once and for all.

**Comment by noc0lour** [comment]
@helgee are your failing tests (with JIT) covered with the fixes in the latest commits? If not, it might be worth to open a separate issue on these.

**Comment by helgee** [comment]
Works fine now 👍

**Comment by mkbrewer** [comment]
@astrojuanlu Just a followup on your comment regarding issue #10230 in the astropy repository. I'm posting here as it seems like a more appropriate place to do so. Your comment was a bit concerning as I have a program that is in use by the control systems at several observatories to track comets in real time given a set of orbital elements. The algorithm that I use is attributed to Danby from the 2nd edition of his book Fundamentals of Celestial Mechanics 1988 (pages 149-154). I don't have a copy of this book, but it uses an accelerated Newton's method very similar to the one I see in `poliastro/core/propigation.py`. The only difference that I see is that: 
```
  fpp = s / 2.0
  fppp = c / 6.0
```
I have never had any problems with this program. From what I see here, the issues that you were seeing only occur when the true anomaly is close to 180 degrees. Is that correct? If that is the case, then I'm not going to worry much as we are only concerned with observing comets as they make their closest approach to Earth.

mkbrewer:
Oh, never mind on the difference in the algorithms. I see that yours has the factors of 1/2 and 1/6 absorbed in to deltas whereas mine doesn't.

astrojuanlu:
Thanks for chiming in @mkbrewer! (xref https://github.com/astropy/astropy/issues/10230)

The conversation in this particular pull request is a mess. The important thing is the end result, based on the Farnocchia et al. paper. Let me know if you have any comments about it.

mkbrewer:
No comments, just my question above.

mkbrewer:
@astrojuanlu An answer to this question would be appreciated when you have the time.

>From what I see here, the issues that you were seeing only occur when the true anomaly is close to 180 degrees. Is that correct?

astrojuanlu:
Hello @mkbrewer, I'm so sorry for the huge delay - I was in a bad moment when you asked, and then when I recovered I have been postponing it. Trying to get back on track.

> From what I see here, the issues that you were seeing only occur when the true anomaly is close to 180 degrees. Is that correct?

Yes, that's correct. The issues that we _were_ seeing in this algorithm were for the case where the true anomaly is close to 180 degrees. If you think about it, that condition is impossible to happen in true parabolic motion, and for near-parabolic motion, the closer we are to 180 the more evidence we have that the orbit is, well, elliptic. That's why in the chart by Farnocchia et al. the "near parabolic" region shrinks towards 180 degrees:

Screenshot from 2020-04-13 04-40-44

Fortunately, we managed to fix our implementation (it contained several conceptual errors) and now our `farnocchia` propagator can handle orbits with eccentricity arbitrarily close to 1.0, and do a smooth transition in the frontiers of the regions.

Hope this answers your question, please let me know otherwise.

PR #913, opened 2020-04-15
Propagation & angles refactor

astrojuanlu:
Departs from #908.

The objective is to use this starting point to work on some of the problems around the `propagation` and `angles` modules. I'll see how far can I go without breaking everything. If I am lucky, I will find the culprit of all the problems and end up absorbing #908. If not, at least I will try to move things around and find out what the next step should be.

### DIFF
--- a/src/poliastro/core/angles.py
+++ b/src/poliastro/core/angles.py
@@ -5,7 +5,7 @@ from ._jit import jit
 
 @jit
 def _kepler_equation(E, M, ecc):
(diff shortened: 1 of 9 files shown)


astrojuanlu:
On second thought, I prefer to have these uncontroversial changes on `master` as soon as possible, so people start noticing the alarming `FIXME: ` comments. Also, I don't like having too many open pull requests at the same time.

Issue #814, opened 2019-12-04
czmlExtractor: allow passing custom arguments to propagate


noc0lour:
Currently the propagator call in CZML limits use to only Keplerian orbits without artificial perturbations. For propagation of orbits with e.g. J2 perturbation it would be necessary to pass custom arguments

🎯 **Goal**

Create CZML packets out of different trajectory data.
I could think of a utility function which converts Coordinate data to CZML packets without doing the propagation itself since Coordinates have a obstime property attached anyhow.

Juanlu001:
Perhaps we should have a similar API than the `OrbitPlotter`s: `plot(orbit)` and then the Keplerian orbit is plotted, and `plot_trajectory(...)` and then the users chooses how to propagate.

noc0lour:
Sounds like a good idea. More or less a simple function which converts a coordinate time series to a CZML packet.

Syed-Osama-Hussain:
So basically we need to construct a packet by it's positional attributes with respect to time?

astrojuanlu:
@Syed-Osama-Hussain The `CZMLExtractor` class has an `add_orbit` method. Let's have an `add_trajectory` that takes a `CartesianRepresentation`, similar to `plot` vs `plot_trajectory` in the plotters.

Syed-Osama-Hussain:
Okay @astrojuanlu . Do we have to provide ground tracking and all the other optional parameters as in add_orbit?

astrojuanlu:
Let's start with `if groundtrack_show: raise NotImplementedError` to simplify.

Syed-Osama-Hussain:
Okay.

astrojuanlu:
The `add_trajectory` method addressed this issue: now users can propagate it "outside" of the `CZMLExtractor`. It would be nice to have an `add_ephem` too, but I'll let others open an issue for that if they really want it. Closing!

PR #915, opened 2020-04-16
add_trajectory and test added.

Syed-Osama-Hussain:
Attempt to solve #814

### DIFF
--- a/src/poliastro/czml/extract_czml.py
+++ b/src/poliastro/czml/extract_czml.py
@@ -70,7 +70,7 @@ class CZMLExtractor:
             projection.
         """
         self.packets = []  # type: List[Packet]
-        self.trajectories = [] #type: List[Any]
+        self.trajectories = []  # type: List[Any]
         self.attractor = attractor
         self.orbits = []  # type: List[Any]
         self.N = N
@@ -472,6 +472,7 @@ class CZMLExtractor:
     def add_trajectory(
         self,
         positions,
+        epochs,
         groundtrack_show=False,
         groundtrack_lead_time=None,
         groundtrack_trail_time=None,
@@ -495,10 +496,8 @@ class CZMLExtractor:
         ----------
         positions: ~astropy.coordinates.CartesianRepresentation
             Trajectory to plot.
-
-        Groundtrack parameters:
-        -----------------------
-
+        epochs: ~astropy.time.core.Time
+            Epochs for positions.
         groundtrack_show: bool
             If set to true, the groundtrack is
             displayed.
@@ -510,28 +509,16 @@ class CZMLExtractor:
             Groundtrack width
         groundtrack_color: list (int)
             Rgba groundtrack color. By default, it is set to the path color
-
-        Id parameters:
-        --------------
-
         id_name: str
             Set orbit name
         id_description: str
             Set orbit description
-
-        Path parameters
-        ---------------
-
         path_width: int
             Path width
         path_show: bool
             Indicates whether the path is visible
         path_color: list (int)
             Rgba path color
-
-        Label parameters
-        ----------
-
         label_fill_color: list (int)
             Fill Color in rgba format
         label_outline_color: list (int)
@@ -546,11 +533,27 @@ class CZMLExtractor:
         """
 
         if self.attractor is None:
-            raise ValueError(
-                "An attractor must be set up first."
-            )
+            raise ValueError("An attractor must be set up first.")
 
-        self.trajectories.append([positions, None, label_text,path_color])
+        positions = (
+            positions.represent_as(CartesianRepresentation).get_xyz(1).to(u.meter).value
+        )
+
+        epochs = Time(epochs, format="isot")
+
+        if len(epochs) != len(positions):
+            raise ValueError("Number of Points and Epochs must be equal.")
+
+        epochs = np.fromiter(
+            map(lambda epoch: (epoch - epochs[0]).to(u.second).value, epochs),
+            dtype=np.float,
+        )
+
+        positions = np.around(
+            np.concatenate([epochs[..., None], positions], axis=1).ravel(), 1
+        ).tolist()
+
+        self.trajectories.append([positions, None, label_text, path_color])
 
         start_epoch = Time(self.start_epoch, format="isot")
 
@@ -596,7 +599,9 @@ class CZMLExtractor:
         self.packets.append(pckt)
 
         if groundtrack_show:
-            raise NotImplementedError("Ground tracking for trajectory not implemented yet")
+            raise NotImplementedError(
+                "Ground tracking for trajectory not implemented yet"
+            )
 
         self.i += 1
 
--- a/src/poliastro/czml/extract_czml.py
+++ b/src/poliastro/czml/extract_czml.py
@@ -70,7 +70,7 @@ class CZMLExtractor:
             projection.
         """
         self.packets = []  # type: List[Packet]
-
+        self.trajectories = [] #type: List[Any]
         self.attractor = attractor
         self.orbits = []  # type: List[Any]
         self.N = N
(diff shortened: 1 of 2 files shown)


tests/test_czml.py
```diff
@@ -329,6 +329,117 @@ def test_czml_add_orbit():
...
 
+@pytest.mark.skipif("czml3" not in sys.modules, reason="requires czml3")
+def test_czml_add_trajectory():
+    start_epoch = iss.epoch
+    end_epoch = iss.epoch + molniya.period
+
+    sample_points = 10
+    color = [255, 255, 0]
+
+    positions = [0.0, 1.0, 2.0, 3.0, 60.0, 4.0, 5.0, 6.0, 120.0, 7.0, 8.0, 9.0]
```
astrojuanlu, line 340: Rather than specifying the positions directly like this, `add_trajectory` should receive a `CartesianRepresentation` with the positions, and an `epochs` list (an `astropy.time.Time` object with several values in it).

Syed-Osama-Hussain: Okay. You mean the format in the packet should be like this?
`"2012-04-30T12:00Z", 1.0, 2.0, 3.0,
  "2012-04-30T12:01Z", 4.0, 5.0, 6.0,
  "2012-04-30T12:02Z", 7.0, 8.0, 9.0`
I had used the seconds from start_epoch format earlier.

astrojuanlu: The format in the packet is correct as it is, but the user should not introduce the plain seconds as it is written now. The method should take care of the formatting.

Syed-Osama-Hussain: Okay.

Syed-Osama-Hussain: Should we also take care of the format ? Or we can trust the user to enter points in ```CartesianRepresentation``` and epochs in ```Time (isot)``` formats?
I think we should take care of that by casting them.

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+        """
+        Adds trajectory
+
+        Parameters
+        ----------
+        positions: ~astropy.coordinates.CartesianRepresentation
+            Trajectory to plot.
+        epochs: ~astropy.time.core.Time
+            Epochs for positions.
+        Groundtrack parameters:
```
astrojuanlu, line 526: Let's remove these intermediate headers, they are not rendered correctly in the documentation.

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+        groundtrack_lead_time: double
+            The time the animation is ahead of the real-time groundtrack
+        groundtrack_trail_time: double
+            The time the animation is behind the real-time groundtrack
+        groundtrack_width: int
+            Groundtrack width
+        groundtrack_color: list (int)
+            Rgba groundtrack color. By default, it is set to the path color
+
+        Id parameters:
```
astrojuanlu, line 541: See above.

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+        --------------
+
+        id_name: str
+            Set orbit name
+        id_description: str
+            Set orbit description
+
+        Path parameters
```
astrojuanlu, line 549: See above.

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+        ---------------
+
+        path_width: int
+            Path width
+        path_show: bool
+            Indicates whether the path is visible
+        path_color: list (int)
+            Rgba path color
+
+        Label parameters
```
astrojuanlu, line 559: See above.

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+        label_text: str
+            Set label text
+        label_show: bool
+            Indicates whether the label is visible
+
+        """
+        positions = (
+            positions.represent_as(CartesianRepresentation).get_xyz(1).to(u.meter).value
+        )
+        epochs = Time(epochs, format="isot")
```
astrojuanlu, line 577: Can you try `epochs.isot` instead?

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+
+        if epochs.__len__() != positions.__len__():
```
astrojuanlu, line 579: Let's use `len(epochs)` instead.

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+            position = np.insert(positions[i], 0, epoch)
+
+            if positionList.__len__() > 0:
+                positionList = np.concatenate((positionList, position))
+            else:
+                positionList = position
+
+        positionList = np.around(positionList, 1).tolist()
+
+        if self.attractor is None:
```
astrojuanlu, line 595: Could we move this to the top? So we don't do all the calculations if we are about to throw an error.

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+        )
+        epochs = Time(epochs, format="isot")
+
+        if epochs.__len__() != positions.__len__():
+            raise ValueError("Number of Points and Epochs must be equal.")
+
+        positionList = np.array([])
+
+        for i in range(positions.__len__()):
+            epoch = (epochs[i] - epochs[0]).to(u.second).value
```
astrojuanlu, line 585: I wish there was a way to avoid this `for` loop... If we are doing `epochs[i] - epochs[0]` every time, isn't it easier to store `epochs - epochs[0]` somewhere?

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+            positions.represent_as(CartesianRepresentation).get_xyz(1).to(u.meter).value
+        )
+        epochs = Time(epochs, format="isot")
+
+        if epochs.__len__() != positions.__len__():
+            raise ValueError("Number of Points and Epochs must be equal.")
+
+        positionList = np.array([])
+
+        for i in range(positions.__len__()):
```
astrojuanlu, line 584: `len(positions)`

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+            epoch = (epochs[i] - epochs[0]).to(u.second).value
+            position = np.insert(positions[i], 0, epoch)
+
+            if positionList.__len__() > 0:
```
astrojuanlu, line 588: `len(positionList)`

src/poliastro/czml/extract_czml.py
```diff
@@ -494,5 +494,158 @@ def add_orbit(
...
+                positionList = np.concatenate((positionList, position))
+            else:
+                positionList = position
+
+        positionList = np.around(positionList, 1).tolist()
```
astrojuanlu, line 593: Is this to reduce the number of decimal places?

tests/test_czml.py
```diff
@@ -329,6 +331,123 @@ def test_czml_add_orbit():
     assert repr(extractor.packets) == expected_doc
 
 
+@pytest.mark.skipif("czml3" not in sys.modules, reason="requires czml3")
+def test_czml_add_trajectory():
```
astrojuanlu, line 335: Nice! I look how the test looks now

src/poliastro/czml/extract_czml.py
Syed-Osama-Hussain: Okay.

src/poliastro/czml/extract_czml.py
Syed-Osama-Hussain: Yes

src/poliastro/czml/extract_czml.py
Syed-Osama-Hussain: The ```for``` loop was mainly to append the epochs and positions into a single array. A vectorized utility function should work maybe for epoch calculation?

src/poliastro/czml/extract_czml.py
Syed-Osama-Hussain: Okay.

src/poliastro/czml/extract_czml.py
Syed-Osama-Hussain: @astrojuanlu I have used a ```lambda``` for epoch calculation and ```zip``` to arrange epochs and positions. How does this sound?

Syed-Osama-Hussain:
I tried epochs.isot but it doesn't work.

Syed-Osama-Hussain:
@astrojuanlu is any other improvement needed? Awaiting review.

Syed-Osama-Hussain:
Test failing due to a numba_config import. Don't know exactly why it was imported.

astrojuanlu:
Merged, thanks @Syed-Osama-Hussain! :clap:

Issue #925, opened 2020-04-20
Tisserand plots


jorgepiloto:
📕 **About Tisserand plots**

Imagine you are observing asteroids which are orbiting near Jupiter. The big
planet might create perturbations among these small bodies orbits but you want
to make sure if an actual observed body was previously observed, althought their
COE parameters might be suffered some variations.

Previous question arised to Félix Tisserand, who was a French astronomer that in
the 19th century developed a method based on an invariant constant (which today
carries his name). You can understand the tisserand parameters as some way to
measure the "energy" of a body. In fact, its units are *distance2 / time2* which
is also known as "characteristic energy".

Nowadays, Tisserand graphs can be plotted in many different ways:

* Tisserand parameter VS apoapsis || eccentricity || inclination
* Heliocentric periapsis VS Heliocentric apoapsis
* Heliocentric orbital period VS Heliocentric apoapsis

The most common ones are the last two, though the first one might be plotted by
making use of a contourf plot as seen in this page:

* [All Tisserand Solutions by contourf plots](https://armchairastronautics.blogspot.com/search?q=tisserand)

The most powerful feature this kind of plots show is related to mission design,
in particular to gravity assist maneuvers. They would lead to the solution to
the question: if I want to go from body A to body B, what would be the most
optimal set of flybys to achieve the trip?

![poliastro_tiss](https://user-images.githubusercontent.com/28702884/79787422-40b8f000-8347-11ea-8285-d8d8cba2fe39.png)

💡 **Feature request**

The algorithm on how to obtain these kind of plots is very well explained in the
following [document](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=22&ved=2ahUKEwje9d2ZzvfoAhUQLBoKHeYXAbMQFjAVegQIBhAB&url=https%3A%2F%2Ffenix.tecnico.ulisboa.pt%2FdownloadFile%2F1407770020545591%2FMSc_DavidPalma_Final_13Dec.pdf&usg=AOvVaw2-9skb6UG2tix-UieDnAFT). Furthermore, it includes lots of solved cases which could be used as unit tests.

With the implementation of the function `get_mean_elements()` it is now possible
to obtain the necessary parameters for any desired body.

🎯 **Goal**

Although the working algorithm is pretty clear, the problem might arise on how
to implement this feature inside *poliastro*. I am thinking about the following
usage:

```python
# File poliastro/plotting/tisserand.py

class Tisserand:
[...]

PR #927, opened 2020-04-22
Tisserand plotting utilities

jorgepiloto:
Solves for #925 as suggested from feedback. Opening this as draft since the following tasks still require to be solved:

- [x] Solve for small plotting bugs regarding upper bounding limits
- [x] Add proper image tests
- [x] Document this module and add notebook examples


src/poliastro/plotting/tisserand.py
```diff
@@ -0,0 +1,142 @@
...
+    # Generate Tisserand
+    tp = TisserandPlotter()
+
+    for ss in SS_BODIES:
+
+        # This example works fine
+        tp.plot(ss, (1, 10) * u.km / u.s, num_contours=10)
+
+        # TODO: BUG APPEARS IF THE UPPER BOUND FOR VINF > 10 KM/S
+        #tp.plot(ss, (1, 11) * u.km / u.s, num_contours=11)
```
astrojuanlu, line 139: I can reproduce the issue. It looks like there's an x value going to 0, or similar.

astrojuanlu:
I think the problem is this:

![Screenshot from 2020-04-27 00-52-35](https://user-images.githubusercontent.com/316517/80321912-6d767700-8821-11ea-893a-9d77b82c84bd.png)

For a given value of α, the semimajor axis starts positive (elliptic orbit), then tends to infinity when `1 - v_∞^2 - 2 v_∞ cos(α)` approaches zero, then comes from minus infinity (hyperbolic orbit). I think it does not make sense to go past the "limit" v_∞, which for α = 0 is, in non-dimensional variables, is:

```
In [8]: solve(Eq(x ** 2 + 2 * x - 1, 0))                                                                                                                                                                                                      
Out[8]: [-1 + sqrt(2), -sqrt(2) - 1]

In [9]: N(_[0])                                                                                                                                                                                                                               
Out[9]: 0.414213562373095
```

Or around ~11 km / s :wink:

jorgepiloto:
After a little bit of testing, I found that for angle values near zero, the plot
starts to show this behaviour. However, the original paper claims that the
method is valid for any angle within [0, pi]. Hope to continue working on this
and discover what is going on...

jorgepiloto:
I see... But then, how is it possible for the author of the paper to plot the lines for v_inf=11km without running into singularities? I also noticed small differences between the values obtained with this and the ones from the paper but my reasoning is that the values for the position and velocity of the planets are just "slightly different".

astrojuanlu:
> But then, how is it possible for the author of the paper to plot the lines for v_inf=11km without running into singularities? 

Which paper exactly? I'm looking at figure 3.4 from David Palma thesis and it goes up to 9 km/s.

jorgepiloto:
Ops! Sorry for this, I should have posted a pic of the figure I am looking at. All credits go to David Falcato Fialho Palma:

tisserand

Previous figure is located in the appendix A together with other tisserand images.

astrojuanlu:
Thinking about this again. The text confirms that hyperbolic orbits are out of scope:

> Tisserand graphs come in many shapes, depending on their purpose, but the one that is most useful to us is the apoapsis vs periapsis (R a vs R p ) formulation (see figure 3.4). It allows for a clearer vision of the spacecraft orbit, despite not being suitable for representing hyperbolic trajectories but these won’t be studied in this work.

I wonder if you are right:

> my reasoning is that the values for the position and velocity of the planets are just "slightly different".

jorgepiloto:
As discussed privately, there seems to be some negative value coming from the computation of the `a_sc` in which the relationship between `V_inf` and `cos(alpha)` returns a negative value. By forcing the denominator to be always positive with `np.abs()` the issue seems to be solved.

However, previous "hack" requires further investigation. In addition, although some figures look like the ones from different papers, values are close but not the same.

jorgepiloto:
I have been working on this during the last days' but got stuck while thinking about the following: imagine as user I want to plot all the three main Tisserand configurations (`E_vs_rp`, `T_vs_rp` and `rp_vs_ra`). I thought about the possibility of having three internal axes, one for each kind of plot, but although this works fine, all of them are displayed one created...

Algorithm seems to be working fine and outputs same lines for different examples found in academical papers:

<p float="left">
  
   
</p>

astrojuanlu:
Sorry for the long delay reviewing this! I will leave some comments. In the meanwhile, this will also need a rebase.

src/poliastro/plotting/tisserand.py
```diff
@@ -0,0 +1,214 @@
...
+        # Tisserand kind collection
+        self.type_collection = ["energy", "period", "apsis"]
+
+        # Prellocate internal data lines
+        self.energy_lines = []
+        self.period_lines = []
+        self.apsis_lines = []
+
+    @property
+    def rp_vs_ra(self):
```
astrojuanlu, line 40: I suggest removing all the logic that mutates the axes, and instead offering something like this:

```
tisserand = TisserandPlotter()
tisserand = TisserandPlotter(kind=TisserandKind.ENERGY)  # Same as above, default kind
tisserand = TisserandPlotter(kind=TisserandKind.PERIOD)
tisserand = TisserandPlotter(kind=TisserandKind.APSIS)
```

How does that sound @jorgepiloto?

jorgepiloto: I tried this implementation and really like it! Let me push new changes before cleaning git commit history :+1:

jorgepiloto:
Sorry for this late contribution. I just added some new changes:

* A new `plot_line` method for better showing the final desired gravity assist path. This would be the "`plot_trajectory`" but in Tisserand figures.
* Documented the module by adding it to the API reference and including a jupytext notebook example.

astrojuanlu:
Needs `tox -e reformat`

astrojuanlu:
Both tests failures come from our dear `rv2coe` docstring. Merging!

Issue #935, opened 2020-04-28
RFC: atmospheric_drag API change


wumpus:
At the moment we have 2 functions for atmospheric_drag: `atmospheric_drag()`, which uses a naive exponential atmosphere, and `atmospheric_drag_model()`, which uses more sophisticated models from `poliastro.atmosphere`.

I'd like to rename `atmospheric_drag` to `atmospheric_drag_exponential` so that users are less likely to think that this routine is a good default.

While I'm in there, these routines take a separate `A` (area) and `m` (mass) parameters. The literature uses `A/m`. I'd like to change the calling signature to take A_over_m instead of two separate arguments.

cc @iamabhishek0

wumpus:
While I'm here I should note that it's very confusing that the user calls atmospheric_drag() with kw arguments but the function doesn't have any kw arguments. I know python 3.6+ preserves the order of kwargs but it sure seems fragile. @astrojuanlu

astrojuanlu:
The first idea is totally reasonable to me, let's follow up in #936 now that @priyanshurohilla started the PR.

Regarding your second comment, I don't quite see what are you referring to - could you write an example?

wumpus:
For the second think about

```
    rr, _ = cowell(
        Earth.k,
        orbit.r,
        orbit.v,
        [tof] * u.s,
        ad=atmospheric_drag_exponential,
        R=R,
        C_D=C_D,
        A_over_m=A_over_m,
        H0=H0,
	rho0=rho0,
    )
```

What is the function signature of `atmospheric_drag_exponential`? It's

```
def atmospheric_drag_exponential(t0, state, k, R, C_D, A_over_m, H0, rho0):
```

and it's called as

```
    ax, ay, az = ad(t0, u_, k, **ad_kwargs)
```

... so in short, it's turning kwargs into named args. And it depends on the order of the kwargs being correct. Python programmers aren't really used to having to pay attention to the order of kwargs when calling functions like `cowell()`.

wumpus:
We either need to document this everywhere and still expect user confusion, or, change all of the ad() functions to have kwargs instead of args.

astrojuanlu:
Very true, now I see the issue. What about

```
def atmospheric_drag_exponential(t0, state, k, *, R, C_D, A_over_m, H0, rho0):
```

and in general have every perturbation function use `*` for everything after `(t0, state, k)`. (Notice that [this won't be respected in numba though](https://github.com/numba/numba/issues/5655)).

[...]

Issue #938, opened 2020-04-29
Interface infelicity: third_body()


wumpus:
`poliastro.core.perturbations` has a method `third_body` that has a kw argument named `third_body`.

Think for a moment what that means when you try to wrap the function:
```
from poliastro.core.perturbations import third_body
def wrapper(**kwargs):
   foo = third_body(..., third_body)
```

jorgepiloto:
Wow! Good catch @wumpus :rocket: 

What about `perturbation_body` as function parameter? :thinking:

wumpus:
That's a great name.

astrojuanlu:
"infelicity" made me chuckle :) Agreed with @jorgepiloto suggestion

priyanshurohilla:
I have fixed this issue. Just waiting for my previous PR #940 to merge and then i'll open a PR for this:)

priyanshurohilla:
@astrojuanlu  This can be closed now:)

PR #940, opened 2020-04-29
fixed drag api

priyanshurohilla:
Again opened for #935


astrojuanlu:
@wumpus If you approve, I will merge this one :)

tests/tests_twobody/test_perturbations.py
```diff
@@ -191,9 +191,8 @@ def test_atmospheric_drag():
 
     # parameters of a body
     C_D = 2.2  # dimentionless (any value would do)
-    A = ((np.pi / 4.0) * (u.m ** 2)).to(u.km ** 2).value  # km^2
-    m = 100  # kg
-    B = C_D * A / m
+    A_over_m = ((np.pi / 4.0) * (u.m ** 2)).to(u.km ** 2).value / 100  # km^2/kg
```
wumpus, line 194: These units are missing u.kg -- won't change the test result, but a future reader might be confused.

tests/tests_twobody/test_perturbations.py
```diff
@@ -234,8 +232,7 @@ def test_atmospheric_demise():
 
     # parameters of a body
     C_D = 2.2  # dimentionless (any value would do)
-    A = ((np.pi / 4.0) * (u.m ** 2)).to(u.km ** 2).value  # km^2
-    m = 100  # kg
+    A_over_m = ((np.pi / 4.0) * (u.m ** 2)).to(u.km ** 2).value / 100  # km^2/kg
```
wumpus, line 235: units, again

tests/tests_twobody/test_perturbations.py
```diff
@@ -277,8 +273,7 @@ def test_atmospheric_demise_coesa76():
...
```
wumpus, line 276: units

priyanshurohilla:
For time being it would be better to merge this PR. I will make these necessary changes in the next PR which i will be doing for #938 as these issues are similar and will be affecting the same files

astrojuanlu:
> For time being it would be better to merge this PR. I will make these necessary changes in the next PR which i will be doing for #938 as these issues are similar and will be affecting the same files.

What about doing all the changes here instead?

priyanshurohilla:
> > For time being it would be better to merge this PR. I will make these necessary changes in the next PR which i will be doing for #938 as these issues are similar and will be affecting the same files.
> 
> What about doing all the changes here instead?

@astrojuanlu  That would be great 🚀 but I would like to keep them separate as i have already fixed that issue taking `poliastro/master` as the base repository :)

astrojuanlu:
@priyanshurohilla I don't quite understand if the problem is in your git setup and I am not entirely comfortable merging a PR where there are some reasonable improvements to be made on it, but on the other hand I don't think discussing it any longer is worth the keystrokes :) The PR is OK, so I merged it. If you have any problems with git, please let us know in the appropriate channels so we can guide you without affecting the workflow of the project. Thanks!

priyanshurohilla:
> @priyanshurohilla I don't quite understand if the problem is in your git setup and I am not entirely comfortable merging a PR where there are some reasonable improvements to be made on it, but on the other hand I don't think discussing it any longer is worth the keystrokes :) The PR is OK, so I merged it. If you have any problems with git, please let us know in the appropriate channels so we can guide you without affecting the workflow of the project. Thanks!

Actually I was unable to make you understand😅. The problem is not with the git setup, but with the way i approached it to solve the issue. I also myself don't want the PR to be merged if there are some improvements to be made. Anyways, whatever i did, i have learnt a lot now from my mistakes, so from next time onwards this thing won't happen😊. 
 
And yeah the next PR is coming soon:) 🚀.

PR #943, opened 2020-04-29
Changed third_body function parameter and added u.kg wherever applicable

priyanshurohilla:
Addresses #938

### DIFF
--- a/tests/tests_twobody/test_perturbations.py
+++ b/tests/tests_twobody/test_perturbations.py
@@ -191,7 +191,9 @@ def test_atmospheric_drag_exponential():
 
     # parameters of a body
     C_D = 2.2  # dimentionless (any value would do)
-    A_over_m = ((np.pi / 4.0) * (u.m ** 2)).to(u.km ** 2).value / 100 * u.kg  # km^2/kg
+    A_over_m = ((np.pi / 4.0) * (u.m ** 2) / (100 * u.kg)).to_value(
+        u.km ** 2 / u.kg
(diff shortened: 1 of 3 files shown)


tests/tests_twobody/test_perturbations.py
```diff
@@ -191,7 +191,7 @@ def test_atmospheric_drag_exponential():
 
     # parameters of a body
     C_D = 2.2  # dimentionless (any value would do)
-    A_over_m = ((np.pi / 4.0) * (u.m ** 2)).to(u.km ** 2).value / 100  # km^2/kg
+    A_over_m = ((np.pi / 4.0) * (u.m ** 2)).to(u.km ** 2).value / 100 * u.kg  # km^2/kg
```
astrojuanlu, line 194: Isn't it better to do something like

```python
A_over_m = ((np.pi / 4 * u.m ** 2) / (100 * u.kg)).to_value(u.km ** 2 / u.kg)
```

? (This applies everywhere)

priyanshurohilla: Yes i was also doing this only. Let me see if it works or not

priyanshurohilla:
Why are the tests failing now @astrojuanlu?

astrojuanlu:
The tests aren't failing - the coverage check is failing. There is a slightly larger number of lines and no new tests, therefore the coverage decreased. However, I am not very strict with these things, so I am merging strainght away. Thanks @priyanshurohilla!

Issue #497, opened 2018-11-20
Ground Track Plotter


jorgepiloto:
🎯 **Objective**
Hi everyone,

I thought it would be interesting to include a "Ground Track Plotter" in the module of `poliastro.plotting`. It works in the same way as the other functions of this module: you just have to call a function called `groundTrack`that takes as arguments an `orbit`, `label`with the possibility of using the `dark` mode. 

🖥 Previous figure tries to plot the following orbit:
```bash
#PLOTING GROUND TRACK
from poliastro.plotting import groundTrack 

r = [-6045, -3490, 2500] * u.km
v = [-3.457, 6.618, 2.533] * u.km / u.s
ss = Orbit.from_vectors(Earth, r, v)

groundTrack(ss, dark=True)
plt.show()

```

Till here everything works as it should. However,  once I pass the function an `Orbit`object, I obtain really strange ground tracks:

![pic](https://user-images.githubusercontent.com/28702884/48805189-fe851c80-ed16-11e8-9148-933c98b3e4e2.png)

Notice that horizontal dashed lines are just for plotting Cancer and Capricorn Tropics.

💡 **Possible solutions**

If the idea is interesting for `poliastro` community, I will make a pull request but probably will need some help with the conversions between reference systems. I think the problem may be here, since if you try to plot just a latitude and longitude it works perfectly, showing that exact point on map.

Cape Canaveral Coordinates
lat = 28.396837
lon = -80.605659

![plot](https://user-images.githubusercontent.com/28702884/48806265-11e5b700-ed1a-11e8-9b4b-2790234ab14b.png)

Juanlu001:
Hi @IngenieroDeAviones, thanks a lot for the suggestion! I recommend you to check out this notebook of mine:

http://nbviewer.jupyter.org/github/Juanlu001/catching-asteroids-tracking-satellites/blob/master/Talk.ipynb#Plot-groundtrack

In particular, Cartopy will help you a lot projecting lat/lon data in the right way, avoiding all sorts of singularities.

Also, avoid basemap because it's abandoned: https://matplotlib.org/basemap/users/intro.html#cartopy-new-management-and-eol-announcement

And finally, to create a really cool 3D version:

https://plot.ly/python/lines-on-maps/#contour-lines-on-globe

IngenieroDeAviones:
[...]

PR #953, opened 2020-05-03
Added groundtrack plotting functionality

priyanshurohilla:
Addresses #497 

I have added a groundtrack plotting functionality to poliastro.

Opening this as a draft because if maintainers want me to add some more features, then i can do that

- [x] Added the groundtrack plotter

- [x] Added some test cases

- [ ] Add some new test cases if any

- [ ] Add some new features if any 

- [ ] Add this feature to docs

Also i would like to thank @jorgepiloto and @noc0lour for mentoring me:)


priyanshurohilla:
@astrojuanlu Any comments on the tests?

src/poliastro/plotting/groundtrack.py
```diff
@@ -0,0 +1,143 @@
...
+        time_values = TimeDelta(np.arange(1, values) * tof / values * u.h)
+    else:
+        time_values = TimeDelta(np.arange(1, values) * orbit.period / values)
+
+    orbit_gcrs = propagation.propagate(orbit, time_values)
+
+    time = {}
+    time = orbit.epoch + time_values
+
+    # Conversion of GCRS into ITRS
```
jorgepiloto, line 104: I think this would be more readable if we split both GCRS and ITRS positions:

```python
orbit_gcrs = GCRS(orbit_gcrs, obstime=time, representation_type=CartesianRepresentation)
orbit_itrs = orbit_gcrs.transform_to(ITRS(obstime=time))
```

src/poliastro/plotting/groundtrack.py
```diff
@@ -0,0 +1,143 @@
...
+            mode="lines",
+            name=label,
+            showlegend=True,
+            line=dict(width=width, color=color,),
+        )
+    )
+
+    fig.update_layout(
+        title_text=title,
+        geo=dict(
```
jorgepiloto, line 127: I am conscious about the fact that we should provide user the highest degree of plotting customization but don't think passing such a bunch of arguments is the best way to solve this problem. From my point of view, we should pass a dictionary under the name `geo` so user can keep all the plotly plotting parameters.

Previous suggestion might become simpler if moving this to a class under the name "GroundtrackPlotter". This could lead to the possibility of having several groundtracks in the same figure, showing the same behavior as the rest of the API plotters.

src/poliastro/plotting/groundtrack.py
```diff
@@ -0,0 +1,143 @@
...
+    CartesianRepresentation,
+    SphericalRepresentation,
+)
+from astropy.time import TimeDelta
+
+from poliastro.examples import iss
+from poliastro.twobody import propagation
+
+
+def groundtrack(
```
jorgepiloto, line 17: This huge amount of function parameters can be avoided while keeping the same customization degree. Many of those variables end up being the keys of a dictionary one (which is the idea behind all plotly plotters). Check out below comments for a deep explanation.

Furthermore, in the original issue:

* What about if a user wants to know where was a satellite some time ago? Is it possible to pass a negative time of flight value and get the old groundtrack?

jorgepiloto:
Thank you for your time implementing this @priyanshurohilla, but finally this was merged in #1004. Since the opening of this pull request a new module under the name `poliastro.earth` was added and we decided to add this functionality to it. Best regards! :rocket:

PR #954, opened 2020-05-07
Upgrade docs theme

astrojuanlu:
Will this be the solution to all our problems????


astrojuanlu:
Yesssss the Plotly issue is gone :heart_eyes: We no longer need a for of the RTD Sphinx Theme, and also the typographies of the headings are fixed :heart:

Issue #959, opened 2020-05-13
Intermittent CI failures on Windows


astrojuanlu:
* [Small numerical differences](https://dev.azure.com/poliastro/poliastro/_build/results?buildId=325&view=logs&j=f32fc568-3d58-5c2f-5c7e-c01e8366ccfc&t=4340fcf2-110e-5f6e-b978-eb43f9f7ccae&l=1059)
* [Tcl installation issues](https://dev.azure.com/poliastro/poliastro/_build/results?buildId=324&view=logs&j=60a3625c-55b6-532a-bb2b-3a6a478dd649&t=36931b45-8a73-5a08-d711-6c01b0fd09e2&l=1121)

astrojuanlu:
The doctest failure happens also on Linux: https://dev.azure.com/poliastro/poliastro/_build/results?buildId=387&view=logs&j=1117f030-3e3a-5822-cfd7-97e776c1f4cb&t=af77bb0c-2bf7-50ca-9891-2682490c119e&l=81

```
___________________ [doctest] poliastro.core.elements.rv2coe ___________________
307     --------
308     >>> from poliastro.constants import GM_earth
309     >>> from astropy import units as u
310     >>> k = GM_earth.to(u.km ** 3 / u.s ** 2).value  # Earth gravitational parameter
311     >>> r = np.array([-6045., -3490., 2500.])
312     >>> v = np.array([-3.457, 6.618, 2.533])
313     >>> p, ecc, inc, raan, argp, nu = rv2coe(k, r, v)
314     >>> print("p:", p, "[km]")
315     p: 8530.47436396927 [km]
316     >>> print("ecc:", ecc)
Expected:
    ecc: 0.17121118195416898
Got:
    ecc: 0.17121118195416907
```

astrojuanlu:
Looks like this one was fixed!

astrojuanlu:
This still happens. Reopening and pinning.

PR #963, opened 2020-05-19
Mark few more tests as remote

olebole:
The `test_plot_ephem_epoch()` and `test_plot_ephem_no_epoch()` actually cause a remote access to download the ephemerides. Without the mark, this causes a failure on systems that have no connection (like in Debian during build time).

This PR adds the required markers for these tests.


astrojuanlu:
Thanks @olebole! The Windows failure is a case of #959, so I'm merging this.

Issue #968, opened 2020-05-23
Tests against Horizons still too fragile


astrojuanlu:
Despite the work done in #867, tests against Horizons still give occasional numerical differences. See for example:

https://dev.azure.com/poliastro/poliastro/_build/results?buildId=337&view=logs&j=538a0c56-c006-575f-81ef-6a7057413c77&t=b32c51f5-9aff-5da6-8e1a-744748933fbc&l=79
https://dev.azure.com/poliastro/poliastro/_build/results?buildId=337&view=logs&j=538a0c56-c006-575f-81ef-6a7057413c77&t=b32c51f5-9aff-5da6-8e1a-744748933fbc&l=104

We should abandon this approach and mock the response. Perhaps we can still have a remote test, but definitely with much larger tolerances.

astrojuanlu:
Closed in #969

PR #969, opened 2020-05-26
Fix for the flaky Horizon Test

Meuge:
There was an error with the epoch.
This addresses the issue #968

### DIFF
--- a/tests/tests_twobody/test_orbit.py
+++ b/tests/tests_twobody/test_orbit.py
@@ -1046,6 +1046,7 @@ def test_from_sbdb_and_from_horizons_give_similar_results(target_name):
             test_elm, ref_elm, rtol=1e-3
         )  # Maximum error of 0.1% (chosen arbitrarily)
 
+
 def test_propagate_to_anomaly_gives_expected_result():
     # From "Going to Jupiter with Python using Jupyter and poliastro.ipynb"
     ic1 = Orbit.from_vectors(
--- a/tests/tests_twobody/test_orbit.py
+++ b/tests/tests_twobody/test_orbit.py
@@ -1034,21 +1034,17 @@ def test_from_coord_if_coord_is_not_of_shape_zero():
 def test_from_sbdb_and_from_horizons_give_similar_results(target_name):
     ss_target = Orbit.from_sbdb(target_name)
     ss_classical = ss_target.classical()
-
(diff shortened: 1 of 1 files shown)


tests/tests_twobody/test_orbit.py
```diff
@@ -1034,21 +1034,17 @@ def test_from_coord_if_coord_is_not_of_shape_zero():
 def test_from_sbdb_and_from_horizons_give_similar_results(target_name):
     ss_target = Orbit.from_sbdb(target_name)
     ss_classical = ss_target.classical()
-
-    ss_ref = Orbit.from_horizons(
-        name=target_name, attractor=Sun, plane=Planes.EARTH_ECLIPTIC
-    )
-
-    ss_ref = ss_ref.propagate_to_anomaly(
```
jorgepiloto, line 1042: Which was the aim of propagating the orbit in the old version? I mean, the method returns the orbit with that specific true anomaly, isn't it?

Meuge: What I understood, this test was to double check that the data from sddb and from horizon gave the same results. In order to do so, I send the same epoch in both methods and the result was positive.

tests/tests_twobody/test_orbit.py
```diff
@@ -629,8 +629,8 @@ def test_orbit_from_horizons_has_expected_elements():
     )
     ss1 = Orbit.from_horizons(name="Ceres", attractor=Sun, epoch=epoch)
     assert ss.pqw()[0].value.all() == ss1.pqw()[0].value.all()
-    assert ss.r_a == ss1.r_a
-    assert ss.a == ss1.a
+    assert_quantity_allclose(ss.r_a, ss1.r_a, rtol=1.0e-4)
+    assert_quantity_allclose(ss.a, ss1.a, rtol=1.0e-4)
```
jorgepiloto, line 633: This assertion looks good! Good job @Meuge :rocket:

tests/tests_twobody/test_orbit.py
astrojuanlu: > What I understood, this test was to double check that the data from sddb and from horizon gave the same results. In order to do so, I send the same epoch in both methods and the result was positive.

Yes, that was the idea of #867!

PR #970, opened 2020-06-11
Earth Satellite Implementation

Meuge:


src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+
+            Returns
+            -------
+            Orbit
+                New 'Orbit' position and velocity vectors.
+
+
+        """
+
+        orbit = Orbit.from_vectors(Earth, r, v)
```
Meuge, line 64: What do you think about using static method instead of using a @classmethod ? The same applies to from_classical

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+            at a given time (epoch).
+
+        Raises
+        ------
+        TypeError
+            If the orbit is not an Orbit's instance.
+
+        """
+
+        if not isinstance(orbit, Orbit):
```
astrojuanlu, line 34: Nahhhh, duck typing :) Also, we might introduce gradual typing in the future to avoid these kind of errors

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+            v : ~astropy.units.Quantity
+                   Velocity vector.
+
+            Returns
+            -------
+            Orbit
+                New 'Orbit' position and velocity vectors.
+
+
+        """
```
astrojuanlu, line 62: Misalignment of the docstring?

```suggestion
        """Return `Orbit` from position and velocity vectors.

        Parameters
        ----------
        r : ~astropy.units.
            Position vector wrt attractor center.
        v : ~astropy.units.Quantity
            Velocity vector.

        Returns
        -------
        Orbit
            New 'Orbit' position and velocity vectors.

        """
```

src/poliastro/earth/__init__.py
astrojuanlu: Hmm it's weird because we are using `poliastro.earth.EarthSatellite` as some sort of "factory" instead of returning a proper `EarthSatellite` object. Remember that now we only have one property, but in the future we might add metadata, satellite mass, orientation...

Let's change this to be a `@classmethod`

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        orbit = Orbit.from_vectors(Earth, r, v)
+        return orbit
+
+    @classmethod
+    @u.quantity_input(a=u.m, ecc=u.one, inc=u.rad, raan=u.rad, argp=u.rad, nu=u.rad)
+    def from_classical(
+        cls, a, ecc, inc, raan, argp, nu,
+    ):
+
+        """Return ` Orbit` from classical orbital elements.
```
astrojuanlu, line 73: Same misalignment as above

Also:
```suggestion
        """Return `EarthSatellite` from classical orbital elements.
```

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+
+            Returns
+            -------
+            Orbit
+                New 'Orbit' from classical orbital elements.
+
+            """
+
+        orbit = Orbit.from_classical(Earth, a, ecc, inc, raan, argp, nu)
+        return orbit
```
astrojuanlu, line 98: Similar to what we had above, we should return `cls(orbit)`

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def propagate(self, tof, *, atmosphere=None, gravity=None):
+
```
astrojuanlu, line 101:
```suggestion
```

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+        """Propagates an Earth orbit a specified time.
+
+            If value is true anomaly, propagate orbit to this anomaly and return the result.
```
astrojuanlu, line 104: Same misalignment as above

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+                New 'Orbit' from classical orbital elements.
+
+            """
+
+        orbit = Orbit.from_classical(Earth, a, ecc, inc, raan, argp, nu)
+        return orbit
+
+    def propagate(self, tof, *, atmosphere=None, gravity=None):
+
+        """Propagates an Earth orbit a specified time.
```
astrojuanlu, line 102: Let's be consistent:

```suggestion
        """Propagates an Earth satellite a specified time.
```

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+            -------
+            Orbit
+                New 'Orbit' from classical orbital elements.
+
+            """
+
+        orbit = Orbit.from_classical(Earth, a, ecc, inc, raan, argp, nu)
+        return orbit
+
+    def propagate(self, tof, *, atmosphere=None, gravity=None):
```
astrojuanlu, line 100: Even if we don't have gravity models yet, let's introduce something light, like an `Enum` with two possible values: `SPHERICAL` and `J2`. If `SPHERICAL` we don't do anything, if `J2` we take into account `Earth.J2`. How does that sound?

src/poliastro/earth/__init__.py
astrojuanlu: However! We can check if `orbit.attractor is Earth`, that would be great

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+        -------
+        Orbit
+            New 'Orbit' from classical orbital elements.
+
+        """
+
+        orbit = Orbit.from_classical(Earth, a, ecc, inc, raan, argp, nu)
+        return cls(orbit)
+
+    def propagate(self, tof, *, atmosphere=None, perturbation=None):
```
astrojuanlu, line 99: Could we call this `gravity` again? Otherwise I think it clashes with `atmosphere`

src/poliastro/earth/enums.py
```diff
@@ -0,0 +1,10 @@
+""" Earth Satellite perturbations.
+
+"""
+
+from enum import Enum, auto
+
+
+class EarthSatellitePerturbations(Enum):
```
astrojuanlu, line 8: And this could be called `EarthGravity` then

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+                    J2=Earth.J2.value,
+                    R=Earth.R.to(u.km).value,
+                )
+            return self.orbit.propagate(
+                tof,
+                method=cowell,
+                ad=J2_perturbation,
+                J2=Earth.J2.value,
+                R=Earth.R.to(u.km).value,
+            )
```
astrojuanlu, line 141: I propose that we rework the logic of this by starting with an empty `kwargs = {}` and an empty list of perturbation functions, and then adding stuff if we got propagations. Something like:

```python
kwargs = {}
perturbations = []
if gravity is EarthGravity.J2:
    kwargs.update(J2=Earth.J2.value, ...)
    perturbations.append(J2_perturbation)

if atmosphere is not None:
    kwargs.update(...)
    perturbations.append(partial(atmospheric_drag_model), model=atmosphere)

```

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+        return self.orbit.propagate(tof, method=cowell)
```
astrojuanlu, line 142: I think we should return a new `EarthSatellite` with the propagated `Orbit`, instead of an `Orbit`. With the comment above, we could do

```python
def ad(t0, u, k):
    return sum(f(t0, u, k) for f in perturbations)

new_orbit = self.orbit.propagate(tof, method=cowell, ad=ad, **kwargs)
```

astrojuanlu:
By the way, this will need a rebase :arrow_forward:

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+            at a given time (epoch).
+
+        Raises
+        ------
+        ValueError
+            If the orbit's attractor is not Earth
+
+        """
+
+        if orbit.attractor.name != "Earth":
```
astrojuanlu, line 36: Perhaps we can check for identity instead? `if orbit.attractor is not Earth`

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+            parameters used in perturbation models.
+
+        Returns
+        -------
+        EarthSatellite
+            A new EarthSatellite with the propagated Orbit
+
+        """
+
+        ad_kwargs: Dict[object, dict] = {}
```
astrojuanlu, line 128: I was not super convinced about adding PEP 526 variable annotations... But what the hell, we're in 2020 and poliastro requires Python 3.6 already :)

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+        perturbations: Dict[object, dict] = {}
+
+        def ad(t0, state, k, perturbations):
+            if len(perturbations):
```
astrojuanlu, line 132: `len` not necessary here?

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+                return np.sum(
+                    [f(t0=t0, state=state, k=k, **p) for f, p in perturbations.items()],
+                    axis=0,
+                )
+            return np.array([0, 0, 0])
```
astrojuanlu, line 137: Can we add an `else` clause here? Just so it's more explicit

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if gravity is EarthGravity.J2:
+            perturbations[J2_perturbation] = {
```
astrojuanlu, line 140: Using the functions as keys, good idea!

src/poliastro/earth/__init__.py
```diff
@@ -0,0 +1,153 @@
...
+                "J2": Earth.J2.value,
+                "R": Earth.R.to(u.km).value,
+            }
+        if atmosphere is not None and A_over_m is not None:
+            perturbations[atmospheric_drag_model] = {
+                "R": Earth.R.to(u.km).value,
+                "C_D": 2.2,  # dimensionless (any value would do)
```
astrojuanlu, line 147: Does this mean that `C_D` is not used then?

astrojuanlu:
@jorgepiloto do you want to have a look before we merge?

src/poliastro/earth/__init__.py
astrojuanlu: Let's leave this for a future PR.

astrojuanlu:
@Meuge last thing: please rebase this on top of current `master` - `git fetch && git rebase upstream/origin` will also remove the merge commits

src/poliastro/earth/__init__.py
astrojuanlu: Actually, now I understood: we need `C_D` as an extra parameter given by the user.

I propose that we add this as a parameter of the `EarthSatellite` object itself. But I don't want to block this PR any longer, so let's leave a `FIXME` comment and deal with it later:

```suggestion
                "C_D": 2.2,  # FIXME: Do not hardcode parameter
```

astrojuanlu:
Pipelines succeeded, but got stuck here somehow. Merging!

PR #972, opened 2020-06-13
Make poliastro compatible with recent numba releases

astrojuanlu:


astrojuanlu:
Merging despite gh-959

PR #971, opened 2020-06-12
Docs captions

jorgepiloto:
This pull requests add some captions to the documentation so it becomes visually easier to differentiate between sections. Although this feature is not linked to any open issue, I see it as a first step for other documentation related tasks.

I thought this could be an interesting simple feature. **Caption names were selected without previous discussion, so let me know if some change is required.**

With captions (PR)             |  Without captions (Current)
:-------------------------:|:-------------------------:
ex_1 |  ex_2

PR #973, opened 2020-06-14
Thumbnail examples gallery

jorgepiloto:
Some time ago I opened #828 in order to fix for #501. Implementation was difficult due to big refactoring and review reached the same complexity level.

However, some months later **the nbsphinx project announced a new feature for generating thumbnail example galleries.** I've tried it and implementation was easy and highly customizable. **Only drawback is not capturing plotly output, but this might be fixed by adding custom thumbnails.** A new directory called `_static/thumbnails/` was created hosting small screenshots for the generated plotly outputs.

Notebooks modifications are just the addition of a cell tag `nbsphinx-gallery` within the plotting cell in order to select final desired thumbnail. If not possible before previously explained conditions, thumbnail is selected through `conf.py` file.

 Result looks like this:

example
 
Hope you like this new feature and let us discuss anything which should be modified :rocket:


jorgepiloto:
*

docs/source/jupyter.rst
```diff
@@ -1,8 +1,15 @@
...
-   :maxdepth: 1
+The following page contains a collection of practical examples and problem
+solutions solved with **poliastro**. From interplanetary transfers, to asteroid
+catching missions or even propagation under external perturbations. In addition,
+user might find other kind of tutorials related on how to use and customize
+plotting utilities, API capabilities and many other orbital mechanics topics.
+
+
+.. nbgallery::
+   :caption: Examples gallery
```
astrojuanlu, line 12: I have a little problem here: the directory is called `examples`, the file is called `jupyter`, the page `Jupyter notebooks`, the TOC is `Tutorials` (after #971), and here you call them `Examples`. Do you think we can unify all this terminology? I think "Examples" everywhere is good enough. People reading the docs online don't necessarily care that they are Jupyter notebooks.

On the other hand, using @evildmp (hi! :blush:) [classification for documentation](https://documentation.divio.com/), these would be How-To Guides. But perhaps "examples" is good enough to keep. Your call :)

pyproject.toml
```diff
@@ -67,6 +67,7 @@ dev = [
     "pytest-mypy",
     "pytest-remotedata",
     "sphinx",
+    "sphinx-gallery",
```
astrojuanlu, line 70: Do we need this?

docs/source/jupyter.rst
jorgepiloto: Sure, I got the same feeling about this `example/tutorial/notebooks namespace`. **In fact, the caption after the `.. nbsphinx-gallery::` is not necessary at all, so we can simply get rid of it.**

Regarding previously shared talk, it looks very interesting. We could also adapt the new `Install, use and contribute` to `How to` section, maybe for another pull?

pyproject.toml
jorgepiloto: Yes, it is **required**... Although `sphinx-gallery` achieves same result, after working with it, I see the following disadvantages:

* It requires a lot of time to build docs because of figure output captures.
* A big refactor is required to switch. Customization and maintenance are more difficult also.

astrojuanlu: Sorry, perhaps I'm confused - don't we need `nbsphinx-gallery` here instead of `sphinx-gallery`?

docs/source/jupyter.rst
astrojuanlu: Sure thing! Let's open an issue and revisit the classification later.

pyproject.toml
astrojuanlu: No, `nbsphinx-gallery` does not exist, it's just `nbsphinx`. But that's already declared, so I don't understand why do we need `sphinx-gallery` if we decided not to use it 😇

jorgepiloto: As claimed in [official nbsphinx docs](https://nbsphinx.readthedocs.io/en/0.7.0/subdir/gallery.html):

> nbsphinx does not provide any gallery styles, but you can easily use the styles from Sphinx-Gallery by installing it. However, you can also create your own CSS styles if you prefer (then you don’t need to install Sphinx-Gallery). You can load your CSS files with html_css_files.

I am not sure how to build the gallery by making use of custom CSS but that option would lead to not depend on sphinx-gallery :thinking:

docs/source/jupyter.rst
jorgepiloto: Sounds good! :rocket:

pyproject.toml
astrojuanlu: Oh, I see! Now I understand, all good then.

docs/source/jupyter.rst
astrojuanlu: Done: https://github.com/poliastro/poliastro/issues/975

astrojuanlu:
Merging! Thanks @jorgepiloto :heart:

PR #984, opened 2020-06-26
Repeat Ground Track Orbit Implementation

Meuge:


src/poliastro/earth/__init__.py
```diff
@@ -104,3 +105,66 @@ def ad(t0, state, k, perturbations):
...
+        The Journal of the Astronautical Sciences, Vol. 39, No. 3, July-September 1991, pp. 313-326.
+
+        """
+        µ = GM_earth.to(u.km ** 3 / u.s ** 2)
+        a = self.orbit.a
+        e = self.orbit.ecc
+        J2 = Earth.J2.value
+        R = Earth.R.to(u.km)
+        i = self.orbit.inc
+        we = 7.292115855306589e-5 * (1 / u.s)
```
astrojuanlu, line 142: We can use `.angular_velocity` from `Body` class instead

astrojuanlu:
```
>>> Earth.angular_velocity
<Quantity 7.29211415e-05 rad / s>
```

src/poliastro/earth/__init__.py
```diff
@@ -104,3 +105,66 @@ def ad(t0, state, k, perturbations):
         ad_kwargs.update(perturbations=perturbations)
         new_orbit = self.orbit.propagate(value=tof, method=cowell, ad=ad, **ad_kwargs)
         return EarthSatellite(new_orbit, self.spacecraft)
+
+    def rgt(self, ndays, norbits, atol=1e-10, iter=float("+inf")):
```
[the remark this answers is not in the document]
astrojuanlu: Ideas:

- Transform this into a `@classmethod` receiving: `(a, ecc, inc, ndays, norbits, spacecraft)`. But we already thought that classmethods were ugly, so...
- Add this as a `@classmethod` in `Orbit`

Can we attempt the second one?

astrojuanlu:
@Meuge added some extra information to https://github.com/poliastro/poliastro/wiki/Repeating-Ground-Track-Orbit we want to compare the algorithms for similarities and differences.

astrojuanlu:
Some ideas after a first pass:

- The numerical model by David Eagle (`repeat2.m`) is not really equivalent to the other ones: it uses a gravitational model of degree and order 18 as well as Solar perturbations, therefore it's much more sophisticated than what we need.
- `repeat1.m` solves the _direct problem_ (how much time does it take for this orbit to repeat its groundtrack?) while `repeat3.m` solves the _inverse problem_ (which orbit do I need to fulfill these repeating groundtrack requirements?). We are interested in the latter.
- `repeat3.m` uses _mean Kozai elements_, while `repeat4.m` uses _osculating elements_
- Curtis algorithm doesn't seem to take any perturbations into account (not even nodal regression), therefore it's probably not very useful.
- It's too early to mplement `repeat2.m` or `repeat4.m` in poliastro, since we don't have good support for high order gravitational models

Therefore, what's left is:

- Make 100 % sure that Vallado and Wagner algorithms are the same or equivalent, which means that we would be using Kozai mean elements (this is not stated in Vallado)
- Find out why Vallado/Wagner and Noreen et al are different, and in particular determine if the latter is a strict superset of the former

Meuge:
> * Noreen

I’ve been reading Wagner’s algorithm in detail along with the one in Vallado, and I haven’t noticed a major difference. 
As for the Noreen case, it shares some similarities with Vallado’s algorithm. Nevertheless, the algorithm looks for repeating ground track orbits that pass the same locations at the same each solar day, so basically daily. Also, another big difference is that the Period takes into consideration the sidereal and solar period, and it distinguishes between prograde and retrograde orbits. 
It’s worth stressing that I haven’t read any mention for retrograde orbits in Vallado for this algorithm, and when I tried to test it with Mars and Venus examples of repeating ground- track orbits, the semimajor axis differs approximately 10/20 km and the eccentricity wasn’t as accurate as Earth’s case.

Meuge:
Another approach worth considering might be using Newton-Raphson, see this [paper](https://arc.aiaa.org/doi/10.2514/1.A32038), for further information. Still, I am dubious if it's going to solve our problem to generalize to all the bodies since it doesn't have any reference other than Earth. It's my understanding that this method and the one implemented in the PR only works for prograde orbits.

astrojuanlu:
Vtipil (2012) looks interesting, it introduces more methods and adds some theoretical background about the assumptions behind them. I added it to the wiki, and expanded a little bit with the information we discussed.

I think our requirements should be:

1. Approximate method. Let's consider only J2 perturbations. Otherwise it's (a) a rabbit hole and (b) much more difficult to implement.
2. Should work for all eccentric orbits, not only near-circular ones.

astrojuanlu:
@Meuge To avoid blocking you any further, if you see that Noreen's algorithm (a) gives same results for orbits that Vallado & Wagner already solve and (b) works for Mars and Venus, let's go for it.

astrojuanlu:
...although now I remembered that Noreen contains several cases: circular equatorial orbits, equatorial eccentric orbits, circular Sun-synchronous orbits, and critically inclined eccentric orbits. Therefore, there is not single formula that would work for all orbits, as far as I understand. @Meuge which case in particular were you using for your comparison?

Meuge:
> ...although now I remembered that Noreen contains several cases: circular equatorial orbits, equatorial eccentric orbits, circular Sun-synchronous orbits, and critically inclined eccentric orbits. Therefore, there is not single formula that would work for all orbits, as far as I understand. @Meuge which case in particular were you using for your comparison?

That's right Noreen details several cases, but I compared the _circular equatorial orbits_ since in the fourth edition of Vallado's book, page 870, it's stated the following "However, consider near-Earth satellites routinely used to study Earth resources, such as Landsat and RADARSAT, with nearly circular orbits, e ≈ 0.0011. Because the eccentricity is of the 22 same order as J2, e ≈ J2 « 1 . As a result, we can simplify the secular rate expressions (deleting terms of e2) and add them together:". So if I understood correctly, it's taking into consideration only circular orbits. 🤔

astrojuanlu:
Gotcha. Perhaps my requirement of having a method that works with all eccentric orbits is too stringent. I'll give this more thought today.

astrojuanlu:
Decision (for now): https://github.com/poliastro/poliastro/wiki/Repeating-Ground-Track-Orbit#algorithm

PR #987, opened 2020-07-02
Pericenter maneuver implementation

Meuge:


src/poliastro/maneuver.py
```diff
@@ -306,3 +306,68 @@ def get_total_cost(self):
...
+        The opposite holds for a delta_v applied against the velocity vector, which decreases the satellite’s velocity.
+        Perigee drift are mainly due to the zonal harmonics, which cause variations in the altitude by changing the
+        argument of perigee.
+        Please note that ecc ≈ 0.001, so the error incurred by assuming a small eccentricity is on the order of 0.1%.
+        This is smaller than typical variations in thruster performance between burns.
+
+        """
+        J2 = orbit.attractor.J2.value
+        if J2 == 0.0:
+            raise ValueError("This orbit has no Periapsis drift")
```
astrojuanlu, line 354: Nitpick:

```suggestion
            raise ValueError("This orbit has no periapsis drift")
```

src/poliastro/maneuver.py
```diff
@@ -306,3 +306,68 @@ def get_total_cost(self):
...
+        elif orbit.ecc > 0.001:
+            raise ValueError(
+                "This orbit has no Periapsis drift, eccentricity should be less than or equal to 0.001"
```
astrojuanlu, line 357: What about a `NotImplementedError` instead? There might be actual periapsis drift, but we don't have an analytical expression for the correction maneuver.

src/poliastro/maneuver.py
```diff
@@ -306,3 +306,68 @@ def get_total_cost(self):
...
+
+        dw = ((3 * n * R ** 2 * J2) / (4 * p ** 2)) * (4 - 5 * np.sin(orbit.inc) ** 2)
+        delta_w = 2 * (1 + orbit.ecc) * max_delta_r
+        delta_w /= orbit.a * orbit.ecc * (1 - orbit.ecc)
+        delta_w **= 0.5
+        delta_t = abs(delta_w / dw)
+        delta_v = 0.5 * n * orbit.a * orbit.ecc * abs(delta_w)
+        vf_ = orbit.v / norm(orbit.v) * delta_v
+
+        return delta_t, vf_
```
astrojuanlu, line 373: Let's return `cls((delta_t, vf_))` instead, to be consistent with the other classmethods

Issue #951, opened 2020-05-02
Exclude docs from source distributions?


astrojuanlu:
Notebooks are quite heavy.

jorgepiloto:
Before #978 was merged, notebooks directory where `14MB` but now, after moving to Jupytext they have reduced half that size, around `7.4MB`. Are we still interested in excluding docs from source?

astrojuanlu:
Yeah, even `7.4 MB` is quite a lot. The only consumer of those source distributions is `pip`, I bet the rest of the users check out the documentation or the repository.

jorgepiloto:
Sorry, I took into account some cache files since the actual disk usage of `docs/source` seems to be around `4.3MB`. However, I agree with you about excluding this from source :+1:

astrojuanlu:
Closed in #988

PR #988, opened 2020-07-03
Exclude docs/ from sdist

jorgepiloto:
Solves for issue #951 by excluding documentation from `sdist` distribution.


astrojuanlu:
Worked! I just checked locally. Merging!

PR #993, opened 2020-07-11
Fix unit for gas constant in COESA models

### LINKED ISSUE #992: Units error in COESA amtospheric models
🐞 **Problem**

I just noticed that the universal gas constant defined in both `coesa62.py` and `coesa76.py` hold wrong units: they make use of `u.J / u.kg / u.K` while proper units should be `u.J / u.kmol / u.K`. **Though defined with wrong units, no errors were raised since code does not make use of those variables.** 

https://github.com/poliastro/poliastro/blob/817fac96365e3848348052d159f6258e502357b1/src/poliastro/earth/atmosphere/coesa62.py#L76
https://github.com/poliastro/poliastro/blob/817fac96365e3848348052d159f6258e502357b1/src/poliastro/earth/atmosphere/coesa76.py#L55

When the universal gas constant is divided by the mean molecular weight of the gas (with `u.kg / u.kmol`), the specific gas constant is obtained with units `u.J / u.kg / u.K`. 

💡 **Possible solutions**

Either delete de variables or correct their units will solve for this problem.

rafaellehmkuhl:
Fixes #992 

COESA atmospheric models 62 and 76 had wrong units for R gas constant (J/kg/K instead of J/kmol/K).

### DIFF
--- a/src/poliastro/earth/atmosphere/coesa62.py
+++ b/src/poliastro/earth/atmosphere/coesa62.py
@@ -181,8 +181,7 @@ class COESA62(COESA):
             # Putting g = (g0*(r0/(r0 +z))**2) in (g * dz / z - zb + Tb/Lb)
             # and integrating it.
             integrand = quad(
-                lambda x: (g0_v * (r0_v / (r0_v + x)) ** 2)
-                / (x - zb_v + Tb_v / Lb_v),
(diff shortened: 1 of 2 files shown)


astrojuanlu:
Amazing! As the change is very simple and it was not tested before, I'll merge straight away :rocket:

Issue #957, opened 2020-05-08
poliastro description on PyPI is wrong


astrojuanlu:
See https://pypi.org/search/?q=poliastro

![Screenshot_2020-05-08 Search results](https://user-images.githubusercontent.com/316517/81390986-aadddd00-911c-11ea-933d-d655c858e98c.png)

astrojuanlu:
Oops, I found it! flit is using `__init__.py` docstring to complete the description:

https://github.com/poliastro/poliastro/blob/817fac96365e3848348052d159f6258e502357b1/src/poliastro/__init__.py#L1-L10

Therefore, we should replace all those extra lines with just the one-line sentence at the end.

astrojuanlu:
Fixed in #994!

PR #994, opened 2020-07-11
Deleted text in doc string  of src/poliastro/__init__.py

vmos1:
This fixes the issue https://github.com/poliastro/poliastro/issues/957

Fixed by deleting the first 3 lines of the doc string in `src/poliastro/__init__.py`


astrojuanlu:
As the fix is very simple and there are no tests for it, I will merge it straight away.

PR #995, opened 2020-07-11
Changes to src/poliastro/czml/extract_czml.py. replaced with gd2gce. …

vmos1:
…Currently failing

changes to src/poliastro/czml/extract_czml.py 
Added gd2gce and removed ellipsoidal_to_cartesian

Not running correctly. issues with variable type in gd2gce


astrojuanlu:
xref: #995

src/poliastro/czml/extract_czml.py
```diff
@@ -276,7 +277,11 @@ def add_ground_station(
                 )  # get semi-major and semi-minor axises
             else:
                 a, b = Earth.R.to(u.m).value, Earth.R_polar.to(u.m).value
-            pos = list(map(lambda x: x.value, ellipsoidal_to_cartesian(a, b, u0, v0)))
+            
+            f=1-(b/a) #Flattenning 
+            
+            pos = list(map(lambda x: x, gd2gce(a, f, u0.to_value(u.rad), v0.to_value(u.rad), 0)))
```
astrojuanlu, line 283: I think we can remove the `map(lambda x: x` part and leave only `list(gd2gce(...))`.

astrojuanlu: Also, this will make the current tests fail because `ellipsoidal_to_cartesian` expects **geocentric** latitude, and `gd2gce` uses **geodetic** latitude.

In case of doubt, I would say that people expect geodetic latitude. So our choice was unfortunate.

We'll have to change the test.

src/poliastro/czml/extract_czml.py
```diff
@@ -276,7 +277,11 @@ def add_ground_station(
                 )  # get semi-major and semi-minor axises
             else:
                 a, b = Earth.R.to(u.m).value, Earth.R_polar.to(u.m).value
```
astrojuanlu, line 279: While you're at it, we can do:

```suggestion
                a, b = Earth.R.to_value(u.m), Earth.R_polar.to_value(u.m)
```

astrojuanlu:
Woops, we were missing the explicit declaration of `pyerfa` as a dependency in our `pyproject.toml`, which will be required until Astropy 4.2 is out. @vmos1 do you want to complete it?

astrojuanlu:
Thanks @vmos1!

Issue #365, opened 2018-05-03
Clarify system requirements to build documentation


OrbitalMechanic:
Attempting to build the pdf document for poliastro version 0.90. 

The output I get from executing the command _make latexpdf_ in the docs subdirectory of the distribution

Samuels-Mac-Pro:~ user$ cd poliastro-master
Samuels-Mac-Pro:poliastro-master user$ cd docs
Samuels-Mac-Pro:docs user$ make latexpdf
sphinx-build -b latex -d build/doctrees   source build/latex
Running Sphinx v1.7.4

Extension error:
Could not import extension nbsphinx (exception: No module named 'nbsphinx')
make: *** [latexpdf] Error 2
Samuels-Mac-Pro:docs user$ 

🖥 **Please paste the output of following commands**

`conda info -a`
[conda_info.txt](https://github.com/poliastro/poliastro/files/1969374/conda_info.txt)

`python -c "import poliastro.testing; poliastro.testing.test()"`
[poliastro.testing.txt](https://github.com/poliastro/poliastro/files/1969375/poliastro.testing.txt)

`pip freeze | grep astropy`
[grep_astropy.txt](https://github.com/poliastro/poliastro/files/1969379/grep_astropy.txt)

`pip freeze | grep poliastro`
[grep_poliastro.txt](https://github.com/poliastro/poliastro/files/1969380/grep_poliastro.txt)

Please advise.

Juanlu001:
Hi @OrbitalMechanic and thanks for your renowed interest! It should be better explained in the documentation but to build the docs you need to install all the devekdeoendencies

Juanlu001:
(Sorry, wrong button)

you need to install the development dependencies:

```
pip install poliastro[dev]
```

or

```
pip install ~/poliastro-master/[dev]
```

(Depending on how you installed poliastro originally)

Can you try this to see it that fixed the issue? I see a strange result in the tests so let's see if everything works.

Juanlu001:
@OrbitalMechanic did you succeed in building the docs?

OrbitalMechanic:
I was able to get a document to build using the command /make latex/ as 
opposed to /make latexpdf/. I'm going through the user's guide at 
present. I've attached the resulted PDF to this note.

Hope this helps.

Sam Dupree.

Juanlu001:
[...]

PR #997, opened 2020-07-13
Add option to skip compiling of jupyter notebooks on documentation

rafaellehmkuhl:
Jupyter notebooks take a long time to compile. This change adds the option for the contributor to set the following environment variable to skip them while generatting docs:

POLIASTRO_SKIP_NOTEBOOKS=True

Discussion about it on #365

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -140,12 +140,12 @@ intersphinx_mapping = {
     'matplotlib': ('https://matplotlib.org', None)
 }
 
-"""
-if os.environ.get('READTHEDOCS') == 'True':
+if os.environ.get('POLIASTRO_SKIP_NOTEBOOKS') == 'True':
     nbsphinx_execute = 'never'
 else:
     nbsphinx_execute = 'always'
 
+"""
     # Controls when a cell will time out (defaults to 30; use -1 for no timeout):
     nbsphinx_timeout = 60
 """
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -140,12 +140,12 @@
...
 }
 
-"""
-if os.environ.get('READTHEDOCS') == 'True':
+if os.environ.get('POLIASTRO_SKIP_NOTEBOOKS') == 'True':
     nbsphinx_execute = 'never'
 else:
     nbsphinx_execute = 'always'
 
+"""
```
astrojuanlu, line 148: While we are at it, can we remove these lines of code that are commented out? If we want to bring them back, we will always have git :)

rafaellehmkuhl: Will remove them. 

I also don't understand when we do that. I think it's a habit we keep from the times when we didn't know how to use git.

astrojuanlu: > I also don't understand when we do that. I think it's a habit we keep from the times when we didn't know how to use git.

Hahaha absolutely!

rafaellehmkuhl: > While we are at it, can we remove these lines of code that are commented out? If we want to bring them back, we will always have git :)

Done

astrojuanlu:
Read the Docs build passed and so did the fast tests, so I'm merging this. Thanks again!

PR #998, opened 2020-07-19
Field of View Implementation

Meuge:


src/poliastro/earth/sensors.py
```diff
@@ -0,0 +1,121 @@
...
+        This formula always gives the answer for the short way to the target ot the acute angle,:math:`{\beta}`,
+        which must be greater than 0º and less than 180º.
+
+    Notes
+    -----
+    For further information, please take a look at "Fundamentals of Astrodynamics and Applications", 4th ed (2013)"
+    by David A. vallado, pages 853-860.
+
+    """
+    if 0 * u.rad < β > np.pi * u.rad:
```
astrojuanlu, line 101: What about `if not (0 <= beta <= pi)`? More legible in my opinion

tests/tests_earth/tests_sensors.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+@pytest.mark.parametrize(
+    "h, n_fov, n_center,expected_Λ_max,expected_Λ_min ",
+    [
+        (
+            800 * u.km,
+            (25 * u.deg).to(u.rad),
+            (40 * u.deg).to(u.rad),
+            0.18736414 * u.rad,
```
astrojuanlu, line 18: Do these test cases come from Vallado, or are they hardcoded values?

Meuge: It's the example 11.1 in the Vallado, with all the decimals values.

astrojuanlu: Let's add a comment stating that then!

astrojuanlu: (also for the other cases below)

src/poliastro/earth/sensors.py
```diff
@@ -0,0 +1,121 @@
...
+    by David A. vallado, pages 853-860.
+
+    """
+
+    R = Earth.R.to(u.km)
+    r_sat = R + h
+    n_max = n_center + n_fov / 2
+    n_min = n_center - n_fov / 2
+    γ_max = np.arcsin(r_sat * np.sin(n_max) / R)
+
```
astrojuanlu, line 40:
```suggestion
```

astrojuanlu: I know it's silly, but I was mentally grouping both intermediate angle γ calculations and this white line was distracting me :sweat_smile:

src/poliastro/earth/sensors.py
```diff
@@ -0,0 +1,121 @@
...
+    if abs(γ_max) <= np.pi / 2 * u.rad:
+        γ_max = np.pi * u.rad - γ_max
+    γ_min = np.arcsin(r_sat * np.sin(n_min) / R)
+    if abs(γ_min) <= np.pi / 2 * u.rad:
+        γ_min = np.pi * u.rad - γ_min
+
+    p_max = R * np.cos(γ_max) + r_sat * np.cos(n_max)
```
astrojuanlu, line 47: Let's add a comment here saying "maximum and minimum slant ranges" (since `ρ` is not defined in the inputs nor the outputs)

src/poliastro/earth/sensors.py
```diff
@@ -0,0 +1,121 @@
...
+    p_min = R * np.cos(γ_min) + r_sat * np.cos(n_min)
+    Λ_max = np.arcsin(p_max * np.sin(n_max) / R)
+    Λ_min = np.arcsin(p_min * np.sin(n_min) / R)
+
+    return Λ_max, Λ_min
```
astrojuanlu, line 52: Good choice returning `Λ_max` and `Λ_min`, with these we can trivially compute `Λ_FOV` while retaining all the information :+1:

src/poliastro/earth/sensors.py
```diff
@@ -0,0 +1,121 @@
...
+    Λ = np.arcsin(p * np.sin(n_center) / R)
+    φ_ctr = np.arcsin(
+        np.cos(β) * np.cos(φ_nadir) * np.sin(Λ) + np.sin(φ_nadir) * np.cos(Λ)
+    )
+    delta_Λ = np.arcsin(np.sin(β) * np.sin(Λ) / np.cos(φ_ctr))
+    λ_ctr = λ_nadir + delta_Λ
+    Λ_max, Λ_min = max_and_min_ground_range(h, n_fov, n_center)
+    delta_λ = (Λ_max - Λ_min) / 2
+
+    return delta_λ, φ_ctr, λ_ctr
```
astrojuanlu, line 121: I'm following Example 11-1 from Vallado, shouldn't we call these `φ_tgt` and `λ_tgt`?

src/poliastro/earth/sensors.py
```diff
@@ -0,0 +1,119 @@
...
+    Parameters
+    ----------
+    h: ~astropy.units.Quantity
+        Altitude over surface.
+    η_fov: ~astropy.units.Quantity
+        Angle of the total area that a sensor can observe.
+    η_center: ~astropy.units.Quantity
+        Center boresight angle.
+    R: ~astropy.units.Quantity
+        Earth equatorial radius
```
astrojuanlu, line 19: What about `Body` or `Attractor` instead of `Earth`?

tests/tests_earth/tests_sensors.py
```diff
@@ -0,0 +1,112 @@
+import pytest
+from astropy import units as u
+from astropy.tests.helper import assert_quantity_allclose
+from poliastro.bodies import Earth
```
astrojuanlu, line 4: `isort` is complaining about this line

astrojuanlu:
:warning: Also, I think the tests are not running because the file should be called `test_sensors.py` instead of `tests_sensors.py`, that's why the patch coverage is so low.

tests/tests_earth/tests_sensors.py
Meuge: I checked running tox in master to find out if the error was on any recent change, and it seems the warning is still in master.

astrojuanlu: Confirmed it was an old `isort` version

astrojuanlu:
All the tests passed before, so I'm merging this already. Thanks again @Meuge! :rocket:

PR #1005, opened 2020-08-26
Reformat everything with new Black version

astrojuanlu:


astrojuanlu:
I hand't noticed we were so inconsistent with docstrings :see_no_evil:

PR #1015, opened 2020-09-09
Renaming max_and_min_ground_range method

nzare:
closes #1014 

- Changed the name of the method from max_and_min_ground_range to min_and_max_ground_range in all relevant files.
- Changed the order of the parameters in the return statement.

PR #1016, opened 2020-09-10
min and max function

### LINKED ISSUE #1014: Docstring error in max_and_min_ground_range method
🐞 **Problem**

We’ve just noticed a minor error in the doctstring from ```max_and_min_ground_range``` method in the ```sensors.py``` file which is in ```Earth``` module. So after given some thought, we realised that it would be better if we rename the method ```min_and_max_ground_range ``` just to clarify that we are returning an interval. 

https://github.com/poliastro/poliastro/blob/c4d0f80c98fb2a0ec59a07f4664cfebb0b1a6cf9/src/poliastro/earth/sensors.py#L6-L8

💡 **Possible solutions**

- Change the name of the method to ```min_and_max_ground_range```. You should perform the rename refactoring to avoid any problem caused by making this change, since this method is called in another files.
- Change the order of the parameters in the return statement.

nzare:
Linked to #1015 
closes #1014

Changed the name of the method from max_and_min_ground_range to min_and_max_ground_range in all relevant files.
Changed the order of the parameters in the return statement.

### DIFF
--- a/src/poliastro/earth/sensors.py
+++ b/src/poliastro/earth/sensors.py
@@ -3,7 +3,7 @@ from astropy import units as u
 
 
 @u.quantity_input(h=u.km, η_fov=u.rad, η_center=u.rad, R=u.km)
-def max_and_min_ground_range(h, η_fov, η_center, R):
+def min_and_max_ground_range(h, η_fov, η_center, R):
     """
     Calculates the minimum and maximum values of ground-range angles
     Parameters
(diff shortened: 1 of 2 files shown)


astrojuanlu (approved):
Excellent!

astrojuanlu:
Investigating the failing tests

astrojuanlu:
They are network failures, unrelated to this pull request. Merging!

Also @nzare, for future pull requests, if you push to the same branch, the pull request gets updated. Therefore, you don't need to open a new one :)

Congratulations on your first contribution! :rocket:

nzare:
Thanks  @astrojuanlu will keep that in mind

Issue #1012, opened 2020-09-07
Remove mention to `poliastro.testing` from issue template


astrojuanlu:
We no longer ship tests as part of poliastro.

(Noticed in #1011)

priyanshurohilla:
Hey @astrojuanlu what i have understood from this issue, if i am rigth is that you want `poliastro.testing`l abel to be removed from issue templates. But this is only possible who has the write access to this repository

astrojuanlu:
Sorry for the delay, the template is on `.github/` directory, see #1024

astrojuanlu:
Closed in #1024!

PR #1024, opened 2020-10-02
Was removed mention to `poliastro.testing` from issue template #Hackt…

hramirezq:
…oberfest

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -36,10 +36,6 @@ We will ask you for your installation details in the following sections.
 ```
 </details>
 
-<!--- Please do `pip install pytest`, run this command and paste its output: -->
-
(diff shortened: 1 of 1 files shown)


jorgepiloto:
Hi @hramirezq, thanks for your pull request (solves #1012). Since `tests/` were removed from `src/poliastro/`, the `testing` module will not be found, so whole line must be removed.

astrojuanlu (changes requested):
What @jorgepiloto said :)

We need to make a decision:

- Just remove the mention to `poliastro.testing`, because it's not there anymore, or
- Tell users to run the tests by checking out the repository

I think the latter is too much to ask for a humble bug report, so I say that we just remove it.

astrojuanlu (approved):
Looks good! Merging straight away

astrojuanlu:
Thanks @hramirezq! :raised_hands:

hramirezq:
Thanks for the advice, I hope to be more useful in the next days!

PR #1026, opened 2020-10-04
Avoid unnecessary element conversions in propagation

andrea-carballo:


src/poliastro/core/propagation/farnocchia.py
```diff
@@ -283,6 +283,17 @@ def nu_from_delta_t(delta_t, ecc, k=1.0, q=1.0, delta=1e-2):
     return nu
 
 
+@jit
+def farnocchia_coe(k, p, ecc, inc, raan, argp, nu):
+    
+    q = p / (1 + ecc)
+
+    delta_t0 = delta_t_from_nu(nu, ecc, k, q)
+    delta_t = delta_t0 + tof
```
astrojuanlu, line 292: Woops, I forgot to specify in the issue that we're missing a `tof` argument for `farnocchia_coe`, at the end

astrojuanlu (changes requested):
Thanks for the pull request @andrea-carballo! There is one variable missing and you need to do `tox -e reformat` to adapt the code for the quality checks, other than that this is on the right direction!

src/poliastro/core/propagation/farnocchia.py
andrea-carballo: ohh! My bad! I fixed it and pushed the changes

astrojuanlu (approved):
This looks good! @andrea-carballo do you want to make a similar change for the rest of propagators, or do you prefer to leave that for another pull request?

astrojuanlu:
Also, there is a bit of a mess in the git history, as can be seen in https://github.com/poliastro/poliastro/network

Screenshot_2020-10-05 poliastro poliastro

You might want to try to `rebase` your commits on top of current `master`, ping me in private in case you need a hand with this (the process can be a bit messy the first time)

astrojuanlu (commented):
My suggestion for `danby` (not sure if this works with numba):

```
while n <= numiter:
    ...
    nu = nu.arctan2(sta, cta)
    break
else:
    raise ValueError("Maximum ...")

return coe2rv(k, p, ...)
```

About `while ... else`: https://docs.python.org/3/reference/compound_stmts.html#the-while-statement

> if the expression is false [...] the suite of the else clause, if present, is executed and the loop terminates.
> A break statement executed in the first suite terminates the loop without executing the else clause’s suite.

Therefore, the code inside `else` is only ran if the maximum number of iterations is surpassed.

src/poliastro/core/propagation/__init__.py
```diff
@@ -838,38 +839,8 @@ def gooding(k, r0, v0, tof, numiter=150, rtol=1e-8):
     return coe2rv(k, p, ecc, inc, raan, argp, nu)
 
 
-@jit
-def danby(k, r0, v0, tof, numiter=20, rtol=1e-8):
-    """Kepler solver for both elliptic and parabolic orbits based on Danby's
-    algorithm.
+def danby_coe(k, p, ecc, inc, raan, argp, nu, tof, numiter, rtol):
```
astrojuanlu, line 842: Missing `@jit` on top! That's the cause of this error message:

```
E           numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
E           Untyped global name 'danby_coe': cannot determine Numba type of <class 'function'>
```

astrojuanlu (changes requested):

astrojuanlu (approved):
🎉

astrojuanlu:
Hmmm not so fast, there's one numba failure... Investigating

src/poliastro/core/propagation/__init__.py
```diff
@@ -893,13 +895,51 @@ def danby(k, r0, v0, tof, numiter=20, rtol=1e-8):
                 cta = ecc - np.cosh(E)
 
             nu = np.arctan2(sta, cta)
-            return coe2rv(k, p, ecc, inc, raan, argp, nu)
-
+            break
```
astrojuanlu, line 898: The diff got a bit messed up and it's difficult to see, but inside `danby_coe` there's a leftover `return coe2rv` on line 853 (of the current commit):

https://github.com/poliastro/poliastro/pull/1026/files#diff-2fa7459268fa44ca071c0369b2cdde7dR853

Also, notice that removing the `return` statements makes the current logic slightly wrong, because it ends up being something like this:

```python
    if ecc == 0:
        # Solving for circular orbit
        ...
        nu = ...  # We could stop here

    elif ecc < 1.0:
        # For elliptical orbit
        ...
        E = ...  # nu will be computed later

    else:
        # For parabolic and hyperbolic
        ...
        E = ...  # nu will be computed later

    # Iterations begin
    n = 0
    while n <= numiter:
        ...
        if np.abs(f) <= rtol:
            ...
            nu = ...  # FIXME: We compute nu with the `while` loop even if we have it already!
            break

    else:  # Good addition on your side!
        raise ValueError("Maximum number of iterations has been reached.")

    return nu
```

astrojuanlu (changes requested):

src/poliastro/core/propagation/__init__.py
astrojuanlu: Well, I left this long explanation because at first I thought that fixing this logic would be more difficult. But doing `return nu` when it's first available fixes the problem in one line!

astrojuanlu:
Documentation failures are unrelated, code and tests look good!

astrojuanlu (approved):

astrojuanlu (approved):
Code looks perfect now! 🙌🏼 Windows tests had an unrelated failure, I retried the job but will merge anyway when it finishes

tests/tests_twobody/test_propagation.py
```diff
@@ -406,3 +414,28 @@ def test_propagation_custom_body_works():
...
+    [danby_coe, markley_coe, pimienta_coe, mikkola_coe, farnocchia_coe, gooding_coe],
+)
+def test_propagate_with_coe(propagator_coe):
+
+    period = iss.period
+    a, ecc, inc, raan, argp, nu = iss.classical()
+    p = a * (1 - ecc ** 2)
+
+    # Delete de units
+    a = a.to_value(u.km)
```
astrojuanlu, line 430: It's `p` the value we need to strip the units to call the función later! That's why the coverage check is failing

astrojuanlu (changes requested):
Sorry again, I didn't see the legitimate coverage failure. We need to change two letters, and we're done!

astrojuanlu:
RTD failure is unrelated

astrojuanlu (approved):

astrojuanlu:
Merged! Thanks @andrea-carballo , and congratulations for your first pull request! 🙌🏼

PR #1030, opened 2020-10-08
Publish packages from CI

### LINKED ISSUE #950: Publish packages from CI
See https://openastronomy-azure-pipelines.readthedocs.io/en/latest/publish.html#usage-and-options.

epassaro:
- [ ] Use the OpenAstronomy `publish` template
- [ ] Set PyPI endpoint 
- [ ] Use `test_extras` and `test_commads` in template to test wheels

Closes #950

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -30,56 +30,73 @@ trigger:
     - '*post*'
 
 jobs:
-- template: run-tox-env.yml@OpenAstronomy
-  parameters:
-    submodules: false
-    env:
-      - PIP_PREFER_BINARY: true
-      - PIP_PROGRESS_BAR: "off"
-      - MPLBACKEND: Agg
-      - FLIT_ROOT_INSTALL: 1
-    posargs: '-vv'
-    libraries:
-      apt:
-        - pandoc texlive texlive-latex-extra texlive-fonts-recommended dvipng graphviz cm-super-minimal
-    envs:
-      - linux: check
-        name : quality
-        libraries: {}
+- job: test
+  steps:
+  - template: run-tox-env.yml@OpenAstronomy
+    parameters:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -29,57 +29,74 @@ trigger:
...
+            - macos: py38
+              name: py38_test_macos
+              libraries: {}
+
+  - stage: publish
+    jobs:
+      - template: publish.yml@OpenAstronomy
+        parameters:
+          ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:
+            pypi_connection_name: 'pypi_endpoint'
```
astrojuanlu, line 94:
```suggestion
            pypi_connection_name: 'pypi_upload'
```

azure-pipelines.yml
```diff
@@ -29,57 +29,74 @@ trigger:
...
+          libraries:
+            apt:
+              - pandoc texlive texlive-latex-extra texlive-fonts-recommended dvipng graphviz cm-super-minimal
```
astrojuanlu, line 97: I don't think we need the LaTeX dependencies to create the sdists and wheels, but I might be wrong?

astrojuanlu (changes requested):
Apart from a couple of minor comments, this is almost good to go!

I noticed that these commits don't have an associated email though, and therefore they are not linked to @epassaro. Would you like to

- Squash all these commits into one, and
- Configure your email so the commit is properly attributed?

azure-pipelines.yml
epassaro: Seems right!

epassaro:
I messed up this PR trying to squash, I'll make a new one, sorry @astrojuanlu.

PR #1028, opened 2020-10-08
my attempt to solve issue #1021

clacri:
Hi! This is my first attempt to solve the issue, I'm happy to take any feedback about it :)

Closes #1021

PR #1032, opened 2020-10-10
Fix MyPy issues

astrojuanlu:
Seen in #1028.


astrojuanlu:
Quick tests passed, merging straight away.

PR #1035, opened 2020-10-12
Publish packages from CI

### LINKED ISSUE #950: Publish packages from CI

epassaro:
## Summary

- [x] Use the OpenAstronomy `publish` template
- [x] Set PyPI endpoint 
- [x] Use `test_extras` and `test_commadns` in template to test wheels

Closes #950

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,7 +12,7 @@ resources:
   - repository: OpenAstronomy
     type: github
     endpoint: poliastro
-    name: OpenAstronomy/azure-pipelines-templates
+    name: epassaro/azure-pipelines-templates
     ref: master
 
 trigger:
@@ -83,3 +83,13 @@ jobs:
       - macos: py38
         name: py38_test_macos
         libraries: {}
+
+- template: publish.yml@OpenAstronomy
+  parameters:
+    ${{ if startsWith(variables.Build.SourceBranch, 'refs/tags/v') }}:
+      pypi_connection_name: 'pypi_upload'
+    test_extras: 'dev'
+    test_command: 'cd $(Build.SourcesDirectory)/tests; pytest -m "not slow and not mpl_image_compare"'
+    targets:
+    - sdist
+    - wheels_universal


epassaro:
@astrojuanlu In order to get this working I needed to do some tweaks:

- The conditional statement described in OpenAstronomy: `${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags/v') }}:` does not work anymore. I used `variables.Build.SourceBranch` instead. I should open an issue on OpenAstronomy about this.
- Forked the `OpenAstronomy/azure-pipelines-templates` because current template does not work with `flit` (see my changes [here](https://github.com/epassaro/azure-pipelines-templates/commit/ee2365a26c7a7f82db93cb7478b969d3796ad61b)). But maybe I'm wrong, because I do not have so much experience with `flit` or `poetry` (I'm more like a hardcore `conda` user).  Tried everything but the pipeline keeps complaining about missing `flit_core.buildapi`, so my approach was to install it through the template.
- Finally, I used `wheels_universal` instead of making wheels for different OSes. Does @poliastro use compiled extensions?

epassaro:
> > Does @poliastro use compiled extensions?
> 
> Nope!
> 
> > I should open an issue on OpenAstronomy about this.
> 
> Good catch, please ping me if you do.
> 
> > Tried everything but the pipeline keeps complaining about missing flit_core.buildapi, so my approach was to install it through the template.
> 
> Do you have a link to any of those failing pipelines?


1. Ok, so universal wheels should be enough, right?

2. Opened an issue [here](https://github.com/OpenAstronomy/azure-pipelines-templates/issues/57).

3. Failed run log [here](https://dev.azure.com/poliastro/poliastro/_build/results?buildId=530&view=logs&j=f7656e4a-aab1-5409-7a28-08f0d640eba6&t=80e3b503-b0d8-577b-8cb0-afa4b369e9c8&l=16).

astrojuanlu:
1. Yes
2. :heavy_check_mark: 
3. https://github.com/FFY00/python-build/issues/98 (let's wait for a new release of `build`)

epassaro:
Ok, so we can start testing the wheel in the pipeline, right? Is `pytest poliastro` enough or should I pass some extra args?

astrojuanlu:
> Is `pytest poliastro` enough or should I pass some extra args?

In fact, `pytest` should be enough! Let's test only the fast tests though, `pytest -m "not slow and not mpl_image_compare"`

astrojuanlu:
Damn, logs for old builds disappear? :flushed: 

I found the `--pyargs poliastro` part weird, but it was also failing without it... not sure what's happening here.

epassaro:
> Damn, logs for old builds disappear? 
> 
> I found the `--pyargs poliastro` part weird, but it was also failing without it... not sure what's happening here.

Old logs can be found under `Runs` in the `Recent | All | Runs` menu.

I don't know why the testing fails. I'll keep looking for a solution.

astrojuanlu:
Perhaps the tests are not in the working directory where the `test` step takes place, you might want to do `ls` as well, just in case

epassaro:
> Perhaps the tests are not in the working directory where the `test` step takes place, you might want to do `ls` as well, just in case

The test command runs in a temporary directory, according to [this line](https://github.com/OpenAstronomy/azure-pipelines-templates/blob/42425f9cb0be22664460660cf572be15bfdf5ff6/publish.yml#L106) of the template.

I tried using the test command from Poliastro documentation: `'python3 -c "import poliastro.testing; poliastro.testing.test()"'` but failed with `ModuleNotFoundError: No module named 'poliastro.testing'`.

astrojuanlu:
Woops, poliastro does not ship its tests anymore. Perhaps you read outdated documentation (either because it's from an older version or because we have outdated instructions somewhere), therefore `poliastro.testing` won't work.

astrojuanlu:
Perhaps SunPy does ship their tests? @Cadair ?

epassaro:
Oh, do you mean that Poliastro tests currently are located outside the package? In that case does have sense ttesting the wheels? That's why `pytest` does not find any test.

astrojuanlu:
Yeah, I assumed that we could `pip install poliastro.whl` and then let `pytest` run the tests against the installed code. After all, when the test does `import poliastro`, this would be testing the wheel, right?

In our case, the trick of the temporary directory is not even needed because we use a `src` layout. In other words: as we don't have a `poliastro` directory, but `src/poliastro`, `import poliastro` only uses the installed version, never the local files.

epassaro:
I think maybe I can use the predefined variables to make it work, but how we make sure we're running the tests on the wheel and not on the source dir?

epassaro:
@astrojuanlu The test are passing now! I just `cd` into the `poliastro/tests` folder and run `pytest -m ...`. Are we testing the wheels correctly, right? Since the wheel was installed in a previous step and `src/poliastro` is on another dir level.

astrojuanlu (changes requested):
Finally! :rocket: Can you squash these into 1 commit?

astrojuanlu:
> Are we testing the wheels correctly, right?

Yes, absolutely! Thanks a lot for your patience, debugging CI can be a pain in the neck :muscle:

epassaro:
1. What about the `flit` issue? Will you keep using my fork until changes got merged to the other repos?
2. Can you squash and merge with the green button below instead? Otherwise I'll try to squash but I can't guarantee it will work :joy:

astrojuanlu:
About the fork, I don't know the `build` release schedule but I don't want it to stop us from merging this. 

About the squash, I just made a local checkout, so no commits will be lost :wink: try this:

1. Make sure `git remote` gives two remotes. If it only gives one, it's `origin`, linked to your fork, and you'll also need to set an `upstream` fork linked to `poliastro/poliastro`
2. Update `master` with respect to `upstream/master` using `git fetch upstream && git checkout master && git branch --set-upstream-to=upstream/master && git merge`
3. Go back to your branch and do an interactive rebase: `git checkout publish-template && git rebase --interactive master`

You will see this:

```
pick 0ac5e374 Initial commit
pick ede5887e add newline
pick e552dddb add test command
pick 70357a53 add test extra param
pick a2a1e43b change test command
pick 7a764c48 change test command
pick 9e1389bb using quotes
pick 9453e45d change test command
pick cd7af1c1 change test command
pick d3873cee change test command
pick a601cf26 change test command
pick fa0a684d change test command
```

replace it with this:

```
r 0ac5e374 Initial commit
f ede5887e add newline
f e552dddb add test command
f 70357a53 add test extra param
f a2a1e43b change test command
f 7a764c48 change test command
f 9e1389bb using quotes
f 9453e45d change test command
f cd7af1c1 change test command
f d3873cee change test command
f a601cf26 change test command
f fa0a684d change test command
```

then save the file, exit the editor, and choose a good name for the commit.

4. Force push the changes: `git push --force origin publish-template`

Let me know if you get stuck at these steps :)
```
```

epassaro:
Done! @astrojuanlu

astrojuanlu (approved):

astrojuanlu:
Merging straight away! Thanks a lot @epassaro, and congratulations! :tada:

epassaro:
Thank you! See you later.

Issue #922, opened 2020-04-19
Properly vectorize propagators


astrojuanlu:
At the moment, the "vectorization" step for most propagators happens outside of numba, harming performance. For example:

https://github.com/poliastro/poliastro/blob/e57d8c4330e45aa4d2f7d2bd1131d806dc95a373/src/poliastro/twobody/propagation.py#L146-L151

This is related to #921 and perhaps should be tackled afterwards.

cc @Josvth

astrojuanlu:
Prospective contributors: consider looking at #921 first, then we can work on this.

astrojuanlu:
@andrea-carballo expressed interest in working on this! The steps should be:

1. Move `test_propagate_with_coe` to a newly created `tests/tests_core/test_propagation.py`
2. Move the functions in `core/propagation/__init__.py` to `core/propagation/vallado.py`, `core/propagation/mikkola.py`... as done with `core/propagation/farnocchia.py` and adapt tests and imports accordingly
3. Write `farnocchia_coe_tofs`, `mikkola_coe_tofs`, `markley_coe_tofs`, `pimienta_coe_tofs`, `gooding_coe_tofs`, `danby_coe_tofs` doing this:

```
@jit
def propagator_tofs(k, ..., tofs, ...):
    nus = np.zeros_like(tofs)
    for tof in tofs:  # pseudocode
        # call corresponding propagator with single tof (time of flight) and store nu
    return nus
```

This doesn't fully fix the issue, but it's a first step.

astrojuanlu:
Postponing this to the next release.

astrojuanlu:
Blocking this on #921.

astrojuanlu:
I think #921 will need a significant refactor, so I'm postponing this issue. At least will try to tackle #921 for the next release.

astrojuanlu:
Now propagators have a `propagate_many` method that still is iterating on the Python side rather than on the numba side:

https://github.com/poliastro/poliastro/blob/3abced1b27ee7a9ac9ae51b2bdf4723b1c3239e4/src/poliastro/twobody/propagation/farnocchia.py#L59-L65

PR #1037, opened 2020-10-19
Vectorize propagators

andrea-carballo:
Working in https://github.com/poliastro/poliastro/issues/922.
Starting with mikkola


src/poliastro/core/propagation/mikkola.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Solving for the classical elements
+    p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
+    nu = mikkola_coe(k, p, ecc, inc, raan, argp, nu, tof)
+
+    return coe2rv(k, p, ecc, inc, raan, argp, nu)
+
+
+@jit
+def mikkola_tofs_coe(k, p, ecc, inc, raan, argp, nu, tofs):
```
astrojuanlu, line 126: This looks good!

src/poliastro/core/propagation/mikkola.py
```diff
@@ -0,0 +1,136 @@
...
+    nus = np.zeros_like(tofs)
+    for i, tof in tofs.ndenumerate():
+        nus[i] = mikkola_coe(k, p, ecc, inc, raan, argp, nu, tof)
+    return nus
+
+
+@jit
+def mikkola_tofs(k, r0, v0, tofs, rtol=None):
+    p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
+    return mikkola_tofs_coe(k, p, ecc, inc, raan, argp, nu, tofs)
```
astrojuanlu, line 136: If we add a function like this, it should return an array of (r, v) vectors. Let's not do it for now, we would have to deal with two dimensions.

astrojuanlu (commented):
This is on the right path!

astrojuanlu:
Ended up doing part of this in #1430. Thanks @andrea-carballo for laying the early work here!

PR #1039, opened 2020-11-03
Update __init__.py

giuseppedipasquale:
Use of attribute instead of property decorator

### DIFF
--- a/src/poliastro/spacecraft/__init__.py
+++ b/src/poliastro/spacecraft/__init__.py
@@ -64,5 +64,5 @@ class Spacecraft:
 
         """
 
-        B = self._C_D * (self.A / self.m)
+        B = self._C_D * (self._A / self._m)
         return B


astrojuanlu:
Hi @giuseppedipasquale, thanks for contributing to poliastro! :rocket: Would you please explain the rationale behind using the private properties instead of the public ones?

giuseppedipasquale:
Hello @astrojuanlu , thanks! I am new to programming at this level and I want to improve my Python skills so I am starting from something that seemed accessible. I am proposing this change for two reasons:
1. Since I am learning about decorators, I understood that somehow they "call" a function so I thought that using directly the attributes would be more efficient. I wrote a simple script [https://github.com/giuseppedipasquale/DecoratorsVsAttributes](probably not in the optimal way, but again, I am here to learn) to assess if this was true, and it appears indeed that, at least on my PC, not going through decorators makes the code more efficient.
1. Part of the ballistic coefficient was written using attributes already (drag coefficient) while area and mass were used through the property decorator, so I thought it would have been nicer to have everything in the same form, and I went for the attribute for the above-mentioned reason. </a>

I hope the reply is clear, however I am available for any further clarification.

Giuseppe

astrojuanlu:
Hi again @giuseppedipasquale! Sorry for the long delay, we are in low-activity mode these days because we are juggling another commitments, but I didn't forget about your contribution.

The argument about the performance is interesting, I thought the overhead of calling `@property` would be low. We normally don't do these kind of micro-optimizations without first performing some benchmarks, but this one is low-hanging fruit.

On the other hand, your point about consistency is also good - we should either use private properties everywhere or public properties everywhere, unless there's a good reason for the former.

And finally, even though we don't have that many "good first issues" at the moment, I'm happy to see that you were proactive and found one easy task to do by reading the source code yourself! I want to reward your proactivity and encourage you to stay around and keep contributing, so I am going to merge your pull request :)

astrojuanlu:
Congratulations on your first contribution to poliastro :rocket: Don't hesitate to keep looking around, and join [our chat](http://chat.poliastro.space)!

giuseppedipasquale:
Hello @astrojuanlu! This is my first merge so I'm quite excited about it 😃 I am checking the "good first issues", I hope I will manage to contribute more and in a more useful way. Until then, thanks for the merge and have a nice evening!

Giu

PR #1044, opened 2020-11-29
Minor change removing if-statement allows for the markley_coe propagator to be called using 1-N arrays.

Josvth:
The if-statement in the markley_coe function is the only line that doesn't allow the function to be properly vectorized.

I've changed these lines to be solely arithmetic, therefore working on 1-N arrays.

With this changes the variables k, p, ecc, nu and tof may be single values (original capability), all equal length 1-N arrays, or a mix of single values & equal length 1-N arrays.

### DIFF
--- a/src/poliastro/core/propagation/__init__.py
+++ b/src/poliastro/core/propagation/__init__.py
@@ -324,9 +324,7 @@ def markley_coe(k, p, ecc, inc, raan, argp, nu, tof):
     M = M0 + n * tof
 
     # Range between -pi and pi
-    M = M % (2 * np.pi)
-    if M > np.pi:
-        M = -(2 * np.pi - M)
+    M = (M + np.pi) % (2 * np.pi) - np.pi
 
     # Equation (20)
     alpha = (3 * np.pi ** 2 + 1.6 * (np.pi - np.abs(M)) / (1 + ecc)) / (np.pi ** 2 - 6)


Josvth:
Code snippet showing a test:

`from poliastro.core.propagation import markley_coe

# Individual propagation
import numpy as np

k = 3.98600442e+14  # k Earth
R = 6378136.6       # radius Earth [m]

# Periapsis
p_a = R + 1000e3
p_b = R + 1500e3
p_c = R + 2000e3

ecc_a = 0.0
ecc_b = 0.1
ecc_c = 0.5
nu0_a = 0.0
nu0_b = np.pi/8
nu0_c = 3*np.pi/25

tof_a = 10
tof_b = 15
tof_c = 379

nu_a = markley_coe(k=k, p=p_a, ecc=0.0, inc=0.0, raan=0.0, argp=0.0, nu=nu0_a, tof=tof_a)
nu_b = markley_coe(k=k, p=p_b, ecc=0.1, inc=0.0, raan=0.0, argp=0.0, nu=nu0_b, tof=tof_b)
nu_c = markley_coe(k=k, p=p_c, ecc=0.5, inc=0.0, raan=0.0, argp=0.0, nu=nu0_c, tof=tof_c)

print("nu_a:", nu_a)
print("nu_b:", nu_b)
print("nu_c:", nu_c)

# Vectorized propagation of orbital elements
# May be called having variables as 1-N arrays or single values
# vectorized tof not supported
nnu = markley_coe(k=np.array([k] * 3),
                  p=np.array([p_a, p_b, p_c]),
                  ecc=np.array([ecc_a, ecc_b, ecc_c]),
                  inc=0.0, raan=0.0, argp=0.0,
                  nu=np.array([nu0_a, nu0_b, nu0_c]),
                  tof=np.array([tof_a, tof_b, tof_c]))

print("nnu:", nnu)`

Josvth:
Because it confused me too after writing this change a while ago:

The lines
```python
M = M % (2 * np.pi)`
if M > np.pi:
        M = -(2 * np.pi - M)
````
First maps _M = (-∞, ∞)_ to _M = [0, 2π)_
Then it maps values [π, 2π) to [-π, 0)

My edit:
```python
M = (M + np.pi) % (2 * np.pi) - np.pi
````

can be understood by the modulo property:
_(a + b) mod n = ((a mod n)+(b mod n))mod n_

Substituting _a = M_, _b = π_ and _n = 2π_:

_(M + π) mod 2π  - π_
_(M mod 2π + π mod 2π) mod 2π  - π_
_(M mod 2π + π) mod 2π  - π_

The first _M mod 2π_ does the same as the original mapping _M = (-∞, ∞)_ to _M = [0, 2π)_.
Then by adding π to the result you get [π, 3π). 
Then the second _mod 2π_ leaves values [π, 2π) unchanged,
but the values [2π, 3π) get mapped to [0, π) resulting in a range [0, 2π)
Then by subtracting π we get the range [-π, π)

Visually in ASCII:
```
   -π    0    π   2π   3π
1.	|    |--->|===>|    | # first modulo
2.	|    |    |--->|===>| # add pi
3.	|    |===>|--->|    | # second modulo
4.	|--->|===>|    |    | # subtract pi
```

Hope that helps somebody understand this minor change!

**Comment by Josvth** [comment]
On the topic of 'poliastro' vectorization. Its great to see a more effort for this in other tickets, i.e  #1037 and #380!

My use of poliastro has pushed me to make some strange edits. Short story, I want to propagate >40000 objects to the next tof. instead of 1 object to 4000 tof's (i.e. orbit sampling).

I chose the markley propagator because it doesn't require iteration (and the accuracy seemed good enough for me) making vectorization easier.

Ideally, you would like to have the function capable of doing a 1-N array of coe's and a 1-M array of tof's and get a N-M matrix as a result.

This however requires some effort making every line capable of dealing with two dimensions (coe's and tof's).

My take to keep things understandable and easy to check is to only support 1-N vectorization:
- In my case I would do 1-N coe's and 1 tof at the time. But you could also do 1 coe and 1-N tof's with the modification in this PR to remove a `for tof in tofs` loop.
- If you want an N coe's times M tof's computation, you could do that by repeating each coe M times, resulting into each variable passed to the function to be of size 1-(N*M)

That has worked pretty good for me so far! Hope this is useful and understandable for somebody!

**Comment by jorgepiloto** [comment]
Hi @Josvth, sorry for this late review and thanks for the simplification in the logic of the algorithm. I wrote a small test and your implementation works as expected:

```python
""" Tests M conversion """

import numpy as np
from numpy.testing import assert_allclose
import pytest

@pytest.mark.parametrize("M", np.linspace(0, 2*np.pi, 3))
def test_M_proper_conversion(M):

    # Solving for M_new
    M_new = (M + np.pi) % (2 * np.pi) - np.pi
    
    # Solving for M_old
    M_old = M % (2 * np.pi)
    if M_old >= np.pi:
        M_old = -(2 * np.pi - M_old)

    assert_allclose(M_old, M_new)
```

This implementation looks good by my side :+1: 

Failing tests are related to AzurePipelines, since for some reason the virtual machines are trying to install poliastro with a Python 3.9 version, something which is still not supported.

astrojuanlu (approved):
Thanks @Josvth for this update and for the patience! I totally agree that this kind of arithmetic requires a bit of head scratching, but I have seen and used this pattern many times (and also in other places in poliastro, I believe) and, with @jorgepiloto double check, this is good to go.

astrojuanlu:
About the pipelines: yes, we are in a very sorry state at the moment. tox is not being maintained because its author is rewriting it from scratch, therefore it has some bugs that they will not fix in the current version (see https://github.com/tox-dev/tox/issues/1738 in particular). As a result, I gave up and will just wait until either numba is available on Python 3.9 ([early 2021](https://github.com/numba/numba/issues/6345#issuecomment-738696458)) or a new tox version (?). There might be a better option but I haven't had time to work around these issues in the CI.

astrojuanlu:
Regarding [your comment]() about vectorization, I discussed a lot a couple of months ago with @s-m-e about these use cases (namely: propagating a lot of orbits, for a lot of epochs). However, although I am all-in about simplifications like the one you made in this pull request, I no longer think making poliastro functions easy to vectorize in the sense of "being able to work with NumPy arrays" is the way to go. Given that all the "core" functions are JITted with numba, I think we should embrace the power of plain for-loops. I will continue this on #1046.

PR #1046, opened 2020-12-06
Coe2rv vectorized

Josvth:
This PR contains a couple of things:
- the vectorized version of the markley propagator (as in #1044)
- removal in the for loops in `propagation.markley(...)`
- vectorized versions of `elements.rv_pqw(...)` and `elements.coe2rv(...)` 
- removed unnecessary (I think) conversions in `Orbit.sample(...)` and use vectorized computation
- some catches in other functions as `elements.coe2rv(...)` now always returns a 2d array (caveat)

### DIFF
--- a/src/poliastro/core/elements.py
+++ b/src/poliastro/core/elements.py
@@ -85,8 +85,8 @@ def rv_pqw(k, p, ecc, nu):
     u_q = np.array([[0, 1, 0]]).T
     u_w = np.array([[0, 0, 1]]).T
 
-    r = ((u_p * cos(nu) + u_q * sin(nu) + u_w * 0.) * (p / (1 + ecc * cos(nu)))).T
-    v = ((u_p * -sin(nu) + u_q *(ecc + cos(nu))  + u_w * 0.) * sqrt(k / p)).T
+    r = ((u_p * cos(nu) + u_q * sin(nu) + u_w * 0.0) * (p / (1 + ecc * cos(nu)))).T
+    v = ((u_p * -sin(nu) + u_q * (ecc + cos(nu)) + u_w * 0.0) * sqrt(k / p)).T
 
     return r, v
 
@@ -156,6 +156,7 @@ def coe2rv(k, p, ecc, inc, raan, argp, nu):
         \end{bmatrix}
 
     """
+
     pqw = rv_pqw(k, p, ecc, nu)
     rm = coe_rotation_matrix(inc, raan, argp)
 
--- a/src/poliastro/core/elements.py
+++ b/src/poliastro/core/elements.py
@@ -68,11 +68,27 @@ def rv_pqw(k, p, ecc, nu):
     r = [-5312706.25105345  9201877.15251336    0] [m]
     v = [-5753.30180931 -1328.66813933  0] [m]/[s]
 
+    Also works on 1-N arrays:
+    >>> r, v = rv_pqw(np.full((4,1),k), np.full((4,1),p), np.full((4,1),ecc), np.full((4,1),nu))
+    r = [[-5312706.25105345  9201877.15251336    0]
+         [-5312706.25105345  9201877.15251336    0]
+         [-5312706.25105345  9201877.15251336    0]
+         [-5312706.25105345  9201877.15251336    0]] [m]
+    v = [[-5753.30180931 -1328.66813933  0]
+         [-5753.30180931 -1328.66813933  0]
+         [-5753.30180931 -1328.66813933  0]
+         [-5753.30180931 -1328.66813933  0]] [m]/[s]
     """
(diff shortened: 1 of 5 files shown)


Josvth:
Some code to check if everything still works:

```
import numpy as np
from poliastro.examples import molniya
from poliastro.twobody.propagation import propagate
from poliastro.twobody.propagation import markley
from astropy import time, units as u

molniya.plot() # Now uses vectorized way for sampling the orbit instead of calling a propagator

# Generate positions at several tofs using vectorized markley propagator
tofs = np.arange(0,100,20) * u.s
cartesian = propagate(molniya, tofs) # computes tofs in for loop using default propagator
print(cartesian)

cartesian_vec = propagate(molniya, tofs, method=markley) # computes tofs vectorized at once
print(cartesian_vec)

# Check if results are similar (should give a small number)
print(np.sum(np.abs(cartesian.get_xyz() - cartesian_vec.get_xyz())))
```

Josvth:
@astrojuanlu I might need some help here with the test failures. Any way I can run the unit test locally?

jorgepiloto:
Hi @Josvth, thanks for your interest in the vectorization of the software. From my point of view, it would be better if you split this pull request into two different one: a `coe2rv` vectorization upgrade and a `markley propagator` one. Same logic addresses behind the new modifications within the `Orbit.sample()` method.

Let us keep this PR only for the `coe2rv` new features.

astrojuanlu (commented):
Hi @Josvth - taking over what I said in #1044, I am not 100 % sure I would merge this pull request as it is.

The reason is that making the functions array-friendly complicates the code a little bit, and ties the implementation to NumPy and its quirks. Before going on, would you like to do a quick comparison between

a) Your approach in this pull request, and
b) What we have in `master`, + a function like `coe2rv_many` that effectively takes an array of values, and then calls `coe2rv` in a `for` loop?

I think we could even leverage https://numba.readthedocs.io/en/stable/user/parallel.html to make this blazing fast.

Josvth:
Hi @astrojuanlu,

Yeah, my opinion on whether or not vectorization is the way to go for multi-object propagation has changed since writing my implementation (I think a year ago now). This PR is therefore mainly intended to show what it would take and what mess it would make trying to merge the changes into poliastro master.

I do agree with you on what the way forward would be, with a slight nuance. Embracing for-loops and leveraging the parallelization ability of numba is 100% the way to go for the propagators. I got lucky with the markley propagator that it could be vectorized, give enough accuracy for my application, and would give me the performance I need without going the way of parallelization. Parallelization of for loops around the propagators would likely give you 90% of the same performance but it would work instantly across all of them without complicating the code.

That said, I think I would make the case for having a vectorized capability for geometry functions like `rv_pqw(...)` and `coe2rv(...)` (preferably also `rv2coe(...)` to have the full set). I was kind of hoping you would see a nicer jit compatible solution and I was missing something being not too experienced in numba. What we need is a way to let the functions return a (3,) 1d-vector in case of scalar inputs and a (n,3) 2d-array in case of (n,) 1d-vector inputs. I've looked at https://numba.pydata.org/numba-doc/latest/user/vectorize.html, I don't think `@vectorize` would work in this case and `@guvectorize` is more appropriate. but that would completely change the function appearance. My point for vectorization is that if you want to do something simple as sampling a multitude of orbits (or one orbit at multiple anomalies), calling multiple threads to do so (although automated by numba) seems a bit too much lazy relying on parallaziation to increase performance. I am also thinking of use-cases such as platforms that do not have multiple threads available, or platforms that could harness vectorization (i.e. gpu's) but both might be totally out of scope here.

I remember when starting my implementation that the main performance hit was a back and forth conversion from rv->coe->propagate nu->coe->rv. Seems like that problem is now fixed by https://github.com/poliastro/poliastro/pull/1037 So it might not be as bad going the way of for loops. Embracing for-loops and keeping performance by parallaziation keeps code clean and readable. And clean understandable code is what got me to poliastro in the first place. 

Having separate API that can harness parallaziation while keeping the master code clean would be the right path forward. Also for cases like mine. It is just beyond my scope unfortunately.

Nevertheless, the `Orbit.sample(...)` method needs some clean-up. I can write a separate PR with a for loop for that one.

Josvth:
@astrojuanlu I think something like this is what you imagined: https://github.com/Josvth/poliastro/commit/a46a4b3f4493b89179b29f46c7ef3f7505c8a1c5

I actually really like it. haven't done a speed test. But the syntax very simple and it worked straight away with `parallel=True`

astrojuanlu:
"Parallelization might not be that bad actually." 😄 

Yes, indeed! I see you started working on a new branch, will wait for you to make the final call.

I have just started using `@njit(parallel=True)` for another project, and I'm so far loving it!

Josvth:
@astrojuanlu I made a PR for the parallel implementation in #1048. I suggest we close this one as its not the way to go.

PR #1048, opened 2020-12-20
Parallelized coe2rv_many

Josvth:
As discussed in #1046 here is the slick and clean implementation of a parallelized coev2rv_many.

`orbit.sample(...)` uses the new parallelized implemenation.

### DIFF
--- a/src/poliastro/twobody/orbit.py
+++ b/src/poliastro/twobody/orbit.py
@@ -21,6 +21,7 @@ from poliastro.frames.util import get_frame
 from poliastro.threebody.soi import laplace_radius
 from poliastro.twobody.propagation import farnocchia, propagate
 
+from ..core.elements import coe2rv_many
 from ..core.propagation.farnocchia import delta_t_from_nu as delta_t_from_nu_fast
 from ..util import find_closest_value, norm
 from ..warnings import OrbitSamplingWarning, PatchedConicsWarning, TimeScaleWarning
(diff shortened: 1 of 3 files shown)


src/poliastro/twobody/orbit.py
```diff
@@ -1416,16 +1417,24 @@ def sample(self, values=100, *, min_anomaly=None, max_anomaly=None):
         else:
             nu_values = self._sample_open(values, min_anomaly, max_anomaly)
 
-        # FIXME: Refactor this insanity to remove a lot of back and forth
```
astrojuanlu, line 1419: Yessssssssssssss 😀

src/poliastro/twobody/orbit.py
```diff
@@ -1416,16 +1417,24 @@ def sample(self, values=100, *, min_anomaly=None, max_anomaly=None):
...
+            np.full(n, self.ecc.value),
+            np.full(n, self.inc.to(u.rad).value),
+            np.full(n, self.raan.to(u.rad).value),
+            np.full(n, self.argp.to(u.rad).value),
+            nu_values.to(u.rad).value,
+        )
+
+        # Add units
+        rr = rr * u.m
+        vv = vv * u.m / u.s
```
astrojuanlu, line 1433: The doctest in line 1405 fails because `rr` now is in `m` instead of `km`. I'd rather convert this to `km` than to change the docstring, because it's what most propagators return already:

https://github.com/poliastro/poliastro/blob/bdd0dd5bc7936774da8ca9d494f914888a761877/src/poliastro/twobody/propagation.py#L159-L161

```suggestion
        rr = (rr << u.m).to(u.km)
        vv = (vv << (u.m / u.s)).to(u.km / u.s)
```

astrojuanlu (approved):
This looks very good! And I'm sure that it accelerated `Orbit.sample` noticeably. I left one tiny comment on a failing docstring, but apart from that this is ready to merge!

astrojuanlu:
Relevant tests passed. Merging!

astrojuanlu:
Thanks a lot @Josvth! 🚀

Issue #1042, opened 2020-11-13
Replace SciPy DOP853 with numbakit-ode


astrojuanlu:
https://github.com/hgrecco/numbakit-ode/blob/main/nbkode/dop853.py

Martinviv:
Hello, I discovered your project, and I find it very interesting. It would be my pleasure to contribute to this project. I am therefore interested in this issue.

astrojuanlu:
Hi @Martinviv, welcome! Take your time to read the code, especially the `cowell` function in poliastro, which is the one using `DOP853`, and don't hesitate to ask questions.

Martinviv:
That's what I'm already doing and I think I already have something that works (it passes the tests "test_propagation.py" on my computer). But I prefer to be sure and take the time to better understand the code before sending the PR

astrojuanlu:
Great to know! :)

Martinviv:
In fact I was mistaken the tests were working on the library and not on my code. But I'm sorry, I'm blocking pretty hard, so I'm going to **abort**  this issue and take time to better understand this project

astrojuanlu:
No rush! :)

hgrecco:
I am really happy to see `numbakit-ode` being adopted. I am really interested to know how it works for other people in other domains. Have you try to benchmark it? If you do, I usually split the benchmark in 3 parts

1. Solver instantiation
2. First step
3. Rest of the steps

The first two are still slow due to numba's inability at the moment to cache LLVM code and have function points. But you should get a speed boost in (3) which pays off for long simulations.

astrojuanlu:
We haven't tried to benchmark it yet, but we'll keep your tips in mind :) Thanks!

jorgepiloto:
For newcomers who want to try this labeled good first issue, here is more information about it:

* In order to propagate an orbit inside an environment where natural or artificial perturbations exist (air drag, J2, solar radiation pressure...) the Cowell method is applied. This integrates the equations of motion using some numerical method, in this case DOP853, which is a Runge-Kutta based one.
* This propagator is located under `poliastro/twobody/propagation.py`. Notice that the `DOP853` is the one coming from `scipy`. You should modify the code in such a way that, the `DOP853` class from `numbakit-ode` is the one being used inside the `cowell` propagator.
* Finally, remember to add the `numbakit-ode` library as a dependency within `pyproject.toml`. Do not forget to run the tests locally, so you can be nothing broke.

[...]

PR #1049, opened 2020-12-27
Changes for accommodating Numbakit-ODE as a replacement for Scipy.integrate module

DhruvSondhi:
Hello, 
This is the changes that I have made to the code to accommodate `Numbakit-ode`. Issues with this code is as follows:
- `events` parameter is inaccessible while execution of the code
- Majority of the test are failing because of differences in the passing of parameters around {numbakit vs scipy}

### DIFF
--- a/src/poliastro/twobody/propagation.py
+++ b/src/poliastro/twobody/propagation.py
@@ -25,6 +25,7 @@ different propagators available at poliastro:
 import numpy as np
 from astropy import units as u
 from astropy.coordinates import CartesianDifferential, CartesianRepresentation
+from scipy.integrate import DOP853, solve_ivp
 
 from poliastro.core.propagation import (
     danby as danby_fast,
@@ -85,7 +86,6 @@ def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, f=func_twobody):
     tofs = tofs.to(u.s).value
 
     u0 = np.array([x, y, z, vx, vy, vz])
-    t0 = 0
 
     result = solve_ivp(
         f,
@@ -111,7 +111,7 @@ def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, f=func_twobody):
         t = tofs[i]
         if t_end is not None and t > t_end:
             t = t_end
-        t, y = result_solver.run(t)
+        y = result.sol(t)
         rrs.append(y[:3])
         vvs.append(y[3:])
 
--- a/src/poliastro/twobody/propagation.py
+++ b/src/poliastro/twobody/propagation.py
@@ -25,7 +25,6 @@ different propagators available at poliastro:
 import numpy as np
 from astropy import units as u
 from astropy.coordinates import CartesianDifferential, CartesianRepresentation
-from scipy.integrate import DOP853, solve_ivp
 
 from poliastro.core.propagation import (
     danby as danby_fast,
@@ -86,6 +85,7 @@ def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, f=func_twobody):
     tofs = tofs.to(u.s).value
 
     u0 = np.array([x, y, z, vx, vy, vz])
+    t0 = 0
 
     result = solve_ivp(
         f,
@@ -111,7 +111,7 @@ def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, f=func_twobody):
         t = tofs[i]
         if t_end is not None and t > t_end:
             t = t_end
-        y = result.sol(t)
+        t, y = result_solver.run(t)
         rrs.append(y[:3])
         vvs.append(y[3:])
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,9 +43,7 @@ requires = [
     "plotly ~=4.0",
     "pyerfa",
     "scipy >=1.4.0",
-    "numbakit-ode",
 ]
-
 
 [tool.flit.metadata.requires-extra]
 jupyter = ["notebook"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,9 @@ requires = [
     "plotly ~=4.0",
     "pyerfa",
     "scipy >=1.4.0",
+    "numbakit-ode",
 ]
+
 
 [tool.flit.metadata.requires-extra]
 jupyter = ["notebook"]


src/poliastro/twobody/propagation.py
```diff
@@ -88,24 +88,23 @@ def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, ad=None, **ad_kwargs):
...
-        rtol=rtol,
-        atol=1e-12,
-        method=DOP853,
-        dense_output=True,
-        events=events,
+        params = p,
+        rtol = rtol,
+        atol = 1e-12,
     )
     if not result.success:
```
DhruvSondhi, line 109: Should be `result_solver` rather than `result`

astrojuanlu:
Hi @DhruvSondhi! Did you attempt to solve the Lotka-Volterra equations first to understand the differences of the two interfaces? My advice is that you publish a notebook in https://gist.github.com/ so we can take a look.

DhruvSondhi:
> Hi @DhruvSondhi! Did you attempt to solve the Lotka-Volterra equations first to understand the differences of the two interfaces? My advice is that you publish a notebook in https://gist.github.com/ so we can take a look.

Yes ... I have tried solving them though I have having some issues which can be discussed after you go through the code here https://gist.github.com/DhruvSondhi/29cc2df6518ed960788255fb6179e44e

src/poliastro/twobody/propagation.py
```diff
@@ -88,24 +88,23 @@ def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, ad=None, **ad_kwargs):
...
+    t0 = 0
 
     # Set the non Keplerian acceleration
     if ad is None:
 
         def ad(t0, u_, k_):
             return 0, 0, 0
 
-    f_with_ad = functools.partial(func_twobody, k=k, ad=ad, ad_kwargs=ad_kwargs)
+    p = [k, ad, ad_kwargs]
```
astrojuanlu, line 99: All the test failures are because of this:

```
E           numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
E           Internal error at <numba.core.typeinfer.CallConstraint object at 0x7f12ad537e80>.
E           Failed in nopython mode pipeline (step: analyzing bytecode)
E           CALL_FUNCTION_EX with **kwargs not supported
E           During: resolving callee type: type(CPUDispatcher(<function func_twobody at 0x7f12ce010af0>))
E           During: typing of call at /home/vsts/work/1/s/.tox/py38-fast/lib/python3.8/site-packages/nbkode/core.py (124)
E           
E           Enable logging at debug level for details.
E           
E           File ".tox/py38-fast/lib/python3.8/site-packages/nbkode/core.py", line 124:
E                       def _rhs(t, y):
E                           return rhs(t, y, params)
E                           ^
```

And I suspect it's because of the presence of `ad_kwargs` there, which is a dictionary.

astrojuanlu (commented):
Thanks @DhruvSondhi for this pull request! However, it looks like we're a little bit blocked here, since `**ad_kwargs` are not supported. I think we should refactor the code even more and remove the ability to pass such kwargs, instructing users to craft their own lambdas or equivalent.

src/poliastro/twobody/propagation.py
DhruvSondhi: This is invalid for the current implementation... will update in the next commit

src/poliastro/twobody/propagation.py
DhruvSondhi: Yes , you are absolutely correct ... the `ad_kwargs` is incompatible with the function signature needed for correct execution of the function `nbkode.DOP853` ... I don't have the exact function call signature of the `ad` function so I will need to check the data type of `ad_kwargs` to confirm it being a dictionary...

DhruvSondhi: While looking in the documentation for `nbkode.DOP853` we can implement a nested list or nested array such as `p = [k, ad, [parameters for ad_kwargs]]` which can be resolved as `parameters` passed in the function is of `np.ndarray` the nature ... Yes, we can create a function for `ad` such that if the `ad_kwargs` doesnt have value we can pass `0` for the same & if it does have something we can instead pass those values but will make it a bit complicated & needs refactoring of `two_body` & tests for the same ...
A suggestion is, if we can execute the `ad` function outside of the execution it may be easier to understand & debug because I am doubting that `nbkode.DOP853` can pass a function as a parameter but need to check the same...

astrojuanlu: Perhaps we should explore a whole different approach than the current one of passing `ad` and `ad_kwargs` to `func_twobody`. For example, we could change the signature of `cowell` to be something like this:

```
def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, f=func_twobody):
```

(let's ignore the `events` for a moment)

In this scenario, the user could do something like:

```
custom_params = ...

def custom_f(t0, u_, k):  # Function signature expected by DOP853, only k allowed as extra parameter
    du_kep = func_twobody(t0, u_, k)  # Assume func_twobody knows nothing about perturbation accelerations

    # custom_ad and custom_params are defined in the outer scope
    # so custom_f can be a closure
    ax, ay, az = custom_ad(t0, u_, k, custom_params)
    du_ad = np.array([0, 0, 0, ax, ay, az])

    # Here we do the same that func_twobody does now with ad
    return du_kep + du_ad

...
cowell(k, r0, v0, tofs, f=custom_f)  # This should work!
```

astrojuanlu: Copied this to #1052

DhruvSondhi: Yes looks promising ... One question ... Where is actually `ad` function defined ? ... Yes this is what I was proposing & can be done easily ... Will start on the same as directed :smile:

astrojuanlu: `ad` can be any of the perturbation accelerations, check out https://docs.poliastro.space/en/latest/examples/Propagation%20using%20Cowells%20formulation.html and https://docs.poliastro.space/en/latest/examples/Natural%20and%20artificial%20perturbations.html

astrojuanlu: This refactor was finally done in gh-1053.

astrojuanlu:
This will need a rebase after gh-1053, or perhaps you want to start over!

astrojuanlu:
As I said in gh-1042, let's pause this effort https://github.com/poliastro/poliastro/issues/1042#issuecomment-753514905

astrojuanlu:
Closing until we resume

DhruvSondhi:
Hello @astrojuanlu, It does seem so that **numbakit-ode** devs have implemented an *Events* parameter & have changed some stuff since the last time we looked into this ... Should we resume the work on this PR? [ I am going through their changes & implementation to get a closer look at how we will be doing the same 😄 ]

astrojuanlu:
Indeed, there have been a few changes in numbakit-ode that we should try out! We can do it in another pull request. In any case, that will be for poliastro 0.16.

PR #1051, opened 2020-12-29
Add Github actions for triggering validation repo

jorgepiloto:
This is related with https://github.com/poliastro/validation/issues/5. It adds a GitHub actions file for triggering the [validation repository](https://github.com/poliastro/validation/) every time a new merge in main branch is done or a release is published.

### DIFF
--- a/.github/workflows/validation_trigger.yml
+++ b/.github/workflows/validation_trigger.yml
@@ -1,0 +1,30 @@
+# This workflow triggers poliastro/validation CI from poliastro/poliastro
+# everytime a new commit is introduced in poliastro:master branch
+name: Validation test cases
+
+on:
+  # A collection of events which trigger the validation action
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+  release:
+
+  # Enable manually to execute the actions
+  workflow_dispatch:
+
+jobs:
+  
+  # Validate poliastro's new features
+  validation:
+    runs-on: ubuntu-latest
+    environment: validation-env
+
+    # Steps to be followed during validation job
+    steps:
+      - name: Trigger poliastro/validation
+        run: |
+          curl -XPOST -u "${{ secrets.PAT_VALIDATION}}:${{secrets.PAT_VALIDATION_TOKEN}}" \
+          -H "Accept: application/vnd.github.everest-preview+json" \
+          -H "Content-Type: application/json" https://api.github.com/repos/poliastro/validation/actions/workflows/build.yaml/dispatches \
+          --data '{"ref": "master"}'
--- a/.github/workflows/validation_trigger.yml
+++ b/.github/workflows/validation_trigger.yml
@@ -26,5 +26,5 @@ jobs:
         run: |
           curl -XPOST -u "${{ secrets.PAT_VALIDATION}}:${{secrets.PAT_VALIDATION_TOKEN}}" \
           -H "Accept: application/vnd.github.everest-preview+json" \
-          -H "Content-Type: application/json" https://api.github.com/repos/poliastro/validation/actions/workflows/build.yaml/dispatches \
+          -H "Content-Type: application/json" https://api.github.com/repos/poliastro/validation/actions/workflows/ci_actions.yml/dispatches \
           --data '{"ref": "master"}'


jorgepiloto:
A couple of things on this:

* [ ]  Tokens still need to be added
* [ ]  I will add the `on: pull_request` for debugging

jorgepiloto:
Alright, the `on pull_request` was added just for debugging purposes. Tokens were not added and thus an error is being raised.

jorgepiloto:
The [validation repo](https://github.com/poliastro/validation/) is [ready](https://github.com/poliastro/validation/pull/11) to run actions when a `repository_dispatch` is recieved.

.github/workflows/validation_trigger.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+
+    # Steps to be followed during validation job
+    steps:
+      - name: Trigger poliastro/validation
+        run: |
+          curl -X POST https://api.github.com/repos/poliastro/validation/dispatches \
+          -H 'Accept: application/vnd.github.everest-preview+json' \
+          -u ${{ secrets.VALIDATION_TRIGGER_TOKEN }} \
+          --data '{"event_type": "ping-validation", "client_payload": { "repository": "'"$GITHUB_REPOSITORY"'" }}'
```
astrojuanlu, line 26: I'm not sure how to proceed. What value should `VALIDATION_TRIGGER_TOKEN` have? Is it something that https://github.com/poliastro/validation/ is expecting? If so, how did you configure it? Is the `ping-validation` something that you should have added in https://github.com/poliastro/validation/pull/11 ? I guess you're using this API https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#create-a-repository-dispatch-event, but I don't see the `-u` parameter.

astrojuanlu (changes requested):
I have lots of questions :)

.github/workflows/validation_trigger.yml
jorgepiloto: The `secrets.TOKEN_NAME` variables are used to reference stored secrets without explicitly calling its value. Thus, a `VALIDATION_TRIGGER_TOKEN` should be created. However, how I see your link to official docs, I am not sure if we require a `token` for this simple task of "pinging" the validation repository.

Regarding the "ping-validation" variable, it is defined within the JSON data one. When complex workflows are defined, those might be useful for debugging or reducing non-useful tasks. For example: if the validation repo actions is triggered by a change in poliastro repo, we now the `code_quality` job should not be applied (since no code has been modified within validation). Thus, we might include add the following to the validation actions file:

```YAML
if github.event.action == "ping-validation"
# do not run the code quality test
```

Finally, parameter -u is the one used for passing user credentials (name + password) within the `curl` command. 

Let me try to refactor this without tokens at all in a new commit and see what happens.

jorgepiloto: This comment might be interesting: https://github.community/t/triggering-by-other-repository/16163/6. They claim that token minimum permission is `public_repo`

astrojuanlu: Hmm so we indeed have to "fake" that the webhook is being called by a user (it can be either you and I) and then we have to create a personal access token as you suggested earlier... https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token okay, I didn't expect that but it makes sense I guess!

jorgepiloto:
Yeah! This is better! See [the message](https://github.com/poliastro/poliastro/pull/1051/checks?check_run_id=1625330286) now under `Trigger poliastro/validation`:

```bash
 {
  "message": "Must have admin rights to Repository.",
  "documentation_url": "https://docs.github.com/rest/reference/actions#create-a-workflow-dispatch-event"
}
```

jorgepiloto:
However I am not sure which is causing this: the security I imposed to the token (only repo access) when it was defined or my user, as it is linked to the token... :thinking:

astrojuanlu:
I think you missed this https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idenvironment on b69115ce5f20a, that's why the variables were empty and the command did not have enough permissions https://github.com/poliastro/poliastro/runs/1625330286#step:2:2

jorgepiloto:
It's working!!! :rocket: 

I forgot to properly point to the actual YAML file hosting the actions in the validation repo. This was the last reason behind the actions not being triggered. Also, notice in the [logbook of validation actions](https://github.com/poliastro/validation/actions) those appear as "manually executed by me".

Now that the backbone of this issue has been completed, I would like to check the following before this gets merged:

* poliastro's actions succeed even before the validation test have been properly executed. This is due to the fact that  they only need to make a simple `curl`. If the command succeeds, everything is assumed to be alright. We could explore the possibility of waiting for a confirmation message from the validation repo and raise some kind of error if those.

.github/workflows/validation_trigger.yml
astrojuanlu: Resolved!

astrojuanlu (approved):
I like your idea @jorgepiloto of trying to make the job fail if the validation pipeline fails. But I still think this is a good incremental improvement. If you agree, let's merge this and you can continue in another PR :)

jorgepiloto:
Sure! Let us merge this and expand its capabilities in a future PR :rocket:

jorgepiloto:
Cool! As soon as I merged this into master, the `validation` was triggered again. Working as expected :wink:

Issue #1052, opened 2020-12-29
Proposal for cowell refactor


astrojuanlu:
_(Motivated by #1042 and #1049, and copy-pasting from https://github.com/poliastro/poliastro/pull/1049#discussion_r549859041)_

Perhaps we should explore a whole different approach than the current one of passing `ad` and `ad_kwargs` to `func_twobody`. For example, we could change the signature of `cowell` to be something like this:

```python
def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, f=func_twobody):
```

(let's ignore the `events` for a moment)

In this scenario, the user could do something like:

```python
custom_params = ...

def custom_f(t0, u_, k):  # Function signature expected by DOP853, only k allowed as extra parameter
    du_kep = func_twobody(t0, u_, k)  # Assume func_twobody knows nothing about perturbation accelerations

    # custom_ad and custom_params are defined in the outer scope
    # so custom_f can be a closure
    ax, ay, az = custom_ad(t0, u_, k, custom_params)
    du_ad = np.array([0, 0, 0, ax, ay, az])

    # Here we do the same that func_twobody does now with ad
    return du_kep + du_ad

...
cowell(k, r0, v0, tofs, f=custom_f)  # This should work!
```

This refactor can be done regardless of #1049 if we think it's a good idea anyway. However, it mostly makes sense if it makes #1049 easier.

What do you think @jorgepiloto?

astrojuanlu:
Done in gh-1053.

PR #1053, opened 2020-12-30
Refactor cowell

astrojuanlu:
See discussion in gh-1052. Did not run the "slow" tests or the notebooks, they will probably fail.

### DIFF
--- a/src/poliastro/twobody/events.py
+++ b/src/poliastro/twobody/events.py
@@ -25,7 +25,7 @@ class LithobrakeEvent:
     def last_t(self):
         return self._last_t * u.s
 
-    def __call__(self, t, u):
+    def __call__(self, t, u, k):
         self._last_t = t
         H = norm(u[:3])
         # SciPy will search for H - R = 0
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,7 +9,7 @@ norecursedirs =
 python_files =
     test_*.py
 doctest_plus = disabled
-addopts = --strict
+addopts = --strict-markers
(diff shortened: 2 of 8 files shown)


astrojuanlu:
As expected, MyPy and the slow tests (including the docs) failed. But I actually like how this simplifies the code, and the burden on the user side is only moderate. We could add a helper to assemble several perturbations together (or just leave it like that).

jorgepiloto:
Congrats on this, the implementation is very clean. I really like this new approach, as now the `func_twobody` is the "true" two-body function. In addition, looks like all the test passed :clap:

src/poliastro/twobody/events.py
```diff
@@ -25,7 +25,7 @@ def terminal(self):
     def last_t(self):
         return self._last_t * u.s
 
-    def __call__(self, t, u):
+    def __call__(self, t, u, k):
```
jorgepiloto, line 28: Is this `k` ever used? :thinking:

jorgepiloto (changes requested):
Just left one comment on a parameter which seems not to be used inside a `__call__` method. All the rest modifications look good by my side :+1:

src/poliastro/twobody/events.py
astrojuanlu: Not in the `LithobrakeEvent`, but potentially in other events! Also, makes the calling convention of the objective function and the events consistent.

astrojuanlu:
Merging! Welcome to the future 🚀

Issue #1009, opened 2020-09-05
Create `Ephem` object from an `Orbit`


astrojuanlu:
> Actually there is one thing "missing" from my perspective, but not in the docs: A clean, fast and out-of-the-box way of computing an `Ephem` object from an `Orbit` object, e.g. `Ephem.from_orbit(orb, time_stamps, propagator, ...)`.

_Originally posted by @s-m-e in https://github.com/poliastro/poliastro/issues/1008#issuecomment-685949410_

Ibnmardanis24:
Hi, I am new to poliastro but I have had some interest on python and mission analysis for a while now. Getting interested on your repository is only natural. I was wondering if I could solve this issue and get to know your project a bit more in the process.

I'm not a huge advocate for OOP so I'm having a bit of a hard time grasping the exact change that you are proposing here. I get from my own formation in celestial mechanics that the objects 'Orbit' and 'Ephem' only differ by the amount of points in the orbit that you sample. But essentialy an 'Ephem' object is really a colection of 'Orbit' objects, right? I get that the object 'Orbit' already features a method to create it from an 'Ephem' object. So what you would be hoping is for the 'reverse' process. This is possible since a state vector produces a single posible motion.

Following this argument, there are still 2 issues that intringue me: I'm not entirely certain on how many attributes I would need to produce from the info in the Orbit object to fully define an Ephem object. Additionally, my initial approach to generate an Ephem from an Orbit is to propagate the trajectory (either forwards or backwards) a given number of time steps. I guess the parameters of this propagation should be user defined right? Even though they would refer to the same family of Ephem.

Regards.

astrojuanlu:
Hello @Ibnmardanis24 , welcome! Let me answer your questions inline.

> I get from my own formation in celestial mechanics that the objects 'Orbit' and 'Ephem' only differ by the amount of points in the orbit that you sample. But essentialy an 'Ephem' object is really a colection of 'Orbit' objects, right?

You could say that... if you knew what the attractor is :) But in reality `Ephem` objects are a list of coordinates without any attractor attached. Given an attractor, you can assume that it's a collection of **osculating** orbits.

> I get that the object 'Orbit' already features a method to create it from an 'Ephem' object. So what you would be hoping is for the 'reverse' process.

Exactly, and this would be equivalent to `propagate`-ing the `Orbit`, sampling it for the specified `timestamps`, and storing the result in an `Ephem`.

[...]

PR #1056, opened 2021-01-01
Add function to create `Ephem` from `Orbit`

Yash-10:
Addresses #1009 

Added a function to create an `Ephem` object from an `Orbit` object.

- Propagates the specified `orbit` using the `epochs` argument.
- Samples the orbit with interval points same as the size of `epochs`.
- Uses the returned coordinates to create an `Ephem` object.
- Checks for scalar epochs.
- Not added corresponding tests.

### DIFF
--- a/src/poliastro/ephem.py
+++ b/src/poliastro/ephem.py
@@ -16,11 +16,11 @@ from astropy.time import Time
 from astroquery.jplhorizons import Horizons
 from scipy.interpolate import interp1d
 
+from poliastro.twobody.propagation import propagate
+
 from .frames import Planes
 from .frames.util import get_frame
 from .warnings import TimeScaleWarning
-
-from poliastro.twobody.propagation import propagate
 
 EPHEM_FORMAT = (
     "Ephemerides at {num} epochs from {start} ({start_scale}) to {end} ({end_scale})"
@@ -311,7 +311,12 @@ class Ephem:
         return cls(coordinates, epochs, plane)
 
     @classmethod
-    def from_orbit(cls, orbit, epochs, plane=Planes.EARTH_EQUATOR,):
+    def from_orbit(
+        cls,
+        orbit,
+        epochs,
+        plane=Planes.EARTH_EQUATOR,
+    ):
         """Return 'Ephem` from an `Orbit` at certain epochs.
         Parameters
         ----------
--- a/src/poliastro/ephem.py
+++ b/src/poliastro/ephem.py
@@ -20,6 +20,8 @@ from .frames import Planes
 from .frames.util import get_frame
 from .warnings import TimeScaleWarning
 
+from poliastro.twobody.propagation import propagate
+
 EPHEM_FORMAT = (
     "Ephemerides at {num} epochs from {start} ({start_scale}) to {end} ({end_scale})"
 )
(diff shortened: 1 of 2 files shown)


src/poliastro/ephem.py
```diff
@@ -308,6 +308,27 @@ def from_horizons(
...
+            Epochs to sample the orbit positions.
+        plane: ~poliastro.frames.Planes, optional
+            Fundamental plane of the frame, default to Earth Equator.
+
+        """
+        if epochs.isscalar:
+            epochs = epochs.reshape(1)
+
+        new_orbit = orbit.propagate(epochs)
+        coordinates = new_orbit.sample(epochs.size)
```
astrojuanlu, line 328: I don't think `orbit.propagate(epochs)` works. The effect of this `propagate + sample` is essentially using `poliastro.twobody.propagation.propagate` with `epochs`.

astrojuanlu (changes requested):
Thanks for this pull request @Yash-10! I added one comment that needs to be addressed, and also we will need tests for this.

src/poliastro/ephem.py
Yash-10: Thank you @astrojuanlu for your reply! I am slightly confused as to why `orbit.propagate(epochs)` wouldn't work since it would create a new orbit with parameters corresponding to the passed in `epochs` and that `propagate(...)` will find out the `time_of_flight`. 

What I suspect is that `new_orbit.sample(epochs.size)` must return the `coordinates` along with the velocities (as per the constructor for `Ephem`) but my current implementation seems to miss that.

I am sorry if I am missing something quite trivial. I must figure it out. Also, I shall soon work on building the tests as well!

Thank you.

astrojuanlu: > I am slightly confused as to why orbit.propagate(epochs) wouldn't work since it would create a new orbit with parameters corresponding to the passed in epochs

Your assumption is incorrect - an `Orbit` corresponds to a single `epoch`, and can't have an array of `epochs`.

Yash-10: Thank you for correcting me! I should instead propagate the `orbit` according to each epoch in `epochs` and then `sample` the corresponding resulting coordinates (in total `epochs.size` coordinates) from all the propagated orbits which can then be used to create the required `Ephem` object, isn't it?

astrojuanlu: Instead of using `Orbit.sample` or `Orbit.propagate`, I think your best bet will be `poliastro.twobody.propagation.propagate`

astrojuanlu:
This looks really good! Before I do a proper review, please rebase & force push on top of current `master`, I merged a couple of fixes for the Continuous Integration

Yash-10:
> This looks really good! Before I do a proper review, please rebase & force push on top of current `master`, I merged a couple of fixes for the Continuous Integration

Thanks for your suggestion on using `twobody.propagate`. I have done the rebasing and force pushing. Please let me know any improvements/suggestions on the tests that can be done and also we would need to fix the 2 failing checks.

Thanks!

astrojuanlu (approved):
This is good to go! But the quality check must pass. This can be done by running `tox -e reformat`

Yash-10:
Thanks for the reformatting command! As a result, the quality checks have passed!

astrojuanlu:
Merged! Thanks @Yash-10 !

Yash-10:
Thank you for merging!! Learning a lot of things and am having a great experience.

PR #1057, opened 2021-01-01
Use tox 4

astrojuanlu:
This should fix a handful of bugs and problems we have with the current (unmaintained) version of tox 🎉

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ pip-delete-this-directory.txt
 # Unit test / coverage reports
 htmlcov/
 .tox/
+.tox4/
 .nox/
 .coverage
 .coverage.*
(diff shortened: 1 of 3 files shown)


astrojuanlu:
Looks like at least the basic commands are working!

astrojuanlu:
Two caveats:

- `docs` is marked as `SKIP` (what does that mean?)
- "factors" https://tox.readthedocs.io/en/latest/config.html#factors-and-factor-conditional-settings don't seem to work like before

astrojuanlu:
`docs` was being skipped because the interpreter was missing. Fixed! factors still seem not to work.

However, it looks like I can backport some of these improvements with the current version of tox and fix the ongoing issues. I'll try in a separate PR.

astrojuanlu:
Pending: https://github.com/tox-dev/tox/issues/1779

Also, this is not urgent anymore, because we've solved our CI problems in a different way. No rush for trying tox 4, will wait for the stable version to appear.

Issue #1036, opened 2020-10-16
Wheel test broken


astrojuanlu:
> Hmm `wheels_universal` failure unrelated, investigating. @epassaro what do you think?

Seems Azure (the OA template to be precise) is installing and running Python 3.9 but poliastro is set <3.9. See [this line](https://dev.azure.com/poliastro/poliastro/_build/results?buildId=549&view=logs&j=05361be5-3d16-5149-5133-60d95c410e79&t=14ba121e-1fd8-5056-ef2a-bfb13bd77529&l=9).

_Originally posted by @epassaro in https://github.com/poliastro/poliastro/pull/1034#issuecomment-710503295_

astrojuanlu:
In principle there is nothing making poliastro incompatible with Python 3.9... but we didn't test it yet, so I wanted to make sure. I am especially worried about numba support.

epassaro:
I don't know why OA template tries to install latest Python, maybe it's configurable (I don't remember).

saintaardvark:
Hey good poliastro folks -- just wanted to let you know that we've seen this issue in [Polaris as well](https://gitlab.com/librespacefoundation/polaris/polaris/-/issues/124), where we have poliastro as one of our upstream dependencies.  If there is anything we can do to help test poliastro on 3.9, please let me know -- I'd be more than happy to help!

astrojuanlu:
Hi @saintaardvark ! 

>  If there is anything we can do to help test poliastro on 3.9, please let me know

In principle, if you get Numba to install on Python 3.9, poliastro should work without problems. We were waiting for the project to provide manylinux wheels (that way we avoid having to compile llvmlite against LLVM), which should happen around mid-December as per https://github.com/numba/numba/issues/6345#issuecomment-721738251.

The other reason why we set poliastro to require Python 3.9 is... because people don't know how to use pip :) They will install Python 3.9, try `pip install poliastro`, see a huge red wall of text, and complain here. I've seen this in other projects, and unfortunately pip (or numba, or llvmlite, whatever) do not provide a nice error message saying " :sparkles: Howdy! You are trying to install a package that requires compiling some binary extensions, but apparently you don't have the necessary dependencies in your operative system, and this failed :sparkles: ".

I'm tracking 3.9 support here https://github.com/poliastro/poliastro/pull/1027 when this pull request passes, I will declare that poliastro is compatible with Python 3.9.

[...]

PR #1060, opened 2021-01-04
Fix CI

astrojuanlu:
With the lessons learned from gh-1057, we try to _finally_ fix the CI, once and for all.

Aims to fix gh-1036 as well.


astrojuanlu:
- sdist and wheels are not broken anymore 🎉 
- docs issue finally surfaced

Merging this, will fix the latter in another PR.

PR #277, opened 2017-10-27
Test on PyPy3

Juanlu001:
Addressed the procedure proposed in https://github.com/poliastro/poliastro/issues/98#issuecomment-339494217. Will allow failures if there are upstream failures that cannot be fixed by us.

Issue #98, opened 2016-01-12
Test on PyPy


astrojuanlu:
Recent PyPy versions are not directly available on Travis CI but there is an easy way to install them in https://github.com/travis-ci/travis-ci/issues/5027#issuecomment-170939193.

Juanlu001:
(Low priority and definitely shouldn't make the build fail, because probably poliastro won't work well on PyPy due to NumPy)

Juanlu001:
To install PyPy with conda: https://github.com/rvianello/conda-pypy

Juanlu001:
Closing this until PyPy reaches Python 3.5 support. https://morepypy.blogspot.com.es/2016/08/pypy-gets-funding-from-mozilla-for.html

Juanlu001:
Reopening this, as PyPy3.5 has just been released:

https://morepypy.blogspot.com.es/2017/03/pypy27-and-pypy35-v57-two-in-one-release.html

While this version gains official support on Travis CI (see https://github.com/travis-ci/travis-ci/issues/7507), we can probably download a precompiled binary:

http://pypy.org/download.html#python-3-5-3-compatible-pypy3-5-v5-7

or a portable version, as suggested in https://github.com/travis-ci/travis-ci/issues/6727#issuecomment-279763549.

Another option would be to use pyenv, as it is about to support the latest version:

https://github.com/pyenv/pyenv/pull/871

Juanlu001:
To set up pyenv in Travis CI see

github.com/praekeltfoundation/travis-pyenv

And to allow some Travis CI configurations to fail, see

https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

Juanlu001:
Perhaps @humitos can lend me a hand with this one? :)

humitos:
Sorry, I got lost here.

What do you want me to do?

* Install pypy with conda in my own pc and run the tests

```
[18:16] humitos: ~> conda create -n pypy --channel ric/channel/pypy --override-channels python pip
Fetching package metadata ...........
Solving package specifications: .

Package plan for installation in environment /home/humitos/.miniconda3/envs/pypy:

The following NEW packages will be INSTALLED:

    bzip2:      1.0.6-1            ric/channel/pypy
    expat:      2.1.0-0            ric/channel/pypy
    libffi:     3.0.13-2           ric/channel/pypy
    ncurses:    5.9-1              ric/channel/pypy
[...]

PR #1061, opened 2021-01-04
Make numba mandatory

astrojuanlu:
At the beginning we were keeping numba optional because it was difficult to install. It's not anymore, wheels are easily available for all platforms shortly after release. We acknowledged this in https://github.com/poliastro/poliastro/pull/221, but then we restored the conditional nature of numba on https://github.com/poliastro/poliastro/pull/277/ to be able to test it on PyPy.

However, the performance of poliastro on PyPy was much slower than CPython + numba https://github.com/poliastro/poliastro/issues/98#issuecomment-469597401 therefore it's not realistic to keep the extra burden because of PyPy at the moment.

In fact, numba might run on PyPy one day https://github.com/numba/numba/issues/5766 and therefore making it conditional makes even less sense.

As a result, this reverts again to having numba as a required dependency everywhere, hence simplifying the code.

Besides, I also removed an unused module and moved it to a separate gist, and I expect to fix the documentation builds with this PR.


astrojuanlu:
Notebooks are failing because of https://github.com/poliastro/poliastro/pull/1053

astrojuanlu:
Docs are fixed, finally 🎉 Will merge straight away, and CI should be green afterwards.

PR #1067, opened 2021-01-08
Fixed the failing docs 

DhruvSondhi:
Hello, 
This PR aims to fix the failing docs. Checked on Local machine, Works as intended !!

PR #1069, opened 2021-01-08
Fix docs, 2nd attempt

astrojuanlu:
Debugging, based on https://github.com/poliastro/poliastro/pull/1067


astrojuanlu:
RTD passed. Merging... But I suspect `master` is going to fail.

PR #1080, opened 2021-01-20
Fixed the binder notebook links

DhruvSondhi:
Hello, This PR fixes the broken binder links to the modified notebooks 😄


DhruvSondhi:
Triggering pipeline for an unrelated issue :)

astrojuanlu:
Yes, this fixes the links. But does it fix _opening_ the notebooks in Binder? I tried it and I think it doesn't. Please confirm.

DhruvSondhi:
I can confirm that the notebooks are not even being recognised by *binder* & hence they are not even opening. Also, checking the URL of the notebooks in notebooks suggest that it is still finding the notebooks in `.md` format & it is not updating the changed references. Very well could be the problem with `Jupytext` extension (nice catch 😉) 🤔

astrojuanlu:
Closing in favor of #1082.

Issue #1074, opened 2021-01-12
Reduce number of warnings in tests


astrojuanlu:
The number of warnings at the end of our tests is very large. We should either:

- Address them. For example, there are a few legitimate Astropy deprecation warnings that will bite us soon.
- Rework our tests so they don't incur in warnings.
- Ignore them explicitly, if they are not important.

And then we could turn warnings into errors.

astrojuanlu:
This was partially addressed in #1092, but there are more warnings left.

DhruvSondhi:
I had checked out a thread regarding the same (would link it in an edit) ... I do think so the remaining warning (all relating to the ERFA `utctai` & `taiutc`) can be suppressed as the time is in the future so it doesn't account for the leap second addition or deletion that may happen ( this year we may have a negative leap second 😉)

astrojuanlu:
Indeed! We need to stop using UTC timescale for times in the distant future.

Yash-10:
Also, one quick fix to solve this warning which I realized afterwards:
```
AstropyDeprecationWarning: The truth value of a Quantity is ambiguous. In the future this will raise a ValueError.
```
is to change
```diff
- elif not (self.attractor.R and self.attractor.R_polar):
+ elif not (hasattr(self.attractor, "R") and hasattr(self.attractor, "R_polar")):
```
in [`extract_czml.py`](https://github.com/poliastro/poliastro/blob/master/src/poliastro/czml/extract_czml.py#L80) since performing logical operations on `Quantity` objects gives that warning.

Thanks!

astrojuanlu:
Or rather, to check that they're non None or non zero 👍🏽

Yash-10:
Oh yes, sorry. Thank you for correcting!

ishanSrt:
I'd like to work upon this 😉

astrojuanlu:
@ishanSrt Any doubts? :)

astrojuanlu:
@ishanSrt started #1137, but it's unclear if they still intend to work on it. If anyone wants to reboot this work, please consider cloning their branch so attribution is preserved.

astrojuanlu:
Notice that, to turn the warnings into errors, one can do this:

```
juanlu@arion2:~/Projects/LSF/poliastro/library$ tox -e py38 -- -W error::erfa.ErfaWarning -k "plot_outer_solar_system"
[...]

PR #1092, opened 2021-01-27
Reduce test warnings

Yash-10:
In reference to #1074, this PR aims to reduce test warnings. Some warnings might need further treatment to mitigate/suppress them but this implementation fixes some of the obvious ones.

Thanks!

### DIFF
--- a/tests/tests_plotting/test_static.py
+++ b/tests/tests_plotting/test_static.py
@@ -62,7 +62,7 @@ def test_plot_trajectory_sets_label():
 
     op = StaticOrbitPlotter()
     trajectory = churi.sample()
-    op.plot_body_orbit(Mars, J2000, label="Mars")
+    op.plot_body_orbit(Mars, Time(J2000, scale="tdb"), label="Mars")
 
     op.plot_trajectory(trajectory, label=expected_label)
 
@@ -88,7 +88,7 @@ def test_redraw_makes_attractor_none():
 def test_set_frame_plots_same_colors():
     # TODO: Review
     op = StaticOrbitPlotter()
-    op.plot_body_orbit(Jupiter, J2000)
+    op.plot_body_orbit(Jupiter, Time(J2000, scale="tdb"))
     colors1 = [orb[2] for orb in op.trajectories]
     op.set_body_frame(Jupiter)
     colors2 = [orb[2] for orb in op.trajectories]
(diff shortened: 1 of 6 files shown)


Yash-10:
I am sorry for a bunch of small pushes..

tests/tests_plotting/test_static.py
```diff
@@ -62,7 +62,7 @@ def test_plot_trajectory_sets_label():
 
     op = StaticOrbitPlotter()
     trajectory = churi.sample()
-    op.plot_body_orbit(Mars, J2000, label="Mars")
+    op.plot_body_orbit(Mars, Time(J2000, scale="tdb"), label="Mars")
```
astrojuanlu, line 65: Let's use `J2000_TDB` instead https://github.com/poliastro/poliastro/blob/a43006b0cf6b9671128ff4e4ed82f45bea6d5205/src/poliastro/constants/general.py#L104-L106

astrojuanlu, line 65: Let's use `J2000_TDB` instead https://github.com/poliastro/poliastro/blob/a43006b0cf6b9671128ff4e4ed82f45bea6d5205/src/poliastro/constants/general.py#L104-L106

tests/tests_twobody/test_orbit.py
```diff
@@ -173,7 +173,9 @@ def test_apply_maneuver_changes_epoch():
 def test_orbit_from_ephem_with_no_epoch_is_today():
     # This is not that obvious http://stackoverflow.com/q/6407362/554319
     body = Earth
-    ss = Orbit.from_body_ephem(body)
+    epoch = Time.now().tdb
+    ephem = Ephem.from_body(body, epoch)
+    ss = Orbit.from_ephem(Sun, ephem, epoch)
```
astrojuanlu, line 178: The tests of `Orbit.from_body_ephem` require further discussion.

If we are still using it internally, even if it's deprecated, it should be tested. And if it has to be tested, then we should silence the deprecation warnings.

But if we are not using it internally anymore, then this approach is on the right path, but I would try not to call `Ephem.from_body` (and use a custom, dummy `Ephem`).

tests/tests_twobody/test_orbit.py
```diff
@@ -1173,6 +1178,7 @@ def test_orbit_change_attractor_out_of_SOI():
     assert "ValueError: Orbit is out of new attractor's SOI" in excinfo.exconly()
 
 
+@pytest.mark.filterwarnings("ignore::poliastro.twobody.orbit.PatchedConicsWarning")
```
astrojuanlu, line 1181: Instead of filtering the warning, should we assert that it's emitted?

astrojuanlu (changes requested):
Thanks a lot @Yash-10 ! This is on the right path. Some fixes are good as they are and I didn't say anything, for some others I left a few comments.

astrojuanlu (approved):
Looks good to me!

astrojuanlu:
As you say, there are still some warnings left but this is strictly better than before. Thanks a lot!

Yash-10:
Thanks @astrojuanlu! Yes, those warnings which include `ERFA`, `Orbit.pqw` would need further treatment but can be dealt in the future too!

Thanks!!

Issue #1029, opened 2020-10-08
peturbation_body and star arguments in third_body and radiation_pressure peturbations


janstruzinski:
Hi! 
I am Aerospace student working on systems engineering project - the subject is spacecraft orbiting Ceres. I wanted to calculate the magnitudes of third body and pressure radiation perturbations. However, it is not clear what peturbation_body and star arguments should be. Would You mind elaborating on it?

PS The tool is great!

astrojuanlu:
Hi @janstruzinski, and sorry for the delay! I totally missed this notification.

Take a look at these:

https://github.com/poliastro/poliastro/blob/c3836dcee6cbda034d8cf6390b386d0a9f455d8c/tests/tests_twobody/test_perturbations.py#L613-L655

https://github.com/poliastro/poliastro/blob/a41c141bbc3013209a4418df22ef311aa7f30f53/docs/source/examples/Natural%20and%20artificial%20perturbations.md#3rd-body

Let us know if you need further assistance!

DhruvSondhi:
Hello, 
After checking through the issue I was thinking what is needed here is a better explanation of the `third_body` & `radiation_pressure` through the docs so that it's easier for anyone to get the correct sense of the way the function tends to behave 🤔 ... What is the outcome that we are aiming for?

astrojuanlu:
We are referring to these two functions:

- https://docs.poliastro.space/en/latest/autoapi/poliastro/core/perturbations/index.html?highlight=perturbation_body#poliastro.core.perturbations.third_body
- https://docs.poliastro.space/en/latest/autoapi/poliastro/core/perturbations/index.html?highlight=star#poliastro.core.perturbations.radiation_pressure

We need one (or both!) of the following:

- A way to say in Sphinx what the signature of the function should be. I don't know if there are any standards for this, maybe there aren't. We could check what other projects do.
- A short example so the user sees the function signature.

DhruvSondhi:
The following comes into my mind (please do suggest alternates)
- To tackle the latter point, we can extend the same functionality in the here in this [notebook](https://docs.poliastro.space/en/latest/examples/Natural%20and%20artificial%20perturbations.html#3rd-body).
- And, for the former point, I think its better that we keep a simple _**Note**_ for the exact meaning of function in consideration & then, for example, link the user to the same example or mode of a demonstration for checking out the implementation in this fashion so that it's clear that the user has to tinker around with the code to get the desired result 🤔

[...]

PR #1096, opened 2021-01-29
Clarification for callable in perturbations module 

DhruvSondhi:
Hello,
This PR aims to document the main idea behind the callables in #1029 
This is a work in progress effort :)

### DIFF
--- a/src/poliastro/core/perturbations.py
+++ b/src/poliastro/core/perturbations.py
@@ -218,6 +218,29 @@ def third_body(t0, state, k, k_third, perturbation_body):
         gravitational constant, (km^3/s^2)
     perturbation_body: a callable object returning the position of the pertubation body that causes the perturbation
 
+    Example
+    -------
+    For clarifying the nature of **perturbation_body** argument in **third_body()** function, we can look into the following example:
+
+    >>> from poliastro.ephem import build_ephem_interpolant
+    >>> from poliastro.bodies import Moon
+    >>> from astropy.time import Time, TimeDelta
+    >>> epoch = Time(
+    ...    2454283.0, format="jd", scale="tdb"
+    ...     )  # setting the exact event date is important
+    >>> body_r = build_ephem_interpolant(
+    ...   Moon, 28 * u.day, (epoch.value * u.day, epoch.value * u.day + 60 * u.day), rtol=1e-2
+    ... )
+    >>> type(body_r)
+    <class 'scipy.interpolate.interpolate.interp1d'>
+
+    Here, this implementation shows that we can create a **function** which returns the position of the
+    perturbation body (in this case the *Moon*) whose effect we need to take into consideration so as
+    get the accurate value of the acceleration.
+
+    This functionality can also be extended with **class** or **class method** callable objects, just keeping
+    in mind that it should return the position of the perturbation body taken into consideration.
+
     Note
     ----
     This formula is taken from Howard Curtis, section 12.10. As an example, a third body could be
@@ -254,7 +277,29 @@ def radiation_pressure(t0, state, k, R, C_R, A_over_m, Wdivc_s, star):
     Wdivc_s : float
         total star emitted power divided by the speed of light (W * s / km)
     star: a callable object returning the position of star in attractor frame
-        star position
+
+    Example
+    -------
+    For clarifying the nature of **star** argument in **radiation_pressure()** function, we can look into the following example:
+
+    >>> import functools
+    >>> def normalize_to_Curtis(t0, sun_r):
+    ...     r = sun_r(t0)
+    ...     return 149600000 * r / norm(r)
+    >>> def sun_r():
+    ...     j_date = 2_438_400.5 * u.day
+    ...     tof = 600 * u.day
+    ...     return build_ephem_interpolant(Sun, 365 * u.day, (j_date, j_date + tof), rtol=1e-2)
+    >>> sun_normalized = functools.partial(normalize_to_Curtis, sun_r=sun_r)
+    >>> type(sun_normalized)
+    <class 'functools.partial'>
+
+    Here, this implementation shows that we can create a **function** which returns the position of the
+    star (in this case the *Sun*) whose effect we need to take into consideration so as to
+    calculate the radiation pressure from the star in attractor frame.
+
+    This functionality can also be extended with **class** or **class method** callable objects, just keeping
+    in mind that it should return the position of the star {frame dependent} taken into consideration.
 
     Note
     ----
--- a/src/poliastro/core/perturbations.py
+++ b/src/poliastro/core/perturbations.py
@@ -233,6 +233,19 @@ def third_body(t0, state, k, k_third, perturbation_body):
     ... )
     >>> type(body_r)
     <class 'scipy.interpolate.interpolate.interp1d'>
+    >>> # The created `interpolate` can be used effectively within the `third body` function:
+    >>> def f(t0, state, k):
+    ...     du_kep = func_twobody(t0, state, k)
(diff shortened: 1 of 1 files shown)


DhruvSondhi:
@astrojuanlu check out the documentation thus created ... I think so this is much better ... Though we need to think for the example that is needed for more clarification 😄

src/poliastro/core/perturbations.py
```diff
@@ -217,6 +217,32 @@ def third_body(t0, state, k, k_third, perturbation_body):
     k : float
         gravitational constant, (km^3/s^2)
     perturbation_body: a callable object returning the position of the pertubation body that causes the perturbation
+        .. py:function:: perturbation_body() -> position_of_perturbation_body
+        .. py:class:: Perturbation_body() -> position_of_perturbation_body
```
astrojuanlu, line 221: I don't understand what are these `Perturbation_body` and `Star` classes :/

DhruvSondhi: These can be seen from this [link](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-signatures) ... The normal `function` directive is for representing a function in the documentation for the type expected or more formally, knowing the function signature. I tried to implement a similar thing with the `class` directive so that examples of function & class in callables be stated. 
This is how that `py:function` is represented 
image
This is how that `py:class` is represented 
image

DhruvSondhi: Any updates for your side on how we should proceed so that it becomes more beginner-friendly documentation?

astrojuanlu (changes requested):
Thanks @DhruvSondhi for this pull request and sorry for the delay! I like the examples, but I would remove the `.. py:function` and `.. py:class` lines entirely. With that, we can merge this (please squash the commits into one)

src/poliastro/core/perturbations.py
```diff
@@ -217,6 +217,32 @@ def third_body(t0, state, k, k_third, perturbation_body):
...
+    >>> from poliastro.bodies import Moon
+    >>> from astropy.time import Time, TimeDelta
+    >>> epoch = Time(
+    ...    2454283.0, format="jd", scale="tdb"
+    ...     )  # setting the exact event date is important
+    >>> body_r = build_ephem_interpolant(
+    ...   Moon, 28 * u.day, (epoch.value * u.day, epoch.value * u.day + 60 * u.day), rtol=1e-2
+    ... )
+    >>> perturbation_body = body_r      # passing created callable function to perturbation_body
+    >>> type(perturbation_body)         # for this particular example :)
```
astrojuanlu, line 237:
```suggestion
    >>> type(perturbation_body)
```

src/poliastro/core/perturbations.py
```diff
@@ -254,7 +280,32 @@ def radiation_pressure(t0, state, k, R, C_R, A_over_m, Wdivc_s, star):
...
+    >>> def normalize_to_Curtis(t0, sun_r):
+    ...     r = sun_r(t0)
+    ...     return 149600000 * r / norm(r)
+    >>> def sun_r():
+    ...     j_date = 2_438_400.5 * u.day
+    ...     tof = 600 * u.day
+    ...     return build_ephem_interpolant(Sun, 365 * u.day, (j_date, j_date + tof), rtol=1e-2)
+    >>> sun_normalized = functools.partial(normalize_to_Curtis, sun_r=sun_r)
+    >>> star=sun_normalized     # passing created callable function to star
+    >>> type(star)              # for this particular example :)
```
astrojuanlu, line 300:
```suggestion
    >>> type(star)
```

src/poliastro/core/perturbations.py
```diff
@@ -254,7 +280,32 @@ def radiation_pressure(t0, state, k, R, C_R, A_over_m, Wdivc_s, star):
...
+    >>> import functools
+    >>> def normalize_to_Curtis(t0, sun_r):
+    ...     r = sun_r(t0)
+    ...     return 149600000 * r / norm(r)
+    >>> def sun_r():
+    ...     j_date = 2_438_400.5 * u.day
+    ...     tof = 600 * u.day
+    ...     return build_ephem_interpolant(Sun, 365 * u.day, (j_date, j_date + tof), rtol=1e-2)
+    >>> sun_normalized = functools.partial(normalize_to_Curtis, sun_r=sun_r)
+    >>> star=sun_normalized     # passing created callable function to star
```
astrojuanlu, line 299: This line is not needed? It's just renaming a variable

src/poliastro/core/perturbations.py
```diff
@@ -217,6 +217,32 @@ def third_body(t0, state, k, k_third, perturbation_body):
...
+    >>> from poliastro.ephem import build_ephem_interpolant
+    >>> from poliastro.bodies import Moon
+    >>> from astropy.time import Time, TimeDelta
+    >>> epoch = Time(
+    ...    2454283.0, format="jd", scale="tdb"
+    ...     )  # setting the exact event date is important
+    >>> body_r = build_ephem_interpolant(
+    ...   Moon, 28 * u.day, (epoch.value * u.day, epoch.value * u.day + 60 * u.day), rtol=1e-2
+    ... )
+    >>> perturbation_body = body_r      # passing created callable function to perturbation_body
```
astrojuanlu, line 236: This line is not needed? It's just renaming a variable

astrojuanlu (changes requested):
On the other hand, I double checked the examples and added some comments to remove stuff. We should also showcase how to use the code in the shortest way possible.

src/poliastro/core/perturbations.py
DhruvSondhi: Yes sure but this was important as the argument passed in the requested function is `star` so that it makes sense to the user reading the example :)

src/poliastro/core/perturbations.py
DhruvSondhi: Same as above ... making it relevant to the user based on the function call signature :)

src/poliastro/core/perturbations.py
```diff
@@ -218,6 +218,29 @@ def third_body(t0, state, k, k_third, perturbation_body):
...
+    >>> from poliastro.ephem import build_ephem_interpolant
+    >>> from poliastro.bodies import Moon
+    >>> from astropy.time import Time, TimeDelta
+    >>> epoch = Time(
+    ...    2454283.0, format="jd", scale="tdb"
+    ...     )  # setting the exact event date is important
+    >>> body_r = build_ephem_interpolant(
+    ...   Moon, 28 * u.day, (epoch.value * u.day, epoch.value * u.day + 60 * u.day), rtol=1e-2
+    ... )
+    >>> type(body_r)
```
astrojuanlu, line 234: Can we add an example of how to use the interpolant?

astrojuanlu (changes requested):

src/poliastro/core/perturbations.py
DhruvSondhi: We can add it but it would be very similar or else exactly what we have in the notebook. 🤔

astrojuanlu: Do you still intend to work on this? Otherwise I can try to write something myself

DhruvSondhi: Yes, I can work upon this... Just need some Guidance for the same or the main essence of the example :)

astrojuanlu:
I marked this pull request as draft, because there are no changes since the last review, and I also unsubscribed. Please ask me for a review when it's ready. And if you are still waiting for my guidance on what to write, I have no other option but to ask you to wait a bit more, sorry.

Issue #921, opened 2020-04-19
Avoid unnecessary element conversions in propagation


astrojuanlu:
Summary:

https://github.com/poliastro/poliastro/blob/e57d8c4330e45aa4d2f7d2bd1131d806dc95a373/src/poliastro/twobody/orbit.py#L1332-L1339

At the moment, all propagators are forced to receive cartesian elements and return cartesian elements. This introduces inaccuracies, especially given the fact that most propagators work internally with classical orbital elements.

We should find a way to introduce more intermediate layers so:

* At the highest level, `Orbit.propagate` and `Orbit.sample` work exactly the same
* At the intermediate level, the function `poliastro.twobody.propagation.propagate` can still receive any method and produce a `CartesianRepresentation` object
* At the lowest level, I have a way to call the propagators based on classical elements without intermediate conversions

astrojuanlu:
I'm wondering if `delta_t_from_nu` and `nu_from_delta_t` would be good enough for all the other propagators...!

astrojuanlu:
Prospective contributors: notice that most of the propagators in https://github.com/poliastro/poliastro/tree/master/src/poliastro/core/propagation do something like this:

```python
@jit
def propagator_name(k, r0, v0, tof):
    p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
    ...
    return coe2rv(k, p, ecc, inc, raan, argp, nu)
```

for all of them, create new functions that do

```python
@jit
def propagator_name_coe(k, p, ecc, inc, raan, argp, nu):
    ...
    return nu
```

so we can replace the current ones with

```python
@jit
def propagator_name(k, r0, v0, tof):
    p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
    nu = propagator_name_coe(k, p, ecc, inc, raan, argp, nu)
    return coe2rv(k, p, ecc, inc, raan, argp, nu)
```

Also, notice that https://github.com/poliastro/poliastro/blob/master/src/poliastro/core/propagation/farnocchia.py, but no https://github.com/poliastro/poliastro/blob/master/src/poliastro/core/propagation/markley.py, https://github.com/poliastro/poliastro/blob/master/src/poliastro/core/propagation/vallado.py... therefore, when doing this change, consider creating new modules for the propagators currently in https://github.com/poliastro/poliastro/blob/master/src/poliastro/core/propagation/__init__.py.

astrojuanlu:
[...]

PR #1027, opened 2020-10-05
Add support for Python 3.9

astrojuanlu:

PR #1098, opened 2021-02-06
Allow propagation to use classical orbital elements

Yash-10:
In reference to #921, this implementation attempts to allow propagators to use classical orbital elements for propagation:

- Implemented on `farnocchia,mikkola,markley,pimienta,gooding,danby` propagators.
- `Orbit.propagate` has been given a keyword argument `classical` (default to `False`) which when set to `True` uses `classical` orbital elements for propagation.

In the test added, the position and velocity vectors returned in both cases (`classical=True` and `classical=False`) do not seem to be in agreement with each other currently. That would certainly need more investigation.

Thanks!!

### DIFF
--- a/src/poliastro/core/propagation/__init__.py
+++ b/src/poliastro/core/propagation/__init__.py
@@ -271,7 +271,7 @@ def mikkola_coe(k, p, ecc, inc, raan, argp, nu, tof):
             # Hyperbolic
             nu = F_to_nu(E, ecc)
 
-    return nu
+    return coe2rv(k, p, ecc, inc, raan, argp, nu)
 
 
 @jit
@@ -304,9 +304,9 @@ def mikkola(k, r0, v0, tof, rtol=None):
 
     # Solving for the classical elements
     p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
-    nu = mikkola_coe(k, p, ecc, inc, raan, argp, nu, tof)
+    coe2rv_ = mikkola_coe(k, p, ecc, inc, raan, argp, nu, tof)
 
-    return coe2rv(k, p, ecc, inc, raan, argp, nu)
+    return coe2rv_
 
 
 @jit
@@ -355,7 +355,7 @@ def markley_coe(k, p, ecc, inc, raan, argp, nu, tof):
     E += delta5
     nu = E_to_nu(E, ecc)
 
-    return nu
+    return coe2rv(k, p, ecc, inc, raan, argp, nu)
 
 
 @jit
@@ -388,9 +388,9 @@ def markley(k, r0, v0, tof):
     """
     # Solve first for eccentricity and mean anomaly
     p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
-    nu = markley_coe(k, p, ecc, inc, raan, argp, nu, tof)
+    coe2rv_ = markley_coe(k, p, ecc, inc, raan, argp, nu, tof)
 
-    return coe2rv(k, p, ecc, inc, raan, argp, nu)
+    return coe2rv_
 
 
 @jit
@@ -722,7 +722,8 @@ def pimienta_coe(k, p, ecc, inc, raan, argp, nu, tof):
         + 15 * w
     )
 
-    return E_to_nu(E, ecc)
+    nu = E_to_nu(E, ecc)
+    return coe2rv(k, p, ecc, inc, raan, argp, nu)
 
 
 @jit
@@ -757,9 +758,9 @@ def pimienta(k, r0, v0, tof):
 
     # Solve first for eccentricity and mean anomaly
     p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
-    nu = pimienta_coe(k, p, ecc, inc, raan, argp, nu, tof)
+    coe2rv_ = pimienta_coe(k, p, ecc, inc, raan, argp, nu, tof)
 
-    return coe2rv(k, p, ecc, inc, raan, argp, nu)
+    return coe2rv_
 
 
 @jit
@@ -791,7 +792,9 @@ def gooding_coe(k, p, ecc, inc, raan, argp, nu, tof, numiter=150, rtol=1e-8):
         n += 1
 
     E = M + psi
-    return E_to_nu(E, ecc)
+
+    nu = E_to_nu(E, ecc)
+    return coe2rv(k, p, ecc, inc, raan, argp, nu)
 
 
 @jit
@@ -826,9 +829,9 @@ def gooding(k, r0, v0, tof, numiter=150, rtol=1e-8):
 
     # Solve first for eccentricity and mean anomaly
     p, ecc, inc, raan, argp, nu = rv2coe(k, r0, v0)
-    nu = gooding_coe(k, p, ecc, inc, raan, argp, nu, tof, numiter, rtol)
+    coe2rv_ = gooding_coe(k, p, ecc, inc, raan, argp, nu, tof, numiter, rtol)
 
-    return coe2rv(k, p, ecc, inc, raan, argp, nu)
+    return coe2rv_
 
 
 @jit
@@ -842,7 +845,7 @@ def danby_coe(k, p, ecc, inc, raan, argp, nu, tof, numiter=20, rtol=1e-8):
         M0 = E_to_M(nu_to_E(nu, ecc), ecc)
         M = M0 + n * tof
         nu = M - 2 * np.pi * np.floor(M / 2 / np.pi)
-        return nu
+        return coe2rv(k, p, ecc, inc, raan, argp, nu)
 
     elif ecc < 1.0:
         # For elliptical orbit
@@ -897,7 +900,7 @@ def danby_coe(k, p, ecc, inc, raan, argp, nu, tof, numiter=20, rtol=1e-8):
     else:
         raise ValueError("Maximum number of iterations has been reached.")
 
-    return nu
+    return coe2rv(k, p, ecc, inc, raan, argp, nu)
 
 
 @jit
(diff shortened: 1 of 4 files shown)


Yash-10:
@astrojuanlu Could you please let me know what do you think about this? 

Thanks!!

astrojuanlu:
Hi @Yash-10, sorry for the delay ! Here I am observing the same weird CI failures I saw in #1027. I will open an issue, we'll have to deal with those separately.

Yash-10:
Hello @astrojuanlu, thank you for your reply! Yes, we would have to deal with those CI failures. There was a small unit conversion problem in my earlier commit, which I have also recently changed.

Thanks!!

astrojuanlu:
Hello @Yash-10 ! Are you able to see the build failures and the documentation warnings? For example, on the Read the Docs build I see this:

```
/home/docs/checkouts/readthedocs.org/user_builds/poliastro/checkouts/1098/docs/source/autoapi/poliastro/core/propagation/farnocchia/index.rst:133: WARNING: Unexpected indentation.
/home/docs/checkouts/readthedocs.org/user_builds/poliastro/checkouts/1098/docs/source/autoapi/poliastro/core/propagation/farnocchia/index.rst:134: WARNING: Literal block ends without a blank line; unexpected unindent.
/home/docs/checkouts/readthedocs.org/user_builds/poliastro/checkouts/1098/docs/source/autoapi/poliastro/core/propagation/index.rst:102: WARNING: Unexpected indentation.
/home/docs/checkouts/readthedocs.org/user_builds/poliastro/checkouts/1098/docs/source/autoapi/poliastro/core/propagation/index.rst:103: WARNING: Literal block ends without a blank line; unexpected unindent.
```

You should not wait for me to say "this needs fixing" 🙃 If the build status is red, you should either fix it or ask for help if you don't know how to.

astrojuanlu:
Thanks! Now that the quality checks were addressed, I observe that some propagation tests give unit errors when disabling numba (the coverage job sets `NUMBA_DISABLE_JIT=1` to actually measure line coverage), for example `test_propagation_cartesian_and_classical_matches`. Could you please have a look?

Yash-10:
Hello @astrojuanlu, I'm sorry for not correcting the build failures that are occuring currently and from now on, I would be more responsible for either asking for help or fixing any failures.

Looking at the codeclimate failure, it seems that Cyclomatic complexity is too high for the propagators implemented in `twobody/propagation.py` ([here](https://codeclimate.com/github/poliastro/poliastro/pull/1098#)). It is most probably due to the code checking if the function expects `classical` or `cartesian` elements. 

Making two separate functions for them might reduce the complexity (and potentially become free from using `*elements` as the argument) but I was wondering whether it would be intelligent to do that since then we would have two separate functions even if their underlying working is very similar.

Could you let me know what would be preferred here?

Thanks!!

astrojuanlu:
Ignore codeclimate for now, we'll decide later :) Focus on the other ones

Yash-10:
I am currently investigating the cause for the unit conversion errors in the absence of numba. It felt like a quick fix but feels like I am missing something trivial. Will get back soon :)

astrojuanlu:
Sure thing @Yash-10 ! 💪🏽

Yash-10:
I was overthinking earlier and tried a bunch of changes. Now the coverage seems to be fine.

The codeclimate failure is probably because the propagator checks for classical and cartesian elements inside the function. Maybe making two separate functions for it can solve that problem (by reducing the complexity). 

Apart from that, two functions give error because it exceeds the max no. of arguments allowed. An initial thought was to harcode `numiter` and `rtol` inside the function to get rid of those extra arguments, but that reduces the users' flexibility, isn't it?

Could you please let me know whether the above is fine or something else is preferred? @astrojuanlu

Yash-10:
I further added some conditions to handle incorrect no. of arguments in the propagator functions. The failing checks (except the codeclimate one) seem to be unrelated, isn't it?

Yash-10:
At the moment, having two separate functions to manage complexity issues in the functions (except `propagate`) might be the solution but for `propagate`, finding a way to reduce its no. of arguments from 6 to 5 is tricky, since we need the additional `classical` argument.

@astrojuanlu Do you have any suggestions?

Thanks!

astrojuanlu:
> The failing checks (except the codeclimate one) seem to be unrelated, isn't it?

Yes! It's an old recurring issue #959

astrojuanlu (changes requested):
Hi @Yash-10 , sorry for the long delay but I am not really convinced by this approach.

On one hand, we have `_classical` and `_coe` propagators, which looks like an unnecessary duplication. I think we should only keep the `_coe` ones, as we were doing after #921 and #1026.

On the other hand, the fact that some propagators now can either accept classical or cartesian elements makes the API and the docstrings more cumbersome, and the implementation more complex.

It is a good attempt at solving the problem, but unfortunately I see a few drawbacks. I appreciate the willingness to put a solution forward 🚀 but for next time, I suggest that we have some discussion in the issue tracker before deciding how to do the work.

At the moment I don't have a specific proposal on how to tackle #921, and unfortunately I will still need a few more days to get up to speed.

Yash-10:
Sorry for the delay in reverting @astrojuanlu, and yes, a discussion beforehand is better. As you had pointed out, the approach surely needs revamping. In the latest approach, some complexity was reduced. This hopefully is a better one but the complexity in `twobody.propagate` still persists. 

It seems that making two separate functions for propagating (in `twobody.propagation`) with classical and cartesian elements, using something like `propagate` and `propagate_classical`, might be a potential cure for the complexity issues.

Could you let me know your thoughts on this?

Thanks!

Yash-10:
There is one idea that could potentially resolve the complexity issues:

Make two separate `propagate` and `propagate_classical` in `twobody.propagation` and handle conditions in the `propagate` method of `Orbit`. This is because cognitive complexity of `Orbit.propagate` (in poliastro's master branch) is not high.

Yash-10:
Just a quick point: \
It turns out that the change from `1e-14` to `1.2e-14` in `test_sampling.py` is a hack to prevent a sampling error that occurs while testing (an issue for the same has also been created to potentially investigate and resolve it).

Yash-10:
**Summarizing the changes made since last reviewed:**
1. The existing `{propagator}_coe` functions are changed such that they return classical orbital elements instead of just returning `nu`. So there are `{propagator}` and `{propagator}_coe` functions in the core propagation section.
2. Twobody propagation section has `{propagator}` and `{propagator}_classical` functions to allow users to be able to use either of them, as per need.
3. `twobody.propagation.propagate` now contains the logic to use classical or cartesian elements based on the `method` user passes in.
4. `twobody.orbit.propagate` uses the twobody propagation to propagate orbits as they were doing earlier.

An equivalently alternative option (if the logic in point 3 above seems to be too complex for a single function) might be:
Make two separate `propagate` and `propagate_classical` functions in `twobody.propagation` and handle conditions in `Orbit.propagate`.

@astrojuanlu Could you let me know your thoughts on this? Additionally, if this PR is merged, it might probably give a head-start to solve similar issues that exist currently, isn't it?

Yash-10:
Hello @astrojuanlu Could you please check the implementation?

Thanks.

Yash-10:
Also, since there are no issues with code complexity, I think there would no need to do (from my earlier comment):

> An equivalent alternative option (if the logic in point 3 above seems to be too complex for a single function) might be:
Make two separate propagate and propagate_classical functions in twobody.propagation and handle conditions in Orbit.propagate.

where point 3 was:

> twobody.propagation.propagate now contains the logic to use classical or cartesian elements based on the method user passes in.

Could you let me know your views @astrojuanlu ?

astrojuanlu:
(Thanks for diligently keeping this branch up to date, I'm catching up now with everything)

astrojuanlu (changes requested):
I am so sorry for being so slow in reviewing this. Unfortunately, I need to be super concentrated to do it properly, and it's something I haven't been able to do in months. At the same time, I think I've made working on this issue too hard, because it's not properly specified (probably because I am not 100 % sure on how to fix it yet).

Please bear with me with a little more of extra patience.

You said

> The existing {propagator}_coe functions are changed such that they return classical orbital elements instead of just returning nu. 

But `coe2rv` returns cartesian elements. Therefore, with this change both `mikkola_coe` and `mikkola` return cartesian elements. Same with all propagators, I think.

`_coe` stands for Classical Orbital Elements, so I'd expect these propagators to either return the true anomaly (current behavior) or all classical elements (unnecessary, in my opinion).

Then you introduced `_classical` methods that actually return cartesian elements as well. Therefore, I don't fully understand where we are going.

Yash-10:
Hello @astrojuanlu I am really sorry for the mess in this PR and thank you so much for the review. I recall that I had initiated this PR without any discussion beforehand but I strongly feel something can be done here!

About this:
> The existing {propagator}_coe functions are changed such that they return classical orbital elements instead of just returning nu.

Sorry for that, I meant that the function now takes classical orbital elements as arguments instead of cartesian elements and at that time, the motivation for doing this was this comment of yours in #921:

>  However, we will need to make `poliastro.twobody.propagation.propagate` smarter now, by adding some sort of mapping that tells whether the function expects classical elements or cartesian elements.

The naming of those functions probably would have confused us. I am not exactly sure whether I interpreted your statement correctly at that time.

`_classical` methods were introduced for the same reason: to be able to take in classical orbital elements as arguments. A separate function was created because the complexity inside a single function was getting too high. If `_classical` methods are not preferred, we could as well handle conditions inside a single function.

If the current implementation is not satisfying (which I suppose since you have requested changes), I would go more deep into the topic and ask questions to be more clear about the bigger objective. If you have any suggestions, please let me know.

Again, thanks for reviewing this PR!!

astrojuanlu:
Since we didn't reach an agreement and there's lots of outdated back and forth in this thread, I think it's better that we close this PR and start over at some other time. Thanks for the work @Yash-10 !

Issue #1100, opened 2021-02-09
CI broken on master


astrojuanlu:
See https://github.com/poliastro/poliastro/pull/1027, https://github.com/poliastro/poliastro/pull/1098.

- Several `mypy` errors due to the latest NumPy version actually shipping types:

```
2021-02-09T22:43:02.2980955Z =================================== FAILURES ===================================
2021-02-09T22:43:02.2981458Z _________________________________ test session _________________________________
2021-02-09T22:43:02.2981817Z mypy exited with status 1.
2021-02-09T22:43:02.2992268Z ____________________________ src/poliastro/ephem.py ____________________________
2021-02-09T22:43:02.2992738Z 60: error: No overload variant of "linspace" matches argument types "Any", "int"
2021-02-09T22:43:02.2993204Z 60: note: Possible overload variants:
2021-02-09T22:43:02.2999138Z 60: note:     def linspace(start: Union[Union[int, float, complex, number[Any], bool_], Sequence[Union[int, float, complex, number[Any], bool_]], ndarray, _SupportsArray, Sequence[Sequence[Any]]], stop: Union[Union[int, float, complex, number[Any], bool_], Sequence[Union[int, float, complex, number[Any], bool_]], ndarray, _SupportsArray, Sequence[Sequence[Any]]], num: SupportsIndex = ..., endpoint: bool = ..., retstep: Literal[False] = ..., dtype: Union[dtype[Any], None, type, _SupportsDType, str, Union[Tuple[Any, int], Tuple[Any, Union[int, Sequence[int]]], List[Any], _DTypeDict, Tuple[Any, Any]]] = ..., axis: SupportsIndex = ...) -> ndarray
[...]

PR #1101, opened 2021-02-12
Resolving CI failures due to latest numpy versions

Yash-10:
Resolving the `mypy` errors (#1100) due to the latest versions of numpy (`1.20.*`).

Thanks!!

### DIFF
--- a/src/poliastro/core/elements.py
+++ b/src/poliastro/core/elements.py
@@ -5,7 +5,7 @@ convert between different elements that define the orbit of a body.
 import numpy as np
 from numba import njit as jit, prange
 from numpy import cross
-from numpy.core.umath import cos, sin, sqrt
+from numpy.core.umath import cos, sin, sqrt # type: ignore
 from numpy.linalg import norm
 
 from .angles import E_to_nu, F_to_nu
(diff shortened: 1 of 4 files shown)


src/poliastro/core/elements.py
```diff
@@ -5,7 +5,7 @@
 import numpy as np
 from numba import njit as jit, prange
 from numpy import cross
-from numpy.core.umath import cos, sin, sqrt
+from numpy.core.umath import cos, sin, sqrt  # type: ignore
```
astrojuanlu, line 8: Does

```
from numpy import cos, sin, sqrt
```

fix the problem?

astrojuanlu (commented):
Thanks for this @Yash-10 ! Left only one comment

src/poliastro/core/elements.py
Yash-10: Oh yes, thank you for the correction!

astrojuanlu (approved):

astrojuanlu:
This addresses part of the CI breakage and the other part is on the works, thanks @Yash-10 !

Issue #1055, opened 2020-12-31
Discussion on missing description for __init__.py in thrust sub-package


DhruvSondhi:
The `__init__.py` file in the `thrust` subpackage doesn't have code or description.
The required file can be found here `src/poliastro/core/thrust/__init__.py`
A `docstring` description of the file will help out a lot.

DhruvSondhi:
I was thinking of addressing this issue as this has been open for quite some time ... Kind of `docstring` description that is needed may include why this file is important for the functioning of the module (basic `__init__.py` structure explanation may be a way to proceed) or something on the similar lines 🤔

DhruvSondhi:
> I was thinking of addressing this issue as this has been open for quite some time ... Kind of `docstring` description that is needed may include why this file is important for the functioning of the module (basic `__init__.py` structure explanation may be a way to proceed) or something on the similar lines

If we need we can go with something describing the `Thrust` sub-package? 🤔

astrojuanlu:
> The __init__.py file in the thrust subpackage doesn't have code or description.
> The required file can be found here src/poliastro/core/thrust/__init__.py
> A docstring description of the file will help out a lot.

This still stands, we should not close this issue.

PR #1107, opened 2021-02-14
Description for modules & sub-modules in Core package

DhruvSondhi:
Hello, 
This PR aims to implement #1055.
Cheers 😄

### DIFF
--- a/src/poliastro/core/__init__.py
+++ b/src/poliastro/core/__init__.py
@@ -1,0 +1,20 @@
+"""
+This is the __init__ file for initialisation of the core module.
+This module contains the following sub-modules:
+- angles.py
+- elements.py
+- hyper.py
+- iod.py
+- perturbations.py
+- stumpff.py
+- util.py
+- Propagation
+--- __init__.py
+--- farnocchia.py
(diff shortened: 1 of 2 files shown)


astrojuanlu (changes requested):
I don't think this has a place in poliastro, sorry. We either add a description of the package or we leave it blank, but explaining basic Python concepts should not be done here.

DhruvSondhi:
No worries ... We can go with the description of the subpackage instead if that's more acceptable or else leave it out 🤔

astrojuanlu (changes requested):
I don't think stating the directory structure is very useful. At the same time, I am unable to find good examples of module docstrings that I like. I'm sorry for not being able to provide useful guidance here, but honestly I am not sure what I expect to see here, until someone proposes something and I say "oh, that looks nice!"

astrojuanlu:
> I marked this pull request as draft, because there are no changes since the last review, and I also unsubscribed. Please ask me for a review when it's ready. And if you are still waiting for my guidance on what to write, I have no other option but to ask you to wait a bit more, sorry.

PR #1114, opened 2021-03-06
Add favicon

Yash-10:
Added favicon for docs, same as for the poliastro website.

Thanks!


Yash-10:
Sorry for the confusion here. As a result, I made a new PR.

Issue #1119, opened 2021-03-15
Discussion on Broken Links & References related to renaming of Master to Main Branch 


DhruvSondhi:
🐞 **Problem**
The current renaming of the `master` branch to the `main` branch lead to different integrated services not working as intended. These have broken some links & services which needs fixing. An example may include Issue #1117.
Current the following things according to me have been affected:
- [x] Binder Repository 
- [x] License on README 
- [x] `index.ipynb` file givings unknown type errors
- [x] `COPYING` file
- [x] Azure pipeline (is redirected accordingly but we need to change it)
- [x] `Poliastro/Validiation` workflow needs to be redirected to the correct branch
- [x] Alternative method of installation via Zip file
- [x] Other 🤔  

💡 **Possible solutions**
Each & Every change needs to be made accordingly. These need to be classified according to the permissions on the repo & can be included to make **Good First Issues**. Others that are not either **GFI** or permission locked, I will be making a PR with the required _FIX_ where ever possible.  

📋  **Steps to solve the problem**
 Each problem needs to be accounted for here & hence be checked accordingly to migrate all `master` branch references to `main` branch for successful renaming.

DhruvSondhi:
@astrojuanlu Please check this out & add/remove if anything is out of the blue 😄

DhruvSondhi:
Thanks, @xlxs4 for fixing `COPYING` file link 😄

astrojuanlu:
What's missing here after #1122 and #1123 were merged?

DhruvSondhi:
Now, according to me, only `License` tag in `README` & changing the build triggers to `main` branch for Azure Pipelines is left ... Everything else has been fixed ... Though I will tinker around a bit to find any unnoticed bug(s) & will update the list ... will wait for a couple of days & then close accordingly :)

abdulmoiz37:
Hi @jorgepiloto @DhruvSondhi I am looking to contribute to this issue in License in Readme to Understand more about the ecosystem and improve my chances for Selection.Can i Work on this issue or someone else is working on this issue currently?

astrojuanlu:
@abdulmoiz37 Nobody else is working on it, feel free to fix the remaining `master` references :)

abdulmoiz37:
#1141

abdulmoiz37:
Hey @astrojuanlu Can u merge this #1141 to the main branch?...Now,I think Changing Building Triggers to main Branch for Azure Pipeline is Left

[...]

PR #1121, opened 2021-03-16
Point README COPYING link to main; not master

xlxs4:
As per https://github.com/poliastro/poliastro/issues/1119.


xlxs4:
Couldn't understand exactly why `requests` failed, shouldn't have to do with the PR

jorgepiloto:
Hi @xlxs4,

thanks for addressing this. You are right, the error is not related to the actual changes but to some "timeout" exception when requesting data from NASA server. Thus, let us merge this 🚀

PR #1122, opened 2021-03-18
Update validation trigger to main branch

jorgepiloto:
Solves for #1119, in particular for the validation trigger.


jorgepiloto:
Yet another classic failure on the validation trigger... I wonder why this requests admin privileges if it was able to work properly without those. May the tokens need to be updated? 🤔

DhruvSondhi:
Where can you see the failure? 🤔

jorgepiloto:
> Where can you see the failure? thinking
@DhruvSondhi 

If you click on "Show environments", you will see that the request was rejected because no admin rights.

astrojuanlu (approved):
Thanks @jorgepiloto ! 🤖

PR #1123, opened 2021-03-18
Fixing Binder Links to point towards Main Branch 

DhruvSondhi:
Hello, 
This PR aims to fix the binder repository that currently points to the `master`, changing to updated `main`
Implements #1119 
Thanks :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,14 +46,14 @@ be read on the wonderful [Read the Docs](https://readthedocs.org/).
 
 # Examples
 
-[![mybinder](https://img.shields.io/badge/launch-binder-e66581.svg?style=flat-square)](https://beta.mybinder.org/v2/gh/poliastro/poliastro/master?filepath=index.ipynb)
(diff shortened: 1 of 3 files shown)


jorgepiloto:
Thank you @DhruvSondhi. This is working again although I noticed that notebooks named "Revisiting Lambert's problem in Python" and "Propagation using Cowell's formulation" can not be opened. The `'` symbol is the reason behind it... May you please also update lines so we can fix all binder stuff in this same pull request?

https://github.com/poliastro/poliastro/blob/56cc3be975fbdc57110286adbbb219533ea58995/index.ipynb#L32

https://github.com/poliastro/poliastro/blob/56cc3be975fbdc57110286adbbb219533ea58995/index.ipynb#L36

DhruvSondhi:
Sure It was also done prior to this but I can do it again :)

jorgepiloto:
Yes, I think we already solved for this some time ago but for some reason it appeared again... Hopefully, this will be the last time fixing those.

astrojuanlu (approved):
Thanks @DhruvSondhi !

Issue #1124, opened 2021-03-18
(Tutorials) documentation revamp


xlxs4:
💡 **Feature request**

Hello, the [Tutorials](https://docs.poliastro.space/en/latest/gallery.html) documentation is a great idea to ease in newcomers to the library! As I was working through the examples, I noticed that, while the texts are well-written and to the point, there's some (maybe inherent/inevitable) inconsistency between different tutorials, as well as small (mostly grammatical) errors here and there.

The inconsistencies mentioned can range from very little nitpicks that probably only bother me (e.g. some sentences before code blocks end with a `.`, others with a `:`; poliastro is sometimes written _poliastro_; `from astropy import units as u` vs `import astropy.units as u`...) to more major differences between tutorials (check [this](https://docs.poliastro.space/en/latest/examples/Going%20to%20Mars%20with%20Python%20using%20poliastro.html) and compare it to [this](https://docs.poliastro.space/en/latest/examples/Comparing%20Hohmann%20and%20bielliptic%20transfers.html)).

Additionally, some extra work could be put into this, for example as per https://github.com/poliastro/poliastro/issues/949. 
An objection to this, however could very well be that:
a) the text is in a pretty good state as it is, and
b) a technical writer will probably get to work on this sooner or later (https://github.com/poliastro/poliastro/issues/1106)

Nonetheless, I believe small, sporadic changes can happen to polish some text sections without requiring significant effort. I could also try to make some iterations, if I can. I intend to hit up a small PR addressing only minor grammatical fixes in a little bit.

🎯 **Goal**

As someone who heavily focuses on the presentation of an idea/project, I think small enhancements such as these mentioned above could take the already great impression `poliastro` leaves on a newcomer and up it just a notch.

More importantly, I think that, at least the more impactful changes proposed, align with the project's [core principles](https://github.com/poliastro/poliastro/wiki/Core-principles), specifically `2` and `7`. If the tutorials could somehow become even more informative in an easy manner, I believe this is a win 🚀. It's a great practice for every FOSS project out there, too; specifically those that depend on long-term community engagement for their development and maintenance.

Let me know what you think and whether this is a sound idea/incoherent rambling/out of the project's scope @jorgepiloto @astrojuanlu

Yash-10:
Hello @xlxs4, it's really true that there are small inconsistencies in some parts of the documentation and I totally agree with your thoughts. I have been making some small grammatical changes throughout the documentation locally, that I felt could be incorporated.

[...]

PR #1125, opened 2021-03-18
Fix minor errors in tutorial docs

xlxs4:
Semi-related to #1124 . Feel free to suggest other, similar changes.

### DIFF
--- a/docs/source/examples/Analyzing NEOs.mystnb
+++ b/docs/source/examples/Analyzing NEOs.mystnb
@@ -64,7 +64,7 @@ except ValueError as err:
     print(err)
 ```
 
-<div class="alert alert-info">Note that epoch is provided by the service itself, so if you need orbit on another epoch, you have to propagate it:</div>
+<div class="alert alert-info">Note that the epoch is provided by the service itself, so if you need orbit on another epoch, you have to propagate it:</div>
 
 ```{code-cell} ipython3
 eros.epoch.iso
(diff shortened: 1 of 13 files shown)


Yash-10:
Thanks @xlxs4 for this. Additionally, I was thinking of these diffs as a starting point:

1. In [index.rst](https://github.com/poliastro/poliastro/blob/main/docs/source/_autoapi_templates/index.rst)

```diff
- * **Core API:** This routines are locate within the `poliastro.core` sub-package
+ * **Core API:** These routines are located within the `poliastro.core` sub-package
```
```diff
- * **High-level API:** All logical files included in this part define main `poliastro` objects such us `poliastro.twobody.Orbit`, `poliastro.maneuver.Maneuver` and how do they interact with each other
+ * **High-level API:** All logical files included in this part define the main `poliastro` objects such us `poliastro.twobody.Orbit`, `poliastro.maneuver.Maneuver` and their interaction with each other
```

2. Docstrings of some functions can also probably be revamped, but I think it will be better to deal with them in a separate PR to avoid clutter. If you feel that we do it right here, then please let me know

Thanks!

xlxs4:
Good catch @Yash-10! For the docstrings, I'd say its best to create a distinct PR. Also create an issue to discuss first, if you think it necessary. For the `index.rst` changes, proceed as you want 🚀

astrojuanlu (approved):
I... can't believe we had so many typos 😂 Thanks again @xlxs4!

PR #1126, opened 2021-03-19
Fix minor documentation, docstring inconsistencies

Yash-10:
As per the discussions in #1124 and #1125.

Please feel free to suggest anything here.

Thanks!!

### DIFF
--- a/docs/source/bibliography.md
+++ b/docs/source/bibliography.md
@@ -26,7 +26,7 @@ order:
 
 ## Software
 
-poliastro wouldn\'t be possible without the tremendous, often unpaid and
+poliastro wouldn\'t have been possible without the tremendous, often unpaid and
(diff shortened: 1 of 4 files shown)


docs/source/bibliography.md
```diff
@@ -26,7 +26,7 @@ order:
 
 ## Software
 
-poliastro wouldn\'t be possible without the tremendous, often unpaid and
+poliastro wouldn\'t have been possible without the tremendous, often unpaid and
```
xlxs4, line 29: `poliastro wouldn't be` is correct here 😄

Yash-10: Thanks for your suggestion @xlxs4. I have reverted that change.

astrojuanlu (approved):
Thanks @Yash-10 ! 🚀

PR #1131, opened 2021-03-21
animated_plots

dark00infinity:
Addresses issue #299
currently implemented for one orbit at a time


DhruvSondhi:
Hello @dark00infinity, If this PR is a Work in Progress effort ... It is better to convert it to a **Draft PR** ... You can always mark it _**Ready for Review**_ whenever you feel this is ready :)

dark00infinity:
> Hello @dark00infinity, If this PR is a Work in Progress effort ... It is better to convert it to a **Draft PR** ... You can always mark it _**Ready for Review**_ whenever you feel this is ready :)

ok @DhruvSondhi

src/poliastro/twobody/orbit.py
```diff
@@ -1509,3 +1509,18 @@ def plot(self, label=None, use_3d=False, interactive=False):
             from poliastro.plotting.core import OrbitPlotter2D
 
             return OrbitPlotter2D().plot(self, label=label)
+
+
+# will be required in future
```
jorgepiloto, line 1514: If you are planing to implement this some time later, then let us dismiss it from now.

tests/tests_animation/test_anim.py
```diff
@@ -0,0 +1,11 @@
+import matplotlib
+
+from poliastro.examples import iss
+from poliastro.plotting.static import StaticOrbitPlotter
+
+
+def test_type_of_anim():
```
jorgepiloto, line 7: We need to figure out how to test animations. I wonder if there exists something like the matplotlib-mpl pytest extension 🤔

src/poliastro/plotting/_base.py
```diff
@@ -159,6 +186,28 @@ def _plot(self, orbit, *, label=None, color=None, trail=False):
             coordinates, orbit.r, label=label, colors=colors, dashed=True
         )
 
+    def _anim(self, orbit, *, label=None, color=None, trail=False):
```
jorgepiloto, line 189: I wonder if we should create something like `AnimatedOrbitPlotter`  which inherits from the `StaticOrbitPlotter` class, as they share most of the methods. In fact, an static plotter should not be able to display animations... Hence, let me ping @astrojuanlu for a further discussion on the topic.

src/poliastro/plotting/static.py
```diff
@@ -139,6 +140,63 @@ def _plot_coordinates(self, coordinates, label, colors, dashed):
...
+        )
+        Animation = animation.FuncAnimation(
+            self.fig,
+            animate,
+            frames=np.arange(0, len(x), 1),
+            interval=30,
+            blit=True,
+            repeat=True,
+        )
+        self._ax.legend(
```
jorgepiloto, line 187: After testing locally, I saw that the legend epoch is not being updated and just displays the initial epoch.

src/poliastro/plotting/_base.py
```diff
@@ -159,6 +186,28 @@ def _plot(self, orbit, *, label=None, color=None, trail=False):
...
+
+        self.set_attractor(orbit.attractor)
+        from poliastro.twobody import propagation
+
+        orbit = orbit.change_plane(self.plane)
+
+        label = generate_label(orbit.epoch, label)
+        time_laps = [i for i in range(int(orbit.period.value))]
+        tofs = time_laps * u.day * 2
+        k = propagation.danby(orbit.attractor.k, orbit.r, orbit.v, tofs)[0]
```
jorgepiloto, line 200: In order to properly visualize the change in velocity, an array of positions and associated times are required. This forces us to propagate the orbit, but, what about if the orbit is a perturbed one? Then, this would force us to make use of the cowell solver. Moreover, you are restricting time propagation to the period and hence, to elliptic orbits. What about open-shape ones like parabolic and hyperbolic?

Again, let me ping @astrojuanlu  for a better discussion on this.

Finally, the farnocchia solver has been proof to be accurate and robust. In fact, it is the default one in the `propagation` function. Is there any particular reason why you selected the danby one?

src/poliastro/plotting/_base.py
```diff
@@ -116,6 +119,20 @@ def __plot_coordinates_and_position(self, trajectory):
 
         return trace_coordinates, trace_position
 
+    def __plot_coordinates_and_position_anim(self, trajectory):
```
jorgepiloto, line 122: Write the private methods using a single under score `_` instead of two `__`.

jorgepiloto (changes requested):
After testing locally, I can confirm that animations are on the right path as now, the change in velocity is evident.

However, some questions started to arise: 

* Should we build an animation plotter class which inherits from the static plotter one?
* How do we propagate perturbed orbits or non-closed ones? 
* Could we have a method to propagate trajectories?

src/poliastro/plotting/_base.py
dark00infinity: yes sure

src/poliastro/plotting/_base.py
dark00infinity: it is not about static and all it is only about how long we can get propagated coordinates from that class. because it is the only thing needed to show the animated plots

src/poliastro/plotting/static.py
dark00infinity: yes, it was just the initial test not modified a lot, will surely look for it thanks for pointing out

tests/tests_animation/test_anim.py
dark00infinity: with my search and all, I feel that there is no plugin for testing animation video or gif. 
we can design our own tester for animation but efficiency is the concern

src/poliastro/twobody/orbit.py
dark00infinity: yes for sure it has nothing to do with animation, it was planned to be an interface like a thing similar to the above plot function which can direct to corresponding 3d or 2d plot as per user input

src/poliastro/plotting/_base.py
dark00infinity: yes, we can implement different solvers for different kind of orbits.
but there should be certain limits on animation like period and all because it is not good to have an animation that is running for a long or infinite time instead of this what we can do is we can give this control to the user for how long they want to see an animated plot for such orbits if no input is given we can put certain default value.
no there is no specific reason I opted for danby we can always change it with the best one as you said farnochia solver

jorgepiloto:
For testing the animations, here is a collection of resources:

* This issue from the official pytetst-mpl plugin about providing support for animations: https://github.com/matplotlib/pytest-mpl/issues/74

* Its original author offers the possibility of using the following testing utilities: https://github.com/t-makaro/animatplot/blob/master/tests/tools.py. However, it would be interesting to implement this under the pytest-mpl.

For the moment, let us try first to implement the animation class. We will carry about tests once everything will be ready merge 👍🏽

src/poliastro/plotting/_base.py
Nirav-Madhani: > 
> 
> I wonder if we should create something like `AnimatedOrbitPlotter` which inherits from the `StaticOrbitPlotter` class, as they share most of the methods. In fact, an static plotter should not be able to display animations... Hence, let me ping @astrojuanlu for a further discussion on the topic.

+1

src/poliastro/plotting/_base.py
```diff
@@ -159,6 +181,28 @@ def _plot(self, orbit, *, label=None, color=None, trail=False):
             coordinates, orbit.r, label=label, colors=colors, dashed=True
         )
 
+    def _anim(self, orbit, *, label=None, color=None, trail=False):
+        colors = self._get_colors(color, trail)
+
+        self.set_attractor(orbit.attractor)
+        from poliastro.twobody import propagation
```
astrojuanlu, line 188: Why can't this import be at the top of the file like all the others?

src/poliastro/plotting/_base.py
```diff
@@ -159,6 +181,28 @@ def _plot(self, orbit, *, label=None, color=None, trail=False):
...
+
+        orbit = orbit.change_plane(self.plane)
+
+        label = generate_label(orbit.epoch, label)
+        time_laps = [i for i in range(800)]
+        tofs = time_laps * u.day * 2
+        k = propagation.farnocchia(orbit.attractor.k, orbit.r, orbit.v, tofs)[0]
```
astrojuanlu, line 195: Why do we need to explicitly call the propagator here, instead of `propagation.propagate` and that's it? The propagation method should be an implementation detail, even at this level.

dark00infinity: for some reason propagation.propagate is not working in animation as expected and further, it is also using farnocchia as default. therefore used the propagator directly. 
testing it ,will soon update this pull request with it. Thanks for review

astrojuanlu: it definitely _should_ work, so please write here the full traceback if it doesn't. Thanks!

src/poliastro/plotting/_base.py
astrojuanlu: Be wary of inheritance: https://en.wikipedia.org/wiki/Composition_over_inheritance

src/poliastro/plotting/_base.py
dark00infinity: it is 👍🏼

Issue #906, opened 2020-04-12
Move maneuver computation to core


astrojuanlu:
Right now there is quite some logic in the `poliastro.maneuvers` module that could probably be moved to `poliastro.core.maneuvers`.

Syed-Osama-Hussain:
@astrojuanlu , any pointers to understand canonical units and to apply them?

priyanshurohilla:
> Right now there is quite some logic in the `poliastro.maneuvers` module that could probably be moved to `poliastro.core.maneuvers`.

@astrojuanlu We currently don't have any module called `poliastro.core.maneuvers'. Do you want me to make one or what shpuld be done. Can you please elaborate a bit

astrojuanlu:
Before tackling this, it would be better to add some extra tests for "3D maneuvers" https://github.com/poliastro/poliastro/issues/905

jorgepiloto:
After https://github.com/poliastro/validation/pull/20 has been merged, poliastro's 3D impulsive maneuvers are now validated against Orekit. Hence, this issue is no longer blocked.

I saw you append this issue to the `Use of canonical units` project board, @astrojuanlu. May people still address this?

astrojuanlu:
True @jorgepiloto ! Despite having this in the board, the issue is no longer blocked.

xlxs4:
Would you guys think this could be a ![](https://img.shields.io/github/labels/poliastro/poliastro/good%20first%20issue)? If so, any pointers?

Yash-10:
I think the functions having numerically intensive calculations (for eg: in `hohmann`, `bielliptic`, `lambert`, and `correct_pericenter`) can be better implemented in the core module and also potentially `@njit`-ing those to speed up the calculations. 

Once we do that, the corresponding high-level API can be configured to use the core implementations.

Please correct me if I am wrong in interpreting anywhere.

Thanks!!

astrojuanlu:
@xlxs4 Hmmm not sure, it might be a bit delicate to get right. I prefer to have more experienced contributors tackle this.

@Yash-10 your assessment is correct!

astrojuanlu:
Closed in #1136!

PR #1136, opened 2021-03-25
Moving maneuver computation to core

Yash-10:
For #906.

@astrojuanlu I have transferred some maneuver computations to `core`. I have a question:

Would it be a better option to use `@jitclass` on `Maneuver` rather than `jit`-ting it's individual functions?

This is because some member methods of `Maneuver` need `self` as an argument, so implementing the jitted version in `core` probably cannot be done.

Could you let me know what do you think?

(The `lambert` method is not jitted in the implementation because it predominantly uses a jitted core method already.)

Thanks!!

### DIFF
--- a/src/poliastro/core/maneuver.py
+++ b/src/poliastro/core/maneuver.py
@@ -1,0 +1,195 @@
+""" Low level manuever implementations """
+
+import numpy as np
+from numba import njit as jit
+from numpy import cross
+from numpy.linalg import norm
+
+from poliastro.core.elements import coe_rotation_matrix, rv2coe, rv_pqw
+
+
+@jit
+def hohmann(k, rv, r_f):
+    r"""Calculate the Hohmann maneuver velocities and the duration of the maneuver.
+
+    By defining the relationship between orbit radius:
+
+    .. math::
+        a_{trans} = \frac{r_{i} + r_{f}}{2}
+
+    The Hohmann maneuver velocities can be expressed as:
+
+    .. math::
+        \begin{align}
+            \Delta v_{a} &= \sqrt{\frac{2\mu}{r_{i}} - \frac{\mu}{a_{trans}}} - v_{i}\\
+            \Delta v_{b} &= \sqrt{\frac{\mu}{r_{f}}} - \sqrt{\frac{2\mu}{r_{f}} - \frac{\mu}{a_{trans}}}
+        \end{align}
+
+    The time that takes to complete the maneuver can be computed as:
+
+    .. math::
+        \tau_{trans} = \pi \sqrt{\frac{(a_{trans})^{3}}{\mu}}
+
+    Parameters
+    ----------
+    k : float
+        Standard Gravitational parameter
+    rv: ~np.array, ~np.array
+        Position and velocity vectors
+    r_f: float
+        Final altitude of the orbit
+
+    """
+    _, ecc, inc, raan, argp, nu = rv2coe(k, *rv)
+    h_i = norm(cross(*rv))
+    p_i = h_i ** 2 / k
+
+    r_i, v_i = rv_pqw(k, p_i, ecc, nu)
+
+    r_i = norm(r_i)
+    v_i = norm(v_i)
+    a_trans = (r_i + r_f) / 2
+
+    dv_a = np.sqrt(2 * k / r_i - k / a_trans) - v_i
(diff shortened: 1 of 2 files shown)


src/poliastro/core/maneuver.py
```diff
@@ -0,0 +1,85 @@
...
+import numpy as np
+from numba import njit as jit
+from numpy import cross
+from numpy.linalg import norm
+
+from poliastro.core.elements import coe_rotation_matrix, rv_pqw
+
+
+@jit
+def hohmann(k, r_i, r_f, v_i, inc, raan, argp, nu, ecc):
```
jorgepiloto, line 12: Let us add also the corresponding docstrings to these functions. Remember that now some of the parameters are no longer `astropy.Quantities` but pure Python types like `float`.

jorgepiloto, line 12: I think we could just pass the following parameters:

`hohmann(k, rv, r_f)`

This simplifies the hihg-level API even more and ensures all computations are carried out inside de jitted functions.

src/poliastro/core/maneuver.py
```diff
@@ -0,0 +1,85 @@
...
+    dv_a = rot_matrix @ dv_a
+    dv_b = rot_matrix @ dv_b
+
+    t_trans = np.pi * np.sqrt(a_trans ** 3 / k)
+
+    return dv_a, dv_b, t_trans
+
+
+@jit
+def bielliptic(k, r_b, r_i, r_f, v_i, inc, raan, argp, nu, ecc):
```
jorgepiloto, line 39: The same applies for this second maneuver type.

jorgepiloto (changes requested):
Thanks for addressing this @Yash-10. I left some requested changes about the implementation 🚀

Yash-10:
Hello @jorgepiloto thanks you so much for your inputs. I have attempted to fix them as you have directed. Can you please have a look?

Moreover, according to me, the `lambert` method need not be jitted since it already computes the solutions using a jitted function. Could you also confirm this?

Thanks!

src/poliastro/maneuver.py
```diff
@@ -202,52 +133,29 @@ def bielliptic(cls, orbit_i, r_b, r_f):
...
-
-        # Bielliptic is defined always from the PQW frame, since it is the
-        # natural plane of the orbit
-        r_i, v_i = rv_pqw(
-            k.to(u.m ** 3 / u.s ** 2).value,
-            p_i.to(u.m).value,
-            orbit_i.ecc.value,
-            orbit_i.nu.to(u.rad).value,
-        )
+        _, ecc, inc, raan, argp, nu = orbit_i.classical()
```
astrojuanlu, line 137: Where are the rest of the variables used? Can't see it after a quick pass

Yash-10: So sorry for that @astrojuanlu and thanks for pointing out. I have changed it! They shouldn't be there since those variables are already created in `core`.

Thanks

Yash-10:
I also see some weird docs error. Investigating the cause...

Yash-10:
After searching online. I think it's because of `docutils`. The latest version of `docutils` (0.17) was released about 6hrs ago. As per sphinx-doc/sphinx/issues/9049, it seems using `docutils==0.16` can solve it.

So it seems that it is not related to this PR and can be ignored?

astrojuanlu:
Thanks @Yash-10 ! I subscribed to https://github.com/sphinx-doc/sphinx/issues/9049, let's see how that goes.

Could you please rebase on top of current `main` and force push?

Yash-10:
It seems there was a timeout error while installing packages....again doesn't seem to be related to this PR.

Is it required to trigger the tests again for a sanity check?

astrojuanlu:
> Is it required to trigger the tests again for a sanity check?

It should be enough to rebase and force push 😉

Yash-10:
Done!

astrojuanlu:
Not yet, it seems!

Screenshot_2021-04-03 Moving maneuver computation to core by Yash-10 · Pull Request #1136 · poliastro poliastro

Yash-10:
Oh, I think now it's done!

Yash-10:
The errors seem to be unrelated to this PR.

astrojuanlu (approved):
Finally! 🚀

astrojuanlu:
Thanks a lot @Yash-10 ! Great to have this in 💪🏽

PR #1137, opened 2021-03-26
Reduce warnings

ishanSrt:
This handles all warnings, the coverage decreases for deprecated functions `Orbit.from_body_ephem()` and `Orbit.pqm()`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,6 +18,10 @@ markers =
     remote_data
     filterwarnings
     mpl_image_compare
+filterwarnings =
+    # This Deprecation warning is not present in numba 0.53.0
+    # and should be removed when the dependency version is upgraded
+    ignore:`np\.long` is a deprecated alias for `np\.compat\.long`:DeprecationWarning
 
 [flake8]
 ignore = E203, E266, E501, W503
--- a/tests/tests_plotting/test_misc.py
+++ b/tests/tests_plotting/test_misc.py
@@ -5,11 +5,19 @@ from poliastro.plotting import OrbitPlotter2D, OrbitPlotter3D
 from poliastro.plotting.misc import plot_solar_system
 
 
+@pytest.mark.filterwarnings(
+    'ignore:ERFA function "epv00" yielded.*of "'
+    'warning. date outsidethe range 1900-2100 AD"'
+)
 @pytest.mark.parametrize("outer,expected", [(True, 8), (False, 4)])
 def test_plot_solar_system_has_expected_number_of_orbits(outer, expected):
     assert len(plot_solar_system(outer).trajectories) == expected
 
 
+@pytest.mark.filterwarnings(
+    'ignore:ERFA function "epv00" yielded.*of "'
+    'warning. date outsidethe range 1900-2100 AD"'
+)
 @pytest.mark.parametrize(
     "use_3d, plotter_class", [(True, OrbitPlotter3D), (False, OrbitPlotter2D)]
 )
@@ -24,6 +32,10 @@ def test_plot_inner_solar_system_static(earth_perihelion):
     return plt.gcf()
 
 
+@pytest.mark.filterwarnings(
+    'ignore:ERFA function "epv00" yielded.*of "'
+    'warning. date outsidethe range 1900-2100 AD"'
+)
 @pytest.mark.mpl_image_compare
 def test_plot_outer_solar_system_static(earth_perihelion):
     plot_solar_system(outer=True, epoch=earth_perihelion)
--- a/tests/tests_twobody/test_orbit.py
+++ b/tests/tests_twobody/test_orbit.py
@@ -231,17 +231,6 @@ def test_parabolic_has_zero_energy():
     assert_allclose(ss.energy.value, 0.0, atol=1e-16)
 
 
-def test_pqw_for_circular_equatorial_orbit():
-    ss = Orbit.circular(Earth, 600 * u.km)
-    expected_p = [1, 0, 0] * u.one
-    expected_q = [0, 1, 0] * u.one
-    expected_w = [0, 0, 1] * u.one
-    p, q, w = ss.pqw()
-    assert_allclose(p, expected_p)
-    assert_allclose(q, expected_q)
-    assert_allclose(w, expected_w)
-
-
 @pytest.mark.parametrize(
     "attractor,alt,argp,expected_argp,expected_inc",
     [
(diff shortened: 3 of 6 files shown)


ishanSrt:
Closes #1074 
@astrojuanlu please review

DhruvSondhi:
I just wanted to ask that why are we filtering the warning? Instead can't we have the fix for these?

ishanSrt:
> I just wanted to ask that why are we filtering the warning? Instead can't we have the fix for these?
> Another thing, why were some tests removed? I would suggest that instead of removing them we find a way around these warnings so as to maintain the functionality & reduce the warnings ... Warnings are not a problem but Errors are :)

@DhruvSondhi The warnings that could be fixed are fixed for eg. in https://github.com/poliastro/poliastro/pull/1137/commits/8010a19be3fd330e38d48f0f2bd329ebcf393461
Other warnings couldn't be fixed for eg. complains about the dates not being withing 1900-2100AD when the code in upstream repositories requires negative date values eg. `erfa.epv00()` https://github.com/poliastro/poliastro/pull/1137/commits/c145603a765df864c2631a81d30f91f090884dfb

ishanSrt:
@DhruvSondhi tests are removed for deprecated functions, which throw a deprecated function warning. Please go through the commits carefully.

DhruvSondhi:
Sure, didn't see them in the first place ... _Filter warnings_ were also proposed by me in #1074 {check discussion}

setup.cfg
```diff
@@ -18,6 +18,10 @@ markers =
     remote_data
     filterwarnings
     mpl_image_compare
+filterwarnings =
+    # This Deprecation warning is not present in numba 0.53.0
+    # and should be removed when the dependency version is upgraded
+    ignore:`np\.long` is a deprecated alias for `np\.compat\.long`:DeprecationWarning
```
astrojuanlu, line 24: Where does this warning come from? If it's not from poliastro, let's remove this

tests/tests_plotting/test_misc.py
```diff
@@ -5,11 +5,19 @@
 from poliastro.plotting.misc import plot_solar_system
 
 
+@pytest.mark.filterwarnings(
```
astrojuanlu, line 8: Instead, let's use a date in the test that doesn't yield this warning. Is this possible?

tests/tests_plotting/test_misc.py
```diff
@@ -5,11 +5,19 @@
...
+    'ignore:ERFA function "epv00" yielded.*of "'
+    'warning. date outsidethe range 1900-2100 AD"'
+)
 @pytest.mark.parametrize("outer,expected", [(True, 8), (False, 4)])
 def test_plot_solar_system_has_expected_number_of_orbits(outer, expected):
     assert len(plot_solar_system(outer).trajectories) == expected
 
 
+@pytest.mark.filterwarnings(
```
astrojuanlu, line 17: See above,

tests/tests_plotting/test_misc.py
```diff
@@ -24,6 +32,10 @@ def test_plot_inner_solar_system_static(earth_perihelion):
     return plt.gcf()
 
 
+@pytest.mark.filterwarnings(
```
astrojuanlu, line 35: See above

tests/tests_twobody/test_orbit.py
```diff
@@ -615,27 +604,6 @@ def test_from_horizons_raise_valueerror():
     )
 
 
-@pytest.mark.remote_data
-def test_orbit_from_horizons_has_expected_elements():
```
astrojuanlu, line 619: Is this already covered by an equivalent `Ephem.from_horizons` test?

tests/tests_twobody/test_propagation.py
```diff
@@ -335,6 +335,12 @@ def test_propagate_to_date_has_proper_epoch():
     assert (ss1.epoch - final_epoch).sec == approx(0.0, abs=1e-6)
 
 
+@pytest.mark.filterwarnings(
```
astrojuanlu, line 338: See my other comments

astrojuanlu (changes requested):
This requires still some work, but it looks good!

tests/tests_plotting/test_misc.py
ishanSrt: @astrojuanlu these warnings can't be removed as they originate upstream in astropy using erfa module where https://pyerfa.readthedocs.io/en/latest/api/erfa.epv00.html can even use negative value of dates.

tests/tests_twobody/test_propagation.py
ishanSrt: @astrojuanlu again these ERFA warnings are generating upstream in astropy

setup.cfg
ishanSrt: @astrojuanlu we can update the numba dependency version. I have confirmed that it doesn't exist in 0.53.0. I can't say the same for the versions between 0.53.0rc1 and 0.53.0 but can check in which version is this warning removed.

tests/tests_twobody/test_orbit.py
ishanSrt: @astrojuanlu removing this test was only decreasing coverage for `Orbit.pqw()` function which is deprecated and was throwing a deprecated function warning, I haven't checked exactly what this test is doing with `Ephem.from_horizons` but can check and confirm whether this is essential and get back to you on this.

setup.cfg
astrojuanlu: In any case, we should not filter this warning, as it's not ours.

tests/tests_plotting/test_misc.py
astrojuanlu: I suspect that `plot_solar_system` is causing this. Some planets might have a period larger than 90 years, therefore more time is needed to compute a full revolution.

tests/tests_twobody/test_propagation.py
astrojuanlu: Yes, ERFA warnings are generated by Astropy, but it's probably our fault. See my comment above.

tests/tests_twobody/test_orbit.py
astrojuanlu: We should consider replacing it by something similar that checks whether `Ephem.from_horizons` for a particular body has the expected data. Kind of like an integration test.

But we might already have those. No need to repeat them.

astrojuanlu:
There were some outstanding review comments that were unaddressed, so I think it's better that we close this and start from scratch at some other time. Thanks @ishanSrt !

PR #1171, opened 2021-04-12
Removed Redundant `precommit.sh` in `bin/` directory

DhruvSondhi:
Hello, 
This PR aims to remove the redundant `precommit.sh` script from the `bin/` directory.
The functionality is already present in the current `tox` automation :)
Thanks :)

### DIFF
--- a/bin/pre_commit.sh
+++ b/bin/pre_commit.sh
@@ -1,5 +1,0 @@
-#!/usr/bin/bash
-
-pep8 .
-pytest --cov-report term-missing --cov . src/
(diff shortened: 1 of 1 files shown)


DhruvSondhi:
Hello @astrojuanlu, Shall I fix the issue with the example as pointed out through [here](https://github.com/poliastro/poliastro/blob/48bf2c6b64f3d35b005d0a60abe8bcd483adbced/src/poliastro/twobody/orbit.py#L428-L432)? or else we could keep this PR only for removing the script 🤔

astrojuanlu:
Merging directly, thanks!

PR #1173, opened 2021-04-13
Moving computations of `spheroid_location` to core

Yash-10:
This implementation attempts to move computations in `spheroid_location.py`.

**Points**

1. Earlier, the `distance` function only returned the value without units even if the inputs are astropy quantities. Since it would be better to get the units as well, the current implementation returns them with units.
2. Tests for checking the execution of `f`, `radius_of_curvature`, and `distance` are added.
3. Two methods are not jitted since they give slight numerical inaccuracies while testing. Probably could be an investigation point. As a result, they are kept unjitted.

Thanks!!

### DIFF
--- a/src/poliastro/core/spheroid_location.py
+++ b/src/poliastro/core/spheroid_location.py
@@ -1,4 +1,4 @@
-""" Low level calculations for spheroid locations """
+""" Low level calculations for oblate spheroid locations """
 
 import numpy as np
 from numba import njit as jit
--- a/src/poliastro/core/spheroid_location.py
+++ b/src/poliastro/core/spheroid_location.py
@@ -1,0 +1,185 @@
+""" Low level calculations for spheroid locations """
+
+import numpy as np
+from numba import njit as jit
+
+
+def cartesian_cords(_a, _c, _lon, _lat, _h):
+    """Calculates cartesian coordinates.
+
+    Parameters
+    ----------
+    _a: ~astropy.units.quantity.Quantity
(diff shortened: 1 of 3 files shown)


Yash-10:
**Question**

According to the docstring in current `main`, the `SpheroidLocation` class handles any arbitrary ellipsoid. But taking a look at how `_a`, `_b`, `_c` are initialized [here](https://github.com/poliastro/poliastro/blob/main/src/poliastro/spheroid_location.py#L28-L30), it seems that the `SpheroidLocation` class only focuses on oblate spheroids and hence is a good choice for astronomical bodies of oblate shape, for example the Earth but not a good choice for Moon, for example. Moreover, the tests only use `Earth` as the attractor.
 
@astrojuanlu Could you let me know if there are there any thoughts on the below: 
1. Could we modify the docstring to specify that the class currently handles only oblate shape ellipsoids and stick with it until it is generalized. 
2. Or, make the class itself generalized to handle other ellipsoid shapes?

If it's the first option temporarily, we could probably change the docstring in this PR itself until we generalize for a general ellipsoid. Am I missing something?

astrojuanlu:
Good catch! Let's do this:

> modify the docstring to specify that the class currently handles only oblate shape ellipsoids and stick with it until it is generalized.

astrojuanlu (approved):
This looks very good! Especially appreciated the new tests and the detailed docstrings. Merging straight away! Thanks @Yash-10 🚀

PR #1176, opened 2021-04-16
Fix validation trigger

jorgepiloto:
Yet another try to fix the validation trigger, hope this to be the last one 🤞🏽

### DIFF
--- a/.github/workflows/validation_trigger.yml
+++ b/.github/workflows/validation_trigger.yml
@@ -27,5 +27,5 @@ jobs:
           curl \
           -XPOST -u "${{ secrets.VALIDATION_USER }}:${{ secrets.VALIDATION_PASS }}" \
           -H "Accept: application/vnd.github.v3+json" \
-          https://api.github.com/repos/poliastro/validation/actions/workflows/ci_actions.yml/dispatches \
+          http://requestbin.net/r/1xh5uu96 \
           --data '{"ref":"main"}'
--- a/.github/workflows/validation_trigger.yml
+++ b/.github/workflows/validation_trigger.yml
@@ -29,7 +29,7 @@ jobs:
           VALIDATION_PASS: ${{ secrets.VALIDATION_PASS }}
         run: |
           curl -v \
-          -XPOST -u "${env:VALIDATION_USER}:${env:VALIDATION_PASS}" \
+          -XPOST -u "$VALIDATION_USER:$VALIDATION_PASS" \
           -H "Accept: application/vnd.github.v3+json" \
           https://requestbin.net/r/1xh5uu96 \
           --data '{"ref":"main"}'
--- a/.github/workflows/validation_trigger.yml
+++ b/.github/workflows/validation_trigger.yml
@@ -24,7 +24,8 @@ jobs:
     steps:
       - name: Trigger poliastro/validation
         run: |
-          curl -XPOST -u "${{ secrets.PAT_VALIDATION}}:${{secrets.PAT_VALIDATION_TOKEN}}" \
+          curl \
+          -XPOST -u "${{ secrets.VALIDATION_USER }}:${{ secrets.VALIDATION_PASS }}" \
           -H "Accept: application/vnd.github.v3+json" \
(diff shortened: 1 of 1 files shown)


.github/workflows/validation_trigger.yml
```diff
@@ -26,5 +26,5 @@ jobs:
         run: |
           curl -XPOST -u "${{ secrets.PAT_VALIDATION}}:${{secrets.PAT_VALIDATION_TOKEN}}" \
           -H "Accept: application/vnd.github.v3+json" \
-          -H "Content-Type: application/json" https://api.github.com/repos/poliastro/validation/actions/workflows/ci_actions.yml/dispatches \
+          -H "Content-Type: application/json" https://api.github.com/repos/poliastro/validation/actions/workflows/ci_actions/dispatches \
```
astrojuanlu, line 29: I think this is the correct workflow: https://api.github.com/repos/poliastro/validation/actions/workflows/ci_actions.yml

astrojuanlu (changes requested):
It's weird that this still doesn't work 🤔 We would need to find a way so that, if the POST request fails, the workflow fails.

jorgepiloto:
This is working through command line but not sure why it fails if triggered from here... 🤔

.github/workflows/validation_trigger.yml
```diff
@@ -18,13 +18,14 @@ jobs:
   # Validate poliastro's new features
   validation:
     runs-on: ubuntu-latest
-    environment: validation-env
```
astrojuanlu, line 21: Shouldn't we keep this? Perhaps if we remove the line, the env vars are not properly read

astrojuanlu:
Idea: use RequestBin or similar services to understand if the `POST` request is being performed correctly. On it.

astrojuanlu:
@jorgepiloto Authorization headers are empty.

.github/workflows/validation_trigger.yml
```diff
@@ -24,7 +24,8 @@ jobs:
     steps:
       - name: Trigger poliastro/validation
         run: |
-          curl -XPOST -u "${{ secrets.PAT_VALIDATION}}:${{secrets.PAT_VALIDATION_TOKEN}}" \
+          curl -v \
+          -XPOST -u "${{ secrets.VALIDATION_USER }}:${{ secrets.VALIDATION_PASS }}" \
```
astrojuanlu, line 28: I think this syntax just doesn't work: https://docs.github.com/en/actions/reference/encrypted-secrets#example-using-bash let me try something different

astrojuanlu:
> Secrets are not passed to workflows that are triggered by a pull request from a fork.

Screenshot_2021-04-18 Build software better, together

I think this is it @jorgepiloto. We should find another way to trigger this, I think.

astrojuanlu (changes requested):
After reading https://blog.teddykatz.com/2021/03/17/github-actions-write-access.html, it makes a lot of sense that secrets are not available for pull request, since an attacker could send a malicious pull request and print them.

I am not sure how was this working before, perhaps they changed something. As the action was not failing when an error message is returned, I am not sure when this change happened.

@jorgepiloto although it's very attractive to know whether a pull request passed the validation, I don't think this is possible with our current structure. Given that the token must have `actions:write` permissions on the `repo` token, it's just dangerous to make it public.

I only see these solutions:

- Bring everything into the same repository. I don't think it's very practical, we could discuss it again as a last resort.
- Create our own bot or action that does this properly on pull requests. A lot of work.
- Assume that we will only trigger validations from `main`, and keep an eye on it especially when a release approaches. I think this is the simplest solution for now.

Therefore, we need to make these changes:

- Make the action fail if curl (✔️  done)
- Rework the CI so these triggers are only sent from `main`, where the environment secrets are available, so we avoid getting failures for every pull request
- Squash all the commits into one

jorgepiloto:
Closing this in favor of #1179. For future modifications and maintenance, it is important to remember that this must be done through branches which belong to the repository, not by ones which belong to forks.

Issue #1158, opened 2021-04-06
Review documentation classification, take 2


astrojuanlu:
Now that I work at a documentation company 🤓 I have been thinking more and more about poliastro docs, and in particular within the Diátaxis framework https://diataxis.fr/ (formerly known as "The Documentation System").

![diataxis](https://user-images.githubusercontent.com/316517/113677223-a10e8200-96bd-11eb-9993-4ac6319aea66.png)

Even though we did restructure the documentation as part of #975, I would like to revisit this with some ideas:

- Add a "Getting started" category that includes the document currently called "Getting started" renamed to "Installation", the "User guide" renamed to "Quickstart guide", and the contributor documentation
- Rename "Tutorials" to "Examples" (in the spirit of How-to Guides)
- Add a new "Background" category that includes the more theoretical examples, like https://docs.poliastro.space/en/latest/examples/Comparing%20Hohmann%20and%20bielliptic%20transfers.html, https://docs.poliastro.space/en/latest/examples/Propagation%20using%20Cowells%20formulation.html, https://docs.poliastro.space/en/latest/examples/Revisiting%20Lamberts%20problem%20in%20Python.html, https://docs.poliastro.space/en/latest/examples/Studying%20Hohmann%20transfers.html, and https://docs.poliastro.space/en/latest/examples/Multirevolutions%20solution%20in%20Lamberts%20problem.html
- Add two TOCs in https://docs.poliastro.space/en/latest/autoapi/index.html, which is something that we had in older versions and I think it's worth bringing back. Not sure how easy it is to do with autoapi.

DhruvSondhi:
I would love to contribute to the same 😄🚀

astrojuanlu:
It's true that these two are related, but #1150 is not necessarily a blocker for this one. In fact, I think it's the other way around :)

astrojuanlu:
pytest also bit the bullet :) https://github.com/pytest-dev/pytest/pull/8455

PR #1182, opened 2021-04-23
Docs reorganization

astrojuanlu:
See discussion at #1158. I also:

- Added proper cross references (I might have missed a few)
- Rewrote the project metadata according to PEP 621 🎉


astrojuanlu:
Docs look good, merging!

PR #1184, opened 2021-04-24
Move flyby computation to core

Yash-10:
This PR tries to move computations for threebody flybys to core.

Thanks!


Yash-10:
@astrojuanlu Could you take a look at this transfer of computation?

It also seems that the codecov/patch test needs a fix.

Thanks!

astrojuanlu (approved):
This is good to go, thanks @Yash-10 and sorry for the delay!

PR #1190, opened 2021-05-01
Moving ra, dec, and W calculations to core

Yash-10:
I noticed the rotational elements were computed in the high-level module. It might be a good choice to move those to core.

Thanks!

### DIFF
--- a/src/poliastro/core/fixed.py
+++ b/src/poliastro/core/fixed.py
@@ -1,0 +1,368 @@
+"""
+Low level ra, dec, and W calculations for bodies.
+
+REFERENCES:
+The values for all bodies except the Moon have been taken from: Archinal, B., \
+Acton, C., A’Hearn, M., Conrad, A., Consolmagno, G., & Duxbury, T. et al. \
+(2018). Report of the IAU Working Group on Cartographic Coordinates and \
+Rotational Elements: 2015. Celestial Mechanics And Dynamical Astronomy, 130(3). \
+doi: 10.1007/s10569-017-9805-5.
+
+The values for Moon have been taken from: Archinal, B., A’Hearn, M., Bowell, E., \
+Conrad, A., Consolmagno, G., & Courtin, R. et al. (2010). Report of the IAU \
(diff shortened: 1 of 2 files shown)


astrojuanlu (approved):
Thanks a lot @Yash-10 ! While I'm not a super fan of having the functions like this, it's something I have struggled with internally for a long time. It's not possible to do it "the smart way" either because each body uses a different equation essentially. And having a list of parameters that are then used in the equations makes everything just more difficult to read. So I am giving my approval on this, since it's an improvement over what we had before.

Yash-10:
Sorry @astrojuanlu ! I had intended to convert this into a draft since I suspected some problems in the units. I will fix it soon!

Again, sorry for the trouble!

astrojuanlu:
No worries! To clarify, #1229 was not because of this 💪🏽

src/poliastro/core/fixed.py
```diff
@@ -0,0 +1,368 @@
...
+    d: float
+        Interval in days from the standard epoch
+
+    Returns
+    -------
+    ra, dec, W: tuple (float)
+        Right ascension and declination of north pole, and angle of the prime meridian.
+
+    """
+    Ja = 99.360714 + 4850.4046 * T
```
jorgepiloto, line 192: This is being computed in degrees and passed in these units to the trigonometric functions below! As pointed out by @Yash-10, the tests within poliastro did not covered all this logic but the validation ones did. Pinging @astrojuanlu, we might discuss this on this week's poliastro meeting (July the 5th, 2021).

jorgepiloto: The same applies to the rest of the bodies, though this was only noticed for the cases of Jupiter and Moon due to the big amount of terms when computing their rotational elements at epoch.

Yash-10: Thanks @jorgepiloto ! I will soon open a pull request to (hopefully) rectify my mistake!

astrojuanlu: Yeah, in general I think having the validation tests is awesome, but also we should have more unit tests :)

PR #1206, opened 2021-05-07
Doc improvements

astrojuanlu:
Several documentation improvements.


docs/source/history.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Future ideas
+
+These are some things that I would love to implement in poliastro to
+expand its capabilities:
+
+- 3D plotting of orbits
+- Continuous thrust maneuvers
+- Tisserand graphs
+- Porkchop plots
```
DhruvSondhi, line 24: Some of these have been implemented in the library ... I think so that we should update it with some newer features that can be added like animated plots, newer atmospheric models, & other complex orbital mechanic related things

DhruvSondhi (commented):
These are some changes that can be made :)

docs/source/history.md
astrojuanlu: Oh, good catch!

DhruvSondhi:
@astrojuanlu another thing that can be done ... Can you please move the `Jacchia` Atmospheric Model to `earth/atmosphere` folder? ... I was thinking of doing it but this can be done here if it doesn't hinder progress, would make much more sense keeping all the atmospheric models together 🤔

PR #1212, opened 2021-05-09
Fix of cowell method calls in user guide

ismaelJimenez:


astrojuanlu:
Thanks @ismaelJimenez for your first pull request, good catch!

PR #1216, opened 2021-05-11
NRLMSISE 2000 Atmospheric Model Implementation

DhruvSondhi:
Hello,
This PR  aims to implement the **NRLMSISE00 Atmospheric Model**.
The current implementation does run <strike>but the results are not accurate ( I am working on fixing the problem ).</strike>
This is a work in progress PR. 
Kindly let me have your opinions on the same, @astrojuanlu @jorgepiloto 🚀 
Thanks :)


DhruvSondhi:
Test do not follow `pytest` method of testing ... that is to be added after the results are accurate enough :)

DhruvSondhi:
I have fixed the code ... Now, the values generated are consistent with those present in the main source 🚀🔥

astrojuanlu (commented):
Hi @DhruvSondhi, I have a couple of concerns about this:

- The code seems to be adapted from https://github.com/DeepHorizons/Python-NRLMSISE-00/, however Joshua Milas is not mentioned here. This goes against the MIT license and must be fixed if we retain this version.
- On the other hand, I am wondering why can't we somehow depend on https://pypi.org/project/msise00/ (source at https://github.com/space-physics/msise00) which is also MIT, instead of maintaining all this data here?

astrojuanlu:
There were some outstanding review comments on this pull request that were not addressed and I didn't quite like the approach anyway, so I think it's better that we close it and start from scratch at some other time.

Issue #1218, opened 2021-05-11
Documentation fails with `jinja2.exceptions.UndefinedError: 'nbformat.notebooknode.NotebookNode object' has no attribute 'tags'`


astrojuanlu:
See for example https://github.com/poliastro/poliastro/pull/1217

```
2021-05-11T22:27:27.1127366Z Traceback (most recent call last):
2021-05-11T22:27:27.1128453Z   File "/home/vsts/work/1/s/.tox/docs/lib/python3.8/site-packages/nbsphinx.py", line 1022, in parse
2021-05-11T22:27:27.1131332Z     rststring, resources = exporter.from_notebook_node(nb, resources)
2021-05-11T22:27:27.1132645Z   File "/home/vsts/work/1/s/.tox/docs/lib/python3.8/site-packages/nbsphinx.py", line 839, in from_notebook_node
2021-05-11T22:27:27.1133377Z     rststr, resources = super(Exporter, self).from_notebook_node(
2021-05-11T22:27:27.1134260Z   File "/home/vsts/work/1/s/.tox/docs/lib/python3.8/site-packages/nbconvert/exporters/templateexporter.py", line 384, in from_notebook_node
2021-05-11T22:27:27.1135194Z     output = self.template.render(nb=nb_copy, resources=resources)
2021-05-11T22:27:27.1136008Z   File "/home/vsts/work/1/s/.tox/docs/lib/python3.8/site-packages/jinja2/environment.py", line 1289, in render
2021-05-11T22:27:27.1136565Z     self.environment.handle_exception()
2021-05-11T22:27:27.1137291Z   File "/home/vsts/work/1/s/.tox/docs/lib/python3.8/site-packages/jinja2/environment.py", line 924, in handle_exception
2021-05-11T22:27:27.1137864Z     raise rewrite_traceback_stack(source=source)
2021-05-11T22:27:27.1138501Z   File "<template>", line 124, in top-level template code
2021-05-11T22:27:27.1139622Z   File "/home/vsts/work/1/s/.tox/docs/share/jupyter/nbconvert/templates/rst/index.rst.j2", line 1, in top-level template code
2021-05-11T22:27:27.1140414Z     {%- extends 'display_priority.j2' -%}
2021-05-11T22:27:27.1141297Z   File "/home/vsts/work/1/s/.tox/docs/share/jupyter/nbconvert/templates/base/display_priority.j2", line 1, in top-level template code
[...]

PR #1219, opened 2021-05-12
Fix Jinja2 version

jorgepiloto:
Solves for #1218 by imposing temporarily `Jinja2<3.0.0`, which was release just a few hours ago. This will help us to keep merging contributions while we find a solution 👍🏽


astrojuanlu (approved):
Thanks a lot for the investigation @jorgepiloto! I'm merging this

PR #1223, opened 2021-05-14
Refactor of bodies

ismaelJimenez:
Hi! As I try to understand some code I like to do some refactoring. I have gone through the bodies and I send my results just in case you would like to include the changes. As a quick summary of the changes:

- Added Type Annotations to Bodies
- Created test for Angular Velocity
- Modified default parameters in method from_relative(...). k and R shall not be defaulted to None, as if they are None the relative multiplications would fail.
- Remove call to _q(...) in SolaSystemPlanet instantiations. _q() is already called in Body.__new__, so there is no need to call it also in the instantiation
- Even out the spaces in the SolaSystemPlanet instantiations

### DIFF
--- a/src/poliastro/bodies.py
+++ b/src/poliastro/bodies.py
@@ -22,7 +22,7 @@ from __future__ import annotations
 
 import math
 from collections import namedtuple
-from typing import Any, Optional, Type
+from typing import Any, Type
 
 from astropy import units as u
 from astropy.constants import G
@@ -63,10 +63,10 @@ class Body(
 
     def __new__(
         cls: Type[Body],
-        parent: Optional[Body],
+        parent: Body | None,
         k: Quantity,
         name: str,
-        symbol: Optional[str] = None,
+        symbol: str | None = None,
         R: Quantity = 0 * u.km,
         R_polar: Quantity = 0 * u.km,
         R_mean: Quantity = 0 * u.km,
@@ -110,7 +110,7 @@ class Body(
         parent,
         k: Quantity,
         name: str,
-        symbol: Optional[str] = None,
+        symbol: str | None = None,
         R: Quantity = 0 * u.km,
         **kwargs,
     ) -> Body:
@@ -120,10 +120,10 @@ class Body(
     def from_relative(
         cls: Type[Body],
         reference: Body,
-        parent: Optional[Body],
+        parent: Body | None,
         k: float,
         name: str,
-        symbol: Optional[None] = None,
+        symbol: str | None = None,
         R: float = 0,
         **kwargs,
     ) -> Body:
@@ -135,8 +135,8 @@ class Body(
 class SolarSystemPlanet(Body):
     def plot(
         self,
-        epoch: Optional[Time] = None,
-        label: Optional[str] = None,
+        epoch: Time | None = None,
+        label: str | None = None,
         use_3d: bool = False,
         interactive: bool = False,
         plane: Planes = Planes.EARTH_ECLIPTIC,
--- a/src/poliastro/bodies.py
+++ b/src/poliastro/bodies.py
@@ -73,7 +73,7 @@ class Body(
         rotational_period: Quantity = 0.0 * u.day,
         J2: Quantity = 0.0 * u.one,
         J3: Quantity = 0.0 * u.one,
-        mass: Quantity = None,
+        mass: Optional[Quantity] = None,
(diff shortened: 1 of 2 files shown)


astrojuanlu (changes requested):
Hi @ismaelJimenez, this looks really good! A couple of high level comments:

- Good catch about the extra `_q` calls! Perhaps they were leftovers from previous refactors.
- Good job on the type annotations. I'm a bit hesitant on how to move forward with them though, since Astropy has its own way of using annotations: https://docs.astropy.org/en/latest/api/astropy.units.quantity_input.html in particular, methods already decorated with `quantity_input` might benefit from this. I don't want to block your PR because of this but I'd like to find the best way of doing this :)
- In case we do include them, I think newer versions of MyPy accept `str | None` instead of `Optional[str]` and other niceties, have a look at https://mypy-lang.blogspot.com/2021/01/mypy-0800-released.html would you update those? (So we can remove the use of `Optional` altogether)

ismaelJimenez:
Hello Juan Luis! 

Thanks for the tip on the new formats accepted by MyPy. I wasn´t aware of the new functionalities and I agree that the "|" is nicer than the Optional, so I have just changed it :-)

With respect to the Type Annotations, I agree that whenever possible, it is better to use "quantity_input" from astropy as it provides an awesome unit validation that we could not get otherwise. That is why I kept the quantity_input decorator in the from_parameters method. In the refactoring that method mixes both standard Python Annotations and quantity_input, although they sometimes don´t play together that well. I have been able to inline "R: u.km" in the method parameters, but wasn´t able to inline "k: u.km ** 3 / u.s ** 2", as mypy likes when one unit is inlined (e.g. u.km) but complains when something more complex is tried (e.g. u.km**2). That is why R is inline but I kept "k" units in the quantity_input decorator.

With respect to the methods that don´t have any astropy units as inputs, like from_relative or plot, I think it would be a good idea to keep the Python Annotations, but that certainly depends on the overall project guidelines of style.

If you don´t think the current type annotations would play well with the overall project, I can remove them without any problem and just keep the _q() removal, the test for angular velocity and the fix of the default Nones in "from_relative". Although I have to admit I am a big fan of type annotations to empower mypy to detect as much problems as possible.

ismaelJimenez:
The build target py38_test_windows didn´t like the refactor from Optional to "|", so I had reverted it back. Maybe in some time this new functionality would work on that target

astrojuanlu:
Looks like the error you observed on Windows is a case of #959:

```

    def __init__(self, imgtype, name=None, cnf={}, master=None, **kw):
        self.name = None
        if not master:
            master = _get_default_root('create image')
        self.tk = getattr(master, 'tk', master)
        if not name:
            Image._last_id += 1
            name = "pyimage%r" % (Image._last_id,) # tk itself would use image<x>
        if kw and cnf: cnf = _cnfmerge((cnf, kw))
        elif kw: cnf = kw
        options = ()
        for k, v in cnf.items():
            if callable(v):
                v = self._register(v)
            options = options + ('-'+k, v)
>       self.tk.call(('image', 'create', imgtype, name,) + options)
E       _tkinter.TclError
```

Could you please revert the reversion? :) And then I'll repeat the review on top of that

astrojuanlu:
I was writing a long response on this thread, but I ended up moving it to its own issue. https://github.com/poliastro/poliastro/issues/1230

Let's use this pull request to only remove the extra `_q` so we can discuss a bit more about the annotation of the library. Thanks @ismaelJimenez for kickstarting this!

src/poliastro/bodies.py
```diff
@@ -105,12 +107,10 @@ def from_parameters(cls, parent, k, name, symbol, R, **kwargs):
...
     @classmethod
-    def from_relative(
-        cls, reference, parent=None, k=None, name=None, symbol=None, R=None, **kwargs
-    ):
-        k = k * reference.k
-        R = R * reference.R
-        return cls(parent, k, name, symbol, R, **kwargs)
+    def from_relative(cls, reference, parent, k, name, symbol=None, R=0, **kwargs):
+        _k: Quantity = k * reference.k
+        _R: Quantity = R * reference.R
```
astrojuanlu, line 112:
```suggestion
        _k = k * reference.k
        _R = R * reference.R
```

astrojuanlu (approved):
Thanks a lot @ismaelJimenez ! 🚀

Issue #1076, opened 2021-01-14
`test_perturbations.py` takes a lot of time


Yash-10:
While testing using `pytest`, the test hangs for a sufficiently long time while `test_perturbations.py` runs.

Running the following shows that it takes ~140s to run `test_solar_pressure` in `test_perturbations.py` which when combined with other durations, gives a significant test time as a whole.
```
pytest --durations=4
```

**Output:**
```
======================================================================== slowest 4 durations =========================================================================
143.81s call     tests/tests_twobody/test_perturbations.py::test_solar_pressure[200-deltas_expected0]
142.41s call     tests/tests_twobody/test_perturbations.py::test_solar_pressure[400-deltas_expected1]
138.72s call     tests/tests_twobody/test_perturbations.py::test_solar_pressure[600-deltas_expected2]
59.31s call     tests/tests_twobody/test_perturbations.py::test_3rd_body_Curtis[test_params2]
```
Specifically, the problem seems to be in the [`test_solar_pressure`](https://github.com/poliastro/poliastro/blob/master/tests/tests_twobody/test_perturbations.py#L613) function. This particularly stood out to me since other tests take much less time.

Thanks!

Yash-10:
It got closed by mistake, I am sorry for that. I suppose the `support` label was added automatically?

DhruvSondhi:
Hello @Yash-10, I would like to add somethings that I had found to this issue. Actually, it is meant to be this way because of the following reasons :
- `pytest.mark.slow` marker is a custom marker set in the test suite signifying that the `test is slow` to run like this 
```
@pytest.mark.slow: mark test as slow.
```
- `pytest.mark.parameterize` is used for testing different parameters set for different values of `t_days`
- The test also uses the strict marker tag for checking the test according to the different attributes.

I don't have the knowledge required for checking out the implementation at the given moment but would get back to you after confirming the same. 
A rather better explanation may be provided by **astrojuanlu** or **jorgepiloto** :)

Yash-10:
Thanks for you reply! But I think that there are other tests which are marked as `slow` but take significantly less time to compute than the `test_solar_pressure` function stated above. Majority of the tests take under 1s to compute. So according to me, there might be a place to investigate here. 

I am currently finding out the reason but after preliminary findings, it seems that

```
rr, vv = cowell(
            Earth.k,
            initial.r,
            initial.v,
            np.linspace(0, (tof).to(u.s).value, 4000) * u.s,
[...]

PR #1227, opened 2021-05-15
Add SciPy.integrate.odeint as option to Cowell propagation

jtegedor:
Includes:

* Add odeint integrator as option in Cowell propagation (See https://github.com/poliastro/poliastro/issues/1076 for additional info)
* Improved `alinspace` to handle angle differences larger than 2*pi and fixed `XFAILED` test


astrojuanlu (changes requested):
Hi @jtegedor, first of all thanks for this pull request and sorry for the delay. After discussing this a bit in our community calls, we reached the conclusion that we are not very keen on this approach:

- The performance gain, although significant, is quite small, and cannot be improved from there.
- `odeint` is the "legacy" interface, written in Fortran, and does not support events and other features. We would like to continue using the new one, and also keep working on switching to numbakit-ode (see #1042)
- I detailed in #1076 other things we can do to make our code faster.

Hopefully we will only have to provide support for `odeint` as last resort, if at all.

jtegedor:
Understood, no worries. Closing this PR.

PR #1228, opened 2021-05-16
Fix units handling in orbit.from_sbdb

jtegedor:
While working on https://github.com/poliastro/poliastro/pull/1227, there was a separate problem in `orbit.from_sdbd`. This shall fix it.


astrojuanlu (approved):
This started failing overnight after https://github.com/astropy/astroquery/pull/2011 was released, see discussion in https://github.com/poliastro/poliastro/issues/1229. Thanks for the catch @jtegedor, merging!

Issue #265, opened 2017-10-12
Propagating long times does not converge


Juanlu001:
🐞 **Problem**

Propagating long times fails in different ways:

```python
In [5]: iss.propagate(100 * u.year)
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-27-503eb53e62e7> in <module>()
----> 1 iss.propagate(100 * u.year)

~/Personal/poliastro/poliastro-library/src/poliastro/twobody/orbit.py in propagate(self, epoch_or_duration, rtol)
    245             time_of_flight = time.TimeDelta(epoch_or_duration)
    246 
--> 247         return propagate(self, time_of_flight, rtol=rtol)
    248 
    249     def sample(self, *, num_points=None, time_values=None):

~/Personal/poliastro/poliastro-library/src/poliastro/twobody/propagation.py in propagate(orbit, time_of_flight, method, rtol, **kwargs)
    156                   time_of_flight.to(u.s).value,
    157                   rtol=rtol,
--> 158                   **kwargs)
    159     return orbit.from_vectors(orbit.attractor, r * u.km, v * u.km / u.s, orbit.epoch + time_of_flight)
    160 

~/Personal/poliastro/poliastro-library/src/poliastro/twobody/propagation.py in kepler(k, r0, v0, tof, rtol, numiter)
    139     f, g, fdot, gdot = _kepler(k, r0, v0, tof, numiter, rtol)
    140 
--> 141     assert np.abs(f * gdot - fdot * g - 1) < 1e-5  # Fixed tolerance
    142 
    143     # Return position and velocity vectors

AssertionError: 

In [6]: halleys[0].propagate(73 * u.year)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-26-9dcacad8e904> in <module>()
----> 1 halleys[0].propagate(73 * u.year)

~/Personal/poliastro/poliastro-library/src/poliastro/twobody/orbit.py in propagate(self, epoch_or_duration, rtol)
    245             time_of_flight = time.TimeDelta(epoch_or_duration)
    246 
--> 247         return propagate(self, time_of_flight, rtol=rtol)
    248 
    249     def sample(self, *, num_points=None, time_values=None):

~/Personal/poliastro/poliastro-library/src/poliastro/twobody/propagation.py in propagate(orbit, time_of_flight, method, rtol, **kwargs)
    156                   time_of_flight.to(u.s).value,
[...]

Issue #557, opened 2019-01-31
Allow passing frames in OrbitPlotter2D.set_frame()


shreyasbapat:
🐞 **Problem**

Currently, we can set frame by passing *Orbit.pqw() in set_frame which is not user friendly and not intuitive.  

🖥 **Please paste the output of following commands**

* `conda info -a` (only if you have conda)
* `conda list` (only if you have conda)
* `pip freeze`

<details>
  
```bash
# Paste your output here:

```
</details>

* `python -c "import poliastro.testing; poliastro.testing.test()"`

<details>

```bash
# Paste your output here:

```
</details>

🎯 **Goal**
The current way of setting frame is not intuitive. Probably directly use set_frame(frame=GSE) for setting frame.

💡 **Possible solutions**

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
Reminder: https://github.com/poliastro/poliastro/wiki/Orbits-and-reference-frames#proposed-steps

Juanlu001:
Notice that all the functionality is already there (I clarified that in the wiki) and this is ready to be implemented now.

shreyasbapat:
Do we still give an option to pass pqw ? Or we directly pass the frame?

Juanlu001:
We should remove this pqw stuff, it's just a hack. Too inconsistent and fragile to be generally useful.

Juanlu001:
I think we probably need some more time to get this in poliastro safely. Release time is approaching and there's nothing to lose by delaying this to the next one, plus we will have more time to think about frames, what to do with the bodies, add planetary constants, improve validation... Therefore, I am changing the milestone.

astrojuanlu:
This was addressed in #891:

> This adds a new method to all orbit plotters, `plot_body_orbit`, that "Plots complete revolution of body and current position if given", and another two for 2D orbit plotters, `set_orbit_frame` and `set_body_frame`. The former avoids calling `Orbit.from_body_ephem` (about to be removed), and the latter avoid the need to call `set_frame(*orbit.pqw())`, which is ugly, impossible to understand, and in general shouldn't be there.

[...]

PR #1235, opened 2021-05-27
Addressing more test warnings

Yash-10:
Continuing work from #1137. Thanks @ishanSrt !

*Some analysis*
- The dubious year warnings (coming from `test_propagate_long_times_keeps_geometry`) due to the dates being far into the future seem to be legitimate, since that test propagates an orbit for 100 years and hence the resultant orbit is mostly going to be out of the non-dubious range defined by `erfa`.
- I thought it might be a good choice to just filter the warning since after looking at [issue 256](https://github.com/poliastro/poliastro/issues/265), I wondered whether it would be good to still check for long propagations like 100 years. Moreover, reducing the propagation time while still keeping it long enough, say, 60 years didn't seem to solve it.
- For the second set of `erfa` warnings, eg:

```
ERFA function "epv00" yielded 8 of "warning: date outsidethe range 1900-2100 AD"
```
Any date after about the year 1935 yields that warning. Here, it might be a good choice to just choose a "safe" date value to prevent these warnings pop up.

Thanks, and would love to get any suggestions!

---
The coverage decrease needs to be addressed and I am working on that...


Yash-10:
Points and questions regarding coverage:

- The test `test_misc.py` is modified such that no warnings are raised up. Doing this has decreased coverage for `misc.py` since it now doesn't test for the case when no `epoch` argument is passed, i.e. the default epoch of `J2000` is used which, as per the above analysis, is bound to generate *"warning: date outsidethe range 1900-2100 AD"*.

Since testing for the case when no epoch argument is passed might be more important than handling warnings, it might make sense to just suppress the warning if no better solution can be reached.

- The reason for ignoring the `dubious year` warnings, again from the PR description might be a good choice since testing for long propagations might be important.

**Question**
Since `Orbit.pqw()` is now deprecated and is going to be removed (as per the warning message) and that tests for it are also removed, would it be fine to remove `Orbit.pqw` completely? I think this might stabilize the coverage because currently the tests for `pqw` are removed but it is still in the source code. Just wanted to clarify this before making any changes :)

Thanks!

Yash-10:
@astrojuanlu Could you please guide me on the above and let me know your thoughts?

Thanks!

astrojuanlu:
Hello @Yash-10! I'm about to review this PR. Would you please rebase on top of `main`?

tests/tests_plotting/test_misc.py
```diff
@@ -26,6 +39,6 @@ def test_plot_inner_solar_system_static(earth_perihelion):
 
 @pytest.mark.mpl_image_compare
 def test_plot_outer_solar_system_static(earth_perihelion):
-    plot_solar_system(outer=True, epoch=earth_perihelion)
+    plot_solar_system(outer=True, epoch=Time("1934-07-14 07:59", scale="tdb"))
```
astrojuanlu, line 42: From `conftest.py`, `earth_perihelion` is `"2020-01-05 07:47:00"`. I get that, by using an epoch in the past, we avoid going outside the range of `epv00`.

However, I think that the correct solution here is to silence such warnings inside `plot_solar_system`. Yes, we know that getting the position of the Earth 100 years from now is unreliable, but for plotting we don't care.

tests/tests_twobody/test_orbit.py
```diff
@@ -197,17 +197,6 @@ def test_parabolic_has_zero_energy():
     assert_allclose(ss.energy.value, 0.0, atol=1e-16)
 
 
-def test_pqw_for_circular_equatorial_orbit():
```
astrojuanlu, line 200: We can't remove `Orbit.pqw` yet, because we need it for the plotting:

https://github.com/poliastro/poliastro/blob/49ae31844b2d59a6b052a42bc64f34efd0f870fb/src/poliastro/plotting/_base.py#L346-L357

We are tracking that in https://github.com/poliastro/poliastro/issues/557.

In this case, let's silence the deprecation warning inside the test.

tests/tests_twobody/test_orbit.py
```diff
@@ -571,27 +560,6 @@ def test_from_horizons_raise_valueerror():
...
-        Sun,
-        2.76710759221651 * u.au,
-        0.07554803091400027 * u.one,
-        27.18502494739172 * u.deg,
-        23.36913218336299 * u.deg,
-        132.2919809219236 * u.deg,
-        21.28957916690369 * u.deg,
-        epoch,
-    )
-    ss1 = Orbit.from_horizons(name="Ceres", attractor=Sun, epoch=epoch)
```
astrojuanlu, line 589: Time to remove `Orbit.from_horizons` :)

tests/tests_twobody/test_orbit.py
```diff
@@ -887,34 +855,13 @@ def test_convert_from_coe_to_rv():
     assert_quantity_allclose(nu, expected_nu, rtol=1e-4)
 
 
-def test_perifocal_points_to_perigee():
```
astrojuanlu, line 890: Let's keep this test, see above

tests/tests_twobody/test_orbit.py
```diff
@@ -887,34 +855,13 @@ def test_convert_from_coe_to_rv():
...
-
-
 def test_arglat_within_range():
     r = [3_539.08827417, 5_310.19903462, 3_066.31301457] * u.km
     v = [-6.49780849, 3.24910291, 1.87521413] * u.km / u.s
     ss = Orbit.from_vectors(Earth, r, v)
     assert 0 * u.deg <= ss.arglat <= 360 * u.deg
 
 
-def test_pqw_returns_dimensionless():
```
astrojuanlu, line 906: Let's keep this test, see above. (But silence the deprecation warning.)

tests/tests_twobody/test_propagation.py
```diff
@@ -335,6 +335,7 @@ def test_propagate_to_date_has_proper_epoch():
     assert (ss1.epoch - final_epoch).sec == approx(0.0, abs=1e-6)
 
 
+@pytest.mark.filterwarnings("ignore::erfa.core.ErfaWarning")
```
astrojuanlu, line 338: Good!

astrojuanlu (changes requested):
Did a thorough review, thanks for your patience!

tests/tests_twobody/test_orbit.py
Yash-10: Just to confirm what you meant: We should remove `Orbit.from_horizons` from `src` and their corresponding tests, right?

Sorry, I was confused with our intention regarding `Orbit.from_horizons`..

astrojuanlu: Yes, let's remove it! It was deprecated in 0.14 I think

astrojuanlu (approved):
This looks good to go! My only remaining question is: did we have to regenerate the images? The changes look like a 1 pixel displacement, and a diff on the tens or hundreds of bytes. Were image tests failing locally because of it? If not, I think we should remove the changes and squash the commit, to avoid increasing the size of the repository.

astrojuanlu:
Merging & squashing after the tests pass 👍🏽

astrojuanlu:
(Not sure the automerge will work because of the coverage check, but let's see)

PR #1237, opened 2021-05-30
Change `.format` to f-strings

Yash-10:
A slight movement from `.format` to f-strings might take up some less space and look better?

Thanks!

### DIFF
--- a/src/poliastro/earth/atmosphere/base.py
+++ b/src/poliastro/earth/atmosphere/base.py
@@ -74,9 +74,7 @@ class COESA:
         # Assert in range
         if z < self.zb_levels[0] or z > self.zb_levels[-1]:
             raise ValueError(
-                "Geometric altitude must be in range [{}, {}]".format(
-                    self.zb_levels[0], self.zb_levels[-1]
-                )
+                f"Geometric altitude must be in range [{self.zb_levels[0]}, {self.zb_levels[-1]}]"
             )
 
         return z, h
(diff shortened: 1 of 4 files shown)


jorgepiloto (changes requested):
Thank you for this little improvement @Yash-10.

I noticed that you decided to kept the strings being formatted with: `MACRO_STRING.format` such us `ORBIT_NO_FRAME_FORMAT.format(...)`. I guess this is because it it is difficult to modify these macros behavior so they can work with f-strings, making the code not so elegant.

Taking advantage of this, let us also fix the following line in one of the tests:
https://github.com/poliastro/poliastro/blob/bebbf617e0ebfb797ba1c593b8241e4cf0f9c466/tests/tests_twobody/test_orbit.py#L1242

Yash-10:
> I noticed that you decided to kept the strings being formatted with: MACRO_STRING.format such us ORBIT_NO_FRAME_FORMAT.format(...). I guess this is because it it is difficult to modify these macros behavior so they can work with f-strings, making the code not so elegant.

Yeah, I had read that it is sometimes better to stick with the older `.format` in cases like the one you proposed.

> Taking advantage of this, let us also fix the following line in one of the tests:

Thanks for this @jorgepiloto ! I think I addressed your comment..

jorgepiloto (approved):
Thanks for the changes @Yash-10. Please, squash the commits and this is ready to be merged if @astrojuanlu approves :+1:

astrojuanlu (approved):

astrojuanlu:
Thanks a lot @Yash-10 !

Issue #1236, opened 2021-05-28
Increase `coverage` for some files?


Yash-10:
Running `pytest --cov` gives some stats about the coverage. For example, it might look like the below

<details>

```bash
Name                                                                 Stmts   Miss Branch BrPart  Cover
------------------------------------------------------------------------------------------------------
src/poliastro/__init__.py                                                1      0      0      0   100%
src/poliastro/bodies.py                                                 54      7      8      2    82%
src/poliastro/constants/__init__.py                                      3      0      0      0   100%
src/poliastro/constants/general.py                                      62      0      0      0   100%
src/poliastro/constants/rotational_elements.py                          12      0      0      0   100%
src/poliastro/core/__init__.py                                           0      0      0      0   100%
src/poliastro/core/angles.py                                            91     62     14      0    28%
src/poliastro/core/elements.py                                          85     65     10      0    21%
src/poliastro/core/fixed.py                                             96     76      0      0    21%
src/poliastro/core/flybys.py                                            23     17      0      0    26%
src/poliastro/core/hyper.py                                             15     11      4      0    21%
src/poliastro/core/iod.py                                              183    147     52      0    15%
src/poliastro/core/maneuver.py                                          58     47      0      0    19%
src/poliastro/core/perturbations.py                                     58     27      2      0    55%
src/poliastro/core/propagation/__init__.py                             300    266     36      0    10%
src/poliastro/core/propagation/farnocchia.py                           146    120     38      0    14%
src/poliastro/core/spheroid_location.py                                 56     38      0      0    32%
src/poliastro/core/stumpff.py                                           33     26     12      0    16%
src/poliastro/core/thrust/__init__.py                                    0      0      0      0   100%
src/poliastro/core/thrust/change_a_inc.py                               27     14      0      0    48%
src/poliastro/core/thrust/change_argp.py                                13      6      0      0    54%
src/poliastro/core/thrust/change_ecc_quasioptimal.py                    12      5      0      0    58%
src/poliastro/core/thrust/change_inc_ecc.py                             16      7      0      0    56%
src/poliastro/core/util.py                                              27     18     12      0    23%
src/poliastro/czml/__init__.py                                           0      0      0      0   100%
src/poliastro/czml/extract_czml.py                                     126      7     40      8    91%
src/poliastro/czml/utils.py                                             12      0      2      0   100%
src/poliastro/earth/__init__.py                                         42      0     10      0   100%
[...]

PR #1239, opened 2021-05-31
Frame computation calculation fixing

Yash-10:
In this pull request, I have tried to rectify the mistake done by me in #1190 .

**Details**

- The `core` module uses the `np.deg2rad` function to address the needed conversion. It is done since `numba` loves `numpy` and the use of `astropy.Quantity` objects in `core`, I think, is discouraged.
- The `epoch` argument in `rot_elements_at_epoch` is made optional since the docstring earlier signified it to be optional, but the code implicitly assumed it to be always passed by the user.
- A test has been added to make sure the implementation works. (This might contribute a small amount to #1236). The expected values for the test were taken from the output of using `rot_elements_at_epoch` on fixed frames of bodies using the source code before the commit of #1190.

@jorgepiloto I would love to get your thoughts on this!

Thanks!

---
If I run the added test but keeping the code in `core` as per the commits in #1190, I get error:

```bash
FAILED tests/test_frames.py::test_fixed_frame_calculation_gives_expected_result[body1-MercuryFixed-radecW1] - AssertionError: 
FAILED tests/test_frames.py::test_fixed_frame_calculation_gives_expected_result[body3-MarsFixed-radecW3] - AssertionError: 
FAILED tests/test_frames.py::test_fixed_frame_calculation_gives_expected_result[body4-JupiterFixed-radecW4] - AssertionError: 
FAILED tests/test_frames.py::test_fixed_frame_calculation_gives_expected_result[body7-NeptuneFixed-radecW7] - AssertionError: 
FAILED tests/test_frames.py::test_fixed_frame_calculation_gives_expected_result[body8-MoonFixed-radecW8] - AssertionError: 
```
Hence, this might be an indication that this pull request would probably resolve the issue.

### DIFF
--- a/src/poliastro/core/fixed.py
+++ b/src/poliastro/core/fixed.py
@@ -60,11 +60,11 @@ def mercury_rot_elements_at_epoch(T, d):
         Right ascension and declination of north pole, and angle of the prime meridian.
 
     """
-    M1 = 174.7910857 + 4.092335 * d
-    M2 = 349.5821714 + 8.184670 * d
-    M3 = 164.3732571 + 12.277005 * d
-    M4 = 339.1643429 + 16.369340 * d
-    M5 = 153.9554286 + 20.461675 * d
+    M1 = np.deg2rad(174.7910857 + 4.092335 * d)
+    M2 = np.deg2rad(349.5821714 + 8.184670 * d)
+    M3 = np.deg2rad(164.3732571 + 12.277005 * d)
+    M4 = np.deg2rad(339.1643429 + 16.369340 * d)
+    M5 = np.deg2rad(153.9554286 + 20.461675 * d)
     ra = 281.0103 - 0.0328 * T
     dec = 61.45 - 0.005 * T
     W = (329.5988 + 6.1385108 * d) + (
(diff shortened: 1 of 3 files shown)


astrojuanlu (approved):
Looks good to me, but waiting for @jorgepiloto to give the final review!

src/poliastro/core/fixed.py
```diff
@@ -60,11 +60,11 @@ def mercury_rot_elements_at_epoch(T, d):
         Right ascension and declination of north pole, and angle of the prime meridian.
 
     """
-    M1 = 174.7910857 + 4.092335 * d
-    M2 = 349.5821714 + 8.184670 * d
-    M3 = 164.3732571 + 12.277005 * d
-    M4 = 339.1643429 + 16.369340 * d
-    M5 = 153.9554286 + 20.461675 * d
+    M1 = np.deg2rad(174.7910857 + 4.092335 * d)
```
jorgepiloto, line 63: Looks good to me too! However, I would like to suggest one thing and know your opinion about it: we are calling several times the `np.deg2rad` function, what about implementing the following at the very beginning of the file:

```python
DEG2RAD = np.deg2rad(1)
```
so we can do:

```python
M1 = number_in_degrees * DEG2RAD
```

What do you think?

astrojuanlu: All these functions are jitted, so I don't expect a big difference. However, I might be wrong :)

Yash-10: The execution times seem to be almost the same with the approach of using `numpy` multiple times being marginally faster (`~1e-2 s`). From a short analysis, this marginal increase might be because

```py
DEG2RAD = np.deg2rad(1)
```
would not be inside any jitted function and hence introduces some more execution time, although using just `DEG2RAD` inside the jitted functions is slightly faster than this PR's implementation.

As @astrojuanlu said, the differences are not that big, so I wonder what option could we choose here?

jorgepiloto: Yeah, let us use the original `np.deg2rad` implementation. Thanks for the comments 🚀

jorgepiloto (approved):
Merging this, thanks for addressing it @Yash-10 👍🏽

jorgepiloto:
Now, paying attention to the latest validation actions run, see https://github.com/poliastro/validation/actions/runs/895277143 🤞🏽

jorgepiloto:
All checks passing 🎉

PR #1241, opened 2021-06-01
Unpin Jinja2

astrojuanlu:


astrojuanlu:
Thanks to the new nbsphinx release we don't need to pin Jinja2 anymore. Docs and general tests are passing. Merging!

PR #1243, opened 2021-06-08
Slight modification to `shadow_function`

Yash-10:
I thought the existing `shadow_function` checked for the condition:

```py
theta < theta_1 + theta_2
```
which, according to the description in Howard Curtis returns `False` if a satellite is in earth's shadow.

Here, I tried to just change the condition so that the function returns `True` if the satellite is in earth's shadow. rather than `False`. This modification would make it more clearer?


astrojuanlu (approved):
Fair enough! Yes I think it's more clear now :) And also thanks for the docstring improvements!

PR #1246, opened 2021-06-11
Eclipse event

Yash-10:
Currently a draft for adding eclipse detection capabilities. Needs some more work for it to work correctly :)

The tests are also not complete. Thinking of some ways to check the implementation...

### DIFF
--- a/src/poliastro/core/events.py
+++ b/src/poliastro/core/events.py
@@ -2,64 +2,45 @@ import numpy as np
 from numba import njit as jit
 from numpy.linalg import norm
 
-from poliastro.core.elements import rv2coe
+from poliastro.core.elements import rv2coe, coe_rotation_matrix
 
 
 @jit
-def continuous_shadow_function_for_penumbra(k, u_, r_sec, R_sec, R_primary):
+def eclipse_function(k, u_, r_sec, R_sec, R_primary, umbra=True):
+    """Calculates a continuous shadow function.
+
+    Parameters
+    ----------
+    k: float
+        Standard gravitational parameter (km^3 / s^2).
+    u_: ~np.array
+        Satellite position and velocity vector with respect to the primary body.
+    r_sec: ~np.array
+        Position vector of the secondary body with respect to the primary body.
+    R_sec: float
+        Radius of the secondary body.
+    R_primary: float
+        Radius of the primary body.
+    umbra: bool
+        Whether to calculate the shadow function for umbra or penumbra, defaults to True
+        i.e. calculates for umbra.
+
+    """
+    # Plus or minus condition
+    pm = 1 if umbra else -1
     p, ecc, inc, raan, argp, nu = rv2coe(k, u_[:3], u_[3:])
 
-    Px = np.cos(raan) * np.cos(argp) - np.sin(raan) * np.sin(argp) * np.cos(inc)
-    Py = np.sin(raan) * np.cos(argp) + np.cos(raan) * np.sin(argp) * np.cos(inc)
-    Pz = np.sin(argp) * np.sin(inc)
-    P_ = np.array([Px, Py, Pz])
-    Qx = -np.cos(raan) * np.sin(argp) - np.sin(raan) * np.cos(argp) * np.cos(inc)
-    Qy = -np.sin(raan) * np.sin(argp) + np.cos(raan) * np.cos(argp) * np.cos(inc)
-    Qz = np.cos(argp) * np.sin(inc)
-    Q_ = np.array([Qx, Qy, Qz])
+    PQW = coe_rotation_matrix(inc, raan, argp)
+    # Make arrays contiguous for faster dot product with numba.
+    P_, Q_ = np.ascontiguousarray(PQW[:, 0]), np.ascontiguousarray(PQW[:, 1])
 
     r_sec_norm = norm(r_sec)
     beta = np.dot(P_, r_sec) / r_sec_norm
     zeta = np.dot(Q_, r_sec) / r_sec_norm
 
-    delta_p = np.arcsin((R_sec + R_primary) / r_sec_norm)
+    sin_delta_shadow = np.sin((R_sec - pm * R_primary) / r_sec_norm)
 
     cos_psi = beta * np.cos(nu) + zeta * np.sin(nu)
-    shadow_function = (
-        ((R_primary ** 2) * (1 + ecc * np.cos(nu)) ** 2)
-        + (p ** 2) * (cos_psi ** 2)
-        - p ** 2
-        - (2 * p * R_primary * cos_psi) * (1 + ecc * np.cos(nu)) * np.sin(delta_p)
-    )
-
-    return shadow_function
-
-
-@jit
-def continuous_shadow_function_for_umbra(k, u_, r_sec, R_sec, R_primary):
-    p, ecc, inc, raan, argp, nu = rv2coe(k, u_[:3], u_[3:])
-
-    Px = np.cos(raan) * np.cos(argp) - np.sin(raan) * np.sin(argp) * np.cos(inc)
-    Py = np.sin(raan) * np.cos(argp) + np.cos(raan) * np.sin(argp) * np.cos(inc)
-    Pz = np.sin(argp) * np.sin(inc)
-    P_ = np.array([Px, Py, Pz])
-    Qx = -np.cos(raan) * np.sin(argp) - np.sin(raan) * np.cos(argp) * np.cos(inc)
-    Qy = -np.sin(raan) * np.sin(argp) + np.cos(raan) * np.cos(argp) * np.cos(inc)
-    Qz = np.cos(argp) * np.sin(inc)
-    Q_ = np.array([Qx, Qy, Qz])
-
-    r_sec_norm = norm(r_sec)
-    beta = np.dot(P_, r_sec) / r_sec_norm
-    zeta = np.dot(Q_, r_sec) / r_sec_norm
-
-    delta_u = np.arcsin((R_sec - R_primary) / r_sec_norm)
-
-    cos_psi = beta * np.cos(nu) + zeta * np.sin(nu)
-    shadow_function = (
-        ((R_primary ** 2) * (1 + ecc * np.cos(nu)) ** 2)
-        + (p ** 2) * (cos_psi ** 2)
-        - p ** 2
-        + (2 * p * R_primary * cos_psi) * (1 + ecc * np.cos(nu)) * np.sin(delta_u)
-    )
+    shadow_function = ((R_primary ** 2) * (1 + ecc * np.cos(nu)) ** 2) + (p ** 2) * (cos_psi ** 2) - p ** 2 + pm * (2 * p * R_primary * cos_psi) * (1 + ecc * np.cos(nu)) * sin_delta_shadow
 
     return shadow_function
--- a/src/poliastro/core/events.py
+++ b/src/poliastro/core/events.py
@@ -1,0 +1,65 @@
+import numpy as np
+from numba import njit as jit
+from numpy.linalg import norm
+
+from poliastro.core.elements import rv2coe
+
+
+@jit
+def continuous_shadow_function_for_penumbra(k, u_, r_sec, R_sec, R_primary):
+    p, ecc, inc, raan, argp, nu = rv2coe(k, u_[:3], u_[3:])
+
+    Px = np.cos(raan) * np.cos(argp) - np.sin(raan) * np.sin(argp) * np.cos(inc)
+    Py = np.sin(raan) * np.cos(argp) + np.cos(raan) * np.sin(argp) * np.cos(inc)
+    Pz = np.sin(argp) * np.sin(inc)
+    P_ = np.array([Px, Py, Pz])
+    Qx = -np.cos(raan) * np.sin(argp) - np.sin(raan) * np.cos(argp) * np.cos(inc)
+    Qy = -np.sin(raan) * np.sin(argp) + np.cos(raan) * np.cos(argp) * np.cos(inc)
+    Qz = np.cos(argp) * np.sin(inc)
+    Q_ = np.array([Qx, Qy, Qz])
+
+    r_sec_norm = norm(r_sec)
+    beta = np.dot(P_, r_sec) / r_sec_norm
+    zeta = np.dot(Q_, r_sec) / r_sec_norm
+
+    delta_p = np.arcsin((R_sec + R_primary) / r_sec_norm)
+
+    cos_psi = beta * np.cos(nu) + zeta * np.sin(nu)
+    shadow_function = (
+        ((R_primary ** 2) * (1 + ecc * np.cos(nu)) ** 2)
+        + (p ** 2) * (cos_psi ** 2)
+        - p ** 2
+        - (2 * p * R_primary * cos_psi) * (1 + ecc * np.cos(nu)) * np.sin(delta_p)
+    )
+
+    return shadow_function
+
+
+@jit
+def continuous_shadow_function_for_umbra(k, u_, r_sec, R_sec, R_primary):
+    p, ecc, inc, raan, argp, nu = rv2coe(k, u_[:3], u_[3:])
+
+    Px = np.cos(raan) * np.cos(argp) - np.sin(raan) * np.sin(argp) * np.cos(inc)
+    Py = np.sin(raan) * np.cos(argp) + np.cos(raan) * np.sin(argp) * np.cos(inc)
+    Pz = np.sin(argp) * np.sin(inc)
+    P_ = np.array([Px, Py, Pz])
+    Qx = -np.cos(raan) * np.sin(argp) - np.sin(raan) * np.cos(argp) * np.cos(inc)
+    Qy = -np.sin(raan) * np.sin(argp) + np.cos(raan) * np.cos(argp) * np.cos(inc)
+    Qz = np.cos(argp) * np.sin(inc)
+    Q_ = np.array([Qx, Qy, Qz])
+
+    r_sec_norm = norm(r_sec)
+    beta = np.dot(P_, r_sec) / r_sec_norm
+    zeta = np.dot(Q_, r_sec) / r_sec_norm
+
+    delta_u = np.arcsin((R_sec - R_primary) / r_sec_norm)
+
+    cos_psi = beta * np.cos(nu) + zeta * np.sin(nu)
+    shadow_function = (
+        ((R_primary ** 2) * (1 + ecc * np.cos(nu)) ** 2)
+        + (p ** 2) * (cos_psi ** 2)
+        - p ** 2
+        + (2 * p * R_primary * cos_psi) * (1 + ecc * np.cos(nu)) * np.sin(delta_u)
+    )
+
+    return shadow_function
--- a/src/poliastro/core/events.py
+++ b/src/poliastro/core/events.py
@@ -2,7 +2,7 @@ import numpy as np
 from numba import njit as jit
 from numpy.linalg import norm
 
-from poliastro.core.elements import rv2coe, coe_rotation_matrix
+from poliastro.core.elements import coe_rotation_matrix, rv2coe
 
 
 @jit
@@ -41,6 +41,11 @@ def eclipse_function(k, u_, r_sec, R_sec, R_primary, umbra=True):
     sin_delta_shadow = np.sin((R_sec - pm * R_primary) / r_sec_norm)
 
     cos_psi = beta * np.cos(nu) + zeta * np.sin(nu)
-    shadow_function = ((R_primary ** 2) * (1 + ecc * np.cos(nu)) ** 2) + (p ** 2) * (cos_psi ** 2) - p ** 2 + pm * (2 * p * R_primary * cos_psi) * (1 + ecc * np.cos(nu)) * sin_delta_shadow
+    shadow_function = (
+        ((R_primary ** 2) * (1 + ecc * np.cos(nu)) ** 2)
+        + (p ** 2) * (cos_psi ** 2)
+        - p ** 2
+        + pm * (2 * p * R_primary * cos_psi) * (1 + ecc * np.cos(nu)) * sin_delta_shadow
+    )
 
     return shadow_function
--- a/src/poliastro/twobody/events.py
+++ b/src/poliastro/twobody/events.py
@@ -1,10 +1,13 @@
 import numpy as np
 from astropy import units as u
+from astropy.coordinates import get_body_barycentric_posvel
 from numpy.linalg import norm
 
+from poliastro.core.elements import rv2coe
 from poliastro.core.spheroid_location import (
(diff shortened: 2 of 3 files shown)


astrojuanlu:
> 48: error: Name "test_penumbra" already defined on line 30

Yash-10:
Currently the `umbral_shadow` and `penumbral_shadow` functions work like this:

They calculate the satellite vertical distance and the umbral or penumbral vertical distances. Since this condition needs to be used inside `solve_ivp` (which checks for zero crossings), these two functions return the differences between the two (using the below geometry from Vallado).

Screenshot from 2021-06-12 16-14-00

The only tricky thing seems to be that  there is one more condition that the dot product of satellite and sun position vectors must be less than zero for it to be an eclipse. Currently, if the satellite is in sunlight, I just return the satellite's position vector instead of some garbage value (this was done since the function whose zero crossings are calculated need to be continuous). So its unification with the `__call__` method of umbra and penumbra events currently still might be suspectful. (*The tests are not complete at the moment.*)

Any thoughts or suggestions @astrojuanlu @jorgepiloto ?

Thanks!

Yash-10:
The cowell propagation works as expected if no eclipse event is detected. Currently there seems to be one problem: The position vector remains the same (i.e. hangs) if it was already in eclipse at the start of propagation or if it enters eclipse during the propagation. 

This behaviour is a bit confusing to me since the `terminal` property for both `penumbra` and `umbra` are set to `False`, so the integration must go on. Something needs investigation here. Maybe it has to do with the return values inside `umbral_shadow` or `penumbral_shadow`...

@astrojuanlu @jorgepiloto Could you have a look at this?

Thanks!

src/poliastro/core/perturbations.py
```diff
@@ -204,6 +204,76 @@ def shadow_function(r_sat, r_sun, R):
     return theta > theta_1 + theta_2
 
 
+@jit
+def umbral_shadow(r_sat, r_sun, R):
```
jorgepiloto, line 208: I think we should call this function `is_in_umbral_shadow`. Therefor, we could take the advantage in the `__call__` method of the `PenumbraEvent` of doing:

```python
return 0 if is_in_umbral_shadow(...) is True else 1
```

Notice that the `__call__` method returns a `0` if the event has happened or a different number if not.

src/poliastro/core/perturbations.py
```diff
@@ -204,6 +204,76 @@ def shadow_function(r_sat, r_sun, R):
...
+    ----------
+    r_sat: numpy.ndarray
+        Position of the satellite in the frame of the attractor (km)
+    r_sun: numpy.ndarray
+        Position of the sun in the frame of the attractor (km)
+    R: float
+        Radius of the body (attractor) responsible for the penumbral shadow (km)
+
+    """
+    alpha_pen = np.deg2rad(0.269007205)
```
jorgepiloto, line 259: Again, this is the case for Sun-Earth system. We need to generalize, see previous comment.

jorgepiloto (changes requested):
This is on the right path @Yash-10. I left you some comments and suggestions which might be the reason behind the errors you claim.

src/poliastro/twobody/events.py
```diff
@@ -25,8 +30,78 @@ def terminal(self):
     def last_t(self):
         return self._last_t * u.s
 
-    def __call__(self, t, u, k):
+    def __call__(self, t, u_, k):
```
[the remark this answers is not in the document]
Yash-10: I noticed the `u` inside `__call__` collided with `astropy.units` which is also imported as `u`. Since `astropy.units` is generally imported as `u`, I changed this variable's name instead...

src/poliastro/twobody/events.py
```diff
@@ -25,8 +30,78 @@ def terminal(self):
...
+        Radius of the attractor.
+
+    """
+
+    def __init__(self, r_sun, R):
+        self._r_sun = r_sun
+        self._R = R
+        self._last_t = None
+
+    @property
```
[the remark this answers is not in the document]
Yash-10: Thanks!

src/poliastro/core/perturbations.py
```diff
@@ -204,6 +204,76 @@ def shadow_function(r_sat, r_sun, R):
...
+    ----------
+    r_sat: numpy.ndarray
+        Position of the satellite in the frame of the attractor (km)
+    r_sun: numpy.ndarray
+        Position of the sun in the frame of the attractor (km)
+    R: float
+        Radius of the body (attractor) responsible for the umbral shadow (km)
+
+    """
+    alpha_um = np.deg2rad(0.264121687)
```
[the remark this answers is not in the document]
Yash-10: Thanks again! I have generalised it now.

src/poliastro/twobody/events.py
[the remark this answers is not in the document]
jorgepiloto: Good catch @Yash-10!

src/poliastro/twobody/events.py
```diff
@@ -25,8 +30,84 @@ def terminal(self):
...
+
+    def __call__(self, t, u_, k):
+        self._last_t = t
+        r_sat = (u_[:3] << u.km).value
+        r_sun = self._r_sun.to(u.km).value
+        r_s = self._r_s.to(u.km).value
+        r_p = self._r_p.to(u.km).value
+        is_in_penumbra = is_in_penumbral_shadow_fast(r_sat, r_sun, r_s, r_p)
+
+        return 0 if is_in_penumbra == 0 else 1
```
Yash-10, line 75: I did this for the `return` condition inside `__call__` since `is_in_penumbra` returns the difference between vertical distances if it is in one of the shadow regions (umbra or penumbra) and returns the satellite's position vector if it is not in one of those shadow regions.

The reason for doing this was `r_sat_norm`, I thought, would never be zero since it is the distance from the center of the primary attractor and satellite would never go to the center of the Earth, hopefully :)

Importantly, doing, for example:

```py
return is_in_penumbra
```
didn't seem to work (the same bug: satellite's position vector hangs) and returning True or False inside `is_in_penumbral_shadow` also didn't seem to work for some reason.

src/poliastro/core/perturbations.py
Yash-10: Hi @jorgepiloto I was thinking of the following:

Would `is_in_umbral_shadow` returning the difference between the vertical distances (as is done in the current commit) be a better choice instead of returning True or False? Since this would allow us to identify whether a satellite is going into or out of a particular shadow type (umbra or penumbra). Am I missing anywhere?

`r_sat_norm` is returned if it is not in any shadow (there we could put any non-zero value but I tried to choose `r_sat_norm` since it was already computed and that it would never be zero since its the distance from the center of the attractor and the satellite can only reach upto the attractor's radius).

src/poliastro/twobody/events.py
```diff
@@ -25,8 +30,84 @@ def terminal(self):
...
+        return self._terminal
+
+    @property
+    def last_t(self):
+        return self._last_t * u.s
+
+    def __call__(self, t, u_, k):
+        self._last_t = t
+        r_sat = (u_[:3] << u.km).value
+        r_sun = self._r_sun.to(u.km).value
```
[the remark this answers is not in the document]
jorgepiloto: This might be a critical point. Vallado only provides the algorithm for a particular instant of time. But in this logic, we are propagating the orbit, which moves together with its attractor. I would like to ping @astrojuanlu here for a further discussion on this.

From my side, I will try to implement validation cases related to eclipse detection so we can be sure if these is properly implemented.

src/poliastro/twobody/events.py
jorgepiloto: I understand the idea behind returning the `r_sat_norm` but do not agree with it. See previous comments on boolean return variables. Notice that the `__call__` method must return a numerical value, hence my previous suggestion on doing:

```python
def __call__(...):
    return 0 if in_penumbra is True else 1
```

jorgepiloto (changes requested):
Hi @Yash-10, I made a more deep review on this, which can be summarized in the following points:

* I think the should implement the core logic of the events within a `core/events.py` file instead of doing so inside the `perturbations.py` module.
* These functions should return boolean types and the naming convention which we could use might be `in_umbra`, `in_perigee`...
* Finally, I raised a question about updating the position of the planet, as it also moves during propagation, introducing a change in the geometry of the problem (the shadow moves with the planet).

src/poliastro/core/perturbations.py
```diff
@@ -204,6 +204,84 @@ def shadow_function(r_sat, r_sun, R):
     return theta > theta_1 + theta_2
 
 
+@jit
```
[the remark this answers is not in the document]
Yash-10: Thanks, I have made that change. I had included it in `core/perturbations` earlier thinking it might be good to modify `radiation_pressure` accordingly once one knows whether the satellite is in sunlight, penumbra, or umbra. But we could anyways import those functions from `core/events`

src/poliastro/core/perturbations.py
```diff
@@ -204,6 +204,84 @@ def shadow_function(r_sat, r_sun, R):
...
+        Radius of the primary body (attractor) responsible for the umbral shadow (km).
+
+    """
+    R_p = norm(r_sun)
+    alpha_um = np.arcsin((r_s - r_p) / R_p)
+    alpha_pen = np.arcsin((r_s + r_p) / R_p)
+
+    dot_sun_sat = np.dot(r_sat, r_sun)
+
+    r_sat_norm = np.sqrt(np.sum(r_sat ** 2))
```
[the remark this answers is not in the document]
Yash-10: Thanks!

src/poliastro/core/perturbations.py
```diff
@@ -204,6 +204,84 @@ def shadow_function(r_sat, r_sun, R):
     return theta > theta_1 + theta_2
 
 
+@jit
+def is_in_umbral_shadow(r_sat, r_sun, r_s, r_p):
+    r"""Calculate whether a satellite is in umbra or not, follows Algorithm 34 from Vallado.
+
+    Parameters
```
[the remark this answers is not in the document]
Yash-10: Would using "secondary" and "primary" body be a more general term to include in the variable names? 

If yes, would it be good to use `R_s` for radius of the secondary body (the star) and `R_p` for the radius of the primary body (attractor)?

To summarize, I currently used:

```
r_sat: numpy.ndarray
    Position of the satellite in the frame of the attractor (km).
r_sec: numpy.ndarray
    Position of the secondary body in the frame of the attractor (km).
R_s: float
    Radius of the secondary body.
R_p: float
    Radius of the primary body (attractor) responsible for the umbral shadow (km).
```
Any suggestions on these?

src/poliastro/twobody/events.py
```diff
@@ -25,8 +30,84 @@ def terminal(self):
...
+    Parameters
+    ----------
+    r_sun: ~astropy.units.Quantity
+        Position vector of the Sun with respect to the Earth.
+    R: ~astropy.units.Quantity
+        Radius of the attractor.
+
+    """
+
+    def __init__(self, r_sun, r_s, r_p, terminal=False):
```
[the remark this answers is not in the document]
Yash-10: Thanks again. Since `r_sun` is the position of secondary body wrt primary body, it might not be related to the orbit? So it would need to be passed as a separate argument, right?

*(`r_sun` is changed to `r_sec`)*

src/poliastro/core/perturbations.py
```diff
@@ -204,6 +204,84 @@ def shadow_function(r_sat, r_sun, R):
...
+        angle = np.arccos(dot_sun_sat / r_sat_norm / r_sun_norm)
+        sat_horiz = np.abs(r_sat_norm * np.cos(angle))
+        sat_vert = np.abs(r_sat_norm * np.sin(angle))
+        x = r_p / np.sin(alpha_pen)
+        pen_vert = np.tan(alpha_pen) * (x + sat_horiz)
+        if sat_vert <= pen_vert:
+            y = r_p / np.sin(alpha_um)
+            umb_vert = np.tan(alpha_um) * (y - sat_horiz)
+            # Edge condition for entering umbra: sat_vert - pen_vert = 0.
+            return sat_vert - umb_vert  # +ve to -ve direction means entering umbra.
```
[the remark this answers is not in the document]
Yash-10: What I was thinking is `sat_vert - umb_vert` is the difference in the vertical distances and the sign of that expression could tell whether a satellite is going in or out of a particular shadow. Since the `__call__` method looks for zero crossings, I had thought the return value of these two core functions does give information about the event.

The motivation for doing the above was this line from scipy's `solve_ivp` docs:

> Each event occurs at the zeros of a continuous function of time and state

and also the `LithobrakeEvent` which returns `H - R` inside the `__call__` method.

I was wondering, if we return either 0 or 1, would we lose the ability to tell whether the satellite is going in or out of shadow?

src/poliastro/twobody/events.py
Yash-10: Thanks, returning a boolean value does look better...I made those changes.

src/poliastro/core/perturbations.py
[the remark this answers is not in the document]
jorgepiloto: I see your point. In the [scipy solve_ivp docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html) it is said that a `direction` attribute can be used to determine the sense in which the zero is found.

However, I think this `direction` is used to determine the limiting cases: the point entering at eclipse and the point going outside of it. The `direction` attribute is in fact another trigger, similar to `terminal`, but focuses on limiting points.

This mean that you are right. By returning only a `0` or a `1` we are loosing the ability to indicate the current `direction` value... Let me think more about how could we face this problem, I will need some time to develop a solution 🤔

src/poliastro/core/perturbations.py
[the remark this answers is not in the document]
astrojuanlu: What about R_1 and R_2 ? :)

Yash-10:
A question was asked in SO by someone about handling boolean return value for events inside `solve_ivp` but no answer yet: https://stackoverflow.com/questions/62306789/how-to-use-solve-ivp-event-detector-with-boolean-event-function. 

Interestingly, it also deals with eclipse detection :)

Yash-10:
Some ideas that I have been thinking. Below is a rough analysis:

This draft approach tries to allow continuous return values rather than boolean values (this is important because there are approaches like the one mentioned in Vallado itself to compute entry/exit points using pure mathematical equations which we might not aim to do):

Screenshot from 2021-06-20 16-34-31

*(Fig 5-6 in Vallado)*

*Below is for umbra. Similarly can be done for penumbra*

What if we modify the algorithm from Vallado? I noticed that it computes all the angles under a certain condition (that is the dot product of `r` vectors is < 0). 

If we calculate the required angles under all conditions without any if conditions, we may have a way to return continuous values that change sign under certain conditions (i.e. entry or exit points of eclipse).

Screenshot from 2021-06-20 16-58-34


This would mean we might have **six regions** out of which shadow is only in region 3 and **four regions where d = 0**. Also, `sat_horiz < 0` on the right half of the attractor and vice-versa. We might set the return value such that it is of constant sign when moving from 4 to 5/5 to 6/6 to 1/1 to 2  and change sign only when going from 2 to 3 and 3 to 4:

```python
if sat_horiz > 0:
  return_value = (sat_vert - shadow_vert) * -1 * np.sign(sat_vert - shadow_vert)
else:
  return_value = sat_vert - shadow_vert
```
By doing this, if `sat_horiz > 0`, then `return_value` would always be, say, negative, but would change sign otherwise.

I strongly think a similar approach to this might just work. Any thoughts on this would really help.

Thanks!

jorgepiloto:
For the `__call__` method, we are forced to compute the position of the secondary body with respect to (w.r.t.) the primary one (i.e. the Sun as seen from Earth). Thus, we might apply the following logic, @Yash-10:

```python
    def __call__(self, t, u_, k):
        """The callable method for the eclipse detector."""
        self._last_t = t
        # Solve for primary and secondary bodies position w.r.t. solar system
        # barycenter at particular epoch
        (r_wrt_ssb_primary, _), (r_wrt_ssb_secondary) = [
            get_body_barycentric_posvel(body.name, self._epoch0 + t * u.s)
            for body in self._body_primary, self._body_secondary
        ]

        # Express vectors w.r.t. primary body
        r_sec = (r_wrt_ssb_primary - r_wrt_ssb_secondary).to(u.km).value
        r_sat = u_[:3].to(u.km).value

        in_umbra = in_umbral_shadow_fast(
            r_sat, r_sec, self._body_secondary.R, self._body_primary.R
        )

        return in_umbra
```

An the class is initialized as:

```python
    def __init__(self, ss_orbit, terminal=False, direction=1):
        """Initializes the umbra event detector instance."""
        super().__init__(terminal, direction)

        # Store data related to main attractor and the secondary body
        self._body_primary = ss_orbit.attractor
        self._body_secondary = self._body_primary.parent

        # Store also the initial epoch
        self._epoch0 = ss_orbit.epoch
```

Yash-10:
Thanks so much, @jorgepiloto for the suggestion! Since both events have exactly the same structure and code, could we create an `EclipseEvent` class and inherit penumbra and umbra from it?

jorgepiloto:
> Thanks so much, @jorgepiloto for the suggestion! Since both events have exactly the same structure and code, could we create an `EclipseEvent` class and inherit penumbra and umbra from it?

Sure, let us do that! 🚀

Yash-10:
In the implementation, I had to create a dummy orbit so that I could access the `pqw` method, which design-wise might still not be good. I am trying to find a way to prevent making an orbit. If we could get [Px, Py, Pz] and [Qx, Qy, Qz] without creating an orbit, the whole shadow function could potentially be `@jit`ted.

Yash-10:
I realized I had misunderstood the Px, Py, .. equations above. Now the shadow function could be computed in `core`. As a result, the resulting shadow function might be good performance-wise.

src/poliastro/core/events.py
```diff
@@ -0,0 +1,65 @@
+import numpy as np
+from numba import njit as jit
+from numpy.linalg import norm
+
+from poliastro.core.elements import rv2coe
+
+
+@jit
+def continuous_shadow_function_for_penumbra(k, u_, r_sec, R_sec, R_primary):
```
jorgepiloto, line 9: Let us name this `eclipse_shadow` so we do not confuse it with the current `shadow_function` from `perturbations.py`. Because the only difference between the `umbra` and `penumbra` logic is a sign, let us use an extra boolean parameter:

```python
def eclipse_function(k, u_, r_sec, R_sec, R_primary, umbra=True):
    # Plus or minus condition
    pm = 1 if umbra is True else -1
    # The eclipse shadow function becomes
    ... + pm * (2 * p * R_primary * cos_psi) ...
```

Therefore, we do not repeat ourselves by having two functions 👍🏽

src/poliastro/core/events.py
```diff
@@ -0,0 +1,65 @@
...
+    p, ecc, inc, raan, argp, nu = rv2coe(k, u_[:3], u_[3:])
+
+    Px = np.cos(raan) * np.cos(argp) - np.sin(raan) * np.sin(argp) * np.cos(inc)
```
jorgepiloto, line 12: This logic is already implemented in poliastro, see:

```python
def coe_rotation_matrix(inc, raan, argp):
    """Create a rotation matrix for coe transformation"""
    r = rotation_matrix(raan, 2)
    r = r @ rotation_matrix(inc, 0)
    r = r @ rotation_matrix(argp, 2)
    return r
```

where `P` and `Q` are the first and second column vectors of previous output matrix. Therefore, this logic could be simplified to:

```python
PQW = coe_rotation_matrix(inc, raan, argp)
P, Q = PQW[:,0], PQW[:,1]
```

jorgepiloto (changes requested):
This is almost ready to go @Yash-10. I just requested some changes so the code can be simplified and be more easy to maintain in the future 🚀

src/poliastro/core/events.py
Yash-10: Oh, I forgot to use that. Thanks so much for this!

src/poliastro/core/events.py
Yash-10: Thanks!

Yash-10:
Just wondering if it would be a good choice to move:

```py
k = self._primary_body.k.to_value(u.km ** 3 / u.s ** 2)
R_sec = self._secondary_body.R.to_value(u.km)
R_primary = self._primary_body.R.to_value(u.km)
```
inside the `__init__` method of `EclipseEvent`, since they don't change with time. Just that it would add `self.k`, `self.R_sec`, etc.. What do you think?

src/poliastro/core/events.py
```diff
@@ -0,0 +1,53 @@
...
+    p, ecc, inc, raan, argp, nu = rv2coe(k, u_[:3], u_[3:])
+
+    PQW = coe_rotation_matrix(inc, raan, argp)
+    P_, Q_ = PQW[:, 0], PQW[:, 1]
+
+    r_sec_norm = norm(r_sec)
+    beta = np.dot(P_, r_sec) / r_sec_norm
+    zeta = np.dot(Q_, r_sec) / r_sec_norm
+
+    delta_shadow = np.arcsin((R_sec - pm * R_primary) / r_sec_norm)
```
jorgepiloto, line 40: I see you are applying the `arcsin` function and a few lines later the `sin`. Let use the same idea as in `cos_psi` variable, therefore:

```python
sin_delta_shadow = np.sin((R_sec - pm * R_primary) / r_sec_norm)
```

so only one function is being called instead of making round trip conversion.

jorgepiloto (changes requested):
Just left a last changes request. Once solved this can be merged 👍🏽

src/poliastro/core/events.py
jorgepiloto: Yeah, it was a little bit hidden...

src/poliastro/core/events.py
Yash-10: Thanks!

Yash-10:
I did `np.ascontiguousarray(PQW[:, 0])` because otherwise `numba` gave warnings saying that since P_ and Q_ are non-contiguous (due to slicing), the `dot` product might be slower.

Yash-10:
Would you be in favour of moving:

```py
k = self._primary_body.k.to_value(u.km ** 3 / u.s ** 2)
R_sec = self._secondary_body.R.to_value(u.km)
R_primary = self._primary_body.R.to_value(u.km)
```
inside the `__init__` method of `EclipseEvent`, since they don't change with time. Just that it would add self.k, self.R_sec, etc.. What do you think?

astrojuanlu:
Wonderful job @Yash-10 , and super thorough reviews @jorgepiloto 👏🏽 

I resolved some outdated conversations, but there are some of them that I was unsure about.

> Would you be in favour of moving:

+1!

astrojuanlu (approved):
Looks good to me! Ready for re-review from @jorgepiloto

jorgepiloto (approved):
Finally, this is implemented! Congratulations @Yash-10 🚀

Yash-10:
Thanks for your help! @jorgepiloto @astrojuanlu

Issue #1240, opened 2021-06-01
Smarter initial estimates of eccentric and hyperbolic anomaly?


Yash-10:
💡 **Feature request**
It deals with the calculation of eccentric and hyperbolic anomaly from the mean anomaly using the newton-raphson iteration method. Specifically, the [`M_to_E`](https://github.com/poliastro/poliastro/blob/main/src/poliastro/core/angles.py#L262) and [`M_to_F`](https://github.com/poliastro/poliastro/blob/main/src/poliastro/core/angles.py#L294) in `core/angles.py`.

The "feature request" comes from a section in "Fundamentals of Astrodynamics and Applications" by Vallado under the section "Hybrid solutions" in chapter 2.

![hybrid](https://user-images.githubusercontent.com/68844397/120311911-b38bed00-c2f5-11eb-984f-647f68706236.png)

Barring the case of parabolic motion (since the values can be determined deterministically using the cubic equation already), the section states that by using a slightly smarter initial guess for the eccentric and hyperbolic anomaly for elliptic and hyperbolic motion, the computation time can be decreased by ~15%. Algorithm 2 and 4 in the book provides the algorithm for it.

**Further points**
- The functions are jitted and using the `timeit` module gives the following rough analysis:

Note: the below "no. of iterations" are the no. of times both the implementations are run using `timeit` and not the no. of iterations for the newton-raphson method.

    - For 1000 iterations, there is ~0.26 s time improvement.
    - For 10000 iterations, there is ~2.62 s time improvement.
    - For 100000 iterations, there is ~26.08 s improvement.

Summarizing it in a graph:

![image](https://user-images.githubusercontent.com/68844397/120313601-d9b28c80-c2f7-11eb-958a-23918ccbd4b5.png)

A starter implementation can be seen [here](https://github.com/Yash-10/poliastro/commit/7ca4d7e298cdf59e6e07c094b5d7ab2e7c6d7001). Important to note, the linked implementation yields a slight mismatch for the hyperbolic case but works perfectly fine for the elliptical case. Maybe it has to do with the no. of iterations in the newton method.

**Benefits**
This results in a micro-improvement (even if the functions were already jitted) in terms of execution time.

I would love to get any thoughts on this.

Thanks!

jorgepiloto:
Hi @Yash-10. I reviewed your code, and it matches Vallado's one. Could you please develop more about those differences within the Hyperbolic case? I checked that your implementation passes all current tests related with angle computations.

[...]

PR #1247, opened 2021-06-11
Slightly faster method for newton iteration in `E` and `H` calculation

Yash-10:
Hello, as per #1240, this is an attempt to slightly make the computation faster (even thought both functions were already jitted).

Thanks!

### DIFF
--- a/src/poliastro/core/angles.py
+++ b/src/poliastro/core/angles.py
@@ -310,7 +310,7 @@ def M_to_F(M, ecc):
     This uses a Newton iteration on the hyperbolic Kepler equation.
 
     """
-    if ecc < 1.25:
+    if ecc < 1.6:
         if -np.pi < M < 0 or np.pi < M:
             F0 = M - ecc
         else:
--- a/src/poliastro/core/angles.py
+++ b/src/poliastro/core/angles.py
@@ -281,11 +281,10 @@ def M_to_E(M, ecc):
     This uses a Newton iteration on the Kepler equation.
 
     """
-    assert -np.pi <= M <= np.pi
-    if ecc < 0.8:
-        E0 = M
+    if -np.pi < M < 0 or M > np.pi:
+        E0 = M - ecc
     else:
-        E0 = np.pi * np.sign(M)
+        E0 = M + ecc
     E = newton("elliptic", E0, args=(M, ecc))
     return E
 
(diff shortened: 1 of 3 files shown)


Yash-10:
As was seen locally, there are slight mismatches. The eccentricity conditions in the hyperbolic case might need to be investigated...

src/poliastro/core/angles.py
```diff
@@ -311,7 +310,16 @@ def M_to_F(M, ecc):
     This uses a Newton iteration on the hyperbolic Kepler equation.
 
     """
-    F0 = np.arcsinh(M / ecc)
+    if ecc < 1.25:
+        if -np.pi < M < 0 or M > np.pi:
```
astrojuanlu, line 314: My recommendation: "Always use less-than" https://thehardcorecoder.com/2014/03/05/always-use-less-than/ :)

astrojuanlu:
> although of magntiude ~1e-20 or lower

Where exactly was this error appearing? I think it should be acceptable to adapt the test accordingly

Also, could you show a comparison of speed or number of iterations before and after?

src/poliastro/core/angles.py
Yash-10: Thanks! That's interesting to know :)

Yash-10:
> Also, could you show a comparison of speed or number of iterations before and after?

I haven't changed the number of iterations for the newton iteration. For speed improvements, there's a slight improvement. I had done an analysis in #1240 for discussion. Maybe that would give an idea?


> The functions are jitted and using the timeit module gives the following rough analysis:
Note: the below "no. of iterations" are the no. of times both the implementations are run using timeit and not the no. of iterations for the newton-raphson method.

> - For 1000 iterations, there is ~0.26 s time improvement.
> - For 10000 iterations, there is ~2.62 s time improvement.
> - For 100000 iterations, there is ~26.08 s improvement.
> Summarizing it in a graph:

> image

Yash-10:
As you say, we could make the implementation look exactly like Vallado's (i.e. change the eccentricty condition to < 1.6 in the hyperbolic case) and modify the tests to prevent any minor errors.

astrojuanlu:
Yes, let's do it!

astrojuanlu (approved):
Thanks @Yash-10 !

PR #1249, opened 2021-06-12
Alinspace to handle angle differences larger than 2*pi, fix xfailed test

jtegedor:
This is a small change to fix `xfailed` test in `test_core_util.py` due to `alinspace` function not handling angle differences larger than 2*pi.


astrojuanlu (approved):
Looks perfect, thanks @jtegedor!

PR #1252, opened 2021-06-13
`czml` computation to `core`

Yash-10:
This pull request tries to move computations of `czml/utils.py` to `core` since it uses numpy and has some heavy calculations.

Thanks!


astrojuanlu (approved):
Thanks @Yash-10 !

PR #1253, opened 2021-06-16
Some test additions

Yash-10:
This pull request tries to add some tests for increasing coverage. This could be a start for solving #1236 .

Thanks!

### DIFF
--- a/tests/test_bodies.py
+++ b/tests/test_bodies.py
@@ -2,12 +2,12 @@ import pytest
 from astropy import units as u
 from astropy.tests.helper import assert_quantity_allclose
 
-from poliastro import bodies
+from poliastro.bodies import Body, Earth, Sun, Jupiter
 
 
 def test_body_has_k_given_in_constructor():
     k = 3.98e5 * u.km ** 3 / u.s ** 2
-    earth = bodies.Body(None, k, "")
+    earth = Body(None, k, "")
     assert earth.k == k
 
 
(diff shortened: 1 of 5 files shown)


tests/test_bodies.py
```diff
@@ -16,37 +16,45 @@ def test_body_from_parameters_raises_valueerror_if_k_units_not_correct():
...
     assert (
         "UnitsError: Argument 'k' to function 'from_parameters' must be in units convertible to 'km3 / s2'."
         in excinfo.exconly()
     )
 
 
+def test_body_from_parameters_returns_body_object():
+    k = 3.98e5 * u.km ** 3 / u.s ** 2
+    _name = _symbol = "jupiter"
+    jupiter = Body.from_parameters(Sun, k, _name, _symbol, Jupiter.R)
```
astrojuanlu, line 29: I find this test a bit weird. What about checking that `k` has the given value, for example?

tests/test_czml.py
```diff
@@ -442,6 +442,19 @@ def test_czml_add_trajectory():
 
     assert repr(extractor.packets) == expected_doc
 
+    x = u.Quantity([1.0, 2.0, 3.0], u.m)
```
astrojuanlu, line 445: Can we turn `expected_doc` to a `pytest.fixture`, and have these new lines as a new test?

tests/tests_plotting/test_misc.py
```diff
@@ -16,6 +16,11 @@ def test_plot_solar_system_has_expected_number_of_orbits(outer, expected):
 def test_plot_solar_system_uses_expected_orbitplotter(use_3d, plotter_class):
     assert isinstance(plot_solar_system(use_3d=use_3d, interactive=True), plotter_class)
 
+    if use_3d:
```
astrojuanlu, line 19: Can this be a new test?

tests/tests_twobody/test_propagation.py
```diff
@@ -448,3 +448,17 @@ def test_propagate_with_coe(propagator_coe):
     nu_final = propagator_coe(k, p, ecc, inc, raan, argp, nu, period)
 
     assert_quantity_allclose(nu_final, nu)
+
+
+def test_danby_with_zero_eccentricity():
```
astrojuanlu, line 453: Why only `danby` here? Is it because we're testing an `if` branch? If so, I think we could test all propagators with zero eccentricity

astrojuanlu (changes requested):
Thanks @Yash-10 ! Left a few comments

astrojuanlu (approved):
This looks very good! Merging when the tests pass (and also the coverage check!)

tests/test_czml.py
Yash-10: Since we are on it, I have a question:
- The marker `@pytest.mark.skipif("czml3" not in sys.modules, reason="requires czml3")` has been repeated on all the tests. I wonder if there is any way to skip the whole test module if `czml3` is not installed?
- Also, there are a couple more `expected_doc` in the test file. Is there a need to create a fixture for each?

tests/tests_plotting/test_misc.py
Yash-10: Thanks!

tests/tests_twobody/test_propagation.py
Yash-10: Yes, `danby`'s core method had an `if ecc == 1.0` condition which wasn't checked (line 840 [here](https://codecov.io/gh/poliastro/poliastro/src/main/src/poliastro/core/propagation/__init__.py)). But as you say, it would be better to check all propagators.

tests/test_bodies.py
Yash-10: Thanks for this suggestion! The motivation for this test was from here: https://codecov.io/gh/poliastro/poliastro/src/main/src/poliastro/bodies.py at line 105, i.e. to check whether `from_parameters` works properly. I checked for both `k` and `R` here.

Do you think we should use this test for all bodies? I thought checking for any one body would suffice for the purposes of this test.

tests/test_czml.py
Yash-10: @astrojuanlu Could you let me know what do you think about the above questions? If needed, these can be done here itself?

PR #1254, opened 2021-06-17
Altitude crossing

Yash-10:
In this pull request, an attempt has been made to include altitude crossing detection. The structure seems to be similar to `LithobrakeEvent`.

Orekit's implementation stops propagation if the altitude starts to get lower than the threshold but continues propagation while ascending. Here, even though there are `terminal` and `direction` attributes, I immediately couldn't find a way to handle both at the same time, i.e. if `direction` is from up to down, `terminal=False` and if `direction` is down to up, `terminal=True`. But this might be done in some way...

- [ ] Based on the above, handle the `terminal` attribute? 

**Question**

- Is it a good choice to have a default threshold altitude or should the user compulsorily pass it?
- It seems there is currently no way to get an orbit's altitude (eg: `orbit.alt`)? If it looks like a plausible addition, `H - self._R` can be changed to `orbit.alt`.

Thanks!

### DIFF
--- a/src/poliastro/twobody/events.py
+++ b/src/poliastro/twobody/events.py
@@ -2,67 +2,81 @@ from astropy import units as u
 from numpy.linalg import norm
 
 
-class LithobrakeEvent:
+class Event:
+    """Base class for event functionalities.
+
+    Parameters
+    ----------
+    terminal: bool
+        Whether to terminate integration if this event occurs.
+    direction: float
+        Handle triggering of event.
+
+    """
+
+    def __init__(self, terminal, direction):
+        self._terminal, self._direction = terminal, direction
+        self._last_t = None
+
+    @property
+    def terminal(self):
+        return self._terminal
+
+    @property
+    def direction(self):
+        return self._direction
+
+    @property
+    def last_t(self):
+        return self._last_t * u.s
+
+    def __call__(self, t, u, k):
+        raise NotImplementedError
+
+
+class AltitudeCrossEvent(Event):
+    """Detect if a satellite crosses a specific threshold altitude.
+
+    Parameters
+    ----------
+    alt: float
+        Threshold altitude (km).
+    R: float
+        Radius of the attractor (km).
+    terminal: bool
+        Whether to terminate integration if this event occurs.
+    direction: float
+        Handle triggering of event based on whether altitude is crossed from above
+        or below, defaults to -1, i.e., event is triggered only if altitude is
+        crossed from above (decreasing altitude).
+
+    """
+
+    def __init__(self, alt, R, terminal=True, direction=-1):
+        super().__init__(terminal, direction)
+        self._R = R
+        self._alt = alt  # Threshold altitude from the ground.
+
+    def __call__(self, t, u, k):
+        self._last_t = t
+        r_norm = norm(u[:3])
+
+        return (
+            r_norm - self._R - self._alt
+        )  # If this goes from +ve to -ve, altitude is decreasing.
+
+
+class LithobrakeEvent(AltitudeCrossEvent):
     """Terminal event that detects impact with the attractor surface.
 
     Parameters
     ----------
     R : float
-        Radius of the attractor.
-
-    """
-
-    def __init__(self, R):
-        self._R = R
-        self._last_t = None
-
-    @property
-    def terminal(self):
-        # Tell SciPy to stop the integration at H = R (impact)
-        return True
-
-    @property
-    def last_t(self):
-        return self._last_t * u.s
-
-    def __call__(self, t, u, k):
-        self._last_t = t
-        H = norm(u[:3])
-        # SciPy will search for H - R = 0
-        print(H - self._R)
-        return H - self._R
-
-
-class AltitudeCrossEvent:
-    """Detect if a satellite crosses a specific threshold altitude.
-
-    Parameters
-    ----------
-    R: ~astropy.units.Quantity
         Radius of the attractor (km).
-    thresh_H: ~astropy.units.Quantity
-        Threshold altitude (in km), defaults to 100 km.
     terminal: bool
-        Whether to terminate integration if this event occurs, defaults to True.
+        Whether to terminate integration if this event occurs.
 
     """
-    def __init__(self, R, thresh_H=100*u.km, terminal=True):
-        self._R = R.to(u.km).value
-        self._thresh_H = thresh_H.to(u.km).value  # Threshold height from the ground.
-        self._terminal = terminal
-        self._last_t = None
 
-    @property
-    def terminal(self):
-        # Orekit's API stops propagation when descending, but not when ascending.
-        return self._terminal
-
-    @property
-    def last_t(self):
-        return self._last_t * u.s
-
-    def __call__(self, t, u, k):
-        self._last_t = t
-        H = norm(u[:3])
-        # H is from the center of the attractor.
-        return H - self._R - self._thresh_H  # If this goes from +ve to -ve, altitude is decreasing.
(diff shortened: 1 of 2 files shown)


Yash-10:
It looks like `readthedocs` build shows still in progress but the build seems to be successfully completed?

jorgepiloto:
Just to summarize what we discussed today about building an `Event` mother class. I think we could build it like this:

```python
class Event:
    def __init__(self, _terminal, _direction):
        self._terminal, self._direction = terminal, direction

    @property
    def terminal(self):
        return self._terminal
       
    @property
    def direction(self):
        return self._direction


    def  __call__(self):
        return NotImplementedError
```

which will save us a few lines of code per implemented event subclass.

jorgepiloto:
Also, I though it is feasible to have something like:

```python
class AltitudeCrossingEvent(Event):
    ...

class LithobrakeEvent(AltitudeCrossingEvent):
    ...
```

src/poliastro/twobody/events.py
```diff
@@ -2,31 +2,78 @@
 from numpy.linalg import norm
 
 
-class LithobrakeEvent:
-    """Terminal event that detects impact with the attractor surface.
+class Event:
```
jorgepiloto, line 5: Looks good! 🎉

src/poliastro/twobody/events.py
```diff
@@ -2,31 +2,78 @@
...
+    ----------
+    R: ~astropy.units.Quantity
+        Radius of the attractor.
+    thresh_alt: ~astropy.units.Quantity
+        Threshold altitude (in km), defaults to 100 km.
+
+    """
+
+    # Orekit's API stops propagation when descending, but not when ascending.
+    def __init__(self, R, terminal=True, direction=0, thresh_alt=100 * u.km):
```
jorgepiloto, line 50: Because this is named `AltitudeCrossEvent`, let us make the threshold altitude to be the first parameter and impose it not to be optional. Also, I think we should change its name just to `alt`, as it is shorter.

```python
def __init__(self, alt, R, terminal=True, direction=0):
    ...
```

src/poliastro/twobody/events.py
```diff
@@ -2,31 +2,78 @@
...
+        super().__init__(terminal, direction)
+        self._R = R.to(u.km).value
+        self._thresh_alt = thresh_alt.to(
+            u.km
+        ).value  # Threshold height from the ground.
+
     def __call__(self, t, u, k):
         self._last_t = t
         H = norm(u[:3])
```
jorgepiloto, line 59: Let us rename variable `H` to `r_norm` in order no to confuse it with height.

src/poliastro/twobody/events.py
```diff
@@ -2,31 +2,78 @@
...
+    Parameters
+    ----------
+    R : ~astropy.units.Quantity
+        Radius of the attractor.
+    thresh_alt: ~astropy.units.Quantity
+        Hardcoded as 0.
+
+    """
+
+    def __init__(self, R, terminal=True, direction=0, thresh_alt=0 * u.km):
```
jorgepiloto, line 78: There is no need to have an altitude argument in this case. Hence:

```python
def __init__(self, R, terminal=True, direction=0)
    ...
```

jorgepiloto (changes requested):
Hi @Yash-10, this is almost ready to be merged. I just left a couple of comments.

Yash-10:
Would it be a good choice think to terminate integration if altitude decreases but continue it when altitude increases? I think this is the logic with Orekit's AltitudeDetector: https://www.orekit.org/static/apidocs/org/orekit/propagation/events/AltitudeDetector.html.

src/poliastro/twobody/events.py
Yash-10: Thanks!

src/poliastro/twobody/events.py
Yash-10: Thanks for these suggestions!

src/poliastro/twobody/events.py
Yash-10: Thanks!

jorgepiloto:
> Would it be a good choice think to terminate integration if altitude decreases but continue it when altitude increases? I think this is the logic with Orekit's AltitudeDetector: https://www.orekit.org/static/apidocs/org/orekit/propagation/events/AltitudeDetector.html.

Yeah, let us apply a `direction = -1` for the `AltitudeCrossingEvent`. In addition, this behavior must be documented in the class' docstring 👍🏽

docs/source/examples/Natural and artificial perturbations.mystnb
```diff
@@ -44,7 +44,7 @@ pio.renderers.default = "notebook_connected"
 The poliastro package now has several commonly used natural perturbations. One of them is atmospheric drag! See how one can monitor decay of the near-Earth orbit over time using our new module `poliastro.twobody.perturbations`!
 
 ```{code-cell} ipython3
-R = Earth.R.to(u.km).value
+R = Earth.R.to(u.km)
```
jorgepiloto, line 47: Were the changes in this notebook intended? They do not introduced any kind of feature nor increase the performance.

Yash-10: The doctest builds in this PR yielded an error since I had changed the signature of `LithobrakeEvent` by allowing an astropy `Quantity` object instead of a float. So a corresponding change needed to be made in the doc examples.

Yash-10:
Since we are on this, would it be a good choice to also set direction to -1 for Lithobrake? Since a satellite would not crash coming from beneath the surface, isn't it? :)

jorgepiloto:
> Since we are on this, would it be a good choice to also set direction to -1 for Lithobrake? Since a satellite would not crash coming from beneath the surface, isn't it? :)

Sure, let us also use that same value for the `Lithobrake` direction attribute.

docs/source/examples/Natural and artificial perturbations.mystnb
jorgepiloto: Alright, so let us make the `Event` class and the members which inherit from it to work without units. This was the original behavior, so let us keep it this way.

jorgepiloto: Only files modified in this PR should be the `events.py` and associated tests.

jorgepiloto (commented):
Just left some more comments, this is almost finished 👍🏽

Yash-10:
Thanks again! I didn't put the direction attribute in the constructor for Lithobrake since we only want to track going from +ve to -ve and allowing users to trigger event from -ve to +ve wouldn't be good. Could you let me know if this looks ok? Or we should still pass it in the constructor?

astrojuanlu:
Rebuilding Read the Docs, in any case it looked green 🟢

astrojuanlu (approved):
I haven't been following this discussion in depth but this looks ready to me! Deferring to @jorgepiloto in case he has more comments.

jorgepiloto:
> Thanks again! I didn't put the direction attribute in the constructor for Lithobrake since we only want to track going from +ve to -ve and allowing users to trigger event from -ve to +ve wouldn't be good. Could you let me know if this looks ok? Or we should still pass it in the constructor?

Sure, I agree with that @Yash-10, thank you for noticing this.

jorgepiloto:
Everything look by my side @Yash-10. Please, squash the commits and this will be ready to be merged 👍🏽

jorgepiloto (approved):

jorgepiloto:
Merging this. Thank you @Yash-10 👍🏽

PR #1255, opened 2021-06-18
Add some more tests and checks

Yash-10:
This pull request is a continuation of addition of some tests and checks.

- The altitude checks were added to prevent creation of an orbit with negative altitude. This might yield unexpected results at some places, for example, in the lithobrake event.

**Questions**

- It looks like `@pytest.mark.skipif("czml3" not in sys.modules, reason="requires czml3")` gets repeated for each test in `test_czml.py`. Could we instead have a global `pytestmark` to skip the whole module if `czml3` is not installed?

```python
pytestmark = pytest.mark.skipif("czml3" not in sys.modules)
```

- `ALL_PROPAGATORS` in `twobody/propagation.py`  seem to not contain `markley` and `danby`. This seems to be contrary to the name of the variable. Is this expected? If not, should we include all the propagators in `ALL_PROPAGATORS`?

- In #1253, you had advised to make `expected_doc` as a fixture. It seems that there are more `expected_doc`'s inside the tests. Is it necessary to have a fixture for each of them?

@astrojuanlu Could you let me know what do you think about the above questions?

Thanks!

### DIFF
--- a/src/poliastro/twobody/propagation.py
+++ b/src/poliastro/twobody/propagation.py
@@ -74,9 +74,9 @@ def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, f=func_twobody):
 
     Note
     -----
-    This method uses a Dormand & Prince method of order 8(5,3) available
-    in the :py:class:`poliastro.integrators` module. If multiple tofs
-    are provided, the method propagates to the maximum value and
+    This method uses the `solve_ivp` method from `scipy.integrate` using the
+    Dormand & Prince integration method of order 8(5,3) (DOP853).
+    If multiple tofs are provided, the method propagates to the maximum value and
     calculates the other values via dense output
 
     """
@@ -501,5 +501,5 @@ HYPERBOLIC_PROPAGATORS = [
     cowell,
 ]
 ALL_PROPAGATORS = list(
-    set(ELLIPTIC_PROPAGATORS) & set(PARABOLIC_PROPAGATORS) & set(HYPERBOLIC_PROPAGATORS)
+    set(ELLIPTIC_PROPAGATORS + PARABOLIC_PROPAGATORS + HYPERBOLIC_PROPAGATORS)
 )
--- a/src/poliastro/twobody/propagation.py
+++ b/src/poliastro/twobody/propagation.py
@@ -451,7 +451,7 @@ def propagate(orbit, time_of_flight, *, method=farnocchia, rtol=1e-10, **kwargs)
     # Check if propagator fulfills orbit requirements
     if orbit.ecc < 1.0 and method not in ELLIPTIC_PROPAGATORS:
         raise ValueError(
-            "Can not use an parabolic/hyperbolic propagator for elliptical orbits."
+            "Can not use an parabolic/hyperbolic propagator for elliptical/circular orbits."
         )
     elif orbit.ecc == 1.0 and method not in PARABOLIC_PROPAGATORS:
         raise ValueError(
@@ -472,10 +472,6 @@ def propagate(orbit, time_of_flight, *, method=farnocchia, rtol=1e-10, **kwargs)
         rtol=rtol,
         **kwargs
     )
-
-    # TODO: Turn these into unit tests
-    assert rr.ndim == 2
-    assert vv.ndim == 2
 
     cartesian = CartesianRepresentation(
         rr, differentials=CartesianDifferential(vv, xyz_axis=1), xyz_axis=1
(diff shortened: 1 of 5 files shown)


jorgepiloto:
The logic when building the `ALL_PROPAGATORS` macros is wrong! 😱

We are only including the propagators which are shared among the elliptic, parabolic and hyperbolic ones. Thank you for noticing this @Yash-10! 🚀

jorgepiloto:
This could be fixed by doing:

```python
ALL_PROPAGATORS = set(ELLIPTIC_PROPAGATORS + PARABOLIC_PROPAGATORS + HYPERBOLIC_PROPAGATORS)
```

Yash-10:
Thanks @jorgepiloto for your answer! Also, while we were on that module, I noticed that the "Note" inside `cowell`'s docstring mentioned `poliastro.integrators`. I thought it was removed some time back and I hence changed the docstring to mention scipy's `solve_ivp`.

src/poliastro/twobody/propagation.py
```diff
@@ -74,9 +74,9 @@ def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, f=func_twobody):
 
     Note
     -----
-    This method uses a Dormand & Prince method of order 8(5,3) available
-    in the :py:class:`poliastro.integrators` module. If multiple tofs
-    are provided, the method propagates to the maximum value and
+    This method uses the `solve_ivp` method from `scipy.integrate` using the
+    Dormand & Prince integration method of order 8(5,3) (DOP853).
+    If multiple tofs are provided, the method propagates to the maximum value and
```
jorgepiloto, line 79: I would also say in here the following:

```
... propagates to the maximum value (unless a terminal event is defined) and calculates ...
```

tests/tests_twobody/test_propagation.py
```diff
@@ -463,3 +463,22 @@ def test_propagator_with_zero_eccentricity(propagator):
     assert_quantity_allclose(orbit.inc, res.inc)
     assert_quantity_allclose(orbit.raan, res.raan)
     assert_quantity_allclose(orbit.argp, res.argp)
+
+
+@pytest.mark.parametrize("propagator", ALL_PROPAGATORS)
+def test_after_propagation_r_and_v_dimensions(propagator):
```
jorgepiloto, line 469: This is not required, there is a test already checking for `ALL_PROPAGATORS`, see:

https://github.com/poliastro/poliastro/blob/0a0c00559a36ce9779f824738773bc7e98a6697c/tests/tests_twobody/test_propagation.py#L105-L106

jorgepiloto (changes requested):
Just some suggestions. After those, this is ready to be merged by my side 👍🏽

tests/tests_twobody/test_propagation.py
Yash-10: The motivation for adding this test was this line: https://github.com/poliastro/poliastro/blob/main/src/poliastro/twobody/propagation.py#L476. I noticed that below the "TODO"'s, there was an assert condition inside the source code. I though it would be good to turn it into its own test.  Also, this method uses the propagator method directly and not the propagation method of an orbit. Hence, after adding this test, the assert condition in `src` was removed.

What do you think about adding this test?

src/poliastro/twobody/propagation.py
Yash-10: Thanks for this!

astrojuanlu:
About having a global marker: yes, absolutely! I didn't know having a `pytestmark` would "apply marks at the module level" https://docs.pytest.org/en/6.2.x/example/markers.html#marking-whole-classes-or-modules

In #1253 I suggested to make `expected_doc` a fixture in case we wanted to reuse it for more than 1 test. If not, every test can have their own `expected_doc` and that's fine.

tests/tests_twobody/test_propagation.py
astrojuanlu: I'm +1 on adding this test and removing the `assert`!

astrojuanlu (approved):
This looks good to me! Ready to merge, unless @Yash-10 wants to make some extra additions (like the global pytest marker)

astrojuanlu:
(And pending on the comment @jorgepiloto left)

Yash-10:
Thanks @astrojuanlu I added the global pytest marker.

tests/tests_twobody/test_propagation.py
jorgepiloto: Oh, I see. Thanks for the clarification @Yash-10!

jorgepiloto (approved):
All modifications looking good, approving this 👍🏽

PR #1258, opened 2021-06-21
General `LOS` for `LOSEvent`

Yash-10:
The current `shadow_function` used trigonometric manipulation (from Howard Curtis). In this pull request:

- An attempt has been made to add a general line of sight functionality that can be used to detect LOS between any given two position vectors (eg: between two satellites, or between a satellite and the Sun).
- `shadow_function` has been modified to use this general LOS implementation.

Currently, the LOS function allows both ellipsoidal and circular shape of the attractor, but I think bodies are taken to be a general case of ellipsoid. So the existence of the `ellipsoid` parameter in the functions might be in doubt.

---
This was an older comment :)

Thanks!

### DIFF
--- a/src/poliastro/core/events.py
+++ b/src/poliastro/core/events.py
@@ -77,11 +77,6 @@ def line_of_sight(r1, r2, R):
 
     """
     r1_norm = np.linalg.norm(r1)
-    if r1_norm <= R:
-        raise ValueError(
-            "The norm of the position vector r_1 is less than the radius of the attractor."
-        )
-
     r2_norm = np.linalg.norm(r2)
 
     theta = np.arccos(np.dot(r1, r2) / r1_norm / r2_norm)
--- a/src/poliastro/core/events.py
+++ b/src/poliastro/core/events.py
@@ -54,3 +54,15 @@ def eclipse_function(k, u_, r_sec, R_sec, R_primary, umbra=True):
     )
 
     return shadow_function
+
+
+@jit
+def line_of_sight(r1, r2, R):
+    r1_norm = np.linalg.norm(r1)
+    r2_norm = np.linalg.norm(r2)
+
+    theta = np.arccos(np.dot(r1, r2) / r1_norm / r2_norm)
+    theta_1 = np.arccos(R / r1_norm)
+    theta_2 = np.arccos(R / r2_norm)
+
+    return theta <= theta_1 + theta_2
--- a/src/poliastro/core/events.py
+++ b/src/poliastro/core/events.py
@@ -58,11 +58,34 @@ def eclipse_function(k, u_, r_sec, R_sec, R_primary, umbra=True):
 
 @jit
 def line_of_sight(r1, r2, R):
+    """Calculates the line of sight condition between two position vectors, r1 and r2.
+
+    Parameters
+    ----------
+    r1: ~np.array
+        The position vector of the first object with respect to a central attractor.
+    r2: ~np.array
+        The position vector of the second object with respect to a central attractor.
+    R: float
+        The radius of the central attractor.
+
+    Returns
+    -------
+    delta_theta: float
+        Greater than or equal to zero, if there exists a LOS between two objects
+        located by r1 and r2, else negative.
+
+    """
     r1_norm = np.linalg.norm(r1)
+    if r1_norm <= R:
+        raise ValueError(
+            "The norm of the position vector r_1 is less than the radius of the attractor."
+        )
+
     r2_norm = np.linalg.norm(r2)
 
     theta = np.arccos(np.dot(r1, r2) / r1_norm / r2_norm)
     theta_1 = np.arccos(R / r1_norm)
     theta_2 = np.arccos(R / r2_norm)
 
-    return theta <= theta_1 + theta_2
+    return (theta_1 + theta_2) - theta
--- a/src/poliastro/core/perturbations.py
+++ b/src/poliastro/core/perturbations.py
@@ -3,6 +3,8 @@ from astropy import units as u
 from numba import njit as jit
 from numpy.linalg import norm
 
+from poliastro.core.events import line_of_sight as line_of_sight_fast
+
 
 @jit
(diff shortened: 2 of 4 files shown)


Yash-10:
I was thinking of the following:

Remove the `ellipsoidal` argument in `line_of_sight` and forcefully use only the ellipsoidal model of attractor. This would remove the option of a circular body as was assumed earlier in `shadow_function`. If this is done, `solar_radiation_pressure` would also need to be probably modified since it worked with circular body assumptions.

Would it be a good change to do the above? This question arose because I wasn't aware if there would be any reason to sometimes allow someone to use a circular body instead of the actual ellipsoidal one.

@astrojuanlu @jorgepiloto What do you think about this?

Thanks!

Yash-10:
The implementation here was changed for #1284

astrojuanlu (commented):
This looks like a good change so far! We discussed adding a docstring to `line_of_sight` and also not propagate the secondary body as discussed in #1289

Yash-10:
Currently, the position coordinates are converted to a `list` for popping elements for each call to `__call__`. This was done since it would be in place and slightly better in terms of performance. It requires one or two extra conversions from and to `numpy` arrays.

I learned that `numpy` array slicing returns a view instead of a copy, so using:

```py
pos_coord = self._pos_coords[0] if self._pos_coords.size != 0 else self._last_coord
self._pos_coords = self._pos_coords[1:]
```
might still be efficient. Would there be any suggestions on whether to stick with lists or use numpy arrays?

Yash-10:
We had discussed that the length of position coordinates that need to be checked must match the length of `tofs`. However, I realized that the user input `pos_coord` is checked at each call of `__call`, and ensuring that user input position vectors' length and the length of `tofs` is the same, might not work. This is because `__call__` is called `n` number of times which depends on what time intervals does scipy wants to call the `__call__` method.

Yash-10:
As suggested by @jorgepiloto , we should use the position coordinates instead of `Ephem` objects for this LOS event. The only question here is: What restrictions do we keep on the number of the position coordinates passed by the user?

The concern is that `__call__` is called some arbitrary number of times (decided by `scipy` internally) throughout the propagation period. Hence, there doesn't seem to be a direct way to restrict the number of position coordinates passed by the user?

Currently, in this pull request, if the list `pos_coords` gets exhausted, then the last position coordinate gets repeated until the end. Does this seem to be a good choice?

Yash-10:
The errors seem to occur because there are solutions from `solve_ivp` that are probably not physically acceptable:

The norm of the position vectors sometimes go less than the radius of the attractor. Specifically, inside the `cowell` method:


```py
y = result.sol(t)
```
gives `y` that has norm less than R. Due to this,


```py
theta_1 = np.arccos(R / r1_norm)
```
returns `NaN`.

Hence, it looks like sometimes `cowell` returns solutions that are not physically acceptable, like the above? We might another issue for this, or am I missing something?

src/poliastro/core/events.py
```diff
@@ -54,3 +54,33 @@ def eclipse_function(k, u_, r_sec, R_sec, R_primary, umbra=True):
...
+    delta_theta: float
+        Greater than or equal to zero, if there exists a LOS between two objects
+        located by r1 and r2, else negative.
+
+    """
+    r1_norm = np.linalg.norm(r1)
+    r2_norm = np.linalg.norm(r2)
+
+    theta = np.arccos(np.dot(r1, r2) / r1_norm / r2_norm)
+    theta_1 = np.arccos(R / r1_norm)
```
jorgepiloto, line 83: As @Yash-10 pointed out during weekly meeting, if `r1_norm` is lower than `R`, then the argument of the cosine function has a value `>1` and `nan` values are returned.

src/poliastro/core/events.py
```diff
@@ -54,3 +54,33 @@ def eclipse_function(k, u_, r_sec, R_sec, R_primary, umbra=True):
...
+        The radius of the central attractor.
+
+    Returns
+    -------
+    delta_theta: float
+        Greater than or equal to zero, if there exists a LOS between two objects
+        located by r1 and r2, else negative.
+
+    """
+    r1_norm = np.linalg.norm(r1)
```
jorgepiloto, line 79: The condition might be checked here 👍🏽

Yash-10:
The edge case that we discussed when both lithobrake and los event are passed and when the satellite crosses the attractor:

The added test `test_LOS_event_with_lithobrake_event_when_satellite_cuts_attractor` seems to yield errors because somehow, the error condition of LOS event is called before lithobrake should have ended the propagation.

Although, checking the last_t values (with the `tofs` value as `tofs = [0.003, 0.004, 0.01, 0.02, 0.03, 0.04, 0.07, 0.1, 0.2, 0.3, 0.4, 1, 3] << u.s
`):

```
lithobrake_event.last_t: 0.35507097s
los_event: 1.0030613s
```

So, lithobrake does get triggered before, as expected.


This would need more investigation. Apart from this, the individual behaviour of LOS event seems to be fine.

src/poliastro/core/events.py
Yash-10: I noticed a slight problem in the edge case we had discussed in #1307. To resolve the problems with any edge cases, the best solution I could think till now is to raise a warning stating "Orbit is going beneath the surface of attractor, LOS event may not work properly".

But we decided a warning was not a good option, right? It might atleast solve the problem.

astrojuanlu: Discussed at lenght in other comments here and in #1307, we decided for the current solution of a warning to minimize complexity

src/poliastro/core/events.py
astrojuanlu: Noted, and handling it in the event function

astrojuanlu (approved):
I agree, this is good to go. The warning approach is probably the "lesser evil", let's merge this pull request and if it proves to be bad in the real world, we can reconsider! Thanks a lot @Yash-10 for the patience and the persistence 🙏🏽

PR #1073, opened 2021-01-12
Changed Docs to Markdown Format from reStructuredText Format

DhruvSondhi:
Hello, 
This PR aims to propose the idea of migration of the documentation from `.rst` format to `.md` format for easier editing & maintenance :smile: 
Implements #977

PR #978, opened 2020-06-17
Switch to Jupytext

jorgepiloto:
Solves for #727 by switching `*.ipynb` files to `*.md` ones with the help of Jupytext.

PR #1260, opened 2021-06-22
DOC: remove myst_parser extension

mgeier:
I'm not quite sure how `nbsphinx` and `myst_parser` interact, but I think latter isn't needed (because it is used only indirectly by Jupytext).

I haven't installed everything locally, so I'll try it here with the RTD integration.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -52,7 +52,6 @@ extensions = [
     'sphinx.ext.graphviz',  # Dependency diagrams
     'notfound.extension',
     'hoverxref.extension',
-    'myst_parser',
 ]
 
 # Hoverxref Extension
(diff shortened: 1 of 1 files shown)


chrisjsewell:
@mgeier how do you think all the Markdown files get passed lol

mgeier:
OK, it does *not* work, because `myst_parser` is used to parse the `*.md` files (as @chrisjsewell mentioned above). I guess some of them could be handled with `nbsphinx` (see https://nbsphinx.readthedocs.io/a-markdown-file.html), but this would not work with `changelog.md`, which uses MyST syntax.

mgeier:
It also doesn't work with `gallery.md` and `background.md`, because they also use MyST syntax.

This concludes my experiments.

See https://github.com/spatialaudio/nbsphinx/issues/420 for further discussion.

astrojuanlu:
Yep, we use `myst_parser` to have our narrative documentation in MyST (unrelated to the notebook parsing, as far as I understand)

astrojuanlu:
In fact, we did this in two steps: first we used jupytext to store the notebooks in `.md` format https://github.com/poliastro/poliastro/pull/978, then we migrated to MyST https://github.com/poliastro/poliastro/pull/1073

PR #1261, opened 2021-06-22
DOC: parse notebooks with MyST-NB

mgeier:
This is another experiment, see #1260.

Again, I'm not sure whether that works, and it might need a few iterations with RTD.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -83,6 +83,7 @@ templates_path = ['_templates']
 
 # The suffix of source filenames.
 source_suffix = {
+    '.rst': 'restructuredtext',  # https://github.com/readthedocs/sphinx-autoapi/issues/294
     '.mystnb': 'myst-nb',
 }
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -85,6 +85,7 @@ templates_path = ['_templates']
 source_suffix = {
     '.rst': 'restructuredtext',  # https://github.com/readthedocs/sphinx-autoapi/issues/294
     '.mystnb': 'myst-nb',
+    '.ipynb': 'dummy',
 }
 
 # Warning suppresses
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -82,7 +82,9 @@ autodoc_member_order = 'bysource'
 templates_path = ['_templates']
 
 # The suffix of source filenames.
-source_suffix = ['.rst' , '.md']
+source_suffix = {
+    '.mystnb': 'myst-nb',
+}
 
 # Warning suppresses
 suppress_warnings = ['image.nonlocal_uri']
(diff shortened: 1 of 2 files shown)


astrojuanlu:
Oh wow, it failed because of pip trying to resolve some dependencies... I would love to see the continuation of this experiment. I can lend a hand tomorrow.

astrojuanlu:
I waited 90 minutes locally, then both ran out of patience and saw the response to pypa/pip#10092 :) So, this is the culprit:

```
$ pip check
sphinx-rtd-theme 0.5.2 has requirement docutils<0.17, but you have docutils 0.17.1.
myst-nb 0.12.3 has requirement myst-parser~=0.13.5, but you have myst-parser 0.15.1.
myst-nb 0.12.3 has requirement nbconvert~=5.6, but you have nbconvert 6.1.0.
```

According to https://github.com/executablebooks/MyST-NB/issues/333#issuecomment-855382132,

> You can already use myst-nb==0.13.0a1 and yes we will be releasing myst-nb 0.13 "proper" shortly :crossed_fingers:

And, indeed, with that change the installation succeeds locally.

astrojuanlu:
```
Extension error:
source_suffix '.ipynb' is already registered
```

mgeier:
Thanks for fixing the previous error!

The current error is because both `nbsphinx` and `myst_nb` register the `.ipynb` suffix, which in the vast majority of cases totally makes sense.

Here it's causing problems, even though, ironically, no `.ipynb` files are used at all.

astrojuanlu:
I see! Would it make sense to (for the sake of the experiment) replace nbsphinx with myst-nb? However I understand that the example gallery will fail... so I don't know if we will get very far.

mgeier:
> Would it make sense to (for the sake of the experiment) replace nbsphinx with myst-nb?

Yes, that's what I'm trying to do here. I didn't expect there to be that many stumbling blocks.

I'm trying at https://github.com/spatialaudio/nbsphinx/pull/578 to accept an override for `.ipynb`.

> However I understand that the example gallery will fail...

I expect the thumbnails to fail, but maybe the gallery itself will work?

astrojuanlu:
It's working now! Build should be up soon

astrojuanlu:
https://poliastro--1261.org.readthedocs.build/en/1261/

astrojuanlu:
So, if I understand correctly, now we are using MyST-NB to parse the notebooks instead of jupytext through nbsphinx?

mgeier:
Cool!

As expected, the thumbnails are missing (except the once specified in `conf.py`): https://poliastro--1261.org.readthedocs.build/en/1261/gallery.html

mgeier:
> So, if I understand correctly, now we are using MyST-NB to parse the notebooks instead of jupytext through nbsphinx?

Yes. At least that's how I understand it.

mgeier:
MyST doesn't seem to support `<div class="alert alert-info">` boxes, see the bottom of https://poliastro--1261.org.readthedocs.build/en/1261/examples/Analyzing%20NEOs.html.
AFAIK, they have a different markup for admonitions (which doesn't work in JupyterLab, though).

astrojuanlu:
I think this experiment is complete @mgeier ? I summarized my findings here https://docs.readthedocs.io/en/stable/guides/jupyter.html (and we are sticking with nbsphinx in poliastro)

mgeier:
Yes, as far as I'm concerned, this experiment is complete. It led to an improvement in `nbsphinx` (https://github.com/spatialaudio/nbsphinx/pull/578), which is great.

Thanks for writing the guide, I think this is really helpful!

PR #1262, opened 2021-06-27
Allow Plotly 5

astrojuanlu:
This should result in better performance and easier installation.


astrojuanlu:
Locally I get a `Uncaught ReferenceError: require is not defined` on JupyterLab. However, standalone `plotly.graph_objects.Figure` objects work without problems. Investigating.

astrojuanlu:
We were using the wrong renderer in one of the notebooks. Fixing.

astrojuanlu:
Plots look good in the documentation 👍🏽 will let this auto-merge

PR #1264, opened 2021-06-27
Some docstring changes and an error check

Yash-10:
The core propagation docstrings mentioned astropy units instead of floats. 

The `get_mean_elements` accessed the `name` property without checking for the body. As a result, the logic didn't seem to go the `except` block and instead raised an in-built python error if `None` was passed as the `body`, for example. Would having a custom error help in such a case?

Thanks!


Yash-10:
Does the link here: https://github.com/poliastro/poliastro/blob/main/src/poliastro/core/propagation/__init__.py#L753 not work?

tests/tests_twobody/test_mean_elements.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+from poliastro.twobody.mean_elements import get_mean_elements
+
+
+def test_get_mean_elements_raises_error_if_invalid_body():
+    body = "Sun"
```
jorgepiloto, line 7: What about `body = "PlanetNine"`? Just kidding, looks good to me 👍🏽

jorgepiloto (approved):

jorgepiloto:
> Does the link here: https://github.com/poliastro/poliastro/blob/main/src/poliastro/core/propagation/__init__.py#L753 not work?

It seems that the article was removed in that link. I found as an alternative this one: https://www.researchgate.net/publication/288133262_Accurate_Kepler_equation_solver_without_transcendental_function_evaluations

tests/tests_twobody/test_mean_elements.py
Yash-10: Yeah, better to keep the name away from the context of astronomical bodies :)

astrojuanlu:
Let's better use this link for the Pimienta propagator: http://ancs.eng.buffalo.edu/pdf/ancs_papers/2012/ProtoType.pdf

src/poliastro/core/propagation/__init__.py
```diff
@@ -733,24 +735,24 @@ def pimienta(k, r0, v0, tof):
...
-        Final velocity vector
+    rr: ~np.array
+        Final position vector.
+    vv: ~np.array
+        Final velocity vector.
 
     Note
     ----
-    This algorithm was drived from the original paper: http://hdl.handle.net/10477/50522
+    This algorithm was derived from the original paper: https://www.researchgate.net/publication/288133262_Accurate_Kepler_equation_solver_without_transcendental_function_evaluations
```
astrojuanlu, line 755: Or better yet, let's put a proper reference:

Pimienta-Penalver, A., and Crassidis, J. "Accurate Kepler Equation Solver without Transcendental Function Evaluations." AAS Jer-Nan Juang Astrodynamics Symposium. College Station, TX, June 2012, AAS Paper #2012-612

astrojuanlu: Not sure what citation style are we using :) Or whether @jorgepiloto used the thesis or the AAS paper:

Screenshot 2021-07-02 at 07-11-01 Accurate Kepler Equation Solver without Transcendental - Google Scholar

src/poliastro/twobody/mean_elements.py
```diff
@@ -23,6 +25,10 @@ def get_mean_elements(body, epoch=J2000):
     # So we are reverting the conversion in the last line
     # This way at least we avoid copy pasting the values
     try:
+        if body not in SOLAR_SYSTEM_BODIES:
```
astrojuanlu, line 28: I just realized that `Sun` is an instance of `SolarSystemPlanet`... I wonder if we should change that somehow 😅

astrojuanlu: In any case, what's the point of checking this list first? The line

```
body_index = PLAN94_BODY_NAME_TO_PLANET_INDEX[name]
```

tells me that we could replace this check by

```
if name not in PLAN94_BODY_NAME_...
```

and so we wouldn't need `SOLAR_SYSTEM_BODIES`. Right?

astrojuanlu (changes requested):
Left a few comments

src/poliastro/twobody/mean_elements.py
Yash-10: I recall the reason was to handle if someone inputs `body=None` or any other invalid body such that

```python
name = body.name.lower()
```
would fail in the next line, since then we would get the error:

```bash
AttributeError: 'NoneType' object has no attribute 'name'
```

So would it be good to check for `body` first, instead of `name`?

astrojuanlu: I think we shouldn't shield against every possible misuse of the API. The documentation and the signature are very clear:

https://github.com/poliastro/poliastro/blob/bebbf617e0ebfb797ba1c593b8241e4cf0f9c466/src/poliastro/twobody/mean_elements.py#L10-L16

If the user passes `None` as a `body`... `AttributeError: 'NoneType' object has no attribute 'name'` is a perfect explanation of what happened :)

Yash-10: Thanks for this suggestion!

Now, it seems that there might be no need of checking `if name not in ...` since it's already inside a try-except block? Instead we could add to the existing error message stating that it must be an instance of `SolarSystemPlanet` and change `Sun` to not be an instance of `SolarSystemPlanet`?

astrojuanlu (approved):
Looks good to me except for the citation, but not a big deal 👍🏽

Yash-10:
I tried to fix the type hint failures.

jorgepiloto:
Merging, thanks @Yash-10 👍🏽

PR #1268, opened 2021-06-30
Lat and Lon detector

Yash-10:
A draft for detecting latitude and longitude crossing.

- [ ] Change ITRS and GCRS to account for a general attractor and not just the Earth.
- [ ] Find a way to get rid of the warnings from tests.

### DIFF
--- a/src/poliastro/twobody/events.py
+++ b/src/poliastro/twobody/events.py
@@ -8,6 +8,8 @@ from astropy.coordinates import (
 from astropy.time import Time
 from numpy.linalg import norm
 
+from poliastro.bodies import Earth
+
 
 class Event:
     """Base class for event functionalities.
@@ -94,10 +96,10 @@ class LatitudeCrossEvent(Event):
 
     Parameters
     ----------
+    orbit: ~poliastro.twobody.orbit.Orbit
+        Orbit.
     lat: float
-        Threshold latitude (km).
-    R: float
-        Radius of the attractor (km).
+        Threshold latitude (in degrees).
     terminal: bool
         Whether to terminate integration if this event occurs, defaults to True.
     direction: float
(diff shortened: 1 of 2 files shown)


src/poliastro/twobody/events.py
```diff
@@ -80,3 +87,80 @@ class LithobrakeEvent(AltitudeCrossEvent):
...
+        Radius of the attractor (km).
+    terminal: bool
+        Whether to terminate integration if this event occurs, defaults to True.
+    direction: float
+        Handle triggering of event based on whether latitude is crossed from above
+        or below, defaults to 0, i.e., event is triggered while traversing from both directions.
+
+    """
+
+    def __init__(self, lat, R, terminal=True, direction=0):
```
jorgepiloto, line 109: As discussed in the weekly meeting on July the 2nd, let us pass a `~poliastro.bodies.body`, so we can check if body is the `Earth` and raise a `NotImplementedError` till #1269 is better discussed.

src/poliastro/twobody/events.py
```diff
@@ -80,3 +87,80 @@ class LithobrakeEvent(AltitudeCrossEvent):
...
+        super().__init__(terminal, direction)
+        self._R = R
+        self._lat = lat  # Threshold latitude (in degrees).
+
+    def __call__(self, t, u_, k):
+        self._last_t = t
+        xyz = u_[:3]
+
+        obstime = Time(self._last_t, format="jd")
```
jorgepiloto, line 118: The `t` variable is just a float value representing the amount of seconds since the initial orbit epoch. Therefore, we require to store the initial orbit epoch to build the right epoch value after `t` elapsed seconds:

```python
Time(self._last_t = self._epoch0 + t * u.s)
```

src/poliastro/twobody/events.py
jorgepiloto: In fact, let us pass the orbit to be propagated so we can compute:

```python
self._R = orbit.attractor.R
self._epoch0 = orbit.epoch
```
those will be useful for later logic.

Yash-10:
Thanks, @jorgepiloto! I added a check to ensure the attractor is Earth. For the tests, there might be a need to modify them since they are more or less arbitrary.

src/poliastro/twobody/events.py
```diff
@@ -80,3 +85,36 @@ class LithobrakeEvent(AltitudeCrossEvent):
...
+        self._R_polar = orbit.attractor.R_polar
+        self._epoch = orbit.epoch
+        self._lat = lat.to(u.deg).value  # Threshold latitude (in degrees).
+
+    def __call__(self, t, u_, k):
+        self._last_t = t
+        pos_on_body = (u_[:3] / norm(u_[:3])) * self._R
+        lat_, _, _ = cartesian_to_ellipsoidal_fast(self._R, self._R_polar, *pos_on_body)
+
+        return np.rad2deg(lat_) - self._lat
```
astrojuanlu, line 120: There seems to be a problem with units here that only surfaces when disabling numba, see

https://dev.azure.com/poliastro/poliastro/_build/results?buildId=1408&view=logs&j=1117f030-3e3a-5822-cfd7-97e776c1f4cb&t=af77bb0c-2bf7-50ca-9891-2682490c119e&l=119

astrojuanlu: Maybe because we need to do `self._R = orbit.attractor.R.to_value(u.km)`

astrojuanlu (approved):
LGTM!

Issue #1270, opened 2021-06-30
Better error message if `len(dv) != 3` in `Maneuver` constructor


Yash-10:
If I try to create a `Maneuver` object with an incorrect delta-v (i.e. doesn't consist of all three components) using:

```python
dv1 = [0, 0, 0] << u.km / u.s
dv2 = [5, 0] << u.km / u.s  # Incorrect dv

man = Maneuver((0 * u.s, dv1), (2 * u.s, dv2))
```
one gets this error:

```bash
     44         self._dts, self._dvs = self._initialize(
     45             [(_dt * u.one).value for _dt in _dts] * (_dts[0] * u.one).unit,
---> 46             [(_dv * u.one).value for _dv in _dvs] * (_dvs[0] * u.one).unit,
     47         )
     48         try:

TypeError: can't multiply sequence by non-int of type 'CompositeUnit'
```
which seems to come from `astropy` directly.

Whereas only having a single incorrect`(dt, dv)` pair:

```python
dv2 = [5, 0] << u.km / u.s  # Incorrect dv

Maneuver((2 * u.s, dv2))
```
gives a better error message:

```bash
ValueError: Delta-V must be three dimensions vectors
```

Would it be better to create a more better error message in the first case to signify that the input delta-v didn't contain all three components?

[This](https://github.com/poliastro/poliastro/blob/main/src/poliastro/maneuver.py#L42) line also suggested to change the API or error-checking, so it might be a good time to do that along with this?

Thanks!

astrojuanlu:
Woops that's horrible. +1 on having a better error message!

astrojuanlu:
Closed in #1271

PR #1271, opened 2021-07-01
Minor `Maneuver` constructor change

Yash-10:
Could be a quick-fix for #1270 to move the `_initialize` method inside the try block. A test was added to check the change.

Any thoughts on this would help!

Thanks!

### DIFF
--- a/src/poliastro/maneuver.py
+++ b/src/poliastro/maneuver.py
@@ -41,12 +41,13 @@ class Maneuver:
         self.impulses = args
         # HACK: Change API or validation code
         _dts, _dvs = zip(*args)
-        self._dts, self._dvs = self._initialize(
-            [(_dt * u.one).value for _dt in _dts] * (_dts[0] * u.one).unit,
-            [(_dv * u.one).value for _dv in _dvs] * (_dvs[0] * u.one).unit,
-        )
+
         try:
-            if not all(len(dv) == 3 for dv in self._dvs):
+            self._dts, self._dvs = self._initialize(
+                [(_dt * u.one).value for _dt in _dts] * (_dts[0] * u.one).unit,
+                [(_dv * u.one).value for _dv in _dvs] * (_dvs[0] * u.one).unit,
+            )
+            if not all(len(dv) == 3 for dv in _dvs):
                 raise TypeError
         except TypeError:
             raise ValueError("Delta-V must be three dimensions vectors")
--- a/tests/test_maneuver.py
+++ b/tests/test_maneuver.py
@@ -1,3 +1,5 @@
+import warnings
+
 import numpy as np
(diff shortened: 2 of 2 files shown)


src/poliastro/maneuver.py
```diff
@@ -41,12 +41,13 @@ def __init__(self, *args):
...
-            [(_dv * u.one).value for _dv in _dvs] * (_dvs[0] * u.one).unit,
-        )
+
         try:
-            if not all(len(dv) == 3 for dv in self._dvs):
+            self._dts, self._dvs = self._initialize(
+                [(_dt * u.one).value for _dt in _dts] * (_dts[0] * u.one).unit,
+                [(_dv * u.one).value for _dv in _dvs] * (_dvs[0] * u.one).unit,
+            )
+            if not all(len(dv) == 3 for dv in _dvs):
```
astrojuanlu, line 50: If this `TypeError` ever raised? I see it's not covered by the tests. Maybe it's dead code and we should remove it? Otherwise, let's use this opportunity to add a test

astrojuanlu (changes requested):
Good change! Left just one comment

src/poliastro/maneuver.py
Yash-10: The test that was added checks for the line:

```python
if not all(len(dv) == 3 for dv in _dvs):
```

and since

```python
self._dts, self._dvs = self._initialize(
                [(_dt * u.one).value for _dt in _dts] * (_dts[0] * u.one).unit,
                [(_dv * u.one).value for _dv in _dvs] * (_dvs[0] * u.one).unit,
            )
```
seems to raises an error only when `len(dv) != 3`, the error message might handle both. What do you think?

astrojuanlu: Then let's remove that code, looks like that case is covered already

Yash-10: Sorry, but now it looks like the test `test_maneuver_raises_error_if_dvs_are_not_vectors` raises an error. This is because removing the `if not all(len(dv) == 3 for dv in _dvs):` check would also allow delta-v's that are not 3 dimensional, which might not be favourable.

Yash-10: Maybe we could keep the condition:

```py
if not all(len(dv) == 3 for dv in _dvs):
```
unchanged due to the above reason. Could you let me know what do you think @astrojuanlu ?

astrojuanlu: Fair enough, if there is a test checking for it, it's not dead code then. Let's keep it.

astrojuanlu (approved):
I think this is ready to merge despite the coverage diff check. Thanks @Yash-10 !

PR #1272, opened 2021-07-04
Some more tests

Yash-10:
Some tests for `heliosynchronous`.

### Questions

- Would it be a good option to check for the "unphysical" condition where `a` is less than the radius of Earth and if radius of perigee is less than radius of Earth), as stated in Vallado?
Screenshot from 2021-07-04 17-30-28.
- I think [this](https://github.com/poliastro/poliastro/blob/main/src/poliastro/twobody/orbit.py#L1067) exception would never be raised since [this](https://github.com/poliastro/poliastro/blob/main/src/poliastro/twobody/orbit.py#L1028) line ensures that and also given that the condition for `alt < 0` is already checked in one of the above lines. Would it be a good choice to remove raising that exception?

 Currently, if I try to create an orbit with a < radius of earth, then it doesn't give an error.

The test `test_heliosynchronous_orbit_ecc` fails, but I think it should have passed. That might need more investigation.

---
It seems the difference in expected eccentricities is of the order `1e-5` if I use these parameters:

```py
expected_ecc = 0.2 * u.one
expected_inc = 98.6 * u.deg
expected_a = 7346.846 * u.km
```
instead of the current parameters (giving errors of the order `1e-2`, which may not be negligible):

```py
expected_ecc = 0.0 * u.one
expected_inc = 98.6 * u.deg
expected_a = 7178.1363 * u.km
```

It looks like the calculation is not accurate, or am I missing something?

Thanks!

### DIFF
--- a/tests/tests_twobody/test_orbit.py
+++ b/tests/tests_twobody/test_orbit.py
@@ -760,18 +760,6 @@ def test_heliosynchronous_orbit_a():
     assert_quantity_allclose(ss0.ecc, expected_ecc)
 
 
-def test_heliosynchronous_orbit_inc():
-    # Vallado, example 11-2b
-    expected_ecc = 0.0 * u.one
-    expected_inc = 98.6 * u.deg
-    expected_a = 7178.1363 * u.km
-    ss0 = Orbit.heliosynchronous(Earth, a=expected_a, ecc=expected_ecc)
-
-    assert_quantity_allclose(ss0.inc, expected_inc, rtol=1e-4)
-    assert_quantity_allclose(ss0.a, expected_a, rtol=1e-5)
-    assert_quantity_allclose(ss0.ecc, expected_ecc)
-
-
 def test_heliosynchronous_orbit_ecc():
     # Vallado, example 11-2b
     expected_ecc = 0.0 * u.one
--- a/tests/tests_twobody/test_orbit.py
+++ b/tests/tests_twobody/test_orbit.py
@@ -790,7 +790,7 @@ def test_heliosynchronous_orbit_raises_floating_point_error_if_invalid_input():
     inc = 0 * u.rad
 
     with pytest.raises(ValueError) as excinfo:
-        orbit = Orbit.heliosynchronous(Earth, a=a, inc=inc)
+        Orbit.heliosynchronous(Earth, a=a, inc=inc)
     assert "No SSO orbit with given parameters can be found." in excinfo.exconly()
 
 
--- a/tests/tests_twobody/test_orbit.py
+++ b/tests/tests_twobody/test_orbit.py
@@ -769,7 +769,8 @@ def test_heliosynchronous_orbit_ecc():
 
     assert_quantity_allclose(ss0.inc, expected_inc, rtol=1e-4)
     assert_quantity_allclose(ss0.a, expected_a, rtol=1e-5)
-    assert_quantity_allclose(ss0.ecc, expected_ecc)
+    # Vallado uses a slightly different value for n_sunsync, hence `atol` needs to be added.
+    assert_quantity_allclose(ss0.ecc, expected_ecc, atol=1e-1)
 
 
 @pytest.mark.filterwarnings("ignore::RuntimeWarning")
--- a/tests/tests_twobody/test_orbit.py
+++ b/tests/tests_twobody/test_orbit.py
@@ -760,6 +760,40 @@ def test_heliosynchronous_orbit_a():
     assert_quantity_allclose(ss0.ecc, expected_ecc)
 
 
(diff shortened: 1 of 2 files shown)


jorgepiloto:
Hi @Yash-10, we might consider doing something similar to `from_classical`, where the decorator `@u.quantity_input` is applied and the value of `a` is checked to be positive. However, its size is never compared with current attractor's radius.

astrojuanlu:
Bear in mind that creating an orbit with a semi-major axis less than the radius of the planet is totally valid. It just means that, if the object is approaching the attractor, it will impact at some point. But there could be:

- Asteroids, comets, and artificial satellites about to impact
- Rockets and missiles just launched from the Earth (and not circularized their orbits yet)

So I wouldn't add any further checks in `from_classical`.

tests/tests_twobody/test_orbit.py
```diff
@@ -760,6 +760,28 @@ def test_heliosynchronous_orbit_a():
...
+def test_heliosynchronous_orbit_ecc():
+    # Vallado, example 11-2b
+    expected_ecc = 0.0 * u.one
+    expected_inc = 98.6 * u.deg
+    expected_a = 7178.1363 * u.km
+    ss0 = Orbit.heliosynchronous(Earth, a=expected_a, inc=expected_inc)
+
+    assert_quantity_allclose(ss0.inc, expected_inc, rtol=1e-4)
+    assert_quantity_allclose(ss0.a, expected_a, rtol=1e-5)
+    assert_quantity_allclose(ss0.ecc, expected_ecc)
```
astrojuanlu, line 772: If I understand correctly, this corresponds to example 11-2a, right? I'm now trying to reproduce the results and I don't get the same. Am I doing something wrong?

```
>>> math.degrees(math.acos(-((1.12532 ** (7/2) * 2 * 1.991063853e-7) / (3 * 6378.137**2 * 0.00108263 * math.sqrt(398600.4418)))))
90.00000000000043
```

astrojuanlu (changes requested):
Left a comment about the failing test, I am not sure but it looks like an error in the book, or I am doing something wrong.

tests/tests_twobody/test_orbit.py
Yash-10: It seems the equations in the solution of example 11-2:

Screenshot from 2021-07-09 19-47-34

The value of `a` is printed as 1.12532 where it must have been 7178.1363. If I use 7178.1363 instead, in the equation, it gives me an inclination of 98.60308111784582. So maybe that's not a problem?

Yash-10:
It seems the differences arise due to the value of `n_sunsync`. Vallado uses `1.991063853e-7` whereas the value calculated in poliastro is `1.99098176e-7`. Although the difference is small, the calculation of eccentricity might have a considerable impact:

```py
_ecc_0 = np.sqrt(
      -3
      * R_SSO ** 2
      * J2_SSO
      * np.sqrt(k_SSO)
      * np.cos(inc)
      / (2 * a ** (7 / 2) * n_sunsync)
)
ecc = np.sqrt(1 - _ecc_0)
```

- The failing `test_heliosynchronous_orbit_ecc` test uses an inclination of 98.6 degrees but to be precise, it is 98.60308111784582 degrees. If I change `inc` to 98.60308111784582 in the test, we get a floating point error since the ecc calculated is slightly lower than 0 (of the order 1e-5).
- If I use 98.60308111784582 degrees in the test but using the `n_sunsync` value from Vallado (`1.991063853e-7`), I get the eccentricity 0.00025043, which, I think, is what is expected from the test.

Maybe we could pass the test by setting a higher `atol` since the `n_sunsync` value used in vallado is different and hence the result is expected to be different, or does the failing test have some bigger consequences?

astrojuanlu:
Oh, very good investigation @Yash-10 ! I guess we can't help but increase the tolerance if Vallado uses a different value for the rotational speed. Let's fix the test, rebase, and merge 👍🏽

tests/tests_twobody/test_orbit.py
astrojuanlu: Thanks for reproducing it! Yeah, shouldn't be a problem

Yash-10:
Codeclimate complained to shorten the condition:

```py
if (a is None) and (ecc is None) and (inc is None):
```

So tried to change it to:

```py
if all(coe is None for coe in [a, ecc, inc]):
```

but still doesn't fix it.

astrojuanlu:
Meh, we are mostly ignoring Codeclimate. Perhaps we should either have more fine-grained custom roles, or just remove it.

I approved the changes in this PR.

astrojuanlu (approved):
LGTM!

PR #1275, opened 2021-07-10
Convert any relative imports to absolute

### LINKED ISSUE #1259: Absolute vs relative imports in modules
There seem to be some places where the imports are done in a relative fashion whereas in an absolute sense at other places. For example, imports are relative [here](https://github.com/poliastro/poliastro/blob/main/src/poliastro/twobody/mean_elements.py) and absolute [here](https://github.com/poliastro/poliastro/blob/main/src/poliastro/twobody/elements.py)

From my preliminary understanding, relative imports are discouraged for intra-package imports since the modules might change locations during development but sometimes look good if the absolute import gets too long to write. 

So, the question is whether it would be good to:

- Convert all instances of relative imports to absolute, or
- Keep the imports as they are now, since they were kept (as they are) for some reason.

Would love to get any thoughts anyone might have!

Thanks!

Yash-10:
As per the suggestions by @astrojuanlu in #1259 to keep imports consistent, here all imports were made absolute. Fixes #1259 

Thanks!

### DIFF
--- a/src/poliastro/bodies.py
+++ b/src/poliastro/bodies.py
@@ -25,8 +25,8 @@ from astropy import units as u
 from astropy.constants import G
 from astropy.units import Quantity
 
-from . import constants
-from .frames import Planes
+from poliastro import constants
+from poliastro.frames import Planes
 
 
 # HACK: Constants cannot be hashed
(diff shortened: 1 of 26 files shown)


jorgepiloto:
Thanks for addressing this @Yash-10, merging as all test passed and coverage remained the same 👍🏽

PR #1277, opened 2021-07-12
Upgrade to newer Sphinx and sphinx-hoverxref

astrojuanlu:
Now math equations will render inside modals as well!


astrojuanlu:
Building docs with Sphinx 4.1 finally 💪🏽 cc @humitos

Issue #282, opened 2017-11-18
Improve and expand equinoctial elements handling


Juanlu001:
🐞 **Problem**

poliastro supports equinoctial elements, but test coverage is low and some conversions are not implemented.

🎯 **Goal**

Improve equinoctial element handling and showcase its advantages (no singularities, etc).

💡 **Possible solutions**

Some papers to start with:

* Broucke & Cefola "On the equinoctial orbit elements" https://doi.org/10.1007/BF01228432
* Cefola "Equinoctial orbit elements - Application to artificial satellite orbits" https://doi.org/10.2514/6.1972-937

📋  **Steps to solve the problem**

 * Comment below about what you've started working on.
 * Add, commit, push your changes
 * Submit a pull request and add this in comments - `Addresses #<put issue number here>`
 * Ask for a review in comments section of pull request
 * Celebrate your contribution to this project 🎉

Juanlu001:
In Hintz "Survey of Orbit Element Sets" [https://doi.org/10.2514/1.32237](https://doi.org/10.2514/1.32237) the modified equinoctial elements that we use in poliastro are explicitly recommended, citing Battin. However, in table 4 it's shown that there are no explicit formulas to convert from/to cartesian vectors:

![screenshot from 2018-04-30 12-36-55](https://user-images.githubusercontent.com/316517/39423649-3a4fdc18-4c73-11e8-8113-98c6391a3a2c.png)

There are many papers citing the original Walker set, but a quick overview gave nothing simple. Further investigation is needed.

Juanlu001:
The equinoctial elements used in GMAT, according to their Mathematical Specification (link available in [GMAT contributors documentation](http://gmatcentral.org/display/GW/For+Contributors)) are not the ones by Walker.

![screenshot from 2018-04-30 12-43-07](https://user-images.githubusercontent.com/316517/39423821-0fca12fa-4c74-11e8-8350-9a5e48f7b591.png)

Neither Orekit:

https://www.orekit.org/static/apidocs/org/orekit/orbits/EquinoctialOrbit.html

or SPICE:

ftp://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/spicelib/eqncpv.html

nor STK:

http://help.agi.com/stk/index.htm#stk/vehSat_coordType_equinoctial.htm

So, unless I'm mistaken, there's essentially no tooling out there that works with modified equinoctial elements, although they are the only ones that feature no singularities and have lots of applications.

[...]

PR #1279, opened 2021-07-14
States tests

Yash-10:
Some test additions for two-body states. Also includes the error check for 180 degrees inclination in modified equinoctial elements as discussed in #282.

**Question**

In this line: https://github.com/poliastro/poliastro/blob/main/src/poliastro/twobody/states.py#L93, would it be a good choice to check if ecc = 1 and set `a`'s value to `None` or something similar instead of numpy's infinity warning:

```
RuntimeWarning: divide by zero encountered in true_divide
```
since `a` is undefined for parabolic orbits?


tests/tests_twobody/test_states.py
```diff
@@ -35,3 +38,42 @@ def test_rv_state_has_rv_given_in_constructor():
...
+        attractor, a * (1 - _ecc ** 2), _ecc, _inc, _raan, _argp, _nu, None
+    )
+
+    expected_mean_motion = (2 * np.pi / period) * u.rad
+    n = ss.n
+
+    assert_quantity_allclose(n, expected_mean_motion)
+
+
+def test_classical_to_equinoctial_raises_singularity_error_if_inc_180_degrees():
```
jorgepiloto, line 65: Let us explicitly say the elements being tested are `modified_equinoctial` and not just `equinoctial`, since both sets are different. Therefore:

``` python
def test_classical_to_modified_equinoctial_raises_singularity_error_if_inc_180_degrees():
```

another approach:

```python
def test_coe_to_mee_raises_singularity_error_orbit_equatorial_and_retrograde():
```

src/poliastro/core/elements.py
```diff
@@ -220,10 +219,13 @@ def coe2mee(p, ecc, inc, raan, argp, nu):
         \end{align}
 
     """
+    if inc == np.pi:
+        raise ValueError(
+            "Cannot compute the equinoctial parameters, `h` and `k`, due to singularity"
```
jorgepiloto, line 224: We must inform user what parameter caused the singularity. What about the following:

```python
raise ValueError(
    "Cannot compute modified equinoctial set for 180 degrees orbit inclination due to `h` and `k` singularity."
)
```

jorgepiloto (changes requested):
Thanks for addressing this @Yash-10. I just left some minimal requested changes/sugestions 👍🏽

src/poliastro/core/elements.py
Yash-10: Thanks! I used your suggestion verbatim.

tests/tests_twobody/test_states.py
Yash-10: Thanks for this!

astrojuanlu:
Windows error is unrelated (#959)

astrojuanlu:
According to Codecov, most of the code in this pull request is not covered by tests, is that expected?

Yash-10:
I tried to add a `mee2rv` method based on: https://spsweb.fltops.jpl.nasa.gov/portaldataops/mpg/MPG_Docs/Source%20Docs/EquinoctalElements-modified.pdf, but couldn't find any example to test it yet. Maybe I should remove the method and work on it separately?

astrojuanlu:
We agreed to manually generate a single test case with GMAT (apart from having more thorough tests on our validation repository)

astrojuanlu:
Seems like all @jorgepiloto review comments were addressed, and the only missing thing was to add a test case based on GMAT. This is already useful though, so I merged the latest changes from `main` hoping that the tests pass so we can merge it.

astrojuanlu:
Thanks @Yash-10 !

PR #1286, opened 2021-07-21
Avoid iterating results twice in `propagation.py`

Yash-10:
This tries to solve a "TODO" in the `propagation` module where the variable `results` was iterated twice. 

The resulting improvement is ~1s for propagation for 50 seconds.

Thanks!

---

~Re-triggering still shows the tests as canceled. Maybe some remote issue.~ Not now


jorgepiloto (approved):
Looks good to me @Yash-10, no more comments by my side. Let us wait for @astrojuanlu approval too 🚀

astrojuanlu (approved):
Yay, thanks @Yash-10 !

Issue #1285, opened 2021-07-21
Orbit position vectors and the `terminal` attribute of events


Yash-10:
I was experimenting with the added event detectors to create a "how-to" guide, where I found, probably, something strange:

**Expected behavior**

If the `terminal` attribute is set to False, the orbit position vectors must keep on propagating until `tofs[-1]` even if an event is triggered in between.

**Actual behaviour**

Orbit vectors get stuck if an event is triggered even if the `terminal` attribute of events is set to `False`.

**Reproducible example**
```py
import numpy as np

from poliastro.bodies import Earth

from poliastro.twobody.events import LatitudeCrossEvent
from poliastro.twobody.propagation import cowell
from poliastro.twobody.orbit import Orbit

r = [-6142438.668, 3492467.56, -25767.257] << u.km
v = [505.848, 942.781, 7435.922] << u.km / u.s
orbit = Orbit.from_vectors(Earth, r, v)

thresh_lat = 60 * u.deg
latitude_cross_event = LatitudeCrossEvent(orbit, thresh_lat, terminal=False)
events = [latitude_cross_event]

tofs = [1000, 1250, 1500, 1710, 2000] << u.s
rr, _ = cowell(
    Earth.k,
    orbit.r,
    orbit.v,
    tofs,
    events=events,
)
print(f"The threshold latitude was crossed {latitude_cross_event.last_t} s after the orbit's epoch")

print(rr)
```

gives output:

```
The threshold latitude was crossed 1701.7168421304814 s s after the orbit's epoch
[[-5636590.6652444   4435248.55827388  7410154.74204886]
 [-5510128.66403839  4670943.80747798  9269135.24138821]
 [-5383666.66274741  4906639.05661024 11128115.74058572]
 [-5281628.60050468  5096813.86201176 12628066.44401966]
 [-5281628.60050468  5096813.86201176 12628066.44401966]] km
```

The last two position vectors (corresponding to the time instants after the event was detected) are unexpectedly the same even if the `terminal` attribute was set to `False`.

**Possible ideas/solutions**

I think [this](https://github.com/poliastro/poliastro/blob/main/src/poliastro/twobody/propagation.py#L112) condition:

```py
if t_end is not None and t > t_end:
    t = t_end
```
checks if the current `t` instant is greater than the time instance of the event. If it is, then the t instance is brought back to the event instance time. This might be the cause of the above "unexpected" behavior. Interestingly, if I remove the lines:

```py
[...]

PR #1288, opened 2021-07-22
Continue propagation if `event.terminal` is `False`

Yash-10:
From #1285, this prevents hanging of orbit if an event's terminal property is False. I realized this behavior might not have been evident earlier since this condition was not tested.

For the test, would there be any suggestions on whether a single event detector would suffice or should it be parametrized for all events?

Thanks!

### DIFF
--- a/tests/tests_twobody/test_events.py
+++ b/tests/tests_twobody/test_events.py
@@ -249,3 +249,103 @@ def test_node_event_equatorial_orbit():
     )
 
     assert_quantity_allclose(node_event.last_t, 0.0 * u.s, atol=1e-1 * u.s)
+
+
+def test_orbit_propagation_continues_if_events_terminal_is_False():
+    r = [-6142438.668, 3492467.56, -25767.257] << u.km
+    v = [505.848, 942.781, 7435.922] << u.km / u.s
+    orbit = Orbit.from_vectors(Earth, r, v)
+
+    thresh_lat = 60 * u.deg
+    # Event occurs at ~1701.7 s.
+    latitude_cross_event = LatitudeCrossEvent(orbit, thresh_lat, terminal=False)
+    events = [latitude_cross_event]
+
+    # The last two tofs are after the detection of the event.
+    tofs = [1000, 1250, 1500, 1710, 2000] << u.s
+    rr, _ = cowell(
+        Earth.k,
+        orbit.r,
+        orbit.v,
+        tofs,
+        events=events,
+    )
+
+    # Check position vectors don't repeat during propagation.
+    assert not np.allclose(rr[-1], rr[-2])
+
+
+def test_orbit_propagation_position_vector_does_not_repeat_if_events_terminal_is_True():
+    r = [-6142438.668, 3492467.56, -25767.257] << u.km
+    v = [505.848, 942.781, 7435.922] << u.km / u.s
+    orbit = Orbit.from_vectors(Earth, r, v)
+
+    thresh_lat = 60 * u.deg
+    # Event occurs at ~1701.7 s.
+    latitude_cross_event = LatitudeCrossEvent(orbit, thresh_lat, terminal=True)
+    events = [latitude_cross_event]
+
+    # The last two tofs are after the detection of the event.
+    tofs = [1000, 1250, 1500, 1710, 2000] << u.s
+    rr, _ = cowell(
+        Earth.k,
+        orbit.r,
+        orbit.v,
+        tofs,
+        events=events,
+    )
+
+    # Check position vector doesn't repeat if terminal set to True.
+    assert len(rr) == 4  # From the 5th tof in tofs, position vector starts repeating.
+
+
+@pytest.mark.parametrize(
+    "latitude_terminal,penumbra_terminal,rr_length,t_end",
+    [
+        (True, True, 4, 266.15058 * u.s),
+        (True, False, 5, 305.65173 * u.s),
+        (False, True, 4, 266.15058 * u.s),
+        (False, False, 6, 500 * u.s),
+    ],
+)
+def test_propagation_stops_if_atleast_one_event_has_terminal_set_to_True(
+    latitude_terminal, penumbra_terminal, rr_length, t_end
+):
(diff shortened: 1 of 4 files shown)


jorgepiloto:
My local approach to this:

```python
    # Collect only the terminal events
    terminal_events = [event for event in events if event.terminal is True]

    # If there are no terminal events, then the last time of integration is the
    # greatest one from the original array of propagation times
    if len(terminal_events) == 0:
        last_t = max(tofs)
        limiting_event = None
    else:
        # Filter the event which triggered first
        last_t = min([event.last_t for event in events])
        limiting_event = [event for event in events if event.last_t == last_t]

    # Crop the original array of times up to the 
    below_or_equal_last_t = tofs <= last_t
    tofs = tofs[below_or_equal_last_t]
```

Yash-10:
Thanks @jorgepiloto for sharing the approach. Currently, I am trying to figure out a clean way to include the immediate next `tof` after the time instant of the event. This might be necessary since only selecting time of flights less than the time instant of the event, might fail if:

```py
len(tofs) == 1
```

Yash-10:
- We calculate `limiting_event` but currently don't use it anywhere. Since we would like to tell the user which event was triggered, should we additionally return `tofs[-1]` along with `r` and `v` inside `cowell` and tell the user to check each events' `last_t` property to verify which event was triggered?
- There is one case where the current implementation fails: When there are at least two events with one having terminal=False and other having terminal=True, and where the event corresponding to terminal=False would have triggered before the other event had the terminal property of that event be also set to True.

Yash-10:
Changing the original line for the `lithobrake` example in docs:

```py
plt.plot(tofs.value, rr.norm() - Earth.R)
```
 to
```py
plt.plot(tofs[:len(rr)].value, rr.norm() - Earth.R)
```
would solve the documentation errors.

I wonder if it would be better to return `last_t` somehow from `cowell`?

astrojuanlu:
> I wonder if it would be better to return last_t somehow from cowell?

Again, we are kind of limited because all propagators should respect the same interface :/ And, at the moment, we want to treat `cowell` like the others. I think it's a good thing that we keep it.

Yash-10:
This pull request is ready to be reviewed :)

tests/tests_twobody/test_events.py
```diff
@@ -207,3 +207,101 @@ def test_penumbra_event_crossing():
...
+    rr, _ = cowell(
+        Earth.k,
+        orbit.r,
+        orbit.v,
+        tofs,
+        events=events,
+    )
+
+    # Check position vector of orbit doesn't hang.
+    assert not np.allclose(rr[-1], rr[-2])
```
astrojuanlu, line 233: This assertion is a bit confusing. What about checking that the. `len` of `rr` is the same as the original `tofs`? Also, I'm wondering if the norm of `rr` becomes zero after "lithobrake" and if it's something we can use here as well.

tests/tests_twobody/test_events.py
```diff
@@ -207,3 +207,101 @@ def test_penumbra_event_crossing():
...
+        20.0 * u.deg,
+        10.0 * u.deg,
+        0 * u.deg,
+    )
+    orbit = Orbit.from_classical(attractor, *coe, epoch=epoch)
+
+    penumbra_event = PenumbraEvent(orbit, terminal=terminals[-1])
+
+    thresh_lat = 30 * u.deg
+    latitude_cross_event = LatitudeCrossEvent(orbit, thresh_lat, terminal=terminals[0])
```
astrojuanlu, line 290: Rather than `terminals[0]` and `terminals[-1]`, can we use two distinct variable names? Like `penumbra_terminal` and `latitude_terminal`. That will make the conditional at the end more clear as well I think.

astrojuanlu (changes requested):
Looks almost perfect! Just left a couple of comments about the tests

tests/tests_twobody/test_events.py
Yash-10: Thanks!

tests/tests_twobody/test_events.py
Yash-10: I wanted to confirm, through this test, the position vectors don't hang in the case where the event was detected and the terminal was set to False. Earlier, `len(rr) == len(tofs)` was satisfied, but the position vectors hanged (i.e. kept repeating till `tofs[-1]`). Would you recommend making the current assertion condition a bit easier to read?

Yash-10: > I'm wondering if the norm of `rr` becomes zero after "lithobrake" and if it's something we can use here as well.

Does it mean adding some more checks that check something with `rr` for all the added event tests?

Yash-10:
The latitude events' default terminal value was changed to False, as was decided earlier.

tests/tests_twobody/test_events.py
```diff
@@ -207,3 +207,103 @@ def test_penumbra_event_crossing():
...
+    tofs = [1000, 1250, 1500, 1710, 2000] << u.s
+    rr, _ = cowell(
+        Earth.k,
+        orbit.r,
+        orbit.v,
+        tofs,
+        events=events,
+    )
+
+    # Check position vector of orbit doesn't hang.
```
jorgepiloto, line 232: Instead of saying "orbit doesn't hang" let us say here "checking not repeated position vectors during propagation".

tests/tests_twobody/test_events.py
jorgepiloto: Checking the length of vectors, as claimed by @Yash-10, does not ensure no repeated position vectors. The last values of the array need to be checked directly.

jorgepiloto (changes requested):

jorgepiloto:
This looks good to me, pinging @astrojuanlu so he can made a final review about this 👍🏽

tests/tests_twobody/test_events.py
astrojuanlu: I understand now, thanks for the clarification!

astrojuanlu (approved):
Ready to merge! 🚀 Thanks @Yash-10 !

PR #1293, opened 2021-07-26
Node crossing

Yash-10:
This attempts to add a node crossing detector.

- The `atol` value for an already equatorial orbit needed to kept `0.1s` for the tests to pass. I think it's because `scipy` would take some small finite time if the initial conditions themselves match the event condition.

Thanks!

### DIFF
--- a/src/poliastro/twobody/events.py
+++ b/src/poliastro/twobody/events.py
@@ -225,5 +225,5 @@ class NodeCrossEvent(Event):
         super().__init__(terminal, direction)
 
     def __call__(self, t, u_, k):
-        self._last = t
+        self._last_t = t
         return u_[2]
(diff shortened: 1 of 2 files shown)


Yash-10:
I locally tried to validate this event using both sets of parameters, one that was mentioned here: https://gitlab.orekit.org/orekit-labs/python-wrapper/-/blob/master/python_files/test/NodeDetectorTest.py and the other which was already there in `validate_events.py`.

The test locally passes using both parameters.

If it would be ok, I could also open a pull request in the validation repo for this event.

Thanks!

jorgepiloto:
All tests passing and @Yash-10 checked this using the validation utilities. Merging... 🚀

jorgepiloto (approved):
Thanks for implementing this 👍🏽

PR #1295, opened 2021-07-26
Change plotting units

### LINKED ISSUE #1291: Change plotting units
💡 **Feature request**

By default, plots are shown with the axis in kilometers, like this one:

plot

And the reason is that units are hardcoded here:

https://github.com/poliastro/poliastro/blob/19880062af49dbf92147528f1b98de378a3d4e89/src/poliastro/plotting/core.py#L186-L190

It would be nice to be able to change those units easily. This can be manually done manipulating the Plotly figure, but it's far from ideal.

A possible solution would be to add a `unit` parameter to the initializer of `_PlotlyOrbitPlotter`, here:

https://github.com/poliastro/poliastro/blob/19880062af49dbf92147528f1b98de378a3d4e89/src/poliastro/plotting/core.py#L17-L18

so that in `_plot_coordinates` we end up with this code:

```python
    def _plot_coordinates(self, coordinates, label, colors, dashed):
        trace = Scatter3d(
            x=coordinates.x.to_value(self.unit),
            y=coordinates.y.to_value(self.unit),
            z=coordinates.z.to_value(self.unit),
```

Notice that this would require modifying also the initializers of `OrbitPlotter3D` and `OrbitPlotter2D`, as well as the corresponding `super` calls.

Patil2099:
Fixes #1291

### DIFF
--- a/src/poliastro/plotting/core.py
+++ b/src/poliastro/plotting/core.py
@@ -14,9 +14,9 @@ from poliastro.plotting.util import generate_sphere
 
 
 class _PlotlyOrbitPlotter(BaseOrbitPlotter):
-    def __init__(self, figure=None, *, num_points=150, plane=None):
+    def __init__(self, figure=None, *, num_points=150, plane=None, unit):
         super().__init__(num_points=num_points, plane=plane)
-
+        self.unit = unit
         self._figure = figure or Figure()
         self._layout = None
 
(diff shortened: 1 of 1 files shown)


src/poliastro/plotting/core.py
```diff
@@ -14,9 +14,9 @@
 
 
 class _PlotlyOrbitPlotter(BaseOrbitPlotter):
-    def __init__(self, figure=None, *, num_points=150, plane=None):
+    def __init__(self, figure=None, *, num_points=150, plane=None, unit):
```
astrojuanlu, line 17: Hi @Patil2099 , I think this syntax is incorrect

astrojuanlu:
Closing this in favor of #1318 because of lack of response.

PR #1297, opened 2021-07-30
Minor event module and other changes

Yash-10:
- Reference for the eclipse was added.
- Docstring for node event was missing earlier, here it is added.

I also wonder if `EclipseEvent` should be made an abstract class since objects of `EclipseEvent` don't have any useful existence, right? In fact, its `__call__` method just returns the position vector of the secondary body, which doesn't seem to be related to any event.

### DIFF
--- a/src/poliastro/ephem.py
+++ b/src/poliastro/ephem.py
@@ -151,8 +151,7 @@ def _get_destination_frame(attractor, plane, epochs):
 class Ephem:
     """Time history of position and velocity of some object at particular epochs.
 
-    Instead of creating Ephem objects directly,
-    the user is encouraged to use the available classmethods.
+    Instead of creating Ephem objects directly, use the available classmethods.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


src/poliastro/ephem.py
```diff
@@ -152,7 +152,7 @@ class Ephem:
     """Time history of position and velocity of some object at particular epochs.
 
     Instead of creating Ephem objects directly,
-    you are encouraged to use the available classmethods.
+    the user is encouraged to use the available classmethods.
```
astrojuanlu, line 155:
```suggestion
    use the available classmethods.
```

?

src/poliastro/twobody/events.py
```diff
@@ -221,9 +221,23 @@ def __call__(self, t, u_, k):
...
+        i.e. event is triggered during both crossings.
+
+    """
+
     def __init__(self, terminal=False, direction=0):
         super().__init__(terminal, direction)
 
     def __call__(self, t, u_, k):
         self._last_t = t
+        # Check if z-coordinate of the satellite is zero or not.
```
astrojuanlu, line 242:
```suggestion
        # Check if the z coordinate of the satellite is zero
```

?

astrojuanlu (commented):
Good improvements! Added a couple of comments

Yash-10:
Thanks for those suggestions!

astrojuanlu (approved):
🚀

PR #1299, opened 2021-08-05
Satellite visibility

Yash-10:
This draft attempt is for satellite visibility with respect to a ground station on the central attractor. Inspired by section 5.4 on rise and set time of a satellite about an oblate planet.

- Currently figuring out the difference between section 5.2.2 and 5.2.5 and which function is the appropriate one.
- Trying to find if the "PREDICT" formulation from Vallado Algorithm 74 would be a simpler choice instead.
- Could check its connection with `is_visible` method of the `SpheroidLocation` class and possible refactoring.

Thanks!


astrojuanlu:
Would you clarify which "section 5.4" are you looking at?

I would _love_ to see this implemented, because we've been trying for 2 years to get it right:

- https://github.com/poliastro/poliastro/pull/719
- https://space.stackexchange.com/q/49136/10716
- https://github.com/satellogic/orbit-predictor/pull/114

I'm eager to understand more about the differences between the method you propose here and the ones we explored earlier.

Yash-10:
I am sorry, I meant section 5.4 from Escobal's book "Methods for orbit determination". Thank you for the above links, the equation in this pull request (as of now) is the same as the equation you wrote in https://space.stackexchange.com/q/49136/10716. Yesterday, it was modified a bit and hence this pull request (as of now) is aligned with **Section 5.4.5**, to be specific.

I would have to go through those links and understand the potential pitfalls you mentioned.

Also, I had found some implementations that "seemed" to be relatively straightforward:

- https://github.com/JuliaSpace/SatelliteToolbox.jl/blob/441470938af978e9d5653a9c4b36ccc107023960/src/analysis/ground_stations.jl#L373-L386
- https://github.com/JuliaSpace/SatelliteToolbox.jl/blob/441470938af978e9d5653a9c4b36ccc107023960/src/analysis/ground_stations.jl#L397-L410

These are the `ground_station_visible` methods.

Yash-10:
The current implementation gives slight errors in the event detection time (ranging from ~1 s to ~2 min) when compared to the orekit's `ElevationDetector`, while for some `lat` and `lon` values of the ground station, the implementation matches.

This could imply the current addition might be closer to being correct, however there are some intricacies that would need investigation for the above slight errors.

Any suggestions would be helpful here!

Yash-10:
As @astrojuanlu pointed out in the meeting to check the ellipsoid that I used for validating the event, I came to know that I had made a foolish error by comparing event times with the flattened earth from orekit, as was done for the other events. Now I changed that:

- Now the 6-7s mismatches, as stated above, do not occur!
- I only seem to observe mismatches for the ground station (lat, lon) cases (in degrees): (45, 45), (45, 46), (46, 45), (47, 45) whereas the times match for all the other cases (at least from the several lat-lon values I checked).

Is there any particular consequence of having both lat and lon values closer to 45 degrees for the ground stations, in general, in orbital mechanics?

Yash-10:
Although a pull request was made: poliastro/validation#35, the deviations with orekit at two (lat, lon) pairs would need some inspection.

PR #1301, opened 2021-08-10
Test numba == 0.54.0rc3

astrojuanlu:
See https://numba.discourse.group/t/numba-0-54-0-rc3/835?u=astrojuanlu


astrojuanlu:
Stable version is out, closing

PR #1308, opened 2021-08-18
Debug widgets on docs

astrojuanlu:
See #1305.

Issue #1305, opened 2021-08-16
Plotly plots aren't shown in the documentation, again


astrojuanlu:
Confirmed by @Yash-10 and @jorgepiloto , the Plotly plots aren't shown on the `latest` docs:

https://docs.poliastro.space/en/latest/examples/Plotting%20in%203D.html

however, the `stable` docs don't have this problem:

https://docs.poliastro.space/en/stable/examples/Plotting%20in%203D.html

Therefore, something changed.

Issue #1132, opened 2021-03-21
Sampling error in tests


Yash-10:
🐞 **Problem**

I have recently started to get a weird error after running `pytest`.

**Traceback**
```
============================================================================== FAILURES ==============================================================================
___________________________________________ test_sample_closed_starts_and_ends_at_min_anomaly_if_in_range_and_no_max_given ___________________________________________

>   ???

tests/tests_twobody/test_sampling.py:53: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/tests_twobody/test_sampling.py:68: in test_sample_closed_starts_and_ends_at_min_anomaly_if_in_range_and_no_max_given
    assert_quantity_allclose(result[-1], min_nu, atol=1e-14 * u.rad)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

actual = <Quantity -3.99680289e-15 rad>, desired = <Quantity 6.2781633e-15 rad>, rtol = 1e-07, atol = <Quantity 1.e-14 rad>, kwargs = {}
np = <module 'numpy' from '/home/yash/poli_astro/poli/lib/python3.8/site-packages/numpy/__init__.py'>
_unquantify_allclose_arguments = <function _unquantify_allclose_arguments at 0x7f9a4d2caaf0>

    def assert_quantity_allclose(actual, desired, rtol=1.e-7, atol=None,
                                 **kwargs):
        """
        Raise an assertion if two objects are not equal up to desired tolerance.
    
        This is a :class:`~astropy.units.Quantity`-aware version of
        :func:`numpy.testing.assert_allclose`.
        """
        import numpy as np
        from astropy.units.quantity import _unquantify_allclose_arguments
>       np.testing.assert_allclose(*_unquantify_allclose_arguments(
[...]

PR #1310, opened 2021-08-18
Fix widgets on docs

astrojuanlu:
See #1305, and discussion in #1308.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,7 @@ cesium = ["czml3 ~=0.5.3"]
 dev = [
     "black==21.5b0",
     "coverage",
+    "docutils<0.17",  # https://github.com/poliastro/poliastro/issues/1305
     "hypothesis",
     "ipython>=5.0",
     "ipywidgets>=7.6",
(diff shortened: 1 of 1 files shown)


astrojuanlu:
Well, I accidentally failed to properly restrict Sphinx on the first commit. The good news is that we can see now if the problem is with docutils or sphinx-hoverxref.

astrojuanlu:
This commit didn't fix the problem. So... everything points to Sphinx 4 now. Incoming.

astrojuanlu:
Confirmed that Sphinx 4.? has broken interactive widgets on docs.

- Failing build: https://readthedocs.org/projects/poliastro/builds/14500050/
- Successful build: https://readthedocs.org/projects/poliastro/builds/14500101/

Will investigate this later. Now I will check if reverting the constraints of the other dependencies still maintains the widgets.

astrojuanlu:
wheels failed because of #1132.

Notebooks are showing properly but other styling things broke, probably because of the newer docutils. So I will revert the last commit and merge.

PR #1323, opened 2021-09-24
Minor `czml` changes

Yash-10:
Trying to make some minor changes that I had in mind but forgot to do earlier.

I think changing `to(...).value` to `to_value(...)` in all modules, wherever they exist, might be made another issue?


Yash-10:
I wonder if this could be removed?:

https://github.com/poliastro/poliastro/blob/b4ea056db0d5eede573d5bbf7d426658b1db2361/src/poliastro/czml/extract_czml.py#L543-L544

Since in `__init__` it is ensured that if the attractor passed is None, then Earth is kept as the default?

astrojuanlu:
Sorry @Yash-10 I missed this one!

> I wonder if this could be removed?

Yeah, looks like it 👍🏽

Yash-10:
Made that change then :+1:

astrojuanlu (approved):
LGTM!

Issue #1209, opened 2021-05-07
Review packaging of sdist with newer flit


astrojuanlu:
See https://github.com/takluyver/flit/issues/409

matthewfeickert:
I think this Issue can get closed now (PR #1329 fixed it but just didn't use one of GitHub's keywords to autoclose). :+1:

astrojuanlu:
Good catch, closing!

PR #1329, opened 2021-10-01
Update flit core version.

Syed-Osama-Hussain:
Addresses [1209](https://github.com/poliastro/poliastro/issues/1209).


astrojuanlu:
Thanks @Syed-Osama-Hussain !

PR #1338, opened 2021-10-11
build: Use PEP 660 compatible flit to enable pip editable installs

### LINKED ISSUE #1337: Replace `flit install --symlink` by `pip install -e .`
Newest pip and flit support PEP 660! 🎉 This means that `pip install -e .` is now standardized across backends.

We should update the instructions here:

https://github.com/poliastro/poliastro/blob/2cfa8e1fea82c68426f0531ad74d2c6fc7098fc4/CONTRIBUTING.md#L112

and the minimum required flit version here:

https://github.com/poliastro/poliastro/blob/2cfa8e1fea82c68426f0531ad74d2c6fc7098fc4/pyproject.toml#L3

matthewfeickert:
Resolves #1337

Require minimum [`flit-core` of `v3.4`](https://pypi.org/project/flit-core/3.4.0/) to allow for support of [PEP 660](https://www.python.org/dev/peps/pep-0660/) which allows for flit to take advantage of `python -m pip --editable .` for editable installs, as compared to `flit install --symlink`. Additionally update the `CONTRIBUTING` docs to reflect this as well.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,7 +109,7 @@ Next, configure your Python environment:
 7. Create a Python virtual environment using `python -m venv .venv`
 8. Activate it using `source .venv/bin/activate`
 9. Upgrade the development dependencies using `python -m pip install -U pip setuptools wheel flit tox`
-10. Install the code in development mode using `flit install --symlink`
+10. Install the code in development mode using `python -m pip install -e .`
     (this means that the installed code will change as soon as you change it in the
     download location).
 


matthewfeickert:
@astrojuanlu I think this is ready for review. I'm not sure what's happening with regards to Codecov's patch coverage, as when I try to look at the Codecov report I get a 400 Error.

I'm happy to iterate on this though if you'd like that resolved first.

PR #1346, opened 2021-10-16
 Fix #1343: Rename change_inc_ecc to change_ecc_inc 

### LINKED ISSUE #1343: change_inc_ecc but parameter are ecc then inc
🐞 **Problem**

The function `change_inc_ecc` expects parameters `ecc` and then `inc`, contrary to what the name may suggest. With keyword-only arguments this would be clear, but without them, the name is misleading.

💡 **Possible solutions**

I suggest to rename the function to `change_ecc_inc` and deprecate `change_inc_ecc`and keep it only for backwards compatibility as a wrapper:

```python
def change_inc_ecc(ss_0, ecc_f, inc_f, f):
    """Only for backwards compatibility. Use change_ecc_inc for new code."""
    return change_ecc_inc(ss_0=ss_0, ecc_f=ecc_f, inc_f=inc_f, f=f)
```

MLopez-Ibanez:


astrojuanlu:
I think there's one rename missing, I see `No module named 'poliastro.core.thrust.change_ecc_inc`

PR #1351, opened 2021-10-20
Update jacchia.py R gas constant unit 

TommasoPino:
correction of R gas constant dimension unit.

### DIFF
--- a/src/poliastro/earth/atmosphere/jacchia.py
+++ b/src/poliastro/earth/atmosphere/jacchia.py
@@ -7,7 +7,7 @@ from poliastro.core.earth_atmosphere.jacchia import (
     _O_and_O2_correction as _O_and_O2_correction_fast,
 )
 
-R = 8314.32 * u.J / (u.kg * u.mol)
+R = 8314.32 * u.J / (u.K * u.kmol)
 k = 1.380622e-23 * u.J / u.K
 
 


astrojuanlu:
Woops, this is pretty bad, thanks a lot @TommasoPino ! However, it looks like this test is failing:

https://github.com/poliastro/poliastro/blob/68bfb05542589a998a46718146db9c981b4e2f3e/tests/tests_earth/tests_atmosphere/test_jacchia77.py#L259-L267

Requesting a review from @jorgepiloto

jorgepiloto:
Thanks for pinging @astrojuanlu. I ran out of time yesterday while reviewing the effects of this single line of code. 

The quantity `R` is only used for computing the density, which I see its being computed using equation (42) from COESA model, see:

Coesa76 density             |  Jacchia77 density
:-------------------------:|:-------------------------:
  |  

Notice that the only difference be tween the two methods is that Jacchia77 divides by Avogadro's constant. I think this comes from the definition of `R_g = R_u / M_g` maybe, not sure why... Anyway, the problem seems to be related with those definitions.

astrojuanlu:
I see! @TommasoPino does that make sense? Would you change your pull request to add a meaningful code comment instead?

astrojuanlu:
@nosudrum analyzed this issue a bit more in #1509

Nosudrum:
I have analysed this in depth in #1509 [here](https://github.com/poliastro/poliastro/issues/1509#issuecomment-1141130732). Hopefully my conclusions are correct and can help fix this :)

astrojuanlu:
I added @TommasoPino commit to #1511 and included the changes @Nosudrum suggested in #1509, so let's continue the conversation there.

astrojuanlu:
The units problem is now fixed 🎉

Issue #1342, opened 2021-10-13
Requiring keyword only arguments for functions with many arguments


MLopez-Ibanez:
Reading the source code (and using the API) would be easier if functions with many arguments used keyword-only arguments for all but the few most typical ones.

Scikit-learn made this transition some time ago and they detailed how they did it: 

https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep009/proposal.html

An example is:

https://github.com/poliastro/poliastro/blob/237647f6db4d3033ac7dc55056323ccf07f86a54/src/poliastro/twobody/orbit.py#L700-L702

It is not easy to see which arguments corresponds to each keyword. For example, nu=arglat but finding out that, requires looking up the signature of the function.

ibLeDy:
I would love to tacke this issue!

What style of formatting should we use for this? 
Should we use a specific formatting like in `contrib/relative:584`[^1] , or just a standard hanging style[^2]?

[^1]: https://github.com/poliastro/poliastro/blob/237647f6db4d3033ac7dc55056323ccf07f86a54/contrib/relative.py#L584-L590

[^2]: ```py
    satC = Orbit.from_classical(
        attractor = Earth,
        a = 6918.140 * u.km,
        ecc = 1e-6 * u.one,
        inc = 10.0 * u.deg,
        raan = 70.0 * u.deg,
        argp = 90.0 * u.deg,
        nu = 1.65 * u.deg
    )
	```

PR #1353, opened 2021-10-20
Use keyword arguments when invoking `from_classical`

ibLeDy:
Omitting invocation from `test_3rd_body_Curtis`[^1] in
`tests\tests_twobody\test_perturbations.py` because it is
parametrized.

- [x] #1342

[^1]: https://github.com/poliastro/poliastro/blob/d0a13af27e5971e3435c9a762942041201ee13a6/tests/tests_twobody/test_perturbations.py#L538


ibLeDy:
Just waiting for the tests to finish, let me know if I should do something about the omitted invocation.

Issue #1349, opened 2021-10-17
Add plot_maneuver


MLopez-Ibanez:
💡 **Feature request**

Currently sample() accepts a min and max anomaly. However, it is not always clear how to set these values. For example, in the [Going to Jupiter example](https://docs.poliastro.space/en/stable/examples/Going%20to%20Jupiter%20with%20Python%20using%20Jupyter%20and%20poliastro.html) we have code like:

```python
plotter.plot_trajectory(
    ic3.sample(min_anomaly=10 * u.deg, max_anomaly=180 * u.deg),
    label="Inner cruise 3",
    color="C3",
)
```

It is not clear why one should set 10 and 180 here. 

In our particular case, we know the start and end time of the maneuver.

Another alternative could be to provide a `plot_maneuver` function that automatically takes care of plotting the whole maneuver but just that, ideally plotting also any intermediate impulses. This way, one could do:

```python
man_flyby = Maneuver.lambert(ic1_end, ss_efly)
plotter.plot_maneuver(man_flyby, label="Inner cruise 1", color="C1")
```

MLopez-Ibanez:
@jorgepiloto Do you know how this could be implemented?

astrojuanlu:
Fixed in #1359

PR #1359, opened 2021-10-22
New plotting maneuver feature

jorgepiloto:
This pull request solves for #1349 by adding a new method to the `BaseOrbitPlotter` and redefining it for the `StaticOrbitPlotter` (following the same approach used by other similar methods). It was also added to the `_PlotlyOrbitPlotter` so the feature is accessible no matter the plotter being used.

Before merging this, the following tasks list needs to be completed:

- [ ] Add the corresponding tests to cover the logic behind this new feature.
- [ ] Simplify examples using the `plot_trajectory` to draw maneuver paths.


jorgepiloto:
Progress on this:

Solarized dark             |  Solarized Ocean
:-------------------------:|:-------------------------:
  |  

I am happy with the result from the static plotters although not so satisfied with the markers from the interactive plotters. What do you think about those,  @astrojuanlu?

jorgepiloto:
And here it is result for the interactive 2D plotters:

<div align="center">

</div>

astrojuanlu:
I think this looks great! After addressing `E741 ambiguous variable name 'l'` this is ready for review I think

astrojuanlu (approved):
Looks good to me!

jorgepiloto:
Oh! I was going to update some of the notebooks with this new feature but saw you just merged it @astrojuanlu. Let me open another PR then 🚀

astrojuanlu:
Woops, sorry! Sure, go ahead 💪🏽

PR #1367, opened 2021-10-28
Add error messages to assert conditions

### LINKED ISSUE #1294: Converting `assert` statements to separate unit tests?
`grep -r src/ -e assert` gives:

```
src/poliastro/frames/fixed.py:            assert type(equatorial_frame) == HCRS
src/poliastro/frames/fixed.py:            assert fixed_coo.body == equatorial_frame.body
src/poliastro/frames/fixed.py:            assert type(equatorial_coo) == HCRS
src/poliastro/frames/fixed.py:            assert equatorial_coo.body == fixed_frame.body
src/poliastro/twobody/propagation.py:    assert np.abs(f * gdot - fdot * g - 1) < 1e-5  # Fixed tolerance
src/poliastro/twobody/orbit.py:        assert np.any(r.value), "Position vector must be non zero"
src/poliastro/core/propagation/farnocchia.py:    assert abs(x) < 1
src/poliastro/core/propagation/farnocchia.py:    assert abs(x) < 1
src/poliastro/core/propagation/farnocchia.py:    assert abs(x) < 1
src/poliastro/core/propagation/farnocchia.py:    assert abs(x) < 1
src/poliastro/core/propagation/farnocchia.py:    assert abs(x) < 1
src/poliastro/core/propagation/farnocchia.py:    assert -np.pi <= nu < np.pi
src/poliastro/core/iod.py:    assert tof > 0
src/poliastro/core/iod.py:    assert k > 0
src/poliastro/core/iod.py:    assert abs(ll) < 1
src/poliastro/core/iod.py:    assert T > 0  # Mistake on original paper
```

Would it be a good choice to remove the above assert statements and instead convert them into their own unit tests?

For example, I had tried to perform a lambert maneuver where the final orbit's epoch was before the initial orbit's epoch where:

```
src/poliastro/core/iod.py:    assert tof > 0
```
was violated. I got an `AssertionError` with no error message. So maybe in such situations, it would be good to explicitly throw an error message? Also, would there be any reason to keep assert statements in `src` sometimes?

Thanks!

Yash-10:
Trying to resolve #1294 :)

### DIFF
--- a/src/poliastro/core/iod.py
+++ b/src/poliastro/core/iod.py
@@ -189,8 +189,8 @@ def izzo(k, r1, r2, tof, M, numiter, rtol):
     """
 
     # Check preconditions
-    assert tof > 0, "tof must be a non-negative value."
-    assert k > 0, "k must be a non-negative value."
+    assert tof > 0
+    assert k > 0
 
     # Check collinearity of r1 and r2
     if not cross(r1, r2).any():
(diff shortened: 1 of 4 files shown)


Yash-10:
It looks like auto-merge couldn't merge this because of codecov/patch, right?

PR #1368, opened 2021-10-28
Added Core and Safe versions for Thrust Modules

### LINKED ISSUE #1355: Thrust functions accept floats rather than quantities
I just realized that `poliastro.twobody.thrust` functions expect floats that get passed directly to `poliastro.core.thrust` functions, like in this case:

https://github.com/poliastro/poliastro/blob/02588dfc3ec4e6d390947f6691a15f8a618e1022/src/poliastro/twobody/thrust/change_a_inc.py#L13-L45

This is inconsistent with our API promises, since these functions should accept quantities with units.

It might be challenging to njit `poliastro.twobody.thrust.change_a_inc.change_a_inc` since it has a closure inside, and that's likely the case for other guidance laws. In any case, we should

1. Provide a safe function that accepts units
2. Move the unsafe function to `poliastro.core.thrust`

TheBuffer:
Tried to fix #1355 . Changed modules change_argp  change_ecc_icc  change_ecc_quasioptimal.


astrojuanlu:
Hi @TheBuffer , thanks for your pull request! However, as @MLopez-Ibanez pointed out, this is a bit big and difficult to review, and also it introduces units under `poliastro.core`, which we don't want. Let's focus on getting #1366 merged first, and after that we can fix one module at a time.

astrojuanlu:
Parts of this have already been addressed in #1366, so there are a lot of conflicts. I think the best will be to close it and start from scratch at some other time. Thanks @TheBuffer for the work! Don't be discouraged from contributing, you can start with a smaller PR next time.

Issue #1362, opened 2021-10-23
Lambert Izzo fails to converge under certain conditions, while Vallado works


MLopez-Ibanez:
🐞 **Problem**

A Lambert maneuver between two orbits fails to converge but it is unclear what is wrong.

The error seems very fickle and trying to manually recreate the orbits seems to avoid the issue, so I saved the orbits and the error to a pickle file. The error can be reproduced with this short code:

```python
from poliastro.maneuver import Maneuver
from poliastro.bodies import Body
from poliastro.twobody import Orbit
from astropy import units as u
import pickle
import gzip
def from_pickle(filename):
    with gzip.open(filename, 'rb') as f:     
        return pickle.load(f)
        
db = from_pickle("two_shot_transfer_proposalj6y1by8g.pkl.gz")
from_orbit = db['from_orbit']
to_orbit = db['to_orbit']
e = db['e']

MU = 1.32712440018e11 # km^3/s^2

Sun = Body(
        parent=None,
        k = MU * u.km ** 3 / u.s ** 2,
        name="Sun")

man = Maneuver.lambert(from_orbit, to_orbit)

```
And this pickle file:

[two_shot_transfer_proposalj6y1by8g.pkl.gz](https://github.com/poliastro/poliastro/files/7402677/two_shot_transfer_proposalj6y1by8g.pkl.gz)

Relevant packages:

```bash
# Paste your output here:
astropy==4.3.1
astroquery==0.4.3
numba==0.54.1
numpy==1.20.0
pandas==1.2.4
scipy==1.7.1
joblib==1.1.0
```

astrojuanlu:
Closed in #1371!

astrojuanlu:
I confirm that the initial guess for the latest case is now `x_0 = 0.9350...`, which is much closer to the actual one! 💪🏽 

I'm sure this has improved the performance of the algorithm as well! Too bad we're not measuring it 😅

PR #1371, opened 2021-10-29
Fix izzo2015 initial guess

jorgepiloto:
Solves #1362 by properly computing the initial guess for Izzo's 2015 solver. This bug was caused because the piecewise equation right after expression (30) in the original paper is wrong.

### DIFF
--- a/src/poliastro/core/iod.py
+++ b/src/poliastro/core/iod.py
@@ -380,6 +380,9 @@ def _initial_guess(T, ll, M):
         else:
             # This is the real condition, which is not exactly equivalent
             # elif T_1 < T < T_0
+            # Corrected initial guess,
+            # piecewise equation right after expression (30) in the original paper is incorrect
+            # See https://github.com/poliastro/poliastro/issues/1362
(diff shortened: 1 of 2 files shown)


src/poliastro/core/iod.py
```diff
@@ -380,7 +380,7 @@ def _initial_guess(T, ll, M):
         else:
             # This is the real condition, which is not exactly equivalent
             # elif T_1 < T < T_0
-            x_0 = (T_0 / T) ** (np.log2(T_1 / T_0)) - 1
+            x_0 = np.exp(np.log(2) * np.log(T / T_0) / np.log(T_1 / T_0)) - 1
```
astrojuanlu, line 383:
```suggestion
            # Corrected initial guess,
            # piecewise equation right after expression (30) in the original paper is incorrect
            # See https://github.com/poliastro/poliastro/issues/1362
            x_0 = np.exp(np.log(2) * np.log(T / T_0) / np.log(T_1 / T_0)) - 1
```

astrojuanlu (approved):
Thanks a lot!

PR #1373, opened 2021-10-30
Fix #1360 vectorize rotation_matrix and fix axis=1

### LINKED ISSUE #1360: rotation_matrix issues
🐞 **Problem**

Perhaps I'm missing some detail but the implementation of `rotation_matrix` looks strange and it doesn't match the one in astropy (but the one of astropy also looks strange).

According to Wikipedia: https://en.wikipedia.org/wiki/Rotation_matrix#Basic_rotations
the rotation for axis=1 should be

```python
[[c, 0.0, s], [0.0, 1.0, 0.0], [-s, 0.0, c]]
```
instead of  https://github.com/poliastro/poliastro/blob/18a390d4f0b33f30b47da45fc7d91186a0a6bc76/src/poliastro/core/util.py#L33

Moreover, the version in astropy: 

https://github.com/astropy/astropy/blob/39cb148a4bd69368f6f30b9abfe32112a42f58e6/astropy/coordinates/matrix_utilities.py#L92-L99
```python
        a1 = (i + 1) % 3
        a2 = (i + 2) % 3
        R = np.zeros(getattr(angle, 'shape', ()) + (3, 3))
        R[..., i, i] = 1.
        R[..., a1, a1] = c
        R[..., a1, a2] = s
        R[..., a2, a1] = -s
        R[..., a2, a2] = c
```
Actually creates:

which probably is due to the fact that "the rotation sense is counterclockwise looking down the + axis (e.g. positive rotations obey left-hand-rule)." But I don't understand the matrix generated by poliastro.

MLopez-Ibanez:


astrojuanlu (approved):
Thanks @MLopez-Ibanez !

PR #1375, opened 2021-11-01
Remove use of units in `poliastro.core`

### LINKED ISSUE #1256: Avoid units in `poliastro.core`
While reviewing @Yash-10  work in #1254, I noticed that the `astropy.units` are being used within the `core` sub-package by the [atmospheric_drag_model function](https://github.com/poliastro/poliastro/blob/3acb5c8b843786fa874f0ea5b620382cd26ac77f/src/poliastro/core/perturbations.py#L131), see:

```python
rho = model.density((H - R) * u.km).to(u.kg / u.km ** 3).value
```

This line of code breaks one of the key points about the `core` sub-package, which is letting user to manage its own units without requiring astropy ones.

The reasons why I claim a refactor on the atmosphere are listed below these lines:

* We can solve for the issue presented before by moving the logic of the atmospheric models to the core.
* Taking advantage of this shipping, we could `jit` the atmospheric functions by refactoring the class approach into a pure function one.

This is tedious task but otherwise, units are being used within core. I would like other project devs and users to discuss a little bit more about this before working on it.

astrojuanlu:
Fix #1256, and enforce it forever using import-linter.

I had to make several other changes along the way:

- Refactor `atmospheric_drag` to take the density as a float rather than a callable. This introduces a minor redundancy because the user needs to compute the height outside, but the advantage is that the perturbation can be jitted, and no units are needed.
- Disable the atmosphere perturbation in `EarthSatellite`. The design of the class is not very flexible and does not support the new `atmospheric_drag` function easily. Rather than trying to find a way out, since these classes are experimental, I temporarily introduced a `NotImplementedError` until we find a better solution.
- Add a `lint-imports` command to our `test` environment in tox, since import-linter requires the package to be installed.

### DIFF
--- a/.importlinter
+++ b/.importlinter
@@ -1,5 +1,9 @@
 [importlinter]
-root_package = poliastro
+# Include Astropy to properly analyze external submodules,
+# see https://github.com/seddonym/import-linter/issues/111
+root_packages =
+    poliastro
+    astropy
 include_external_packages=True
 
 [importlinter:contract:1]
@@ -7,7 +11,5 @@ name=poliastro.core does not import astropy.units
 type=forbidden
 source_modules=
     poliastro.core
-# astropy.units is not supported,
-# see https://github.com/seddonym/import-linter/issues/111
 forbidden_modules=
-    astropy
+    astropy.units
--- a/.importlinter
+++ b/.importlinter
@@ -1,0 +1,13 @@
+[importlinter]
+root_package = poliastro
+include_external_packages=True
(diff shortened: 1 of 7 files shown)


astrojuanlu:
Also, notice I ended up preventing the whole astropy in `poliastro.core`, see https://github.com/seddonym/import-linter/issues/111

astrojuanlu:
Found a workaround for import-linter, merging!

PR #1376, opened 2021-11-01
Using `to_value` instead of `to(...).value`

Yash-10:
I think we would agree upon this, it's easier to write and slightly faster too.

Thanks!

### DIFF
--- a/src/poliastro/core/earth_atmosphere/util.py
+++ b/src/poliastro/core/earth_atmosphere/util.py
@@ -93,7 +93,7 @@ def _get_index(x, x_levels):
     """
     for i, value in enumerate(x_levels):
         if i < len(x_levels) and value < x:
-            pass
+            continue
         elif x == value:
             return i
         else:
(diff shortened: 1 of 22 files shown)


src/poliastro/core/earth_atmosphere/util.py
```diff
@@ -93,7 +93,7 @@ def _get_index(x, x_levels):
     """
     for i, value in enumerate(x_levels):
         if i < len(x_levels) and value < x:
-            pass
+            continue
```
astrojuanlu, line 96: ✔️

astrojuanlu (approved):
Thanks a lot! Must have taken some time to go through all these :)

src/poliastro/threebody/flybys.py
```diff
@@ -34,11 +34,11 @@ def compute_flyby(v_spacecraft, v_body, k, r_p, theta=0 * u.deg):
         Turn angle.
 
     """
-    v_spacecraft = v_spacecraft.to(u.km / u.s).value
-    v_body = v_body.to(u.km / u.s).value
-    k = k.to(u.km ** 3 / u.s ** 2).value
-    r_p = r_p.to(u.km).value
-    theta = theta.to(u.rad).value
+    v_spacecraft = v_spacecraft.to_value(u.km / u.s)
+    v_body = v_body.to_value(u.km)
```
astrojuanlu, line 38: Woops!

```suggestion
    v_body = v_body.to_value(u.km / u.s)
```

Yash-10: :sweat_smile:

PR #1378, opened 2021-11-01
Fix #1377 plot_maneuver with single impulse.

### LINKED ISSUE #1377: Bug in plot_maneuver
This cannot be correct:

https://github.com/poliastro/poliastro/blob/80c7de132c42c751d47fbb495dc54fb3e3460f12/src/poliastro/plotting/_base.py#L171-L173

If the len() is zero, then one cannot index it.

MLopez-Ibanez:
I also updated the Jupiter example  to exercise the new function. This is how I discovered the bug.

### DIFF
--- a/docs/source/examples/Going to Jupiter with Python using Jupyter and poliastro.mystnb
+++ b/docs/source/examples/Going to Jupiter with Python using Jupyter and poliastro.mystnb
@@ -4,7 +4,7 @@ jupytext:
     extension: .mystnb
     format_name: myst
     format_version: 0.13
-    jupytext_version: 1.11.1
+    jupytext_version: 1.13.0
 kernelspec:
   display_name: Python 3
   language: python
@@ -93,6 +93,7 @@ ss_e_ic1 = Orbit.from_ephem(Sun, earth, epoch=ic1_end.epoch)
 # We can check new bodies positions
 plotter = StaticOrbitPlotter()
 plotter.plot_body_orbit(Earth, ic1_end.epoch, label="Earth at end of flyby")
+plotter.plot_maneuver(ss_e0, man, color="C1", label="Initial impulse")
 plotter.plot_trajectory(
     ic1_end.sample(min_anomaly=0 * u.deg, max_anomaly=180 * u.deg),
     label="Inner cruise 1 full orbit",
@@ -139,13 +140,13 @@ plotter.plot_body_orbit(
     label="Earth's position at flyby",
     trail=True,
 )
+plotter.plot_maneuver(ss_e0, man, color="C1", label="Initial impulse")
 plotter.plot_trajectory(
     ic1.sample(max_anomaly=180 * u.deg), label="Inner cruise 1", color="C1"
 )
-plotter.plot_trajectory(
-    ic2_end.sample(min_anomaly=180 * u.deg, max_anomaly=400 * u.deg),
-    label="Inner cruise 2 full orbit",
-    color="C2",
+plotter.plot_maneuver(ic1_end, man_flyby,
+        label="Inner cruise 2 maneuver",
+        color="C2",
 );
 ```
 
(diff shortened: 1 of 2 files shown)


astrojuanlu:
Requesting a review from @jorgepiloto , because I think he intended to update the notebook as well

docs/source/examples/Going to Jupiter with Python using Jupyter and poliastro.mystnb
```diff
@@ -188,20 +189,23 @@ plotter.plot_body_orbit(
...
-    color="C3",
+plotter.plot_maneuver(ic2_end, man_jupiter,
+        label="Inner cruise 3",
+        color="C3",
 )
 plotter.plot_body_orbit(
     Jupiter, date_arrival, label="Jupiter's position at arrival", trail=True
 );
 ```
+
```
astrojuanlu, line 208: We can remove this trailing empty cell

src/poliastro/plotting/_base.py
```diff
@@ -42,7 +42,7 @@ def plane(self):
     def _set_attractor(self, attractor):
         if self._attractor is None:
             self._attractor = attractor
-        elif attractor is not self._attractor:
+        elif attractor is not self._attractor and attractor != self._attractor:
```
astrojuanlu, line 45: Do we need to check for identity + equality? Bodies are supposed to be singletons so in my mind the identity check should be enough, but perhaps I'm missing something?

src/poliastro/plotting/_base.py
```diff
@@ -170,7 +170,7 @@ def _plot_maneuver(
 
         if len(maneuver_phases) == 0:
             # For single-impulse maneuver only draw the impulse marker
-            self._draw_impulse(color, f"Impulse - {label}", maneuver_phases[0].r)
+            return ([self._draw_impulse(color, f"Impulse - {label}", final_phase.r)],)
```
astrojuanlu, line 173: Right, we need a `return`, otherwise the `coordinates = concatenate_representations(coordinates_list)` line below fails.

astrojuanlu (commented):
Left some comments

src/poliastro/plotting/_base.py
MLopez-Ibanez: This was something I had in my tree. I didn't want to submit it with this PR. However, I had to add it to my tree because we load some orbits from a pickle file and in that case the id() does not match but the objects are equal. The error otherwise are very strange.

docs/source/examples/Going to Jupiter with Python using Jupyter and poliastro.mystnb
MLopez-Ibanez: OK

src/poliastro/plotting/_base.py
MLopez-Ibanez: Also, if there is only one impulse, there is no line to draw.

src/poliastro/plotting/_base.py
astrojuanlu: Oh, I see how that could go wrong. We can discuss this in a separate PR I think.

astrojuanlu: I'd rather revert this change and seek a more robust solution, see https://github.com/poliastro/poliastro/issues/1395

astrojuanlu (changes requested):
Reworking this. I understand that the original point of having a `plot_maneuver` was to avoid having to sample the orbits between anomalies, so I'm trying to figure out if that's possible, taking the Jupiter Jupyter (🥁) notebook as an example.

astrojuanlu:
Oh, I get it now. Still working on it.

astrojuanlu (approved):
I ended up not changing anything, the code looks great to me :+1: I took the time to add better comments on the code to understand how we got there, and also reformatted the notebook. Let's get this merged!

astrojuanlu:
Woops, @jorgepiloto just noticed that all the crosses :x: for the maneuvers are gone. Hope we can restore this soon.

MLopez-Ibanez:
It was working in my version. Did something change from my branch?

astrojuanlu:
I'm not sure, but the rendered plots on the documentation appear without crosses 🤔 @jorgepiloto  is on it

jorgepiloto:
After building the docs from a clean environment, I can confirm that not `x` markers are displayed when drawing the impulses, debugging this... 🔎

jorgepiloto:
The current `_draw_marker` function is not adding neither the `makrer`, `name` nor the `size` arugment:

https://github.com/poliastro/poliastro/blob/c82e492183bdfde689b947bd2b1a0efe55864e12/src/poliastro/plotting/static.py#L80-L89

I will continue the discussion in a new PR as there are a couple of thing I want to improve related with the legend labels order.

astrojuanlu:
Also, mew=0 means no border, right?

MLopez-Ibanez:
There is another issue that may be related: If you do not specify a color to plot_maneuver, then points and lines have different colors. They should have the same color.

PR #1328, opened 2021-10-01
Remove codeclimate.yml.

Syed-Osama-Hussain:
Resolves [1316](https://github.com/poliastro/poliastro/issues/1316).

PR #1379, opened 2021-11-03
Remove codeclimate from README

Yash-10:
Codeclimate was removed in #1328 but not from README. Maybe it's a good choice to remove it from README as well?


astrojuanlu (approved):
Yep! Thanks!

PR #1388, opened 2021-11-04
Fix #421 newton factory for specialized Newton method.

### LINKED ISSUE #421: Generate fast Newton solvers
🐞 **Problem**

numba 0.38 shipped the capability to pass a jitted function to another jitted function as an argument. Unfortunately, it turns out it's slow (https://github.com/numba/numba/issues/2952), which means that we can't reuse code.

However, I discovered that there's a workaround: using closures to generate fast functions. For example, we could have something like this:

```
def newton_generator(func, fprime):
    @njit
    def njit_newton_wrapper(x0):
        for _ in range(50):
            fder = fprime(x0)
            fval = func(x0)
            newton_step = fval / fder
            x = x0 - newton_step
            if abs(x - x0) < 1.48e-8:
                return x
            x0 = x

    return njit_newton_wrapper

newton_func = newton_generator(func, fprime)
```

and this is going to be as fast as it gets. For example, we could replace these pieces of code:

https://github.com/poliastro/poliastro/blob/3295282794f5c710a3593b7fd9469b04d49fd421/src/poliastro/core/angles.py#L122-L140
https://github.com/poliastro/poliastro/blob/3295282794f5c710a3593b7fd9469b04d49fd421/src/poliastro/core/angles.py#L68-L76

By something like:

```
_newton_elliptic = newton_generator(_kepler_equation, _kepler_equation_prime)
_newton_parabolic = newton_generator(_kepler_equation_parabolic, _kepler_equation_prime_parabolic)
_newton_hyperbolic = newton_generator(_kepler_equation_hyper, _kepler_equation_prime_hyper)

@jit
def M_to_E(M, ecc):
    E = _newton_elliptic(M, args=(M, ecc))
    return E
```

Which still doesn't get the 100 % reusability goal, but gets us closer.

🎯 **Goal**

Make core more reusable, and bring us closer to the moment when numba will properly support functions as arguments.

MLopez-Ibanez:


astrojuanlu (approved):
Nothing broke, so let's get this merged! Thanks @MLopez-Ibanez 🚀

Issue #1013, opened 2020-09-07
Converting mean anomaly to true anomaly


s-m-e:
`Orbit.from_classical` "only" accepts a true anomaly. It lacks support for a mean anomaly as an alternative. poliastro also lacks a method for converting from mean anomaly to true anomaly in its [`angle` module](https://github.com/poliastro/poliastro/blob/0.14.x/src/poliastro/core/angles.py).

💡 **Feature request**

A good, working and MIT-licenced implementation for converting from mean to true anomaly can be found [here](https://github.com/RazerM/orbital/blob/73437b627bb2f816612bd4e9c39010483d3c1a29/orbital/utilities.py#L73). It could be adopted and accelerated with numba.

🎯 **Goal**

Virtually every database for solar system objects' orbital elements provides mean anomalies.

astrojuanlu:
> poliastro also lacks a method for converting from mean anomaly to true anomaly in its angle module.

This confinement, mean anomalies have become a favourite topic of mine :smiling_imp: Take a look at this comment:

https://github.com/poliastro/poliastro/pull/908#issuecomment-615889854

In short: the "mean anomaly" is by no means (pun intended) a well defined quantity with a clear geometrical or physical meaning (no pun intended) _for all possible eccentricity values_, and therefore generic conversions should be avoided.

In fact, we _used_ to have `M_to_nu` and `nu_to_M` functions! But I ended up removing them after this realization.

However, if you are analyzing asteroid trajectories and know for real their eccentricity, you can use the well known `M_to_E` and `E_to_nu` that are available in `poliastro.core.angles`, as it's done in these two places in poliastro already:

https://github.com/poliastro/poliastro/blob/3d85e79ea67ec3bf8a575243a7a27de3728940a9/src/poliastro/neos/dastcom5.py#L370-L378

https://github.com/poliastro/poliastro/blob/c516ac476efa00353649077ed63162148cd617ec/src/poliastro/twobody/orbit.py#L683-L691

Let me know what you think.

astrojuanlu:
> Virtually every database for solar system objects' orbital elements provides mean anomalies.

I also observed that, and none of those databases cares to define what's the mean anomaly for the non elliptic case for them, or how near-parabolic cases should be handled (there's a singularity in the vicinity of ecc = 1).

s-m-e:
[...]

PR #1398, opened 2021-11-13
Add M_to_nu_elliptic

MLopez-Ibanez:


astrojuanlu (changes requested):
Not sure about what to do with this pull request, for the reasons stated below.

On one hand, as I stated in #1013, we don't have generic "M from/to nu" _on purpose_ since, as I painfully discovered in https://github.com/poliastro/poliastro/pull/908#issuecomment-615889854, the mean anomaly is far from being a well-defined quantity with a clear geometrical or physical meaning for all eccentricity regimes. Having a `M_to_nu_elliptic` invites the question of "why not having `M_to_nu_{parabolic,hyperbolic}` as well. In any case, we need to document this better, both for users and for potential contributors.

On the other hand, `M_to_nu_elliptic` is not a generic converter, but is only meant to be used for elliptic orbits. There is no eccentricity check though.

And finally, you mention that this is "slightly faster" than calling `E_to_nu(M_to_E())`, probably because of similar reasons as the ones you stated in https://github.com/poliastro/poliastro/issues/1360#issuecomment-955668446. But can we have a measurement of how much faster it is, so we can have an idea of the performance gain?

If we were to merge this, I'd vote to:

- Move a function to a different module, like `poliastro.util`, to make it clear that this is a convenience function.
- Add a short sentence to the docstring pointing out that parabolic and hyperbolic versions are not provided on purpose, and linking to https://github.com/poliastro/poliastro/issues/1013
- Make it raise a `ValueError` if the eccentricity is out of range.

What do you think?

MLopez-Ibanez:
```python
from poliastro.twobody.angles import M_to_nu_elliptic, E_to_nu, M_to_E
%timeit [ M_to_nu_elliptic(M=x * u.deg, ecc=0.1 * u.one) for x in range(0,180)]
%timeit [ E_to_nu(M_to_E(M=x * u.deg, ecc=0.1 * u.one), ecc=0.2 * u.one)  for x in range(0,180)]
```

```python
In [2]: %timeit [ M_to_nu_elliptic(M=x * u.deg, ecc=0.1 * u.one) for x in range(0,180)]
52.8 ms ± 3.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [3]: %timeit [ E_to_nu(M_to_E(M=x * u.deg, ecc=0.1 * u.one), ecc=0.2 * u.one)  for x in range(0,180)]
117 ms ± 19.6 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

```

MLopez-Ibanez:
>   * Add a short sentence to the docstring pointing out that parabolic and hyperbolic versions are not provided on purpose, and linking to #1013

I didn't provide them because I didn't use them. But I noticed that poliastro itself does convert from M to nu for parabolic and hyperbolic in at least two places:

https://github.com/poliastro/poliastro/blob/01a12df75a62a079d2e1bf399052b3a964c3c32c/src/poliastro/core/propagation/farnocchia.py#L267

astrojuanlu:
Yep, we are forced to do that because 2 public NASA APIs use M as the anomaly for potentially parabolic and hyperbolic objects as well. I'm sure there's a good reason for it, but I'm not sure which.

astrojuanlu:
Ah sorry, that belongs to the propagator itself (didn't have my morning coffee yet).

I wonder if the performance gain also happens if the caller is also jitted?

astrojuanlu:
> Yep, we are forced to do that because 2 public NASA APIs use M as the anomaly for potentially parabolic and hyperbolic objects as well. I'm sure there's a good reason for it, but I'm not sure which.

To be clear, the ambiguity appears mainly in the near-parabolic region (not so uncommon for comets), and my beef is with calling M "an angle" for non-elliptic orbits.

MLopez-Ibanez:
> I wonder if the performance gain also happens if the caller is also jitted?

I would assume that if numba does inlining (or when numba does inlining), then there should not be any measurable difference when using just the core versions. But the caller code in orbit.py is not using the core versions.

MLopez-Ibanez:
>     * Move a function to a different module, like `poliastro.util`, to make it clear that this is a convenience function.

Sure. Should the core version also move to poliastro.core.util ?
 
>     * Add a short sentence to the docstring pointing out that parabolic and hyperbolic versions are not provided on purpose, and linking to #1013

What should the sentence say?

>     * Make it raise a `ValueError` if the eccentricity is out of range.

What would be an acceptable range? This is something that could be done for the non-core version for extra safety. However, note that M_to_E does not have such a check (nor any other M_to_X function).

Issue #1364, opened 2021-10-24
Easier way of generating a trajectory from an orbit between two instants in time


astrojuanlu:
In the early days, the `Orbit.sample()` method had a lot of complexity and could take a lot of different inputs. This was cool but also difficult to maintain, and opened the door to overusing it. There is more background in #465, and we removed most of the functionality in #555.

However, there is a very common use case with orbits:

> As a user, I want to retrieve the trajectory described by an object between two periods of time.

At the moment, there is no `Orbit` method capable of doing that:

- `Orbit.sample` can only work with anomalies (see above).
- `Orbit.propagate` returns another `Orbit` with the new time, but not the time history between them.

Our go-to answer for this has been to use the `poliastro.twobody.propagation.propagate` function, which takes an array of times and returns a `CartesianRepresentation` object. It is often exactly what users need, and is more or less what `Orbit.sample` used to do under the hood (not anymore since #1048).

However, there are two problems I think we should solve:

- It's very difficult to discover the `poliastro.twobody.propagation.propagate` function. We have spoiled our users and they expect everything to be an `Orbit.method`, which I'd call success 😃 and it would be cool to support it without having to import an extra module. In fact, even IBM got it wrong https://github.com/IBM/spacetech-ssa/issues/16.
- As discovered while working in #1359, generating a `time_range()` and using it with `poliastro.twobody.propagation.propagate` has the classic problems of uneven point density for highly eccentric orbits. `poliastro.twobody.sampling.sample_closed` does the right thing and generates an array of true anomalies, but to retrieve the corresponding timestamps or elapsed times, one would have to repeatedly call `poliastro.core.propagation.farnocchia.delta_t_from_nu` on those.
  - **Moreover**, `delta_t_from_nu` would convert again those values to eccentric anomaly, doing an unnecessary two-way conversion (since `sample_closed` samples linearly on the eccentric anomaly).

And some non-goals as well:

- As @jorgepiloto noted, animations require uniform time sampling, rather than our smart anomaly sampling. But I think this use case is well served by `poliastro.twobody.propagation.propagate` already, and niche enough to not bring another `Orbit` method.

I'm open to ideas here. One option could be to:

1. Add a `Orbit.sample_anomalies` method that does what `Orbit.sample` does now,
2. Add a `Orbit.sample_times` method that receives minimum and maximum times (so that users don't have to call `time_range` themselves) and that generates a nice time array that avoids oversampling at pericenter,
[...]

PR #1401, opened 2021-11-21
Some plot_manuever improvements

jorgepiloto:
This expands the work done by @MLopez-Ibanez in #1378 by adding:

- [x] Proper cross markers representing the impulse trajectories.
- [x] Legend order according to maneuver impulses execution.
- [ ] Check behavior when no color is provided


jorgepiloto:
The impulse markers are now properly displayed, see [the draft docs for this PR](https://poliastro--1401.org.readthedocs.build/en/1401/examples/Going%20to%20Jupiter%20with%20Python%20using%20Jupyter%20and%20poliastro.html).

However, check the last figure of the notebook: the impulses' labels appear before the trajectory one. I would like to change this  behavior so that the label elements are ordered according to impulse execution.

jorgepiloto:
I fixed the legend's labels order. For achieving this, I made use of a dictionary relating the label and each one of the lines to be displayed. Not sure if other type of object such us `Tuple` one would be more interesting...  

**Non-ordered legend** | **Ordered legend**
:------------------------------:|:-------------------------:
man_disorder | man_order

jorgepiloto:
I deleted some comments made by @MLopez-Ibanez within the code as it would be interesting to have those as an open-issue. I prefer fixing plotting issues for the moment and care later about performance.

astrojuanlu:
The results look good so far! Perhaps I'd made the crosses a little big bigger, so they're visible if there's a dot at the same position.

Maybe the legend could be achieved with a tuple, yes. Worth trying it out before reviewing.

And it was me who made those comments 😛 We are tracking #1364, but can we keep those? I think they're good to communicate intent.

jorgepiloto:
The last problem to be solved is the colors applied by `matplotlib` when no color is provided by user. As an example, see the figure attached below this lines:

<div align="center">

</div>

The current code makes a call to the `axes.plot` function and, since no colors are provided, various are introduced by `matplotlib`.

jorgepiloto:
Sure! Let me apply the suggested changes 👍🏽

astrojuanlu:
> The last problem to be solved is the colors applied by matplotlib when no color is provided by user.

I wonder what would be the desired behavior here?

jorgepiloto:
> > The last problem to be solved is the colors applied by matplotlib when no color is provided by user.
> 
> I wonder what would be the desired behavior here?

Should we leave the random color behavior so that when users get the plot they realize providing a color is a good idea?

src/poliastro/plotting/_base.py
```diff
@@ -170,13 +168,27 @@ def _plot_maneuver(
...
+                # Plot the impulse marker and collect its label and lines
+                impulse_label = f"Impulse {ith_impulse + 1} - {label}"
+                impulse_lines = (
+                    [self._draw_impulse(color, impulse_label, orbit_phase.r)],
+                )
+                lines_list.append((impulse_label, impulse_lines))
+
+                # HACK: if no color is provided, get the one randomly generated
+                # for previous impulse lines
+                color = impulse_lines[0][0].get_color() if color is None else color
```
jorgepiloto, line 190: This line of code executes every time, reducing the performance...

jorgepiloto:
I improved the size of the new markers. Here is the result:

<div align="center">

</div>

src/poliastro/plotting/_base.py
astrojuanlu: I'm not concerned about the performance of plotting. About your question:

> Should we leave the random color behavior so that when users get the plot they realize providing a color is a good idea?

If `plot_maneuver` plots both markers _and_ lines, I think they should have the same color. If the user specifies a color, then that one should be used. If not, then then color that matplotlib chooses should be used, both for markers and lines. (This is not random, but the next color in the cycle - codenamed "C0", "C1", "C2", ...)

Does this make sense?

jorgepiloto: Sure! Then, previous code accomplishes the suggested task 👍🏽

astrojuanlu: Awesome 💪🏽

astrojuanlu (approved):
This looks perfect, thanks a lot @jorgepiloto !

Issue #1405, opened 2021-11-24
Address astroquery deprecation warning


astrojuanlu:
```
tests/tests_twobody/test_orbit.py::test_can_set_iss_attractor_to_earth
  /home/circleci/repo/miniconda3/envs/poliastro/lib/python3.8/site-packages/astroquery/jplhorizons/core.py:113: AstropyDeprecationWarning:
  
  ``id_type``s 'majorbody' and 'id' are deprecated and replaced with ``None``, which has the same functionality.

```

discovered here: https://github.com/numba/numba-integration-testing/pull/77

It's not a bug yet, but it will be very soon.

astrojuanlu:
Closed in #1409

PR #1409, opened 2021-11-29
Resolving astroquery deprecation warnings

Yash-10:
From #1405. I think this would be a small change to set the default `id_type` to `None` as done upstream.

Thanks!


astrojuanlu (approved):
I confirm that the warning is gone, thanks a lot!

Issue #1397, opened 2021-11-12
`Maneuver.lambert` sometimes generates negative times of flight and crashes


astrojuanlu:
Example:

```
In [25]: Maneuver.lambert(gto2, geo)
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-25-f36ab89d597a> in <module>
----> 1 Maneuver.lambert(gto2, geo)

~/Projects/LSF/poliastro/library/src/poliastro/maneuver.py in lambert(cls, orbit_i, orbit_f, method, short, **kwargs)
    188 
    189         # Compute all possible solutions to the Lambert transfer
--> 190         sols = list(method(k, r_i, r_f, tof, **kwargs))
    191 
    192         # Return short or long solution

~/Projects/LSF/poliastro/library/src/poliastro/iod/izzo.py in lambert(k, r0, r, tof, M, numiter, rtol)
     44     sols = izzo_fast(k_, r0_, r_, tof_, M, numiter, rtol)
     45 
---> 46     for v0, v in sols:
     47         yield v0 << kms, v << kms

~/Projects/LSF/poliastro/library/src/poliastro/core/iod.py in izzo()
    190 
    191     # Check preconditions
--> 192     assert tof > 0
    193     assert k > 0
    194 

AssertionError: 
```

To reproduce:

```
import matplotlib.pyplot as plt
from astropy import units as u
from poliastro.bodies import Earth
from poliastro.twobody import Orbit
from poliastro.maneuver import Maneuver
from poliastro.plotting.static import StaticOrbitPlotter
geo = Orbit.circular(Earth, 36_000 * u.km)
man = Maneuver.impulse([0, 1, 0] * u.km / u.s)
gto = Orbit.circular(Earth, 36_000 * u.km, arglat=-180 * u.deg).apply_maneuver(man).propagate_to_anomaly(0 * u.deg)

fr = StaticOrbitPlotter()
fr.plot(geo)
fr.plot(gto)
plt.show()

Maneuver.lambert(gto, geo)
```

The reason for that is in the epochs of the orbits:

```
In [5]: gto.epoch.isot
Out[5]: '2000-01-01T18:16:27.067'

In [6]: geo.epoch.isot
Out[6]: '2000-01-01T12:00:00.000'
```

We should give a better error message rather than failing inside an assertion.

TheBuffer:
Hello, I am new to this Project so please correct me if I'm wrong.
[...]

PR #1413, opened 2021-12-05
raised error with message for negative tof value

TheBuffer:
#1397 Added message to negative tof assert error

### DIFF
--- a/src/poliastro/core/iod.py
+++ b/src/poliastro/core/iod.py
@@ -189,6 +189,7 @@ def izzo(k, r1, r2, tof, M, numiter, rtol):
     """
 
     # Check preconditions
+    assert tof > 0
     assert k > 0
 
     # Check collinearity of r1 and r2
--- a/src/poliastro/core/iod.py
+++ b/src/poliastro/core/iod.py
@@ -189,7 +189,6 @@ def izzo(k, r1, r2, tof, M, numiter, rtol):
     """
 
     # Check preconditions
-    assert tof > 0
     assert k > 0
 
     # Check collinearity of r1 and r2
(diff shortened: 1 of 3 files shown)


astrojuanlu (changes requested):
Thanks for the pull request @TheBuffer!

However, this makes the change only in the `izzo` algorithm (while we also have `vallado`). And besides, the concept of "orbit epochs" doesn't exist at this low level function, it's only the `tof`, so I think the error message might be misleading.

What about moving the check upwards, to the `Maneuver.lambert` classmethod?

astrojuanlu:
Also, this will need a test.

src/poliastro/maneuver.py
```diff
@@ -185,6 +185,8 @@ def lambert(cls, orbit_i, orbit_f, method=lambert_izzo, short=True, **kwargs):
 
         # Time of flight is solved by subtracting both orbit epochs
         tof = orbit_f.epoch - orbit_i.epoch
+        if tof < 0:
+            raise ValueError("Time of Flight must be positive")
```
astrojuanlu, line 189: Can we have a more explanatory error message here? Mentioning the orbit epochs (like you initially did, but now we are on the right function)

src/poliastro/core/iod.py
```diff
@@ -189,7 +189,6 @@ def izzo(k, r1, r2, tof, M, numiter, rtol):
     """
 
     # Check preconditions
-    assert tof > 0
```
astrojuanlu, line 192: Can we keep this assert?

astrojuanlu (changes requested):
Just a couple of minor things and we're ready!

src/poliastro/maneuver.py
TheBuffer: Sure!

astrojuanlu (approved):
Thanks a lot and congratulations for your first pull request! Merging

PR #1418, opened 2021-12-09
Add contrib code to load GP data from Celestrak

astrojuanlu:
As well as other style changes to `contrib/` code.

I'm so happy I don't have to find the right API or module for this, just dump some useful code to a file and move on 😀


astrojuanlu:
Looks like it's working correctly on my tests. Merging :+1:

Issue #1315, opened 2021-08-27
[0.15.2] Test failures on 32-bit systems


olebole:
🐞 **Problem**

I am currently updating the poliastro package on Debian. On 32-bit systems (i385, armhf), I get a number of failures due to missing parallel support, like:

```
___________________ test_plot_sets_attractor[OrbitPlotter2D] ___________________

plotter_class = <class 'poliastro.plotting.core.OrbitPlotter2D'>

    @pytest.mark.parametrize("plotter_class", [OrbitPlotter2D, OrbitPlotter3D])
    def test_plot_sets_attractor(plotter_class):
        frame = plotter_class()
        assert frame._attractor is None
    
>       frame.plot(iss)

tests/tests_plotting/test_core.py:52: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3/dist-packages/poliastro/plotting/core.py:397: in plot
    return super().plot(orbit, label=label, color=color, trail=trail)
/usr/lib/python3/dist-packages/poliastro/plotting/core.py:75: in plot
    super().plot(orbit, label=label, color=color, trail=trail)
/usr/lib/python3/dist-packages/poliastro/plotting/_base.py:252: in plot
    self._plot(orbit, label=label, color=color, trail=trail)
/usr/lib/python3/dist-packages/poliastro/plotting/_base.py:156: in _plot
    coordinates = orbit.sample(self._num_points)
/usr/lib/python3/dist-packages/poliastro/twobody/orbit.py:1421: in sample
    rr, vv = coe2rv_many(
[…]
/usr/lib/python3/dist-packages/numba/parfors/parfor_lowering.py:43: in _lower_parfor_parallel
    ensure_parallel_support()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    def ensure_parallel_support():
        """Check if the platform supports parallel=True and raise if it does not.
        """
        if config.IS_32BITS:
            msg = ("The 'parallel' target is not currently supported on 32 bit "
                   "hardware.")
>           raise errors.UnsupportedParforsError(msg)
E           numba.core.errors.UnsupportedParforsError: Failed in nopython mode pipeline (step: nopython mode backend)
[...]

PR #1426, opened 2021-12-13
Mark tests that potentially fail

olebole:
For the new Debian package, I found two tests that fail under some circumstances:

* test_apply_manuever_correct_plane wants to retrieve some remotel data for Ceres,
* test_plot_maneuver requires the numba "parallel" extension nor-existent on 32-bit (#1315) 

This PR marks them properly.

### DIFF
--- a/tests/test_maneuver.py
+++ b/tests/test_maneuver.py
@@ -225,6 +225,7 @@ def test_correct_pericenter_ecc_exception():
     )
 
 
+@pytest.mark.remote_data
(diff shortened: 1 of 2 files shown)


astrojuanlu:
Thank you @olebole ! Is this worth backporting to a 0.16.1 release if we ever make it?

olebole:
I will apply this as a patch to the 0.16 sources when building the Debian package, so I don't need it myself.

Issue #1215, opened 2021-05-09
Document how to handle TLEs with poliastro


astrojuanlu:
On #1185 it became clear that we should better document how to properly handle TLEs and other General Perturbations orbital data, like OMMs, in poliastro. By the way, `tle` is among the most searched term in our documentation, and it's not mentioned even once❗

![Screenshot from 2021-05-10 00-39-38](https://user-images.githubusercontent.com/316517/117589165-39b67880-b128-11eb-8a67-a10dd9301ff2.png)

Ideally, we should write a How-to (more practical) and a Background section (with an expanded version of this explanation https://github.com/poliastro/poliastro/issues/1185#issuecomment-830601802)

astrojuanlu:
An `Ephem.from_tle` would be perfect!!

astrojuanlu:
Added some notes about potential APIs https://github.com/poliastro/poliastro/wiki/Notes-about-TLEs

astrojuanlu:
Started to add some of these ideas to https://github.com/poliastro/poliastro/pull/1418 and dumped more notes to the wiki

astrojuanlu:
Closing this issue, which was about _documenting_ TLEs in poliastro. Will open a new one for actually adding relevant APIs, which will take a bit more effort.

PR #1428, opened 2021-12-19
Add notebook on OMM and TLE data

astrojuanlu:
Anticipated Christmas present 🎅🏽 See #1215. It's not an actual implementation of a nice API in poliastro, but at least it documents the proper way of doing things.


astrojuanlu:
🤖 Rendered result: https://poliastro--1428.org.readthedocs.build/en/1428/examples/Loading%20OMM%20and%20TLE%20satellite%20data.html

astrojuanlu:
Given that "TLE" is always in the top-4 search queries, and probably the one that returns the most useless result... I hope our users find this much more informative.

Screenshot 2021-12-19 at 02-27-59 Search Analytics Read the Docs

Issue #1425, opened 2021-12-12
Simplify `poliastro.twobody.orbit`


astrojuanlu:
The `Orbit` class in the `poliastro.twobody.orbit` module is a bit of a ["God object"](https://en.m.wikipedia.org/wiki/God_object) and the code of the module is too long. We should try to find ways to simplify it.

At the moment we have:

- 100 lines of imports
- 100 lines of property delegation with States + extra properties
- 800 lines of classmethods
- 500 lines of methods

Therefore, one such way is moving the logic of the classmethods (which are [syntax sugar](https://en.m.wikipedia.org/wiki/Syntactic_sugar)  in the end) to funcions living in a separate module. This would probably result in some docstring duplication, but still worth it.

astrojuanlu:
Partially addressed in #1430. The module is still long, but seems hard to avoid it. I'm closing this issue as solved ✔️

PR #1430, opened 2021-12-20
`poliastro.twobody.orbit` cleanup

astrojuanlu:
Generally working towards #1425, but did a lot of other things along the way. Each commit could be reviewed separately.

Summary:

- Remove almost all instances of `np.dot` and replace them by `@` finally!
- Remove `Orbit.from_body_ephem` at last!
- Move some logic of `Orbit` classmethods to separate functions
- Rename and move some stuff around for consistency

Managed to remove ~200 lines from `orbit.py`, but we can do much more. `Orbit.frozen` in particular is a bit daunting.


astrojuanlu:
Pushed 1 more commit to finish the work we started on #1026 and separate the propagators in modules.

I promise I won't add anything else :) Next I'd love to tackle #921 and #1364.

Merry Christmas! 🎄

astrojuanlu:
> I promise I won't add anything else :) Next I'd love to tackle #921 and #1364.

Next in line: https://github.com/astrojuanlu/poliastro/pull/12

jorgepiloto:
Reviewing this 🔎

contrib/rv2tle.py
```diff
@@ -73,14 +73,14 @@ def rvel(r, v):
 
     vk = 1.0 / XKE * vv2  # smult
     h = np.cross(rr2, vk)  # cross
-    pl = np.dot(h, h)
+    pl = h @ h
     vz = np.array([0.0, 0.0, 1.0])
     n = np.cross(vz, h)
     n = unitv(n)
     rk = np.linalg.norm(rr2)
```
jorgepiloto, line 80: In other lines, the `np.linalg.norm` is replaced by `np.sqrt(vec @ vec)`. The `norm` function is more explicit, but I wonder which one is faster 🤔

jorgepiloto: Another way of doing so is using: `(vec @ vec) ** 0.5`, which is applied in other lines  too.

src/poliastro/core/elements.py
```diff
@@ -11,6 +11,31 @@
 from poliastro.core.util import rotation_matrix
 
 
+@jit
+def eccentricity_vector(k, r, v):
```
jorgepiloto, line 15: What about adding a docstring in here? 📝

src/poliastro/core/util.py
```diff
@@ -1,32 +1,6 @@
...
-from numpy.linalg import norm as norm
-
-
-@jit
-def eccentricity_vector(k, r, v):
-    return ((v.dot(v) - k / (norm(r))) * r - r.dot(v) * v) / k
-
-
-@jit
-def circular_velocity(k, a):
```
jorgepiloto, line 13: Which is the reason behind shipping this to `poliatro.core.elements`?

jorgepiloto (approved):
This was a heavy PR but many things got simplified. Amazing work @astrojuanlu 👏🏽

contrib/rv2tle.py
astrojuanlu: Interestingly, the last approach is faster!

```
In [7]: np.sqrt(u @ u)
Out[7]: 0.9193857259765762

In [8]: norm(u)
Out[8]: 0.9193857259765762

In [9]: (u @ u) ** .5
Out[9]: 0.9193857259765762

In [10]: %timeit np.sqrt(u @ u)
1.32 µs ± 5.52 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [11]: %timeit norm(u)
2.16 µs ± 21.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [12]: %timeit (u @ u) ** .5
708 ns ± 7.26 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [13]: def unorm(u):
    ...:     return (u @ u) ** .5
    ...: 

In [14]: %timeit unorm(u)
753 ns ± 16.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

With numba though, the differences are not so big:

```
In [15]: from numba import njit

In [16]: @njit
    ...: def unorm1(u):
    ...:     return np.sqrt(u @ u)
    ...: 

In [17]: @njit
    ...: def unorm2(u):
    ...:     return norm(u)
    ...: 
    ...: 

In [18]: @njit
    ...: def unorm3(u):
    ...:     return (u @ u) ** .5
    ...: 

In [19]: unorm1(u)
Out[19]: 0.9193857259765762

In [20]: unorm2(u)
Out[20]: 0.9193857259765762

In [21]: unorm3(u)
Out[21]: 0.9193857259765762

In [22]: %timeit unorm1(u)
172 ns ± 1.05 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [23]: %timeit unorm2(u)
299 ns ± 9.15 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [24]: %timeit unorm3(u)
203 ns ± 2.16 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

But I agree we should be consistent:

1. Add a fast `norm` implementation in `poliastro.core.util` (or even the new upcoming `poliastro._math`)
2. Use it everywhere

astrojuanlu: Opened #1433 to continue that discussion

src/poliastro/core/elements.py
astrojuanlu: On it! 💪🏽

src/poliastro/core/util.py
astrojuanlu: I'm using that module for small Astrodynamics formulas, I think it fits there better. Probably we should find a better module name for the other functions in the future.

src/poliastro/core/elements.py
astrojuanlu: Done!

astrojuanlu:
Thanks a lot @jorgepiloto ! 🚀

contrib/rv2tle.py
jorgepiloto: Wow, thanks for the deep review! Yeah, I was thinking about the recently proposed `poliastro._math` module.

PR #1434, opened 2021-12-27
Refactor math (was: `poliastro.ephem` cleanup)

astrojuanlu:
Notable changes:

- ~Rewrite `build_ephem_interpolant` in terms of existing classes. I tried to remove its usage but it will require a bit more care.~ Had to revert this because it caused a massive slowdown, see below.
- Embrace the Astropy builtin ephemeris everywhere in the tests, now that Astropy >= 5.0 can compute the Moon velocity! astropy/astropy#11753 The large diff in the tests is because of the indentation removal, should be mostly a whitespace change.
- As a result, drop support for Python 3.7.
- Create new `poliastro._math` package with all mathematical functions not strictly related to Astrodynamics, so we can easily identify (and possibly refactor) dependencies with SciPy. I decided to make it private but maybe we shouldn't be so overprotective, I'm open to suggestions... 😋


astrojuanlu:
Pro tip:

Screenshot 2021-12-27 at 07-46-57 `poliastro ephem` cleanup by astrojuanlu · Pull Request #1434 · poliastro poliastro

astrojuanlu:
Next in line: https://github.com/astrojuanlu/poliastro/pull/13

astrojuanlu:
Hmmm slow tests got really slow? 🤔

astrojuanlu:
`py38-slow` took ~12 minutes on `main`, let's see if there's a slowdown. I did some tests and I think there shouldn't, but we'll see.

src/poliastro/ephem.py
```diff
@@ -50,46 +49,42 @@ def build_ephem_interpolant(body, period, t_span, rtol=1e-5):
...
-        )
+    epochs = time_range(
+        Time(t_span[0], format="jd", scale="tdb"),
+        end=Time(t_span[1], format="jd", scale="tdb"),
+        periods=int(((t_span[1] - t_span[0]) / (period * rtol)).to_value(u.one)),
+    )
+    ephem = Ephem.from_body(body, epochs, attractor=attractor)
 
-        r_values[i] = r.xyz.to(u.km)
+    def interpolant(t):
```
astrojuanlu, line 64: This probably poses a big performance regression and can't be merged as it stands now 😭

astrojuanlu:
I have some ideas on how I could tackle this but I'll need more time. For now, I'll partially revert this and merge.

astrojuanlu:
Notes: https://github.com/poliastro/poliastro/wiki/Ideas-for-fast-interpolation

astrojuanlu:
Looking much better now.

Issue #1350, opened 2021-10-20
Minimum orbit intersection distance + miss distance (collision)


MLopez-Ibanez:
💡 **Feature request**

Is there a function to calculate the minimum orbit intersection distance between two orbit objects? 

If not, what would be the best strategy to implement it in poliastro without duplicating anything already implemented?

astrojuanlu:
What I called "Time-dependent" above is called "Miss Distance" by Vallado (see @Yash-10 original comment in #1419).

astrojuanlu:
Tracing back from the references of Hedo et al, looks like the classical, robust method of computing the MOID is Dybczyński et al https://doi.org/10.1007/BF01238925. This paper cites faster methods from Lazovič and Hoots that however don't have universal convergence.

Also, Hedo et al mention that "Giovanni Gronchi’s FORTRAN software for MOID computation of 1995 (Gronchi 2005) has been provided by the author". Therefore, maybe we can also ask for it :)

Yash-10:
I have been thinking of a good way to add collision detection keeping in mind the `solve_ivp` interface with which all events are tied to.

- Since it is not possible (at the moment) to detect events while propagating more than one orbit, I remember we had settled on passing (`positions`, `velocities`) of a secondary satellite by the user (we discarded ephemerides as an option but do not recall the exact reason).
- Since one is more interested in the minimum distance between two orbits rather than checking for exact orbit intersection, just returning the relative distance between two satellites inside the `__call__` method of the event class might not be a good idea.
    - One simple option might be to use a known threshold distance (for eg: the distance that classifies an object to be a potentially hazardous object (PHO)) and subtract our current satellite-satellite distance with it to detect a "collision" event.
    - If the above seems not a good idea, maybe we can follow the "Close Approach" algorithm mentioned in Vallado. That would return the minimum distance between two orbits. Using this minimum distance, it's still unknown to me how we could align it with "zero-crossing" for event detection.
- Most importantly, how could we ensure that the coordinates and velocities of the secondary satellite that the user passes correspond to the same time as the primary satellite's coordinates and velocities?

I would like to get some ideas and advice on this. Maybe I am missing somethings along the way.

Thanks!

astrojuanlu:
- Having geometrical (time-independent) MOID is a nice-to-have (if we know MOID > threshold, then we *know* that "collision" will never happen)
- Numba-compatible interpolation would be a nice-to-have
[...]

PR #1439, opened 2022-01-14
Added recursive series Kepler solver for elliptical orbits

JackCrusoe47:
- Added the series solution approximation for elliptical orbits.  
- Small correction to Danby code. Corrected nu to M for circular orbits as M = nu @ e=0

Currently, I only included the manual series solution, where the user provides the order of recursion (instead of tolerance).

Later, I can implement a hybrid variant that uses a lookup table to select appropriate order for a mean anomaly and eccentricity and then uses the initial guess for a Newton-Raphson. Here, the user will be able to specify a tolerance, and hopefully, the table provides an order of recursion that overall reduce the cost of computing the solution. I am currently having trouble with fast lookup.

Hope I didn't mess anything up with the pull request. I have never worked with Github.

Cheers

### DIFF
--- a/src/poliastro/core/propagation/__init__.py
+++ b/src/poliastro/core/propagation/__init__.py
@@ -10,6 +10,7 @@ from poliastro.core.propagation.markley import markley, markley_coe
 from poliastro.core.propagation.mikkola import mikkola, mikkola_coe
 from poliastro.core.propagation.pimienta import pimienta, pimienta_coe
 from poliastro.core.propagation.vallado import vallado
+from poliastro.core.propagation.recSeries import recSeries, recSeries_coe
 
 __all__ = [
     "func_twobody",
@@ -26,6 +27,8 @@ __all__ = [
     "gooding",
     "danby_coe",
     "danby",
+    "recSeries_coe",
+    "recSeries",
 ]
 
 
--- a/src/poliastro/core/propagation/__init__.py
+++ b/src/poliastro/core/propagation/__init__.py
@@ -9,8 +9,8 @@ from poliastro.core.propagation.gooding import gooding, gooding_coe
 from poliastro.core.propagation.markley import markley, markley_coe
 from poliastro.core.propagation.mikkola import mikkola, mikkola_coe
 from poliastro.core.propagation.pimienta import pimienta, pimienta_coe
-from poliastro.core.propagation.vallado import vallado
 from poliastro.core.propagation.recSeries import recSeries, recSeries_coe
+from poliastro.core.propagation.vallado import vallado
 
 __all__ = [
     "func_twobody",
--- a/src/poliastro/core/propagation/__init__.py
+++ b/src/poliastro/core/propagation/__init__.py
@@ -9,7 +9,7 @@ from poliastro.core.propagation.gooding import gooding, gooding_coe
 from poliastro.core.propagation.markley import markley, markley_coe
 from poliastro.core.propagation.mikkola import mikkola, mikkola_coe
 from poliastro.core.propagation.pimienta import pimienta, pimienta_coe
-from poliastro.core.propagation.recSeries import recseries, recseries_coe
+from poliastro.core.propagation.recseries import recseries, recseries_coe
 from poliastro.core.propagation.vallado import vallado
 
 __all__ = [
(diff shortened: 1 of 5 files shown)


astrojuanlu (changes requested):
Amazing @JackCrusoe47 ! That's what I call "from zero to hero" 😃 Good use of GitHub, nothing to worry about there.

I have a few minor requests:

- Could you rename the module and function names to `all_lowercase` instead of `camelCase`? to keep consistency
- Could you add your method here

https://github.com/poliastro/poliastro/blob/9797eb079d7ce86b39aaef379b750d88c3135ce2/src/poliastro/twobody/propagation.py#L488-L497

so it gets tested automatically?

JackCrusoe47:
Thank you @astrojuanlu. 

of course, I'll do the corrections right away. I also saw that two automatic tests are failing. Is it related to not adding my method in the tests, or would I need to modify anything further?

The tests that fail are:

- codecov/patch
- codecov/project

Yash-10:
Wow! This seems to be a great addition. As the paper states,

> Fast approximation of closest approach distance for co-planar orbits.

this approach could also be a good approach to check out for the satellite collision event #1350, isn't it?

JackCrusoe47:
@Yash-10 Well my idea was to use the recursive series (not more than 3-recursions) as a fast initial guess pass over the domain, to get a rough estimate on solution space. Then use this information to target possible collision regions with an accurate algorithm. Unfortunately, besides mentioning this, I really didn't experiment with this in my paper.

JackCrusoe47:
P.s. I apologize in advance for the state of my commits. I still have not figured a way to edit and test code directly, and still couldn't find the option to roll back commits in Github individually.

astrojuanlu (commented):
5 tests have failed:

```
FAILED tests/tests_twobody/test_propagation.py::test_elliptic_near_parabolic[recseries-0.9]
FAILED tests/tests_twobody/test_propagation.py::test_elliptic_near_parabolic[recseries-0.99]
FAILED tests/tests_twobody/test_propagation.py::test_elliptic_near_parabolic[recseries-0.999]
FAILED tests/tests_twobody/test_propagation.py::test_elliptic_near_parabolic[recseries-0.9999]
FAILED tests/tests_twobody/test_propagation.py::test_elliptic_near_parabolic[recseries-0.99999]
```

which means that, with eccentricities of 0.9 and higher, the method loses a bit of accuracy.

I assume this is how the method works, so I'll assist you in relaxing the tolerance for the tests without affecting the other methods.

JackCrusoe47:
> 5 tests have failed:
> 
> ```
> FAILED tests/tests_twobody/test_propagation.py::test_elliptic_near_parabolic[recseries-0.9]
> FAILED tests/tests_twobody/test_propagation.py::test_elliptic_near_parabolic[recseries-0.99]
> FAILED tests/tests_twobody/test_propagation.py::test_elliptic_near_parabolic[recseries-0.999]
> FAILED tests/tests_twobody/test_propagation.py::test_elliptic_near_parabolic[recseries-0.9999]
> FAILED tests/tests_twobody/test_propagation.py::test_elliptic_near_parabolic[recseries-0.99999]
> ```
> 
> which means that, with eccentricities of 0.9 and higher, the method loses a bit of accuracy.
> 
> I assume this is how the method works, so I'll assist you in relaxing the tolerance for the tests without affecting the other methods.

Unfortunately yes. The rate of convergence of the method drops to zero as the eccentricity becomes close to 1. In theory, it should work for cases up to 0.99 with a high enough order (say 20-30 for 0.99 for example). But for now the function I use to convert accuracy to order probably is underpredicting the order for 0.9 and 0.99.

JackCrusoe47:
Maybe I can add an appropriate function to scale order with rtol and eccentricity. It won't probably fix the case of 0.999 and 0.9999, but until 0.99 I think I can make a good prediction.

JackCrusoe47:
All cases more or less should work now. 

Now the recseries( ) solver can be called in the following two ways:
- (rf,vf) = recseries(k, r0, v0, tof, method='order', order=12)   # use a fixed 12th order recursive series
- (rf,vf) = recseries(k, r0, v0, tof, method='rtol', rtol=1e-8)      # recurse until a tolerance of 1e-8

This works the same for recseries_coe( )

astrojuanlu (approved):
Looks good to me! Will merge when the tests pass 💪🏼

JackCrusoe47:
Perfect. Thank you.

astrojuanlu:
A small coverage difference is not worth blocking this PR 😉

astrojuanlu:
Congratulations @JackCrusoe47 on your first contribution to poliastro! 🎉

JackCrusoe47:
Thank you @astrojuanlu for merging my commits 🚀. I am eager to see its use by the community. 😄

Issue #1433, opened 2021-12-27
Compute norm consistently everywhere


astrojuanlu:
Interestingly, the last approach is faster!

```
In [7]: np.sqrt(u @ u)
Out[7]: 0.9193857259765762

In [8]: norm(u)
Out[8]: 0.9193857259765762

In [9]: (u @ u) ** .5
Out[9]: 0.9193857259765762

In [10]: %timeit np.sqrt(u @ u)
1.32 µs ± 5.52 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [11]: %timeit norm(u)
2.16 µs ± 21.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [12]: %timeit (u @ u) ** .5
708 ns ± 7.26 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [13]: def unorm(u):
    ...:     return (u @ u) ** .5
    ...: 

In [14]: %timeit unorm(u)
753 ns ± 16.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

With numba though, the differences are not so big:

```
In [15]: from numba import njit

In [16]: @njit
    ...: def unorm1(u):
    ...:     return np.sqrt(u @ u)
    ...: 

In [17]: @njit
    ...: def unorm2(u):
    ...:     return norm(u)
    ...: 
    ...: 

In [18]: @njit
    ...: def unorm3(u):
    ...:     return (u @ u) ** .5
    ...: 

In [19]: unorm1(u)
Out[19]: 0.9193857259765762

In [20]: unorm2(u)
Out[20]: 0.9193857259765762

In [21]: unorm3(u)
Out[21]: 0.9193857259765762

In [22]: %timeit unorm1(u)
172 ns ± 1.05 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [23]: %timeit unorm2(u)
299 ns ± 9.15 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [24]: %timeit unorm3(u)
203 ns ± 2.16 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

But I agree we should be consistent:

1. Add a fast `norm` implementation in `poliastro.core.util` (or even the new upcoming `poliastro._math`)
2. Use it everywhere

[...]

PR #1442, opened 2022-01-15
Consistent norm computation

Yash-10:
From observations in #1433.

Thanks!

### DIFF
--- a/src/poliastro/_math/linalg.py
+++ b/src/poliastro/_math/linalg.py
@@ -1,0 +1,7 @@
+import numpy as np
+from numba import njit as jit
+
+
+@jit
+def norm(arr):
+    return np.sqrt(arr @ arr)
--- a/docs/source/examples/Analyzing the Parker Solar Probe flybys.mystnb
+++ b/docs/source/examples/Analyzing the Parker Solar Probe flybys.mystnb
@@ -66,7 +66,7 @@ flyby_1_time
 ```
 
 ```{code-cell} ipython3
-r_mag_ref = norm(Ephem.from_body(Venus, flyby_1_time).rv()[0])
+r_mag_ref = norm(Ephem.from_body(Venus, flyby_1_time).rv()[0].squeeze())
 r_mag_ref.to(u.au)
 ```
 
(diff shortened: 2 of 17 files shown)


Yash-10:
Trying to investigate the docs error...

Yash-10:
It seems if the input array contains extra unwanted dimensions, it fails. `np.squeeze()` is not yet an option in `numba`, so maybe something else is needed.

(Maybe support to different types of norms, i.e. L1, L2, etc could also be kept just as in the case of `np.linalg.norm`)

astrojuanlu:
I see that `norm(Ephem.from_body(Venus, flyby_1_time).rv()[0])` fails, because `Ephem.from_body(Venus, flyby_1_time).rv()[0].ndim == 2`. On the other hand, `norm(Ephem.from_body(Venus, flyby_1_time).rv()[0][0])` and `norm(Ephem.from_body(Venus, flyby_1_time).rv(flyby_1_time)[0])` succeed.

We want norms of vectors, not norm of matrices, so I think this failure is correct. It forces the user to put the vector in the correct shape though. Addmittedly, `Ephem.from_body(Venus, flyby_1_time).rv(flyby_1_time)[0]` is not a very convenient way of computing the position of a planet at a given time, we are probably overusing `Ephem` there. But I suggest we make the small change in the notebook and we continue.

src/poliastro/_math/linalg.py
```diff
@@ -0,0 +1,7 @@
+import numpy as np
+from numba import njit as jit
+
+
+@jit
+def norm_fast(arr):
```
astrojuanlu, line 6: Can we perhaps rename this to just `norm`, and import it as such? That way we minimize the lines changed too.

```suggestion
def norm(arr):
```

astrojuanlu (approved):
Looks perfect to me! When we fix the notebook that is failing, we can merge.

src/poliastro/_math/linalg.py
astrojuanlu: (And, in modules where we have units-compatible `norm` and this function, we can do `from poliastro._math.linalg import norm as norm_fast`)

Yash-10: Yeah, this seems to be better!

astrojuanlu:
Thanks a lot @Yash-10 !

Issue #1395, opened 2021-11-12
Attractor identity changes when pickling & unpickling


astrojuanlu:
This manifests itself when doing something like this, as @MLopez-Ibanez hinted in #1377:

```python
import pickle
from poliastro.examples import iss                                                                                                                                                                                                            
from poliastro.plotting.static import StaticOrbitPlotter

with open("/tmp/orbit.pkl", "wb") as fh:
    pickle.dump(iss, fh)
     
with open("/tmp/orbit.pkl", "rb") as fh:
    iss_pickle = pickle.load(fh)

fr = StaticOrbitPlotter()
fr.plot(iss)
fr.plot(iss_pickle)
```

which results in:

```
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
<ipython-input-22-b48ffc98bf85> in <module>
     12 fr = StaticOrbitPlotter()
     13 fr.plot(iss)
---> 14 fr.plot(iss_pickle)

~/Projects/LSF/poliastro/library/src/poliastro/plotting/static.py in plot(self, orbit, label, color, trail)
    256             self.set_orbit_frame(orbit)
    257 
--> 258         lines = self._plot(orbit, label=label, color=color, trail=trail)
    259         lines = lines[0] + [lines[1]]
    260 

~/Projects/LSF/poliastro/library/src/poliastro/plotting/_base.py in _plot(self, orbit, label, color, trail)
    212         colors = self._get_colors(color, trail)
    213 
--> 214         self.set_attractor(orbit.attractor)
    215 
    216         orbit = orbit.change_plane(self.plane)

~/Projects/LSF/poliastro/library/src/poliastro/plotting/_base.py in set_attractor(self, attractor)
     57 
     58         """
---> 59         self._set_attractor(attractor)
     60 
     61     def _clear_attractor(self):

~/Projects/LSF/poliastro/library/src/poliastro/plotting/_base.py in _set_attractor(self, attractor)
     44             self._attractor = attractor
     45         elif attractor is not self._attractor:
---> 46             raise NotImplementedError(
     47                 f"Attractor has already been set to {self._attractor.name}"
     48             )

NotImplementedError: Attractor has already been set to Earth
```

And the underlying reason is that:

```
In [24]: iss
Out[24]: 6772 x 6790 km x 51.6 deg (GCRS) orbit around Earth (♁) at epoch 2013-03-18 12:00:00.000 (UTC)

In [25]: iss_pickle
[...]

PR #1443, opened 2022-01-16
Add `__reduce__` to `Body` class

Yash-10:
This change does seem to solve the problem mentioned in #1395. `__reduce__` returns a string, which I think is commonly done for singleton classes. As a result, I was able to get the plot:
fig

Just wondering if this is really a solution to the issue...

### DIFF
--- a/src/poliastro/bodies.py
+++ b/src/poliastro/bodies.py
@@ -96,6 +96,9 @@ class Body(
     def __str__(self):
         return f"{self.name} ({self.symbol})"
 
+    def __reduce__(self):
+        return self.name
+
     def __repr__(self):
         return self.__str__()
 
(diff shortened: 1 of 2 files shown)


astrojuanlu (changes requested):
Looks like `__reduce__` accepts either a string, and that it's the right thing to do in this case:

> If a string is returned, the string should be interpreted as the name of a global variable. It should be the object’s local name relative to its module; the pickle module searches the module namespace to determine the object’s module. This behaviour is typically useful for singletons.

https://docs.python.org/3/library/pickle.html#object.__reduce__

So, this solution is good! But it needs a test 😄

Yash-10:
I used the `tmp_path` fixture (https://docs.pytest.org/en/latest/how-to/tmp_path.html#the-tmp-path-fixture) to save and load objects into a file for testing. Had not used it before though. Hopefully it's not an overkill :sweat_smile:

astrojuanlu (approved):
Looks perfect!

PR #1445, opened 2022-01-18
[RFC - do not merge] First `OrbitArray` API prototype

s-m-e:


s-m-e:
I have put some of my thinking down into a notebook and actually built a little prototype - which does Brownian motion, not Orbits, to make stuff simple enough to reason about the user-facing API as well as some internal aspects.

prototypes/brownian_01.ipynb
```diff
@@ -0,0 +1,925 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 9,
+   "id": "0727ffd1-14e1-419b-bb78-bddabd290d21",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "@typechecked\n",
+    "class StateArray:\n",
```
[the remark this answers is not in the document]
astrojuanlu: @jorgepiloto raises the idea of using custom dtypes, for example the quaternion implementation here https://github.com/numpy/numpy-dtypes/ or the new generation dtypes here https://github.com/seberg/unitdtype but that is a bit experimental at the moment so we will continue with a simpler approach. Then we can evaluate the status of this later.

prototypes/brownian_01.ipynb
```diff
@@ -0,0 +1,925 @@
...
+    "        return  (\n",
+    "            f'<StateArray shape={self.shape} value=[\\n'\n",
+    "            + '\\n'.join([\n",
+    "                f' (epoch={state.epoch} x={state.x} y={state.y}),'\n",
+    "                for state in self.reshape(self.size)\n",
+    "            ])\n",
+    "            + '\\n]>'\n",
+    "        )\n",
+    "    \n",
+    "    def __getitem__(self, idx) -> 'Union[StateArray, State]':\n",
```
astrojuanlu, line 471: 💯

prototypes/brownian_01.ipynb
```diff
@@ -0,0 +1,925 @@
...
+    "                for state in self._statearray.reshape(np.multiply.reduce(self._statearray.shape))\n",
+    "            ])\n",
+    "            + '\\n]>'\n",
+    "        )\n",
+    "    \n",
+    "    def __getitem__(self, idx) -> 'Union[OrbitArray, Orbit]':\n",
+    "        target = self._statearray[idx]\n",
+    "        if isinstance(target, State):\n",
+    "            return Orbit(state = target)\n",
+    "        return OrbitArray(statearray = target)\n",
```
astrojuanlu, line 675: ✔️

prototypes/brownian_01.ipynb
```diff
@@ -0,0 +1,925 @@
...
+    "\n",
+    "    def propagate(self, timedelta: TimeDelta, inplace: bool = False) -> 'OrbitArray':\n",
+    "        \n",
+    "        if timedelta.ndim == 0:\n",
+    "            timedelta = np.repeat(timedelta.to_value(u.d), self.size).reshape(self.shape) << u.d\n",
```
astrojuanlu, line 680: The broadcasting rules mentioned above ☝🏽

prototypes/brownian_01.ipynb
```diff
@@ -0,0 +1,925 @@
...
+    "            + '\\n]>'\n",
+    "        )\n",
+    "    \n",
+    "    def __getitem__(self, idx) -> 'Union[OrbitArray, Orbit]':\n",
+    "        target = self._statearray[idx]\n",
+    "        if isinstance(target, State):\n",
+    "            return Orbit(state = target)\n",
+    "        return OrbitArray(statearray = target)\n",
+    "\n",
+    "    def propagate(self, timedelta: TimeDelta, inplace: bool = False) -> 'OrbitArray':\n",
```
astrojuanlu, line 677: Same comment as above, let's not close the door on `inplace=True` but let's not do this in this first iteration so we can focus.

But @s-m-e is *veeeeeeeeeeeeeeeeeeeery* stubborn and wants to burn hours writing 2x as many tests 😉 so let's give him the chance!

prototypes/brownian_01.ipynb
```diff
@@ -0,0 +1,925 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "beb5b6bd-199a-45ba-8b68-e81201f68948",
+   "metadata": {},
+   "source": [
+    "## Further thinking\n",
+    "\n",
+    "- [Structured arrays](https://numpy.org/doc/stable/user/basics.rec.html) can actually be n-dimensional. This could be a foundation for `StateArray` classes, opening the back door for alternative libraries which are exposing an [array interface](https://numpy.org/doc/stable/reference/arrays.interface.html). In a bigger picture, the state array could be the place where stuff like `cupy.ndarray` or `dask.array` are transparently accepted.\n",
```
astrojuanlu, line 888: This is very interesting idea - are structured arrays something that has a future in the NumPy roadmap? Maybe we could rethink some things on top of this.

However, epochs are going to be an issue.

astrojuanlu: NumPy datatime is known to have lots of issues, or that's what I recall from reading Twitter threads and GitHub comments. I don't have any pointers at hand.

astrojuanlu: Notice that time of flight is _not_ used by the propagation algorithm, despite the ill-designed poliastro `core` functions:

https://github.com/poliastro/poliastro/blob/726c95d17da3fb6b43bf38ace5b2b17170e1255f/src/poliastro/core/propagation/farnocchia.py#L290-L298

Astropy should take care of adding times, not us. This deserves opening an issue.

astrojuanlu (commented):
Did a joint review with @s-m-e, @jorgepiloto, left a few comments and will follow-up asynchronously

prototypes/brownian_01.ipynb
astrojuanlu: Deleted a highly misleading comment I made: the time of flight is _definitely_ used by the propagation algorithm.

astrojuanlu: To @s-m-e's point: by passing `tof` (elapsed time in seconds) to the propagation algorithms, we have already chosen a unit. Therefore, one could take this approach one step further and work with elapsed seconds everywhere internally, to avoid the alleged performance issues of `astropy.time` and the difficulties to store them.

I _guess_ this approach would work as long as folks don't want to propagate trans-Neptunian objects or the Solar System barycenter with respect to the Milky Way. But, if it doesn't work, it's already slightly wrong in the current implementation.

Important reading: https://docs.astropy.org/en/stable/time/index.html#internal-representation

> The Time object maintains an internal representation of time as a pair of double precision numbers expressing Julian days. The sum of the two numbers is the Julian Date for that time relative to the given time scale. 

And about what would happen by using a single-float representation:

> Users requiring no better than microsecond precision over human time scales (~100 years) can safely ignore the internal representation details and skip this section.

On the other hand, about time scales:

> We have shown in the above that the difference between two UTC times is a TimeDelta with a scale of TAI. This is because a UTC time difference cannot be uniquely defined unless the user knows the two times that were differenced (because of leap seconds, a day does not always have 86400 seconds). For all other time scales, the TimeDelta inherits the scale of the first Time object.

Therefore,

1. There is a way to _do it right_ (in other words, aim for maximum accuracy) with relatively simple tools, but it requires some care.
2. We might (I haven't deeply analyzed it) have tiny precision bugs in the code at the moment, given that we are not being particularly careful with the `TimeDelta` scales and just performing a `to(u.s)` to pass the value to the propagator:

https://github.com/poliastro/poliastro/blob/726c95d17da3fb6b43bf38ace5b2b17170e1255f/src/poliastro/twobody/orbit.py#L1098-L1099

https://github.com/poliastro/poliastro/blob/726c95d17da3fb6b43bf38ace5b2b17170e1255f/src/poliastro/twobody/propagation.py#L516-L520

On the other hand, we might be getting too deep into this topic, and I don't want this to block our efforts.

s-m-e: > The Time object maintains an internal representation of time as a pair of double precision numbers expressing Julian days. The sum of the two numbers is the Julian Date for that time relative to the given time scale.

Then I was not digging deep enough. Interesting. This would actually allow a really clean implementation.

PR #1446, opened 2022-01-19
Break the Orbit class into two parts: Its creation via classmethods and its actual handling

s-m-e:
This changes the structure of the docs - but they still work. Please check.


s-m-e:
@astrojuanlu ping

src/poliastro/twobody/orbit/creation.py
```diff
@@ -1,223 +1,29 @@
...
-class Orbit:
-    """Position and velocity of a body with respect to an attractor
-    at a given time (epoch).
-
-    Regardless of how the Orbit is created, the implicit
-    reference system is an inertial one. For the specific case
-    of the Solar System, this can be assumed to be the
-    International Celestial Reference System or ICRS.
-
+class OrbitCreationMixin:
```
astrojuanlu, line 20: I don't see ourselves reusing this for Orbit arrays (the logic will be different) so maybe we can rename it?

...On the other hand, the class is not called `ScalarOrbit`, but `Orbit`. So, let's keep it.

astrojuanlu (approved):
This looks perfect, thanks @s-m-e !

Issue #1054, opened 2020-12-30
Review Lambert API


astrojuanlu:
Some concerns were raised months ago in the mailing list about the solution pairs, as well as the prograde/retrograde nature of the solutions.

When I implemented the Izzo algorithm I used a generator, and that makes me think that I was probably misunderstanding the paper. This affected how the API is used.

https://github.com/poliastro/poliastro/blob/333a2f77f5c4fa1c652bcdd9bea72ce42531b566/src/poliastro/core/iod.py#L235-L241

We should review this in depth and answer this message on the mailing list, even if months have passed:

https://groups.io/g/poliastro-dev/message/118

jorgepiloto:
I am finally able to properly answer original question raised by Irina Kovalenko and also introduce some ideas about how we could improve the current Lambert's API and associate routines.

**Amount of solutions**

Even if considering non-degenerate, degenerate and multi-revolutions ones, the solution to the Lambert's problem is an elliptic orbit most of the times. Of course, multi-revolution solutions will be for sure elliptic kind, as the orbit needs to be closed. 

Under certain combinations of time of flight and transfer angle, it is possible to find two elliptical orbits as the solution to the transfer. The following figure from the open paper [Optimal Two-Impulse Rendezvous Using Multiple-Revolution Lambert Solutions](https://www.researchgate.net/publication/242182610_Optimal_Two-Impulse_Rendezvous_Using_Multiple-Revolution_Lambert_Solutions) (DOI: 10.2514/2.5014) provides a very illustrative figure:

<div align=center>
<img  src="https://user-images.githubusercontent.com/28702884/109733557-9b55e600-7bbf-11eb-92ce-c68c81dde48e.png"/>
</div>
<br>

*Notice that there exist a short and long paths for prograde transfer but also another short and long arcs for retrograte motion, making thus a total of 4 possible transfer arcs. The final orbit is selected by imposing the sense of motion and selecting the short or long arc.*


**State of the current Lambert API**

At the moment, only two Lambert solvers are available within poliastro: [vallado](https://github.com/poliastro/poliastro/blob/3c1d0eb8382075a8e7a191cd84acf34748c2d649/src/poliastro/core/iod.py#L11) and [izzo](https://github.com/poliastro/poliastro/blob/3c1d0eb8382075a8e7a191cd84acf34748c2d649/src/poliastro/core/iod.py#L164). Although both address the same problem, they require different parameters:

[...]

PR #1450, opened 2022-01-20
Fix lambert API

jorgepiloto:
This PR solves for #1054 by:

* Making sure all Lambert's problem solvers accepts the same amount of solvers. In the case of `iod.vallado` and error is raised when `M>0` since the multi-revolution scenario has not been implemented yet (see  #858).

* Allowing user to select the desired inclination of the transfer orbit via `is_prograde` argument.
* When two solutions are found (only for `M>0`), the `is_lowpath` argument allows user to select between one of those.


jorgepiloto:
I still need to include a test for the new `NotImplementedError` exception in `core.iod.vallado` and fix the notebook titled "Multirevolutions solution in Lamberts problem".

jorgepiloto:
The notebook is in the oven... 👨🏽‍🍳

src/poliastro/core/iod.py
```diff
@@ -7,7 +7,7 @@
 
 
 @jit
-def vallado(k, r0, r, tof, short, numiter, rtol):
+def vallado(k, r0, r, tof, M, is_prograde, is_lowpath, numiter, rtol):
```
astrojuanlu, line 10: What about

```suggestion
def vallado(k, r0, r, tof, M, prograde, lowpath, numiter, rtol):
```

src/poliastro/core/iod.py
```diff
@@ -61,6 +61,13 @@ def vallado(k, r0, r, tof, short, numiter, rtol):
         True for short path, False for long one.
     tof : ~float
         Time of flight
+    M : int
+        Number of revolutions
+    is_prograde: boolean
+        Controls the desired inclination of the transfer orbit.
+    is_lowpath: boolean
+        Decides between high and low path when two possible transfer orbits are
```
astrojuanlu, line 69: What do "high" and "low" mean in this context? (I _think_ I know, but it would be nice to clarify a bit, if it can be done in a short sentence)

astrojuanlu (commented):
This is starting to look better! 👏🏽 Left a couple of preliminary comments

src/poliastro/core/iod.py
jorgepiloto: Sure, no problem 👍🏽

src/poliastro/core/iod.py
jorgepiloto: What about this?

```python
"""
is_lowpath:
    If `True` or `False`, gets the transfer orbit whose vacant focus is below or above the chord line, respectively.
"""
```

astrojuanlu: Looks OK!

jorgepiloto:
The notebook about "Multirevolutions solution..." has been almost written from scratch. However, I faced the following problem:

* The `StaticOrbitPlotter.plot_maneuver` method did not allowed me to customize the `linestyle` for each one of the transfer trajectories. Even after following the [plotter customization tutorial](https://docs.poliastro.space/en/latest/examples/Customising%20static%20orbit%20plots.html) I did ont succeed at this task.

If you think that the new notebook is good enough and the rest of the changes are good to merge, we can proceed and investigate `plot_maneuver` customization in a new issue, @astrojuanlu.

astrojuanlu (changes requested):
The new notebook looks great, congrats @jorgepiloto ! 👏🏽 

My only feedback is whether we could use `docs/source/examples/lambert_paths.png` as the thumbnail, and that the last plot looks a bit small. Perhaps the axes could be displayed vertically instead of horizontally.

About the linestyles, yes feel free to open a new issue if you tried and failed.

Issue #1454, opened 2022-01-31
Documentation typo in Hohmann (Maneuver) functions (altitude/orbital radius)


manweichan:
The documentation for Hohmann transfers incorrectly describes the r_f input parameter as the 'altitude' of the orbit. The following equations support that r_f is actually the orbital radius, not altitude (distance above the surface of the gravitational body). I'm thinking this may trip someone up in the future.

[poliastro/core/maneuver](https://github.com/poliastro/poliastro/blob/726c95d17da3fb6b43bf38ace5b2b17170e1255f/src/poliastro/core/maneuver.py#L39)
[poliastr/maneuvero](https://github.com/poliastro/poliastro/blob/726c95d17da3fb6b43bf38ace5b2b17170e1255f/src/poliastro/maneuver.py#L86)

Note: this should be a quick fix, and I figure this would be the best channel to expose this error, but please let me know if there's another preferred way.

📋  **Steps to solve the problem**

 * Change lines above from `final altitude of the orbit` to `final orbital radius`

astrojuanlu:
Good catch @manweichan ! Do you want to send a pull request?

manweichan:
Looks like I was slow to the pull! 

Thanks @Sync271 for fixing this!

I've never contributed to a project before and was working through the documentation. I'll definitely try to contribute to another issue in the future!

Looks like this one can be closed.

Sync271:
> Looks like I was slow to the pull! 
> 
> Thanks @Sync271 for fixing this!
> 
> I've never contributed to a project before and was working through the documentation. I'll definitely try to contribute to another issue in the future!
> 
> Looks like this one can be closed. 

I just started contributing, thanks to you too in a way for reporting the issue :)

manweichan:
Awesome @Sync271 ! If I may ask, what do you find easiest for contributing? Working from the GitHub browser, cloud, or local?

Sync271:
@manweichan check for issues that have "good first issue", which usually are editable through a browser

Issue #1456, opened 2022-02-01
Migrate away from Azure Pipelines?


astrojuanlu:
We were greeted yesterday by messages like this: `##[error]No hosted parallelism has been purchased or granted`. I quickly requested parallelism to be granted for our project, but it is unclear whether they will reply quickly or not.

I personally consider this to be extremely bad DX. Nothing changed on our end and, references to this date from last year https://devblogs.microsoft.com/devops/change-in-azure-pipelines-grant-for-public-projects/ and, to the best of my knowledge, I didn't receive an email explaining why we suddenly got this. The fact that I needed to set up an Outlook account to handle this was quite annoying as well, so if I can use this as an excuse to delete it, I'll be happy.

I think the only serious alternatives are CircleCI and GitHub Actions. Initally I was considering CircleCI because we were using it in the past and I don't want to tie us to GitHub even more, but at the same time, each alternative platform seems to have its own CI service (GitLab CI, Codeberg-CI), so it seems unlikely that we can avoid rewriting some stuff in the unlikely event that we ever want to do a migration.

If Azure hasn't replied by Thursday, I'll do the migration on Friday. And even if they reply and our pipelines run again, I'd say that we leave anyway at our own pace.

astrojuanlu:
The pipelines are back.

astrojuanlu:
...The pipelines are blocked again! 🤬 Just sent an email asking for clarification.

PR #1457, opened 2022-02-01
Documentation typo in Hohmann (Maneuver) functions (altitude/orbital radius) #1454

Sync271:
Fixed as per #1454 
I've nothing more to add


astrojuanlu (approved):
Thanks a lot @Sync271 ! 🚀 

Notice that the pipelines are broken because of #1456 .

Issue #1431, opened 2021-12-20
Add Gabbard plots


astrojuanlu:
See for example

![savetweetvid_FHA8kccUcAIyIix](https://user-images.githubusercontent.com/316517/146810204-48ce7192-3c20-41ed-b373-e37d90951fe9.gif)

(we can do the static ones first 😉 )

Ln11211:
Hello,
I was just wondering if this issue is open for contribution and if it is beginner-friendly? If yes, Can you please provide more details about this?

astrojuanlu:
Hi @Ln11211 ! This issue is open for contribution. However, we usually use "beginner-friendly" for issues that are a 1 line change, or trivially explained in the issue itself, and they're just for people to get used to git and contributing 😃

This is a bit more involved, but essentially I imagine something like

```python
from poliastro.plotting.gabbard import GabbardPlotter

gd = GabbardPlotter()
gd.plot_orbits([orbit for orbit in orbits])
```

and

```python
import matplotlib.pyplot as plt

fig, ax = plt.subplots()

gd = GabbardPlotter(ax=ax)
```

And a static plot similar to the one in the first comment should show.

Ln11211:
@astrojuanlu, so I understand that we want to produce gabbard plots for the satellite debris (If I am not wrong, it was that same satellite which was destroyed in a Russian missile test which later posed a risk to ISS crew right ? 🤔). How are the debris orbit altitude and time fed as an input for the graphs ?, so, my task is to input altitude and time and produce more such plots on different dates right?

astrojuanlu:
The orbital data would be passed as a list of `poliastro.twobody.orbit.Orbit` objects in this case, so we would only need to write the code to do the plotting.

Ln11211:
ok, I am assuming this will be similar to `porkchop.py` and I am really interested in doing this but I don't know for certain if I can do this(Idk how to do it, I lack the knowledge, I looked at all other files in `plotting` but I feel confused). If it's ok with you, can I write a mail to you?

astrojuanlu:
Yeah, the plotting code is quite nested (I'm somewhat happy with the design, but I don't claim it's easy to understand initially). On the other hand, `porkchop.py` is much more amenable.

You are welcome to send me an email with general advice applied to your specific circumstances but I'd rather have discussion in public if it's applicable to the project :) This issue tracker or [our chat](http://chat.poliastro.space) are good places.

Ln11211:
[...]

PR #1318, opened 2021-09-09
Modification units poliastro_plotting_core.py

JohaMR:
fix #1291

PR #1464, opened 2022-02-04
Added `gabbard.py` in `poliastro.plotting`

TheBuffer:
Tried adding a Static Gabbard Plotter as described in #1431
and tested it out with the COSMOS 1408 DEB data
Screenshot from 2022-02-04 14-12-07


astrojuanlu (commented):
Thanks @TheBuffer for this contribution! 🎉 I rebased your branch and everything looks good initially. I'm going to review this in depth in a few hours.

TheBuffer:
Should I try adding the Dynamic Gabbard Plots and maybe integrating something which takes in the Debris Ephemerides as an input?

src/poliastro/plotting/gabbard.py
```diff
@@ -0,0 +1,74 @@
...
+        if not self._ax:
+            if dark:
+                with plt.style.context("dark_background"):
+                    _, self._ax = plt.subplots(figsize=(6, 6))
+            else:
+                _, self._ax = plt.subplots(figsize=(6, 6))
+
+        self._frame = None
+
+    @staticmethod
```
astrojuanlu, line 20: I'd say we don't need this static method? (See my comment below)

src/poliastro/plotting/gabbard.py
```diff
@@ -0,0 +1,74 @@
...
+    def _get_orbit_properties(orbit):
+        return orbit.r_p, orbit.r_a, orbit.period
+
+    def _get_orbit_property_list(self, orbits):
+        apogees = []
+        perigees = []
+        periods = []
+
+        for orbit in orbits:
+            perigee, apogee, period = self._get_orbit_properties(orbit)
```
astrojuanlu, line 30: We could simplify this to remove the static method:

```suggestion
            perigee, apogee, period = orbit.r_p, orbit.r_a, orbit.period
```

src/poliastro/plotting/gabbard.py
```diff
@@ -0,0 +1,74 @@
...
+
+        apogee_paths = plt.scatter(
+            periods, apogees, marker="o", color="blue", label="Apogee"
+        )
+        perigee_paths = plt.scatter(
+            periods, perigees, marker="o", color="red", label="Perigee"
+        )
+
+        self._ax.set_xlabel("Period(min)")
+        self._ax.set_ylabel("Altitude(km)")
```
astrojuanlu, line 53:
```suggestion
        self._ax.set_xlabel("Period (min)")
        self._ax.set_ylabel("Altitude (km)")
```

(see comment about units above)

src/poliastro/plotting/gabbard.py
```diff
@@ -0,0 +1,74 @@
...
+    def _get_orbit_property_list(self, orbits):
+        apogees = []
+        perigees = []
+        periods = []
+
+        for orbit in orbits:
+            perigee, apogee, period = self._get_orbit_properties(orbit)
+            apogees.append(apogee.value)
+            perigees.append(perigee.value)
+            periods.append(period.to(u.min).value)
```
astrojuanlu, line 33: Probably we should not hardcode `u.min` here and make it a parameter, like we did for other plotters in #1318

src/poliastro/plotting/gabbard.py
```diff
@@ -0,0 +1,74 @@
...
+        return apogees, perigees, periods
+
+    def _static_gabbard_plot(self, orbits):
+        """Plots a Static Gabbard Plot given a list of Orbits
+        Parameters
+        ----------
+        orbits : ~poliastro.twobody.orbit.Orbit List
+            The Orbits whose perigee and apogee will be plotted.
+        """
```
astrojuanlu, line 42:
```suggestion
        """Plots a Static Gabbard Plot given a list of Orbits

        Parameters
        ----------
        orbits : ~poliastro.twobody.orbit.Orbit List
            The Orbits whose perigee and apogee will be plotted.

        """
```

astrojuanlu (changes requested):
This looks very good, thanks @TheBuffer for your contribution! I'd say let's focus on static Gabbard plots for now to make things simpler, we can expand it later.

I left some comments, but most importantly this will need some tests. You can take inspiration from the image tests we have for other plotters, that compare an expected outcome with the actual one:

https://github.com/poliastro/poliastro/tree/main/tests/tests_plotting

src/poliastro/examples.py
```diff
@@ -58,3 +58,186 @@
     epoch=time.Time("2015-11-05 12:00", scale="utc"),
 )
 """Comet 67P/Churyumov–Gerasimenko orbit example"""
+
+debris_orbits_list = []
```
astrojuanlu, line 62: I'd prefer that we put this only in the tests, rather than shipping it with the library.

astrojuanlu (changes requested):
Starting to look good! Left only one comment. With that, I'm 👍🏽 as soon as the tests pass.

astrojuanlu:
There seem to be some subtle (unimportant) typography differences between the baseline image and the result. I downloaded `result.png` via SSH, it's this:

result

and the difference is this:

result-failed-diff

You can probably just download `result.png` and commit it as the baseline.

astrojuanlu (approved):
Looks good to me! Thanks again @TheBuffer !

PR #1473, opened 2022-02-08
Issue1472 - Correct Pimienta Differentiated Coefficients

amacki3:
Update f14 and f15 function coefficients in pimienta propagation to reflect correct differentiation


astrojuanlu (approved):
Thanks @amacki3 and congratulations for your first pull request! 🎉

PR #1476, opened 2022-02-10
Rework API docs

### LINKED ISSUE #1183: Split API documentation into "safe" and "core"
- Split API reference in two, like it was before: https://docs.poliastro.space/en/v0.14.0/api/safe/safe_index.html and https://docs.poliastro.space/en/v0.14.0/api/core/core_index.html
- Put redirects in place: some pages changed their name, and I would like to avoid broken links. Read the Docs allows this easily 🤓

astrojuanlu:
Fixes #1183. I used some Sphinx and sphinx-autoapi hackery to make the API index more useful, by hopefully addressing the most common use cases and making the toctrees more shallow. I think this strikes a good balance between manually written docs and automatically generated ones.

This made me think about what should we consider **public API** though. We have plenty of high level modules and probably we should communicate to users which ones they should use and which ones they can ignore. However, making submodules private would be a breaking change, and since I plan to backport this pull request to 0.16.1, I will deal with that separately.

I also used this opportunity to trim most of the cruft from `conf.py`. Modern versions of `sphinx-quickstart` don't include so much stuff, but this wasn't the case 8 years ago 🙃 Folks willing to change our docs config should read the documentation.

Requesting a review from @jorgepiloto, who did the original work here.


astrojuanlu:
Rendered version: https://poliastro--1476.org.readthedocs.build/en/1476/api.html

docs/source/api.md
```diff
@@ -0,0 +1,68 @@
...
+% but then we re-include the `constants` and `czml`.
+% A more powerful syntax would be desirable but it is not yet supported,
+% see https://github.com/sphinx-doc/sphinx/issues/6650
+
+```{toctree}
+---
+maxdepth: 1
+glob:
+---
+/autoapi/poliastro/[!c_]*/index
```
jorgepiloto, line 51: Sadly, it looks like this is the best approach to exclude the core at this moment. Let us keep an eye on the Sphinx issue you linked.

astrojuanlu: Yep 😢 no way around that for now unfortunately

PR #1483, opened 2022-02-15
Expand numpy in docstring

### LINKED ISSUE #1482: `numpy.ndarray` instead of `np.array` in docstrings?
I observed that sometimes the numpy array docstrings looks like this:

https://github.com/poliastro/poliastro/blob/9797eb079d7ce86b39aaef379b750d88c3135ce2/src/poliastro/core/util.py#L56

whereas sometimes like this:

https://github.com/poliastro/poliastro/blob/ada3fce2d26bdb020db38d54795d0b2a63d05760/src/poliastro/core/propagation/gooding.py#L52

The latter seems to be a better choice since it renders a link when viewing the documentation which is not observed when using the former. I am not sure if the former style was used due to the documentation style followed, but maybe the inconsistency is something that could be worked upon?

jerobado:
Fixes #1482

### DIFF
--- a/src/poliastro/core/elements.py
+++ b/src/poliastro/core/elements.py
@@ -71,9 +71,9 @@ def rv_pqw(k, p, ecc, nu):
 
     Returns
     -------
-    r: ndarray
+    r: numpy.ndarray
         Position. Dimension 3 vector
-    v: ndarray
+    v: numpy.ndarray
         Velocity. Dimension 3 vector
 
     Notes
@@ -151,9 +151,9 @@ def coe2rv(k, p, ecc, inc, raan, argp, nu):
 
     Returns
     -------
-    r_ijk: np.array
+    r_ijk: numpy.ndarray
         Position vector in basis ijk.
-    v_ijk: np.array
+    v_ijk: numpy.ndarray
         Velocity vector in basis ijk.
 
     Notes
@@ -508,9 +508,9 @@ def mee2rv(p, f, g, h, k, L):
 
     Returns
     -------
-    r: np.array
+    r: numpy.ndarray
         Position vector.
-    v: np.array
+    v: numpy.ndarray
         Velocity vector.
 
     Note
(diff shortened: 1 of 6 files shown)


Yash-10:
Thanks, @jerobado for this pull request!

While this solves one case, I see a few more cases in the docstrings where `np.array` is used instead of `numpy.array`, which can be seen by searching for "np.array" in the top-left corner in GitHub, for example. Could you also fix those?

Yash-10:
Also, I think `numpy.ndarray` would be a better choice than `numpy.array`.

jerobado:
Got it @Yash-10. I'll search for other cases. Thank you!

jerobado:
Hello @Yash-10, I was able to find these additional files that needs to updated:

```src/poliastro/core/elements.py```
```src/poliastro/core/util.py```
```src/poliastro/maneuver.py```

Can you kindly confirm if these two lines needs to be included as well?

https://github.com/poliastro/poliastro/blob/ada3fce2d26bdb020db38d54795d0b2a63d05760/src/poliastro/core/elements.py#L74

https://github.com/poliastro/poliastro/blob/ada3fce2d26bdb020db38d54795d0b2a63d05760/src/poliastro/core/elements.py#L76

Yash-10:
> Can you kindly confirm if these two lines needs to be included as well?

Great! Yes, they would also need to be changed to `numpy.ndarray` since just `ndarray` does not seem to help documentation identify it correctly.

Apart from this, I also realized that these files:

[src/poliastro/core/thrust/change_a_inc.py](https://github.com/poliastro/poliastro/blob/041525f130d7f67585149dbd7cd4834cea893016/src/poliastro/core/thrust/change_a_inc.py)
[src/poliastro/core/thrust/change_argp.py](https://github.com/poliastro/poliastro/blob/ada3fce2d26bdb020db38d54795d0b2a63d05760/src/poliastro/core/thrust/change_argp.py)
[src/poliastro/core/iod.py](https://github.com/poliastro/poliastro/blob/ada3fce2d26bdb020db38d54795d0b2a63d05760/src/poliastro/core/iod.py)

also have docstrings that show `numpy.array` which also would be a good thing to change to `numpy.ndarray`

Yash-10:
The documentation build error seems unrelated (not sure, though, about its cause) but the changes look good to me!

astrojuanlu (approved):
LGTM!

astrojuanlu:
Forcing a merge of this one, as @Yash-10 says the docs error seems unrelated

Issue #923, opened 2020-04-19
Return time information from `Orbit.sample()`


priyanshurohilla:
The sample method currently only returns a set of `CartesianRepresentation`, which has no associated frame and therefore no `obstime` attribute.

💡 **Feature request**

Provide a feature which can return a `GCRS` vector array in `sample()` function

priyanshurohilla:
While working on it I realized that the codebase has significatly changed from the previous version and will require some changes in the codebase. While trying to add this feature I realized some points which I need to discuss with you@astrojuanlu-

-  In `__init__()` function there is no `def frame()` which does not help in getting `frame_attributes` . While going through the codebase I realized there is a function called `def get_frame()` which has a similar functionality but does not allow `frame_attributes`

-  While trying to add this feature i realized that now  `__init__()`function does not take `plane` as a parameter ,which is posing a problem for the 'obstime' in `sample()` to work properly. Also i `def plane()` now returns `self.state_plane` so i think maybe it is also creating some problems in evolving with `obstime` of which i am not sure of.

Please guide me @astrojuanlu

priyanshurohilla:
Any updates on this @astrojuanlu ?

astrojuanlu:
Took me a while to respond to this one...

Maybe `sample` should return an `Ephemeris` instead of a `CartesianRepresentation`, so it has time information attached.

astrojuanlu:
gh-1521 addresses this by _not_ changing the `Orbit.sample` return value, and instead introducing a `Orbit.to_ephem` method that returns an `Ephem` object, containing both coordinates and epochs, as it should. We went full circle from returning a `BaseCoordinateFrame` to them returning an `Ephem`, but ours is a bit easier to understand I hope.

astrojuanlu:
Closed in gh-1521.

PR #1492, opened 2022-03-18
[DO NOT MERGE YET] State & orbit array API (iterative implementation)

s-m-e:
This PR aims at providing a full `OrbitArray` API and corresponding tests while relying on a slow iterative implementation based on `Orbit` first. This should allow the implementation of proper tests before cleanup and refactoring of other package internals begins.


astrojuanlu (commented):
Very high-level comments from our weekly meeting:

- Let's avoid allocating memory ourselves and accept an `out=` parameter instead
- @s-m-e will work on the tests before calling for review
- Let's break this down into pieces: first, the low hanging fruit (pytest-xdist, decorators going away, etc)
  - Then, we should open another PR for the `State` signature change (the epoch)

astrojuanlu:
While working on gh-1521, I realized it would have been cool to have a `StateArray` However, progress stalled there because we could not reach full agreement about what to do with the `epoch`.

In summary: @s-m-e advocates for having it in the `State` classes, while I was unsure because it's not needed at all for propagation and thus breaks Separation of Concerns. However, if we don't include it, end up with a data structure without times attached to it, which is related to #923. This forces us to either

a. have this information in a higher-level object, potentially introducing small complications in the implementation of some algorithms, or
b. adding it to the `State` anyway, thus potentially creating a structure that is heavier than needed.

I'll try to devote some thinking to this during SciPy.

astrojuanlu:
More arguments for keeping the epochs outside of StateArrays:

- Creates nice simmetry with Astropy: `State` objects resemble Coordinate objects from Astropy (which famously don't work with Keplerian elements, see https://github.com/astropy/astropy/issues/5067), `Orbit` and `Ephem` would be Frame instances from Astropy
- Creates nice internal simmetry: Orbit = State + epoch, Ephem = StateArray + epoch (which begs the question: to what extent a `OrbitArray` is needed, but again this deserves further discussion)

Issue #1496, opened 2022-04-06
The `astropy<5,>=3.2` dependency makes poliastro not installable on Python 3.10.2 since astropy-4.3.1 does not install on that version of python


danielk333:
🐞 **Problem**

When installing poliastro trough pip on python 3.10.2 it uses `astropy-4.3.1` which fails installation with the error message `ImportError: cannot import name 'soft_unicode' from 'markupsafe'`. 

🖥 **Please paste the output of following commands**

<details>
  
```bash
# Paste your output here:
(poliastro) danielk@IRF033-danielk ~/python_scripts> pip freeze
appdirs==1.4.4
application-utility==1.3.2
arandr==0.1.10
btrfsutil==5.16.2
CacheControl==0.12.6
ceph==1.0.0
ceph-volume==1.0.0
cephfs==2.0.0
cephfs-shell==0.0.1
certifi==2021.10.8
cffi==1.15.0
chardet==4.0.0
colorama==0.4.4
contextlib2==0.6.0.post1
crit==0.0.1
cryptography==36.0.1
Cython==0.29.28
distlib==0.3.4
distro==1.7.0
dnspython==2.2.0
docopt==0.6.2
gufw==21.4.0
html5lib==1.1
idna==3.3
importlib-metadata==4.8.1
keyutils==0.6
lensfun==0.3.3
lightdm-gtk-greeter-settings==1.2.2
lit==13.0.1.dev0
louis==3.21.0
manjaro-sdk==0.1
Markdown==3.3.6
menulibre==2.2.3
more-itertools==8.10.0
msgpack==1.0.3
mugshot==0.4.3
nftables==0.1
npyscreen==4.10.5
numpy==1.22.3
ordered-set==4.0.2
packaging==20.9
pacman-mirrors==4.23.2
pep517==0.12.0
pexpect==4.8.0
ply==3.11
progress==1.6
protobuf==3.19.4
psutil==5.9.0
ptyprocess==0.7.0
pycairo==1.20.1
pycparser==2.21
PyGObject==3.42.0
pyOpenSSL==21.0.0
pyparsing==3.0.1
PyQt5==5.15.6
PyQt5-sip==12.9.1
PySide6==6.2.3
python-distutils-extra==2.39
pyxdg==0.27
PyYAML==6.0
rados==2.0.0
ranger-fm==1.9.3
rbd==2.0.0
requests==2.27.1
resolvelib==0.5.5
retrying==1.3.3
rgw==2.0.0
[...]

Issue #1487, opened 2022-02-25
Unify coding style dependencies


jorgepiloto:
💡 **Feature request**

At the moment, there are three separate files holding coding style requirements:

1. The `pyproject.toml`. In this file, it is possible to find the style requirements in the `dev` group.
2. The `.pre-commit-config.yaml`. This file is not able to parse previous one, so dependencies need to be declared again.
3. The `tox.ini` file. Becasue the `[check]` environment does not install the project, we are forced to pass the required commands.

It is likely that, eventually, we miss the same version for a particular tool or that these three files end up being different from the point of view of the coding style. 

🎯 **Goal**

The goal is to unify all of them following these steps:

* All style requirements must be only defined in the `.pre-commit-config.yaml`.
* In the `tox.ini` file, the `[check]` style should read as follows:

   ```bash
    description = Checks for coding style
    skip_install = True
    deps = pre-commit
    commands = 
        pre-commit install
        pre-commit run --all-files
    ```
* Finally, we could get rid of the code style dependencies in the `pyproject.toml`. Unless those are being parsed by any third tool, I think previous approach unifies and simplifies the problem depicted before.

By the way, if users would like to use `pre-commit`, they can install it by doing `python -m pip install pre-commit`.

astrojuanlu:
We were hit again by this: https://github.com/poliastro/poliastro/pull/1495

PR #1497, opened 2022-04-15
Updated Version Pins to backport Python 3.10 support to 0.16.x

yakkonaut:
Updated tox.ini and pyproject.toml to pin black at 22.3.0 and astropy at >=5.0<6 to address issue 1496
https://github.com/poliastro/poliastro/issues/1496

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Astronomy",
 ]
 dependencies = [
-    "astropy >=3.2,<5",
+    "astropy >=5.0,<6",
     "astroquery >=0.3.9",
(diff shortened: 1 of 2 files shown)


yakkonaut:
As noted in matrix chat- tox is failing here because of black- the updated version wants to modify 52 files. Most of those mods exist in main. Not sure if right move is to pull main in 0.16.x or just submit a PR that includes the black changes. Thoughts?

astrojuanlu:
Hi @yakkonaut, thanks for this! I believe it's OK to push another commit with the fixes. It is not very ergonomic because of https://github.com/poliastro/poliastro/issues/1487, so you will have to change the black version in 3 places.

astrojuanlu:
I believe it will be better to backport #1498 first.

PR #1498, opened 2022-04-17
Unify code style tools and configuration files

### LINKED ISSUE #1487: Unify coding style dependencies

jorgepiloto:
Resolves #1487 by collecting required code style tools under the `pre-commit-config.yaml`. All configuration files have been moved to `pyproject.toml` **except flake8**, see https://github.com/PyCQA/flake8/issues/234.

In addition, `check` and `reformat` Tox environments have been unified under a common `style` one. Now, the code style is executed every time a `tox -e style ` command is applied.


setup.cfg
```diff
@@ -1,24 +1,3 @@
...
-    test_*.py
-doctest_plus = disabled
-addopts = --strict-markers
-markers =
-    slow
-    remote_data
-    filterwarnings
-    mpl_image_compare
-
 [flake8]
```
astrojuanlu, line 1: I see that `flake8` does not support `pyproject.toml` and probably it will never do https://github.com/PyCQA/flake8/issues/234#issuecomment-812800722 I understand the reasoning, but it's a pity anyway

astrojuanlu (approved):
This is excellent, thanks a bunch @jorgepiloto! I love how most of the config converged into `pyproject.toml` and how style checking and reformatting became much simpler.

Issue #1513, opened 2022-05-31
Potential units mismatch in `atmospheric_drag`


Nosudrum:
🐞 **Problem**

Per its [documentation](https://docs.poliastro.space/en/stable/autoapi/poliastro/core/perturbations/index.html#poliastro.core.perturbations.atmospheric_drag), `atmospheric_drag` in `poliastro.core.perturbations` is computed as follows :

$$ \vec{p} = -\frac{1}{2}\rho v\left ( \frac{C_{d}A}{m} \right )\vec{v} $$

with :
- $\rho$ in `kg/m3`
- $v$ in `km/s`
- $C_{d}$ unitless
- $A/m$ in `km2/kg`

Mixing `m` (from $\rho$) and `km` (from $\vec{v}$) obviously cannot lead to correct units for the perturbation acceleration $\vec{p}$, unless the quantities are all associated to a unit and conversion happens automatically.

This is as I understand it not the case, as `poliastro.core` does not work with units anymore (per #1256 ?). Indeed, the cartesian `state` used by `atmospheric_drag` contains values in `km` and `km/s` but without using said units linked in `quantity` objects.

Therefore, when mixed with $\rho$ in `kg/m3` following the documentation there will be a units mismatch with no automatic conversion.

You can see that behaviour for yourself with the following code :

```python
import numpy as np
from astropy import units as u
from numpy.linalg import norm
from poliastro.bodies import Earth
from poliastro.core.perturbations import atmospheric_drag
from poliastro.earth.atmosphere import jacchia

atm_jacchia = jacchia.Jacchia77(1050 * u.K)
alt = 500 * u.km
rho = atm_jacchia.density(alt)

t0 = 0 * u.s
state = np.array([7200, 0, 0, 0, -1.03551869, 7.36809836])  # in km and km/s
k = Earth.k.to(u.km ** 3 / u.s ** 2)  # does not matter here
C_D = 1.2 * u.one
A = 20 * u.m ** 2
m = 2000 * u.kg
A_over_m = (A / m).to(u.km ** 2 / u.kg)  # conversion to km2/kg as per documentation

# Compute the drag acceleration using the poliastro function
acc_drag = atmospheric_drag(t0, state, k, C_D, A_over_m, rho)
print(f'Drag acceleration (poliastro): {acc_drag}')

# Compute the drag acceleration using the same code as the poliastro function
v_vec = state[3:]
v = norm(v_vec)
B = C_D * A_over_m

[...]

PR #1514, opened 2022-05-31
Fix rho unit for atmospheric_drag

Nosudrum:
Changed documentation unit for $\rho$ from `kg/m3` to `kg/km3` to match other units, as suggested in #1513.


astrojuanlu (approved):
Congratulations on your first contribution to poliastro! 🚀

PR #1515, opened 2022-06-02
Fixed unit mismatch in `radiation_pressure`

Nosudrum:
`Wdivc_s` was requested in `W * s / km` (W contains distance units in meters), which did not work with the other parameters requested in km, kg and s.


astrojuanlu:
Woops, good catch!

astrojuanlu:
Thanks a lot!

PR #1521, opened 2022-06-22
Rework Orbit <-> Ephem conversion

astrojuanlu:
This aims to solve #1364, #923, and #921 all at once (risky, I know).

### Summary

- 💥 `poliastro.twobody.propagation.propagate` is gone! Users must now decide between `Orbit.propagate` for a single time of flight or `Orbit.to_ephem(strategy=EpochsArray(...))` for an array of times of flight
  - A (hopefully) helpful error message is generated when the function is called
- 💥 States API changed, now elements are passed as a 6-tuple
  - Old usage will raise "too many parameters" errors
- 💥 Propagation API changed, now all propagators are classes and must be initialized with extra parameters before they can be used
  - A (hopefully) helpful error message is generated when the modules are accidentally called
- ⚠️ `Orbit.sample` with min and max anomaly is deprecated in favor of `Orbit.to_ephem`
- ⚠️ `Orbit.sample` no longer emits `OrbitSamplingWarning` when anomalies are out of range


astrojuanlu:
Working on this, I realize it would be cool to have a `StateArray` @s-m-e, that's the first thing we should extract from gh-1491 and gh-1492. Leaving a comment there for future reference.

astrojuanlu:
21 tests to go. Code is still a bit rough, but getting there.

astrojuanlu:
Many tests failing still, but made significant progress: successfully implemented https://github.com/poliastro/poliastro/issues/1364#issuecomment-1174173401 and I'm happy how it looks. This requires a bit more work though.

astrojuanlu:
Almost all tests fixed locally, although git history is a bit messy. Still working on it.

astrojuanlu:
Fast tests passing :rocket: interestingly, some propagators have lower accuracy now that we're converting from classical elements only for the tests 🤔 which makes me think that maybe `coe2rv` is not as precise as it should be, or that it's conditioning our propagation tests somehow. But we will revisit this after the release.

astrojuanlu:
Added an error message for old propagators and the `propagate` function so that users that upgrade at least have a hint on what to do ✔️

astrojuanlu:
Note to reviewers: commit 98cb5ea applied `black` to all notebooks, so it should probably be ignored.

docs/source/examples/Natural and artificial perturbations.mystnb
```diff
@@ -81,18 +88,15 @@ def f(t0, state, k):
...
 
-rr = propagate(
-    orbit,
-    tofs,
-    method=cowell,
-    f=f,
-)
+rr, _ = orbit.to_ephem(
+    EpochsArray(orbit.epoch + tofs, method=CowellPropagator(f=f)),
+).rv()
```
astrojuanlu, line 93: This is how users are expected to transition from `propagate` to `Orbit.to_ephem`.

docs/source/examples/Natural and artificial perturbations.mystnb
```diff
@@ -81,18 +88,15 @@ def f(t0, state, k):
...
 ```
 
-```{code-cell} ipython3
+```{code-cell}
 plt.ylabel("h(t)")
 plt.xlabel("t, days")
-plt.plot(tofs.value, rr.norm() - Earth.R)
+plt.plot(tofs.value, norm(rr, axis=1) - Earth.R)
```
astrojuanlu, line 99: Since `Ephem.rv()` returns `Quantity` objects rather than `CartesianRepresentation` objects, I extended our `poliastro.util.norm` function to work with `axis=1` too.

docs/source/examples/Natural and artificial perturbations.mystnb
```diff
@@ -223,38 +229,34 @@ def f(t0, state, k):
...
 )
 ```
 
-```{code-cell} ipython3
+```{code-cell}
 frame = OrbitPlotter3D()
 
 frame.set_attractor(Earth)
-frame.plot_trajectory(rr, label="orbit influenced by Moon")
+frame.plot_ephem(ephem, label="orbit influenced by Moon")
```
astrojuanlu, line 241: Notice that sometimes the `Ephem` object can be used directly!

src/poliastro/core/propagation/__init__.py
```diff
@@ -30,26 +33,3 @@
     "recseries_coe",
     "recseries",
 ]
-
-
-@jit
-def func_twobody(t0, u_, k):
-    """Differential equation for the initial value two body problem.
```
astrojuanlu, line 37: This was just moved to another submodule

src/poliastro/core/propagation/cowell.py
```diff
@@ -0,0 +1,49 @@
+import numpy as np
+
+from poliastro._math.ivp import DOP853, solve_ivp
+from poliastro.core.propagation.base import func_twobody
+
+
+def cowell(k, r, v, tofs, rtol=1e-11, *, events=None, f=func_twobody):
```
astrojuanlu, line 7: This is essentially like the old function, but I moved it to `core` and stripped it from unit handling.

src/poliastro/core/propagation/pimienta.py
```diff
@@ -10,7 +10,7 @@ def pimienta_coe(k, p, ecc, inc, raan, argp, nu, tof):
 
     q = p / (1 + ecc)
 
-    # TODO: Do something to increase parabolic accuracy?
+    # TODO: Do something to allow parabolic and hyperbolic orbits?
     n = np.sqrt(k * (1 - ecc) ** 3 / q**3)
```
astrojuanlu, line 14: Notice that, if `ecc == 1`, `n == 0`, which makes all the rest fail.

tests/tests_twobody/test_events.py
```diff
@@ -55,16 +55,13 @@ def f(t0, u_, k):
         du_ad = np.array([0, 0, 0, ax, ay, az])
         return du_kep + du_ad
 
-    rr, _ = cowell(
-        Earth.k,
-        orbit.r,
-        orbit.v,
+    method = CowellPropagator(events=events, f=f)
+    rr, _ = method.propagate_many(
+        orbit._state,
```
astrojuanlu, line 60: Using `method.propagate_many` here to make it more explicit that we're testing a `CowellPropagator` method instead of using `Orbit.to_ephem`. But we could as well change all these for `orbit.to_ephem(method=CowellPropagator(...))`.

astrojuanlu:
I fixed one more bug locally and I think this is good to go already. If I find something else, I'll send another pull request. Let's go 🚀

Issue #1519, opened 2022-06-21
'lat' and 'lon' swapped in core.spheroid_location.py


zfunke7:
This issue is at least affecting the cartesian_cords method:

    e2 = 1 - (_c / _a) ** 2
    N = _a / np.sqrt(1 - e2 * np.sin(_lon) ** 2)

    x = (N + _h) * np.cos(_lon) * np.cos(_lat)
    y = (N + _h) * np.cos(_lon) * np.sin(_lat)
    z = ((1 - e2) * N + _h) * np.sin(_lon)
    return x, y, z

Note that the Z coordinate depends on latitude (N/S value), not longitude.

See [stack overflow](https://stackoverflow.com/questions/1185408/converting-from-longitude-latitude-to-cartesian-coordinates) for the spherical simplification, but illustrates the mix-up:

"
The conversion is:

x = R * cos(lat) * cos(lon)

y = R * cos(lat) * sin(lon)

z = R *sin(lat)
"

WolfsSky:
@astrojuanlu if @zfunke7 does not take this one, I am willing to use it as a task to get back into helping poliastro.

astrojuanlu:
@applysomesky it's been a few days, so feel free! :)

astrojuanlu:
Closed in gh-1537 and gh-1547.

PR #1537, opened 2022-07-14
Removed underscores, fixed lat and lon usage in cartesian_cords

applysomesky:
This was a small fix for issue #1519 . I am not able to validate the `cartesian_cords` at this moment. Could allocate more time for that if desired.

I wasn't sure what `N` was in `cartesian_cords` function and if the `lon` should be `lat`, so I tried to look it up and came across this [PDF](https://www.researchgate.net/publication/330201994_Cartesian_to_geodetic_coordinates_conversion_on_an_oblate_spheroid_using_the_bisection_method). Here it lists `N` in a different format but on [wikipedia](https://en.wikipedia.org/wiki/Earth_radius#Radii_of_curvature) it lists the actual used (simplified) method for `N`, derived from the one in the PDF. So here I also had to change `lon` to `lat`.

The `(c/a)^2` in the PDF is obtained by using `1-e^2`, where `e^2 = 1 - (c / a) ** 2`.

Leaving this info just in case.

### DIFF
--- a/src/poliastro/core/spheroid_location.py
+++ b/src/poliastro/core/spheroid_location.py
@@ -7,45 +7,46 @@ from poliastro._math.linalg import norm
 
 
 @jit
-def cartesian_cords(_a, _c, _lon, _lat, _h):
+def cartesian_cords(a, c, lon, lat, h):
     """Calculates cartesian coordinates.
 
     Parameters
     ----------
-    _a : float
+    a : float
         Semi-major axis
-    _c : float
+    c : float
         Semi-minor axis
-    _lon : float
+    lon : float
         Geodetic longitude
-    _lat : float
+    lat : float
         Geodetic latitude
-    _h : float
+    h : float
         Geodetic height
 
     """
-    e2 = 1 - (_c / _a) ** 2
-    N = _a / np.sqrt(1 - e2 * np.sin(_lon) ** 2)
-
-    x = (N + _h) * np.cos(_lon) * np.cos(_lat)
-    y = (N + _h) * np.cos(_lon) * np.sin(_lat)
-    z = ((1 - e2) * N + _h) * np.sin(_lon)
+    
+    e2 = 1 - (c / a) ** 2
+    N = a / np.sqrt(1 - e2 * np.sin(lat) ** 2)
+    
+    x = (N + h) * np.cos(lat) * np.cos(lon)
+    y = (N + h) * np.cos(lat) * np.sin(lon)
+    z = ((1 - e2) * N + h) * np.sin(lat)
     return x, y, z
 
 
 @jit
-def f(_a, _c):
+def f(a, c):
     """Get first flattening.
 
     Parameters
     ----------
-    _a : float
+    a : float
         Semi-major axis
-    _c : float
+    c : float
         Semi-minor axis
 
     """
-    return 1 - _c / _a
+    return 1 - c / a
 
 
 @jit
@@ -89,21 +90,21 @@ def tangential_vecs(N):
 
 
 @jit
-def radius_of_curvature(_a, _c, _lat):
+def radius_of_curvature(a, c, lat):
     """Radius of curvature of the meridian at the latitude of the given location.
 
     Parameters
     ----------
-    _a : float
+    a : float
         Semi-major axis
-    _c : float
+    c : float
         Semi-minor axis
-    _lat : float
+    lat : float
         Geodetic latitude
 
     """
-    e2 = 1 - (_c / _a) ** 2
-    rc = _a * (1 - e2) / (1 - e2 * np.sin(_lat) ** 2) ** 1.5
+    e2 = 1 - (c / a) ** 2
+    rc = a * (1 - e2) / (1 - e2 * np.sin(lat) ** 2) ** 1.5
     return rc
 
 
@@ -156,7 +157,7 @@ def is_visible(cartesian_cords, px, py, pz, N):
 
 
 @jit
-def cartesian_to_ellipsoidal(_a, _c, x, y, z):
+def cartesian_to_ellipsoidal(a, c, x, y, z):
     """
     Converts cartesian coordinates to ellipsoidal coordinates for the given ellipsoid.
     Instead of the iterative formula, the function uses the approximation introduced in
@@ -164,9 +165,9 @@ def cartesian_to_ellipsoidal(_a, _c, x, y, z):
 
     Parameters
     ----------
-    _a : float
+    a : float
         Semi-major axis
-    _c : float
+    c : float
         Semi-minor axis
     x : float
         x coordinate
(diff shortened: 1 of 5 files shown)


applysomesky:
The first test fail correspond to mix up of lat and lon.

The test expects `lat = 38.43`, `lon = 41.2`. As I can confirm with `orbit_predictor` package `llh2ecef`. Resulting in:
```
(3764.2586478469702, 3295.3593385548143, 3942.9452857930723)
```
The test result somehow corresponds to inputs `lat = 41.2`, `lon = 38.43`. Resulting in:
```
(3764.859301272756, 2987.2016749669842, 4179.160715400216)
```

I am not familiar with your standards for order of `lat` and `lon`. `el_cords` shows 2 `u.deg` values but I am unable to say which is which. Maybe the unit test input is wrong way round and puts it in the `lat`, `lon` order, while `cartesian_cords ` does currently expect `lon`, `lat` order.

astrojuanlu:
Thanks a lot for this pull request @applysomesky!

> I am not familiar with your standards for order of lat and lon.

Let's settle on (lon, lat) forever, since it resembles (x, y) and also because [it's the right way](https://macwright.com/2016/07/15/longitude-latitude-is-the-right-way.html).

applysomesky:
I wrote a small script that runs the corrected `cartesian_cords` versus `llh_to_ecef ` from `orbit_predictor` package.
It produces the same results for the same `lon`, `lat`.

```
__ orbit_predictor llh_to_ecef __
lon: 38.43, lat: 41.2
result op [3764.86, 2987.2, 4179.16]
__ poliastro cartesian_cords __
el_cords (lon,lat): (38.43, 41.2, 0)
result pa [3764.86, 2987.2, 4179.16]
```

However the unit test for the `el_cords` expects something else, which is:
```
expected_cords unit test: [3764258.64785411, 3295359.33856106, 3942945.28570563]
```
Switching the lat and lon on input in both functions results in the same outcome as the unit test expects:
```
__ orbit_predictor llh_to_ecef __
lon: 41.2, lat: 38.43
result op [3764.26, 3295.36, 3942.95]
__ poliastro cartesian_cords __
el_cords (lat,lon): (38.43, 41.2, 0)
result pa [3764.26, 3295.36, 3942.95]
```
This strongly suggest the unit tests switched `lat` and `lon` as input on its creation.

astrojuanlu:
> This strongly suggest the unit tests switched lat and lon as input on its creation.

Sounds like it, yes. Do you want to make the change in the test as well? If I understand correctly, that should fix the CI.

applysomesky:
I will have a go at it.

applysomesky:
The last one failing is `test_h_calculation_near_lat_singularity`.

In the last commit I found a mistake in `cartesian_to_ellipsoidal` when `lat == 0.0`, it listed `x` in the numerator whilst on [wikipedia](https://en.wikipedia.org/wiki/Geographic_coordinate_conversion#Coordinate_system_conversion) and literature (source no. 6 on wiki page) for example it lists `sqrt(x**2 + y**2)`. So I changed that, and on my machine produces the correct result, converting forth and back from Geodetic to Cartesian and back.

It does not fail on the `lat == 0.0`, but as follows:
```
Falsifying example: test_h_calculation_near_lat_singularity(
    lat=<Quantity 5.e-324 rad>,
)
```

I am personally not sure how to solve this at the moment. Maybe I can reiterate on `cartesian_to_ellipsoidal` and see if there is another small mistake. It seems to produce wrong results for values for `lat` smaller than `5e-315`, but correct results at values bigger than that. 

Love to hear your thoughts.

applysomesky:
An option would be to make 2 `test_h_calculation_near_lat_singularity` tests, one for the positive and one for the negative side such that we can limit the smallest number to be bigger than about `5e-315`. Not sure if testing smaller numbers has any use. Problem is that if someone uses a latitude smaller than that the function will return a wrong value for `h`.

Otherwise one can convert latitudes smaller than `5e-315` to 0, with a user warning. Also not sure if that is desirable.

astrojuanlu:
Thank you for your perseverance here. These numerical errors near singularities are tricky. 1e-300 is ridiculously small, maybe we can set the cutoff at 1e-18 or 1e-5 even! I'd be against conversion, if possible.

applysomesky:
So just to be clear, so you want to limit the unit test inputs? 

In which case we need to apply that limit somehow in this part of the test:
```
@settings(deadline=None)
@given(
    lat=with_units(
        elements=st.floats(min_value=-1e-2, max_value=1e-2), unit=u.rad
    ),
)
```

This might mean that the function behaves incorrectly if a user uses a value smaller than `1e-315`, if that ever happens.

src/poliastro/core/spheroid_location.py
```diff
@@ -176,18 +177,22 @@ def cartesian_to_ellipsoidal(_a, _c, x, y, z):
...
-        (z + e2_ * _c * np.sin(th) ** 3) / (p - e2 * _a * np.cos(th) ** 3)
+        (z + e2_ * c * np.sin(th) ** 3) / (p - e2 * a * np.cos(th) ** 3)
     )
 
-    v = _a / np.sqrt(1 - e2 * np.sin(lat) ** 2)
-    h = x / np.cos(lat) - v if lat == 0.0 else z / np.sin(lat) - (1 - e2) * v
+    v = a / np.sqrt(1 - e2 * np.sin(lat) ** 2)
+    h = (
+        np.sqrt(x**2 + y**2) / np.cos(lat) - v
+        if lat == 0.0
```
astrojuanlu, line 194: Instead of `lat == 0.0` we could do `abs(lat) < 1e-18` or something like that?

astrojuanlu:
> So just to be clear, so you want to limit the unit test inputs?

Rather than that, replace the `lat == 0.0` check inside the function with something more flexible, like `abs(lat) < 1e-18`. I am hoping that the function is continuous and that we don't see further test failures? Just left a comment on where exactly we could do the change.

applysomesky:
Using the latitude of `5e-315 rad` did not produce errors on my machine with the proposed change.

I just want to note that the back and forth conversion at these small numbers for latitude results in very small changes in the latitude.

applysomesky:
I also observed that `cartesian_to_ellipsoidal` returns `lat, lon, h`, so in the wrong order that what is to be expected (`lon, lat, h`) 🙃.

Changing this would require probably the unit test to be rewritten as well, and might have implications for backwards compatibility.

astrojuanlu:
Glad it's working finally 🎉 

> Changing this would require probably the unit test to be rewritten as well, and might have implications for backwards compatibility.

I'm happy if you want to pursue that change too (we are not _that_ careful about backwards compatibility outside the `Orbit` and `Ephem` objects, which are the true public API), but I'm also happy to merge already 🙂

applysomesky:
Let me fix it, I like consistency :)

astrojuanlu:
Superseded by gh-1547 (just rebased on top of `main` to solve the merge conflicts), thanks a lot @applysomesky! 🙌🏽

Issue #1535, opened 2022-07-11
Sometimes OMM notebook fails


astrojuanlu:
https://github.com/astropy/astropy/issues/13447

astrojuanlu:
- Failed: https://readthedocs.org/projects/poliastro/builds/17408507/
- Passed: https://readthedocs.org/projects/poliastro/builds/17430049/

At least the `stable` version of the docs is built now. Will add some debugging statements so that the next time it fails we hopefully have more information.

astrojuanlu:
This hasn't happened in a while. Closing until it does again.

jorgepiloto:
This just hitted me suddenly. Docs were rendering without problem until now.

PR #1539, opened 2022-07-16
Add debugging information to OMM notebook

astrojuanlu:
See gh-1535.


astrojuanlu:
I'm going to keep rebuilding this until it fails 😉

Issue #1532, opened 2022-07-10
Restore Sphinx strict mode


astrojuanlu:
We had to disable `-W` mode in Sphinx after turning on inherited members in sphinx-autoapi (comes from gh-1531).

Now there are several `WARNING: Duplicate explicit target name: "1".` because now I'm telling `sphinx-autoapi` to show the inherited properties, and this results in repeated `.. [1] https://en.wikipedia.org/wiki/Great-circle_distance` on the same page because they're inherited from Astropy.

_Comes from https://github.com/poliastro/poliastro/pull/1531#discussion_r917403669_

PR #1545, opened 2022-07-17
New plotters architecture with backend support

### LINKED ISSUE #1544: Refactor plotters
After @sudughonge and @jorgepiloto tried to work on gh-1361, gh-1538, and gh-1525, it became clear that the current implementation of the OrbitPlotters is too difficult to understand. To summarize:

- The architecture is "smart" but for humans reading the code there's lots of back and forth between abstract and inherited methods, making it very difficult to understand where the actual plotting is taking place.
- `OrbitPlotter2D/3D` inherit from `_PlotlyOrbitPlotter`, which makes it difficult to implement any other 2D or 3D plotter that is not based on Plotly.

Last refactoring of the plotting happened in gh-528, 4 years ago. It solved some long-standing issues with plotting, but it's probably time to give it a fresh look.

What @jorgepiloto and I discussed is that

1. The `OrbitPlotter*` interface must remain the same: `.plot()`, `.plot_ephem()`, `.plot_maneuver()`, `.plot_body_orbit()`, `.set_attractor()`, `.set_view()`, `.show()` (maybe we can get rid of `.plot_trajectory()` and `.trajectories` at this point by the way).
2. We should probably have `OrbitPlotter*` objects contain a `Backend` object that does the actual plotting. With this architecture, `OrbitPlotter` would do the "logic", and `Backend` would do the backend-specific tasks (e.g. `fig.add_trace` for Plotly vs `???` for PyVista).

In December 2018 I wrote ["TL;DR: Say "no" to backends."](https://github.com/poliastro/poliastro/wiki/Plotting) but I think this only applies to separating matplotlib and "the rest" (worth noting that ipympl still has some issues, see https://github.com/matplotlib/ipympl/issues/359 and https://github.com/matplotlib/ipympl/issues/473). Plotly and PyVista probably work in a similar way (ipywidgets?), so "it shouldn't be that hard"™ to have them both as interactive backends, while keeping the `StaticOrbitPlotter` separate.

...Having said that, `StaticOrbitPlotter` is probably a bad name for a backend-specific plotter: we could potentially have "static" plots based on Plotly and https://github.com/plotly/Kaleido. Therefore, this deserves some thinking too.

### LINKED ISSUE #1591: Background examples in latest documentation is empty
🐞 **Problem**
The [Background section](https://docs.poliastro.space/en/latest/background.html) in poliastro's latest documentation does not show any of the examples.

### LINKED ISSUE #1603: Can't set orbit plot axis units
When using StaticOrbitPlotter() you can customize certain things such as trajectory colors, markers, title, grid, etc. However, I can't find any documentation on the poliastro site on how, if even possible, to change the units used for the x and y axis, which are always shown in km.  

Having the ability to change from km to astronomical units is essential when working, for example, with asteroid orbits, due to them so much more expansive compared to satellites or crafts.

I tried using ax.yaxis.set_units(u.AU) from matplotlib and astropy, just to see if that could work but it didn't do anything. 

If there is indeed a way to change the units from km to au, please let me know.

jorgepiloto:
Resolves #1544. Fix #1591 too. Fix #1603.

Main advantages of this pull-request are:

* Unify all orbit plotters under a common `OrbitPlotter` that accepts a `backend_name`. This can be any of the `SUPPORTED_ORBIT_PLOTTER_BACKENDS`.
* Unify all backends features and facilitate their maintenance and extension.
* Select the best backend according to the environment in which the code gets executed when running `Orbit.plot`. poliastro decides whether to use matplotlib2D or plotly2D if no backend has been specified.


jorgepiloto:
Opening this in draft for better discussing about the implementation.

astrojuanlu (commented):
Very quick comments:

- Not sure if we need a `orbit_plotters` sub package? (Or maybe yes, given that now we have groundtrack plotters, tisserand, etc)
- In my mind, `OrbitPlotter` should have `plot`, `set_attractor`, etc as discussed in gh-1544, and `OrbitPlotterBackend` should have `draw_point`, `draw_sphere`, `draw_coordinates` (from the current `plot_coordinates`). Then the `OrbitPlotter` methods can call the corresponding `Backend` methods.

astrojuanlu:
About the sub packages, what about

```
plotting
|- orbit/...
|- gabbard.py
|- porkchop.py
|- tisserand.py
```

(so `orbit/` instead of `orbit_plotters/`?)

src/poliastro/plotting/orbit/backends/base.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(self, backend):
+        """Initialize the orbit plotter backend.
+
+        Parameters
+        ----------
+        backend : str
+            Name of the backend.
+
+        """
+        self._backend = backend
```
astrojuanlu, line 16: `backend_name`?

src/poliastro/plotting/orbit/backends/matplotlib_backend.py
```diff
@@ -0,0 +1,509 @@
+"""A module implementing matplotlib backend for orbit plotters."""
+
+from poliastro.plotting.orbit_plotters.backends.base import (
+    Trajectory,
+    _OrbitPlotterBackend,
+)
+
+
+class OrbitPlotterMatplotlib2D(_OrbitPlotterBackend):
```
astrojuanlu, line 9: I was thinking that we could have something like this:

```python
class OrbitPlotterMatplotlib2D:  # Notice it doesn't inherit from the backend
    def __init__(self):
        self._backend = _MatplotlibBackend2D(...)

    def plot_coordinates(self, ...):
        # Prepare data
        ...
        self._backend.draw_line(...)  # What now is called "draw_coordinates"
        # Wrap up
        ...
```

in sum:

- A plotter _has_ a backend (rather than "a plotter _is_ a backend") - i.e. [composition over inheritance](http://en.wikipedia.org/wiki/Composition_over_inheritance)
- The plotter takes care of preparing the inputs and the backend only takes care of drawing primitives (`draw_line` for coordinates, `draw_sphere` for bodies, `draw_arrow` for impulses)

I think this should decouple the logic in a more effective way and avoid having too many methods.

jorgepiloto: Really like the approach you presented in here. I was over complicating things.

jorgepiloto:
Working on this. I would like to have it ready for review by the end of this weekend.

jorgepiloto:
Updates on this:

The layout for the `poliastro.plotters.orbit` looks like this:

```text
poliastro/plotters
└── orbit
    ├── __init__.py
    ├── backends
    │   ├── __init__.py
    │   ├── base_backend.py
    │   ├── matplotlib_backend.py
    │   └── plotly_backend.py
    └── plotter.py
```

The desired backend can be specified using one of the keys of the `poliastro.plotters.orbit.backends.SUPPORTED_ORBIT_PLOTTER_BACKENDS` by doing `OrbitPlotter(backend_name=...)`.

> An `OrbitPlotter` *contains* an instance of the desired backend, using composition as suggested by @astrojuanlu.

The backend instance associated to the plotter gets initialized using a `scene` object. This scene object is particular to each backend:

* Matplotlib backends -> scene = matplotlib.Axes
* Plotly backends -> scene = plotly.Figure

This allows to do:

```python
OrbitPlotter(scene=a_mpl_ax, backend="matplotlib2D")
OrbitPlotter(scene=a_plotly_fig, backend="plotly2D")
```

If no scene gets assigned, each backend will create a default one that is capable to work with.

jorgepiloto:
I bet some of the tests are failing because I renamed those. Thus, their base images need to be renamed too.

jorgepiloto:
The `matplotlib2D` backend is now complete. However, the `plotly2D` and `plotly3D` are still missing.

jorgepiloto:
Note that I had to update some baseline images as the tests got renamed.

jorgepiloto:
Still need to fix some of the tests by updating their baseline images for matplotlib.

jorgepiloto:
Update on the status of this pull-request:

- [ ] Need to implement `plotly2D` and `plotly3D`
- [ ] Update documentation
- [ ] Make sure coverage did not reduce
- [ ] Regenerate baseline images (if required)
- [ ] Check images

jorgepiloto:
This is on the right way.

I discovered various issues with our plotters while working on this. All of those are now solved.

It is much easier to integrate things now. We have to pay more lines of code, but from the point of view of maintenance, it should be much easier now to implement new features in the plotters.

jorgepiloto:
Must confess I am still amazed by how @astrojuanlu was able to integrate everything into a common interface and simplify lots of lines.

astrojuanlu:
Wohooo this looks promising! Amazing job @jorgepiloto 👏🏽

jorgepiloto:
The matplotlib backend returns a trace for coordinates and position and no longer a list. This allows to get rid of those list indices in the [Customizing static plots notebook](https://docs.poliastro.space/en/latest/examples/Customising%20static%20orbit%20plots.html).

jorgepiloto:
All plotters are working as expected. The last feature to implement is the ability to select the distance units.

This feature is only available in the plotly backend but with this new refactor, it is easy to enable it in all the backends!

jorgepiloto:
All plotters support now custom length units! 🎉

jorgepiloto:
Everything is working as expected.

The baseline images that got generated for the new test names show a black colored edge for the position markers. These do not appear in the usual images.

I am going to force the edge color for the markers in matplotlib backend to guarantee that it is the same as the marker one.

Hopefully, the only thing that changes is the filename for the baseline image and not its content.

jorgepiloto:
I am fixing an issue related with the coverage. Apparently, when running `tox.ini` it does no longer generate any report. Making some tests locally.

jorgepiloto:
Fixed. Coverage works as expected locally. Let see how this affects codecov.

jorgepiloto:
I decided to update all the baseline images. All of them look the same and all image tests were passing locally.

However, there was one failing complaining about the size of the images. This test was passing locally. The only difference I saw was the fact that the CI machine runs Python 3.8 while I develop under Python 3.10...

jorgepiloto:
It is absurd because I had to commit all the baseline images. Using the GitHub image diff tool, it does not show any difference at the naked eye. Maybe something changed at metadata level but user experience and the end result is the same.

src/poliastro/bodies.py
```diff
@@ -134,49 +134,30 @@ def plot(
...
-            return OrbitPlotter3D(plane=plane).plot_body_orbit(
-                self, epoch, label=label
-            )
-        else:
-            from poliastro.plotting.interactive import OrbitPlotter2D
-
-            return OrbitPlotter2D(plane=plane).plot_body_orbit(
-                self, epoch, label=label
-            )
+        # HACK: import here the OrbitPlotter to avoid a circular depedency
```
astrojuanlu, line 154:
```suggestion
        # HACK: import here the OrbitPlotter to avoid a circular dependency
```

src/poliastro/plotting/misc.py
```diff
@@ -15,75 +14,61 @@
...
-    orbit_plotter = OrbitPlotter3D(plane=Planes.EARTH_ECLIPTIC)
-
-    _plot_bodies(orbit_plotter, outer, epoch)
-
-    return orbit_plotter
-
-
-def plot_solar_system(outer=True, epoch=None, use_3d=False, interactive=False):
+def plot_solar_system(
+    backend_name="matplotlib2D", epoch=None, labels=None, outer=True
```
astrojuanlu, line 21: Maybe I'd make the `backend_name` the last parameter, to avoid breaking backwards compatibility?

src/poliastro/plotting/orbit/backends/base_backend.py
```diff
@@ -0,0 +1,232 @@
+"""A module implementing the base class for a orbit plotter backend."""
+
+
+class _OrbitPlotterBackend:
```
astrojuanlu, line 4: Maybe rename `backends/base_backend.py` to `backends/_base.py` + rename `_OrbitPlotterBackend` to `OrbitPlotterBackend`? (i.e. make the module private, but the class public)

src/poliastro/plotting/orbit/backends/base_backend.py
```diff
@@ -0,0 +1,232 @@
...
+    """A base class for implementing new orbit plotter backends."""
+
+    def __init__(self, scene, name, ref_units):
+        """Initialize the orbit plotter backend.
+
+        Parameters
+        ----------
+        scene : object
```
astrojuanlu, line 12: What types are allowed for `scene`? (This would help understanding the implementation I believe)

src/poliastro/plotting/orbit/backends/matplotlib_backend.py
```diff
@@ -0,0 +1,282 @@
+"""A module implementing orbit plotter backends based on Matplotlib."""
```
astrojuanlu, line 1: Maybe rename the module from `backends/matplotlib_backend.py` to `backends/matplotlib.py`?

src/poliastro/plotting/orbit/backends/matplotlib_backend.py
```diff
@@ -0,0 +1,282 @@
...
+def _segments_from_arrays(x, y):
+    # Copied pasted from
+    # https://matplotlib.org/3.1.0/gallery/lines_bars_and_markers/multicolored_line.html
+    # because this API is impossible to understand
+    points = np.column_stack([x, y])[:, None, :]
+    segments = np.concatenate([points[:-1], points[1:]], axis=1)
+    return segments
+
+
+class OrbitPlotterBackendMatplotlib2D(_OrbitPlotterBackend):
```
astrojuanlu, line 20: These class names are rather long but I don't have good ideas on how to shorten them 😅

src/poliastro/plotting/orbit/backends/matplotlib_backend.py
```diff
@@ -0,0 +1,282 @@
...
+    """An orbit plotter backend class based on Matplotlib."""
+
+    def __init__(self, ax, use_dark_theme, ref_units):
+        """Initializes a backend instance.
+
+        Parameters
+        ----------
+        scene : object
```
astrojuanlu, line 28: Docstring doesn't match `__init__` signature

src/poliastro/plotting/orbit/backends/plotly_backend.py
```diff
@@ -0,0 +1,476 @@
+"""A module implementing orbit plotter backends based on Plotly."""
```
astrojuanlu, line 1: Same here, maybe rename to `backends/plotly.py`?

src/poliastro/plotting/orbit/backends/plotly_backend.py
```diff
@@ -0,0 +1,476 @@
...
+    """An orbit plotter backend class based on Plotly."""
+
+    def __init__(self, figure, layout, ref_units):
+        """Initializes a backend instance.
+
+        Parameters
+        ----------
+        figure : ~plotly.graph_objects.Figure
+            The plotly ``Figure`` to render the scene.
+        ~plotly.graph_objects.Layout
```
astrojuanlu, line 24:
```suggestion
        layout : ~plotly.graph_objects.Layout
```

src/poliastro/plotting/orbit/plotter.py
```diff
@@ -0,0 +1,758 @@
...
+            Epoch of current position.
+
+        """
+        from warnings import warn
+
+        from astropy import time
+
+        from poliastro.bodies import Sun
+        from poliastro.twobody import Orbit
+        from poliastro.warnings import TimeScaleWarning
```
astrojuanlu, line 177: If these imports cannot be here, maybe worth adding a comment explaining why?

src/poliastro/twobody/orbit/scalar.py
```diff
@@ -619,36 +619,48 @@ def apply_maneuver(self, maneuver, intermediate=False):
...
+                    # Terminal running IPython
+                    is_interactive = False
+                else:
+                    # Other type of shell
+                    is_interactive = False
+            except NameError:
+                # Standard Python interpreter
+                is_interactive = False
+            finally:
+                backend_name = "plotly2D" if is_interactive else "matplotlib2D"
```
astrojuanlu, line 659: Oh wow, nice trick 🤯  does this work with Spyder too?

astrojuanlu (approved):
This is AWESOME @jorgepiloto amazing job!

I left a few comments of small nitpicks and inconsistencies I spotted. My only remaining "big" comment is whether the API should use the backend name + a mapping between names and classes, or use the class directly. Like

```
from poliastro.plotting.orbit.backends import Matplotlib2D

plotter = OrbitPlotter(backend=Matplotlib2D())
```

(I think this will be more flexible in the long run by the way, since it allows the user to configure the backend, for example passing extra args to the initializer or something like that)

Apart from these comments, feel free to merge the PR when you feel it's ready!

src/poliastro/plotting/misc.py
jorgepiloto: I agree.

src/poliastro/plotting/orbit/backends/base_backend.py
jorgepiloto: Sounds good!

src/poliastro/twobody/orbit/scalar.py
jorgepiloto: I just checked. Both the script and the interactive Python shell in Spyder generate a "matplotlib" based figure.

jorgepiloto:
I agree with all the points you suggested here, @astrojuanlu. From my point of view, we can define the two main tasks:

**Backend objects as instances of a `BaseOrbitPlotterBackend` instead of docstrings**.
I like the idea you propose of passing the backend object instead o a string. It is clear that this allows for backend customization while still provides the same user experience when working with the plots.

**Naming convention.**
Let us play with the namespaces to simplify variables names. For instance, we could have:

```python
from poliastro.plotting.orbit import backends as op_backend # We may also have backends for other plotters
# from poliastro.plotting.groundtrack import backends as gt_backend
from poliastro.plotting

# Use a default matplotlib two-dimensional backend
plotter = OrbitPlotter()

# Use a different but still default configured backend
plotter = OrbitPlotter(backend=op_backends.Plotly2D)
plotter = OrbitPlotter(backend=op_backends.Plotly3D)

# Use a custom backend (using PyVista as example due to its highly customizable nature)
# Note this is just an example an does not reflect the future implementation of the PyVista's backends
custom = op_backends.PyVista3D(
        use_planets_textures=True, render_stars=False, notebook_backend="pythreejs"
)
plotter = OrbitPlotter(backend=custom)

**Comment by astrojuanlu** [comment]
I love this naming convention, looks really clean and terse. Go for it 💯

**Comment by astrojuanlu** [comment]
Just one nitpick: the default API could be

```python
OrbitPlotter(backend=op_backends.Plotly2D())
```

rather than

```
OrbitPlotter(backend=op_backends.Plotly2D)
```

hence: always pass a backend instance, never a class. Otherwise, the implementation needs to check whether the class was instantiated or not, for no real benefit (it's just two more characters for the user `()`).

jorgepiloto:
Still working on the suggestions...

I am thinking that it would be great to decouple untis from the backend. This means performing all conversions to the desired reference units of length inside the `plotter.py` instead of doing so inside each backend.

astrojuanlu:
Agreed!

jorgepiloto:
This is weird... All tests are passing locally, even after recreating my tox environent. I will double check again what is going on.

jorgepiloto:
Finally, this is likely to be finished today! 🚀

Here is a brief resume of the status of this pull-request:

* The new class `OrbitPlotter` accepts a `OrbitPlotterBackend` instance.
* It also accepts the desired length scale units.
* All units are managed by the `OrbitPlotter` class. Backends are unit agnostic.
  
  ```python
  backend = Plotly2D(use_dark_theme=True)
  plotter = OrbitPlotter(backend=backend, lenght_scale_units=u.au)
  ```

* For users who want to implement a custom backend, simply override the methods in the `OrbitPlotterBackend`.
* The `OrbitPlotter.show()` method is required to display the plot. This unifies all backends.

jorgepiloto:
Let me try to solve also for those Sphinx warnings. This is related with #1532.

kerel-fs (changes requested):

docs/source/quickstart.md
```diff
@@ -88,8 +88,11 @@ This plot is made in the so called *perifocal frame*, which means:
 The dotted line represents the *osculating orbit*: the instantaneous Keplerian orbit at that point. This is relevant in the context of perturbations, when the object shall deviate from its Keplerian orbit.
 
 ```{note}
-This visualization uses Plotly under the hood and works best in a Jupyter notebook.
-To use the static interface based on matplotlib, which might be more useful for batch jobs and ublication-quality plots, check out the {py:class}`poliastro.plotting.static.StaticOrbitPlotter`.
+This visualization uses Plotly under interactive environments like Jupyter
+Noteook or Jupyter Lab while it switches to Matplotlib otherwise. Nevertheless,
```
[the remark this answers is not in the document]
jorgepiloto: Good catch, @kerel-fs! Let me fix it.

src/poliastro/plotting/misc.py
```diff
@@ -15,75 +15,69 @@
...
-    interactive : bool, optional
-        Produce an interactive (rather than static) image of the orbit, default to False.
-        This option requires Plotly properly installed and configured for your environment.
+    labels : list[str]
+        A list of strings containing the labels of the bodies.
+    outer : bool, optional
+        Whether to print the outer Solar System, default to True.
+    backend : ~poliastro.plotting.orbit.backends._base.OrbitPlotterBackend
+        An instance of ``OrbitPlotterBackend`` for rendendering the scene.
+    lenght_scale_units : ~astropy.units.Unit
```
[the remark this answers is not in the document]
jorgepiloto: Awesome! Thanks for pointing this out, Fabian 🚀

jorgepiloto:
Finally, merging this after 6 months 🚀

astrojuanlu:
Amazing job @jorgepiloto, congratulations!!!!

PR #1552, opened 2022-07-22
fix pylint invalid-all-object / E0604 in propagators

s-m-e:
Fixes [E0604](https://pylint.pycqa.org/en/latest/user_guide/messages/error/invalid-all-object.html). Edge case, but strictly speaking `__all__` is supposed to be a list/tuple of strings and some tools are rather unhappy if they're not.


astrojuanlu:
Good catch, thanks!

PR #1553, opened 2022-07-22
fix pylint#2910 via workaround: ignore not an iterable warning on numba.prange

s-m-e:
Long standing annoyance, see [here](https://github.com/PyCQA/pylint/issues/2910). Simply globally disabling this warning is a rather bad idea, so it should go into the code.


astrojuanlu (approved):
Thanks!

PR #1554, opened 2022-07-22
fix pylint warning for super init not called on mixin class (no init there)

s-m-e:
I plead guilty ...

### DIFF
--- a/src/poliastro/twobody/orbit/scalar.py
+++ b/src/poliastro/twobody/orbit/scalar.py
@@ -40,7 +40,7 @@ class Orbit(OrbitCreationMixin):
 
     """
 
-    def __init__(self, state, epoch):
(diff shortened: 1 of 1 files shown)


astrojuanlu (changes requested):
Rather than "fixing" the warning, this PR ignores the warning. What about actually calling the `__init__()`? That way we also don't break Python algorithm to traverse the MRO (I think it's called "cooperative multiple inheritance"):

s-m-e:
The `Orbit` class inherits from the `OrbitCreationMixin` class that is designed as a "mixin", i.e. in this case a class that "only" collects a bunch of class methods. The only reason for the mixin to exist was to break the giant `Orbit` class into two more manageable parts. That's why it does not have a constructor - why should it? Pylint is not aware of this design pattern and yells at me, hence "fix" ;)

PR #1555, opened 2022-07-23
add missing test dependency: tox

s-m-e:
If this is intentional, please ignore.


astrojuanlu (commented):
This is kind of intentional because:

https://github.com/poliastro/poliastro/blob/b8ae53368d3e91f2d013c03de8279ff76820530f/tox.ini#L38-L40

tox uses the `test` extra to install test dependencies, and with this change tox would install tox itself.

But it's a silly thing anyway, and I see that all the tests passed anyway. So no harm in merging this.

PR #1556, opened 2022-07-23
Enable faster parallel testing

s-m-e:
It automatically scales to the number of cores that a machine has, so it should not break single-core-CI. As a bonus, if you have a laptop, it checks if its cooling is sufficient ;)

For this change to take effect, cached tox environments must be updated or removed and recreated.

Nice to have on a few more cores. `tox -e py310-fast` passes in 1:24 without and in 0:41 with this change applied on my system. "Your mileage may vary."


astrojuanlu (changes requested):
Nice! I recall talking about this some months ago, and @jorgepiloto and I discussed it last week again.

However, this _did_ make CI a bit more than 1.5x slower, so I'd rather find another solution. Besides, it would be nice to have this for `-slow` tests too. I think there are other ways to achieve this in `tox.ini`, I can investigate a bit.

s-m-e:
Quick and dirty: Add a (parallel) `veryfast` target, leave the `fast` target as it is?

PR #1557, opened 2022-07-23
remove twobody decorators

s-m-e:
As per our discussion sometimes earlier this year, they can (and should) go.


astrojuanlu (approved):
Indeed, these must go 🔥

PR #1559, opened 2022-07-25
Create documentation-links.yaml

astrojuanlu:
See https://github.com/readthedocs/readthedocs-preview

cc @humitos

### DIFF
--- a/.github/workflows/documentation-links.yaml
+++ b/.github/workflows/documentation-links.yaml
@@ -1,0 +1,17 @@
+name: Read the Docs Pull Request Preview
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+
+permissions:
+  pull-requests: write
+
+jobs:
+  documentation-links:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: readthedocs/readthedocs-preview@main
+        with:
+          project-slug: "readthedocs-preview"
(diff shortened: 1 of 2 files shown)


astrojuanlu:
Let's merge and see if it works 😏

.github/workflows/documentation-links.yaml
```diff
@@ -0,0 +1,17 @@
...
+permissions:
+  pull-requests: write
+
+jobs:
+  documentation-links:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: readthedocs/readthedocs-preview@main
+        with:
+          project-slug: "readthedocs-preview"
```
humitos, line 17: project-slug: "poliastro"

.github/workflows/documentation-links.yaml
```diff
@@ -0,0 +1,17 @@
+name: Read the Docs Pull Request Preview
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
```
humitos, line 6: This `synchronize` is not required.

humitos (commented):
I think it should work now. I will try to release a `v1` early so you can pin it and do not have breaking changes.

Take into account this is in beta state, tho 😄

.github/workflows/documentation-links.yaml
astrojuanlu: Done ✔️

.github/workflows/documentation-links.yaml
astrojuanlu: Done ✔️

PR #1560, opened 2022-07-25
Strip units within states

s-m-e:
This is a fun one. As discussed, I wanted the states to use `numpy.ndarray` objects instead of tuples for holding their data. Preparation for arrays. Turns out this is a bad idea. But as part of testing this idea, I had to deal with units ...

**The states now strip units internally.** This might be the one place to manage (i.e. strip and add) units. Interestingly, this little change consistently shaves more than 10 seconds off the time required to run the test suite (from 1:24 to 1:11 on average on my system, single threaded).

---

Reasoning: Let's take the `ClassicalState`as an example. In your implementation @astrojuanlu the elements use the following structure internally: `tuple[Quantity, Quantity, Quantity, Quantity, Quantity, Quantity]`, with each quantity being a scalar (the `RVState` is yet another tricky thing, let's ignore it for now). I wanted to go for an `ndarray` of shape `(6,)` and eventually turn it into a shape `(1,6)` to make it compatible with my array thinking. Either way, different astropy units within a single array do not mix well, so they had to go. Unit stripping moved into the state classes. 

Back to the ndarrays themselves. Let's have a look at what would happen in the states's constructors, simplified:

```python
class ClassicalState:
    def __init__(self, attractor, elements, plane):
        self._attractor = attractor
        self._elements = np.array([
            elements[0].to_value(u.km),  # p
            elements[1].value,  # ecc
            elements[2].to_value(u.rad),  # inc
            elements[3].to_value(u.rad),  # raan
            elements[4].to_value(u.rad),  # argp
            elements[5].to_value(u.rad),  # nu
        ])
        self._plane = plane
```

For a single state, this is fine. It's six doubles that get copied into a new array. Nothing fancy. Now let's picture the exact same code for state arrays: `elements` is a tuple of incoming ndarrays (one array per orbital element to keep things clean). Throwing them all into a new `self._elements` array would trigger a big malloc and some annoyingly slow memory copies. So leaving `self._elements` as `tuple[nadarry, nadarry, nadarry, nadarry, nadarry, nadarry]` is way more efficient.

Memory copies? Behold, there is still the unit stripping itself, which triggers some math and at least one memory copy and it happens at least once per conversion of one type of state to another. If there already are default units, why not make the stripping optional in the constructor of the states? This is where the `_units` parameter comes in. Its benefits can also be seen in the propagators, which can now run while messing a lot less with quantities and units. Even for the scalar states that astropy "only" covers at the moment, this little change is responsible for a solid 2 seconds less across the entire test suite.


s-m-e:
I am definitely breaking `tests/tests_twobody/test_perturbations.py::test_solar_pressure`, part of the slow tests. `main` is ok. Need to investigate this. No idea yet how I managed to do this ...

astrojuanlu (changes requested):
Thanks for the detailed explanation. Unfortunately, I see several problems with this - here are my initial thoughts right after reading the PR:

- If you realized that "using `numpy.ndarray` objects instead of tuples for holding their data is a bad idea" for the reasons you explained, what is the motivation for this change?
- The problem with this approach is that the `@property`es are now the ones that define the units, and this is error prone. For example, `a` is `return self.p / (1 - self.ecc**2) << u.km`. What if `p` was not given in kilometers?
- The above means that, by poliastro definitions, this change makes `*State` objects potentially unsafe.
  - I really dislike that an initializer parameter `_units` defines whether the object is unsafe or not. We should either roll back this change, or embrace it with all its consequences.
  - If we are to make `*State` objects unsafe, they should belong to `poliastro.core`. But `poliastro.core` should be self-contained, so there's no way a `*State` object can combine two poliastro objects (`attractor` and `plane`) with the rest of the values.

s-m-e:
> [...] what is the motivation for this change?

Consolidating unit handling into one place, the "core data structures" or state classes. If there is one and ONLY one place which does unit handling, it's (a) a lot less error-prone and (b) simply faster. A lot of unnecessary steps when converting between state types or propagating orbits can be dropped.

> The problem with this approach is that the `@propertyes` are now the ones that define the units, and this is error prone. For example, `a` is return `self.p / (1 - self.ecc**2) << u.km`. What if p was not given in kilometers?

Hey, that's a bug :) `a` should be `self._elements[0] / (1 - self._elements[1]**2) << u.km`. Thanks for pointing this out. It does not make any of the tests fail, so it needs one ... (TODO)

Anyway, unit handling is always a risk. On the other hand, the current implementation (all of poliastro) makes a ton of assumptions on units. Although it isn't stated anywhere explicitly, `core` runs on kilometers, seconds and radians - something that literally invites errors to begin with. From this perspective, I have not changed the surface of attack much.

I'd suggest to introduce something like a `poliastro.const` module, which could contain stuff like:

```python
DISTANCE_UNIT = u.km
```
Across all other modules, one would then in all critical places do:

```python
from poliastro.cost import DISTANCE_UNIT
some_distance = 1.0 << DISTANCE_UNIT
```

> I really dislike that an initializer parameter _units defines whether the object is unsafe or not. We should either roll back this change, or embrace it with all its consequences.

I *usually* prefer class methods for this kind of stuff:

```python
class SomeState:
    def __init__(self, x):
        self._x = x
    @classmethod
    def from_safe(cls, x): # with unit
        return cls(x = x.to_value(DISTANCE_UNIT))
    @classmethod
    def from_unsafe(cls, x): # without unit
        return cls(x = x)

s1 = SomeState.from_safe(1.0 << DISTANCE_UNIT)
s2 = SomeState.from_unsafe(1.0)
```

In the above example, directly using the constructor of `SomeState` should be discouraged, but it can not be prohibited. So my thinking went on to "how about a parameter set to a safe default and prefixed with an underscore to raise red flags when used". This is how `_units` came to be.

> If we are to make *State objects unsafe, they should belong to poliastro.core.

If I recall you correctly, the states are in some kind of limbo between user-facing and core ;) Let's embrace this: They are THE interface, the little layer in between that poliastro is currently missing.

s-m-e:
Although I still consider the way I computed `a` as a bug, astropy appears to be *really* robust. Not sure about all the edge cases, but the following "just works":

```python
x = 2_000.0 << u.m
y = 5.0 << u.one

assert (x / (1 - y) << u.km) == (-0.5 << u.km)
assert (x.to_value(u.km) / (1 - y.to_value(u.one)) << u.km) == (-0.5 << u.km)
```

The class method approach can be made a bit more user-friendly, to a degree, by hiding the responsible parameter in the fine-print of the implementation ... 

```python

DISTANCE_UNIT = u.km

class SomeState:

    def __new__(cls, *args, **kwargs):
        obj = super().__new__(cls)
        obj._unsafe = kwargs.pop('unsafe', False)
        return obj

    def __init__(self, *elements, **_):
        self._elements = elements if self._unsafe else (
            elements[0].to_value(DISTANCE_UNIT),
        )

    @property
    def elements(self):
        return self._elements

    @classmethod
    def _unsafe(cls, *elements):  # with unit
        return cls(*elements, unsafe = True)

s1 = SomeState(1.0 << DISTANCE_UNIT)
s2 = SomeState._unsafe(1.0)

s1.elements, [type(item) for item in s1.elements], s2.elements, [type(item) for item in s1.elements]
```

... will print `((1.0,), [numpy.float64], (1.0,), [numpy.float64])`. Not sure if that's a good idea.

s-m-e:
Ping @astrojuanlu What are the odds of getting this in or having a chat about it (see my comments since your last reply)? I'd love to build on top of this (i.e. rebase some of my current stuff to it).

PR #1561, opened 2022-07-25
Test documentation PR

astrojuanlu:
----------

:books: Documentation preview :books:: https://readthedocs-preview--1561.org.readthedocs.build/en/1561/

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ Name:         | poliastro
 **Author**:   | Juan Luis Cano Rodríguez [![orcid](https://img.shields.io/badge/id-0000--0002--2187--161X-a6ce39.svg)](http://orcid.org/0000-0002-2187-161X)
 **Version**:  | 0.18.dev0
 
+
 [![poliastro](https://img.shields.io/circleci/build/gh/poliastro/poliastro/main?style=flat-square)](https://circleci.com/gh/poliastro/poliastro/?branch=main)
 [![codecov](https://img.shields.io/codecov/c/github/poliastro/poliastro.svg?style=flat-square)](https://codecov.io/github/poliastro/poliastro?branch=main)
 [![pre-commit.ci status](https://results.pre-commit.ci/badge/github/poliastro/poliastro/main.svg)](https://results.pre-commit.ci/latest/github/poliastro/poliastro/main)


astrojuanlu:
@humitos still failing 🤔

humitos:
He he, I think it's because the `pull.body` (description of the PR) is `none`. Can you edit it to add some text and re-execute the action? That will help me to confirm my theory and fix the issue 🙏🏼

astrojuanlu:
Edited the body, reran, and at least now it worked! The RTD build is still pending. I'm not sure if we should have seen a new comment by now.

humitos:
> I'm not sure if we should have seen a new comment by now.

Yeah. The comment is not related to the status of the commit. I didn't do this yet because it seems the "commit status" is tied to _all the checks_. I didn't find a way to check the status of only one of the test and I think it may not be possible. Also, there is a comment about working _only in the default branch_. Maybe I'm wrong, tho...

This is what I've read: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#status

humitos:
Oh, BTW, the description was properly edited.

Also note that I changed to be a edition of the description because adding a comment it's noisy. Both have pros/cons, but for now I'm sticking with the description edit.

Let me know what you think, tho.

humitos:
Also, note that the `project-slug` is still `readthedocs-preview`. _I think_ this is because the GitHub Action is run with the code from the default branch, not the one from the PR... I've hit this while developing the action. Maybe there is a way to force using the code from the PR, but I guess it's not like that to avoid security issues.

astrojuanlu:
> Also, note that the project-slug is still readthedocs-preview. I think this is because the GitHub Action is run with the code from the default branch, not the one from the PR... I've hit this while developing the action. Maybe there is a way to force using the code from the PR, but I guess it's not like that to avoid security issues.

Merghhhhhhh you're right. Fun! I'll make the changes on `main` directly and rebase this.

astrojuanlu:
Hmm the workflow check does not appear anymore, but that might be because it has done its job already. Even closing and reopening doesn't work.

In any case, looks like it's working! I'm closing for now, and we'll see this in action for future PRs. Thanks a lot @humitos 🙏🏽

PR #1567, opened 2022-08-11
Added GM of various moons and mean distances of multiple systems

DhruvJ22:
mean_elements.py -> mean distances of sun-planet and various planet-moon systems
general.py -> added GM of multiple moons, which will be useful for CR3BP.

----
:books: Documentation preview :books:: https://poliastro--1567.org.readthedocs.build/en/1567/

### DIFF
--- a/src/poliastro/constants/general.py
+++ b/src/poliastro/constants/general.py
@@ -49,6 +49,15 @@ __all__ = [
     "GM_neptune",
     "GM_pluto",
     "GM_moon",
+    "GM_phobos",
+    "GM_deimos",
+    "GM_europa",
+    "GM_ganymede",
+    "GM_enceladus",
+    "GM_titan",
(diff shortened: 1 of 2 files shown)


astrojuanlu (changes requested):
Looks good @DhruvJ22! My only comment is to rename all

```
sunmercury
sunvenus
...
saturntitan
...
```

to

```
mean_a_mercury
mean_a_venus
...
mean_a_titan
...
```

even if this makes it less obvious that Titan is orbiting Saturn, at least it makes much more explicit that we are talking about semimajor axis here.

astrojuanlu:
Opened gh-1568 so we can tackle it in another pull request.

astrojuanlu (approved):
This is getting in, thanks @DhruvJ22!

PR #1570, opened 2022-08-13
added cr3bp_char_quant and its tests

DhruvJ22:
1. cr3bp_char_quant -> SystemChars class will serve as the base class for cr3bp model.
2. Added tests for cr3bp_char_quant.py
3.  Corrected "Saturn's..." -> "Uranus's...." in mean_elements.py -> mean_a_uranus

I will work on the read docs for constants in a future PR

----
:books: Documentation preview :books:: https://poliastro--1570.org.readthedocs.build/en/1570/


DhruvJ22:
@astrojuanlu I believe the tox error is due to the latex equation that I have in the doc string for t*. Any suggestions on how to fix it?

src/poliastro/constants/mean_elements.py
```diff
@@ -108,7 +108,7 @@
 
 mean_a_uranus = Constant(
     "mean_a_uranus",
-    "Saturn's orbit mean semimajor axis around Sun",
+    "Uranus's orbit mean semimajor axis around Sun",
```
jorgepiloto, line 111: Good catch, @DhruvJ22 🚀

jorgepiloto, line 111: Good catch, @DhruvJ22. No need to add an extra `s` though:

```suggestion
    "Uranus' orbit mean semimajor axis around Sun",
```

jorgepiloto (changes requested):

src/poliastro/constants/mean_elements.py
DhruvJ22: @jorgepiloto, I think it is fine to have that extra 's' (https://www.grammarly.com/blog/possessive-case/) since Uranus is a singular noun. Do you have any suggestions on how to get around the quality error?

DhruvJ22: Also, do you think the tests in the test_cr3bp_char_quant.py are sufficient tests, or do I need any additional tests?

tests/tests_threebody/test_cr3bp_char_quant.py
```diff
@@ -0,0 +1,27 @@
+"""
+@author: Dhruv Jain, Multi-Body Dynamics Research Group, Purdue University
+"""
+from astropy.tests.helper import assert_quantity_allclose
+
+from poliastro.threebody.cr3bp_char_quant import SystemChars
+
+
+def test_mu(p1, p2, expected_mu):
```
astrojuanlu, line 9: Unfortunately these tests won't run, pytest expects test parameters to be fixtures or parametrized parameters. You can verify this yourself locally running

```
tox -e py39-fast
```

(replace `39` with `38` or `310` depending on your Python version)

or

```
tox -e py39 -- -k threebody
```

(to only run threebody tests)

If you want to run the same test for different values of a parameter, have a look at `@pytest.mark.parametrize`. We are using it in some other places in poliastro.

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,236 @@
...
+        t*: float, sec
+            Characterisitc time of P1-P2 system
+
+        .. math::
+
+            \sqrt{\frac{l*^3}{M1+M2}}
+        """
+        return (lstar**3 / (mu_pi[0] + mu_pi[1])) ** 0.5
+
+    def BodiesCharCompute(self, p1, p2):
```
astrojuanlu, line 113: Method names should follow PEP8 conventions, so I'd expect this to be called `bodies_char_compute` instead

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,236 @@
...
+    def p1(self):
+        """Primary body P1"""
+        return self._p1
+
+    @property
+    def p2(self):
+        """Primary body P2"""
+        return self._p2
+
+    def __mu_calc(self, mu_pi):
```
astrojuanlu, line 77: Any reason to make protected methods? Notice that private methods in Python don't really exist (`_single_underscore` methods can be called anyway) and `__double_underscore` methods exist to avoid collisions with child classes, which is something we don't expect to have here.

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,236 @@
...
+        distances["marsdeimos"] = dist.mean_a_deimos
+        distances["jupitereuropa"] = dist.mean_a_europa
+        distances["jupiterganymede"] = dist.mean_a_ganymede
+        distances["saturnenceladus"] = dist.mean_a_enceladus
+        distances["saturntitan"] = dist.mean_a_titan
+        distances["uranustitania"] = dist.mean_a_titania
+        distances["neptunetriton"] = dist.mean_a_triton
+        distances["plutocharon"] = dist.mean_a_charon
+
+        return mu_body, distances
```
astrojuanlu, line 236: This whole method is needed to translate body names to variable names. What about

- Having the primaries be `poliastro.bodies.Body` objects?
- Add `mean_a` as a property to the `Body` objects? (I know I have opposed to this in the past, but it's just convenience at this point)

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,236 @@
...
+        else:  # if p2 is the bigger primary
+            mu_pi.append(mu_body[p2])
+            mu_pi.append(mu_body[p1])
+
+            # Create a string to get distance of the P1-P2 system
+            bodies = p2 + p1
+            # swap p1 and p2, as p1 should be the more massive body
+            temp_pi = p2
+            p2 = p1
+            p1 = temp_pi
```
astrojuanlu, line 169: Cool trick to swap variables in Python:
```suggestion
            p2, p1 = p1, p2
```

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,236 @@
...
+        Returns
+        -------
+        t*: float, sec
+            Characterisitc time of P1-P2 system
+
+        .. math::
+
+            \sqrt{\frac{l*^3}{M1+M2}}
+        """
+        return (lstar**3 / (mu_pi[0] + mu_pi[1])) ** 0.5
```
astrojuanlu, line 111: Let's create a new `poliastro.core.threebody` module and add `calculate_t_star` as a standalone function there.

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,236 @@
...
+        p2 : string
+           'body_j'. The deafult is Moon
+        """
+
+        mu, lstar, tstar, p1, p2 = self.BodiesCharCompute(
+            p1.lower(), p2.lower()
+        )  # lower() to ensure lowercase characters
+        units_nd = u.def_unit(
+            "[nd]"
+        )  # Define custome units, [nd]: non-dimensional unit
```
astrojuanlu, line 43: Do we need a custom non-dimensional unit though? We have `u.one` for that (aka `u.dimensionless_unscaled`, although [I introduced the `u.one` shorthand in Astropy 8 years ago](https://github.com/astropy/astropy/pull/1980))

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,236 @@
...
+
+        self._mu = mu * units_nd
+        self._lstar = lstar
+        self._tstar = tstar
+        self._p1 = p1
+        self._p2 = p2
```
astrojuanlu, line 49: A pattern we're using in poliastro a lot:

- Have `__init__` methods in classes just store variables:

```suggestion
    def __init__(self, lstar, tstar):
        """
        Constructor

        Parameters
        ----------
        mu
        lstar
        tstar
        """

        self._mu = mu * units_nd
        self._lstar = lstar
        self._tstar = tstar
```

(unsure if the two primaries are needed)

- Add `@classmethod`s to create the class:

```python
@classmethod
def from_primaries(cls, p1, p2):
    mu, lstar, tstar, p1, p2 = self.BodiesCharCompute(
        p1.lower(), p2.lower()
    )  # lower() to ensure lowercase characters
    units_nd = u.def_unit(
        "[nd]"
    )  # Define custome units, [nd]: non-dimensional unit
    return cls(mu, lstar, tstar)
```

(see my comment about `p1` and `p2` possibly being `Body` objects)
(see my above to my question about having `p1` and `p2` in the class itself)
(see my other comment about units)

astrojuanlu (changes requested):
Thanks for this pull request @DhruvJ22! I made several inline comments, addressing them will take you a bit of time so I'm stopping there. I will make some more high level comments:

- We are not experts in the CR3BP, so please if you have any authoritative references on the topic (better if they're books, journal papers otherwise) that would help us a lot in understanding the terminology and the intent.
- I'm not very fond of the `cr3bp_char_quant` name 😇 We already have `threebody/restricted.py`, so that could replace the `cr3bp` part completely. At the moment that module contains functions to compute the Lagrange points - if you want, we can move those to `threebody/lagrange.py`, and use `threebody/restricted.py` for this code. About the `char_quant`, I guess it means "characteristic quantities"?
- In general, we'd like to split (a) pure algorithms (like computing characteristic time, characteristic length) to raw functions in `poliastro.core`, and have more high level interfaces in `poliastro.threebody`. That way we decouple "the science" from the data structures we want to introduce, and also give our users the possibility to keep using the low level functions if those suit their needs.

Looking forward to reviewing the next iteration of this 🚀

DhruvJ22:
Thanks for all the suggestions! I will reply to all the changes that you requested in some time.

DhruvJ22:
> 

1. Some helpful resources on CR3BP:
a. E. Zimovan Spreen, "Trajectory Design and Targeting for Applications to the Exploration Program in Cislunar Space," Ph.D., May 2021
b. V. Szebehely, "Theory of Orbits: The Restricted Problem of Three Bodies", 1967
c. W. Koon, M. Lo, J. Marsden, S. Ross, "Dynamical Systems, The Three-Body Problem, and Space Mission Design", 2006

2. It might be better to explicitly name the file cr3bp as there are various Restricted 3BP models like ER3BP, HR3BP.
For Lagrange point calculation, I will prefer to use my code -> (https://github.com/poliastro/poliastro/blob/main/contrib/cr3bp_DhruvJ/cr3bp_lib_calc.py) with some modifications. In any case, it will make more sense to rename the existing restricted.py to lagrange.py. 

Yes, char_quant means characteristic quantities. 

3. I will on splitting it.

tests/tests_threebody/test_cr3bp_char_quant.py
DhruvJ22: I will focus on the cr3bp_char_quant.py for now and come back to the tests later.

src/poliastro/threebody/cr3bp_char_quant.py
DhruvJ22: Thanks, my bad, I will fix it

src/poliastro/threebody/cr3bp_char_quant.py
DhruvJ22: The whole idea was to make the quantities constants but I think a better might be by using astropy -> Constants (). But I will appreciate if you have any other ideas on making mu, l* and t* constants.

src/poliastro/threebody/cr3bp_char_quant.py
DhruvJ22: We will need to define our own [nd] units. But after reading this piece of code again, I realized that mu is dimensionless, so it will make sense to use u.one.

The idea will be to non-dimensionalize position by l*, so 1 [nd] = l* km. Example: Earth-Moon -> 1 [nd] position = 384400 km. 

We can get rid of this specific unit but I will add a "[nd]" for position and "[nd]" for velocity as the [nd] quantities are still dimensional quantities. They are just scaled by the characteristic quantities which is why the units are called non-dimensional and not dimensionless.

src/poliastro/threebody/cr3bp_char_quant.py
DhruvJ22: I will work on making the p1 and p2 as Body objects then go from there.

src/poliastro/threebody/cr3bp_char_quant.py
DhruvJ22: I can add the primaries be bodies.Body objects and add mean_a as a property. But for the moons like charon, titan, is it alright if I only add their: parent, k, mean_a?

src/poliastro/threebody/cr3bp_char_quant.py
astrojuanlu: I'd say astropy Constants are for global constants. The Python pattern would be to make them private properties:

```python
class System:
    def __init__(self, mu, lstar, tstar):
        self._mu = mu
        self._lstar = lstar
        self._tstar = tstar

    @property
    def mu(self):
        return self._mu

    # Same for lstar, tstar

s = System(1, 1, 1)
print(s.mu)  # 1
s.mu = 1  # AttributeError: can't set attribute 'mu'
```

we don't need to go more "constant" than that I believe.

src/poliastro/threebody/cr3bp_char_quant.py
astrojuanlu: Yes 👍🏽

src/poliastro/threebody/cr3bp_char_quant.py
astrojuanlu: I see. What about something more explicit? This is what I did in an experiment to introduce canonical units everywhere in poliastro https://github.com/poliastro/poliastro/pull/708/:

```python
du = u.def_unit("du", norm(r))
tu = u.def_unit("tu", np.sqrt((1 * du) ** 3 / source_body.k))

r_value = r.to_value(du)
v_value = v.to_value(du / tu)
```

(just an idea)

DhruvJ22:
@astrojuanlu Do you have any suggestions on what I should rename cr3bp_char_quant.py? restricted.py might not be suitable if other models are to be added in the future.

DhruvJ22:
I have implemented most of the changes that you suggested. I just realized that I have not thoroughly tested the new setup. So pls ignore the latest commit till my next message.

I also removed the saved plots in /contrib, so that github does not show Poliastro to be HTML.

src/poliastro/core/threebody/cr3bp_quantities_calculations.py
```diff
@@ -0,0 +1,45 @@
+"""
+@author: Dhruv Jain, Multi-Body Dynamics Research Group, Purdue University
+"""
+
+from astropy import units as u
+
+
+def calculate_mu(mu1, mu2):
+    """Calculate mu of CR3BP
+    Parameters
```
astrojuanlu, line 10:
```suggestion
    """Calculate mu of CR3BP

    Parameters
```

src/poliastro/core/threebody/cr3bp_quantities_calculations.py
```diff
@@ -0,0 +1,45 @@
...
+    -------
+    mu: float, dimensionless
+        mu2/(mu1+mu2), mu2<mu1
+    """
+    return (mu2 / (mu1 + mu2)) * u.one
+
+
+def calculate_tstar(mu1, mu2, lstar):
+    """Calculate t* of CR3BP
+    Parameters
```
astrojuanlu, line 27:
```suggestion
    """Calculate t* of CR3BP

    Parameters
```

src/poliastro/core/threebody/cr3bp_quantities_calculations.py
```diff
@@ -0,0 +1,45 @@
...
+    mu1: float, dimensionless
+        mu of P1
+    mu1: float, dimensionless
+        mu of P2
+    lstar: float, km
+        Characterisitc length of P1 - P2 system
+
+    Returns
+    -------
+    t*: float, sec
```
astrojuanlu, line 38:
```suggestion
    tstar: float, sec
```

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,105 @@
...
+        name: string
+            System name, format: 'Body1Body2', e.g. 'EarthMoon', 'SunEarth'
+        mu: float, dimensionless
+           mass ratio of P1-P2 primary bodies
+        lstar: float, km
+           Characterisitc length of P1-P2 system
+        tstar: float, sec
+            Characterisitc time of P1-P2 system
+        """
+        self.name = name
```
astrojuanlu, line 33: Should we make the `name` read-only too?

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,105 @@
...
+        self._mu = mu
+        self._lstar = lstar
+        self._tstar = tstar
+
+    @classmethod
+    def from_primaries(cls, p1, p2):
+        """
+        Computes and sets the characteristic quanitites based on p1 and p2 bodies
+        Parameters
```
astrojuanlu, line 42:
```suggestion
        Computes and sets the characteristic quanitites based on p1 and p2 bodies

        Parameters
```

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+
+        name, mu, lstar, tstar = cls.bodies_char_compute(p1, p2)
+        return cls(name, mu, lstar, tstar)
+
+    @property
+    def bodies_char_compute(self, p1, p2):
+        """
+        Calculates mu, lstar, and tstar of the 'p1' and 'p2' system
+        Also, if M2>M1, then swaps p1 and p2, so that M1>M2
```
astrojuanlu, line 55:
```suggestion
        Calculates mu, lstar, and tstar of the 'p1' and 'p2' system

        Also, if M2>M1, then swaps p1 and p2, so that M1>M2
```

src/poliastro/threebody/cr3bp_char_quant.py
```diff
@@ -0,0 +1,105 @@
...
+    def from_primaries(cls, p1, p2):
+        """
+        Computes and sets the characteristic quanitites based on p1 and p2 bodies
+        Parameters
+        ----------
+        p1: ~poliastro.bodies.Body
+        p2: ~poliastro.bodies.Body
+        """
+
+        name, mu, lstar, tstar = cls.bodies_char_compute(p1, p2)
```
astrojuanlu, line 48: This is boing to fail, `@property`-decorated methods cannot take parameters. You will have to make `bodies_char_compute` a private module function (or at most a private classmethod of the class)

astrojuanlu (changes requested):
Starting to look much cleaner! I didn't address some of your questions but gave this another pass before you start writing the tests.

astrojuanlu:
Also, please address the pre-commit errors, otherwise the rest of the pipeline won't run.

DhruvJ22:
@astrojuanlu do you have any suggestion on how to fix the pre-commit '/s' error?

src/poliastro/core/threebody/cr3bp_quantities_calculations.py
```diff
@@ -0,0 +1,48 @@
...
+    Returns
+    -------
+    mu: float, dimensionless
+        mu2/(mu1+mu2), mu2<mu1
+    """
+    return (mu2 / (mu1 + mu2)) * u.one
+
+
+def calculate_tstar(mu1, mu2, lstar):
+    """Calculate t* of CR3BP
```
astrojuanlu, line 27:
```suggestion
    r"""Calculate t* of CR3BP
```

This will fix the pre-commit error

DhruvJ22:
Thanks, Juan! Hopefully, the error is fixed and I have removed the test for now. I will add it in a future PR.

astrojuanlu:
The tests are what will tell us if the code is correct, I'd rather have them in this PR

astrojuanlu:
The tests are looking really good! @DhruvJ22 let us know when this is ready, I expect max 1 more round of feedback and we'll be good to go 💪🏽

DhruvJ22:
@astrojuanlu, I believe I have all the required tests, so we should be good to merge. Since this branch is out of date, should I update with rebase?

src/poliastro/bodies.py
```diff
@@ -13,6 +13,16 @@
...
 * Pluto (♇)
+Phobos
+Deimos
+Europa
+Ganyemede
+Enceladus
+Titan
+Titania
+Triton
+Charon
```
astrojuanlu, line 24:
```suggestion
* Phobos
* Deimos
* Europa
* Ganyemede
* Enceladus
* Titan
* Titania
* Triton
* Charon
```

astrojuanlu (changes requested):
Few more things and we are done:

- There's a `Uranuss in mean_elements.py -> mean_a_uranus` file that got added accidentally, please remove it
- The `src/poliastro/core/threebody` package is missing a `__init__.py` file which is mandatory
- Adding that file surfaces an `import-linter` error:

```
$ lint-imports
=============
Import Linter
=============

---------
Contracts
---------

Analyzed 1164 files, 6329 dependencies.
---------------------------------------

poliastro.core does not import astropy.units BROKEN

Contracts: 0 kept, 1 broken.


----------------
Broken contracts
----------------

poliastro.core does not import astropy.units
--------------------------------------------

poliastro.core is not allowed to import astropy.units:

-   poliastro.core.threebody.cr3bp_quantities_calculations -> astropy.units (l.5)


```

In other words: `poliastro.core` functions must

- receive raw Python floats or NumPy arrays,
- assume certain units and document them: (kilometers, seconds),
- not depend on `astropy.units` for anything (all the unit handling should be outside of `poliastro.core`)

I also left 1 style suggestion. Thanks @DhruvJ22 for your patience!

astrojuanlu:
Also, yes please rebase 👍🏽

DhruvJ22:
> Few more things and we are done:
> 
>     * There's a `Uranuss in mean_elements.py -> mean_a_uranus` file that got added accidentally, please remove it
Can you expand on that? Do you mean the s after Uranus'?

astrojuanlu:
See this empty file: https://github.com/dhruvj22/poliastro/blob/add_cr3bp_char_quant/Uranuss%20in%20mean_elements.py%20-%3E%20mean_a_uranus

astrojuanlu (approved):
This has received enough rounds of feedback already and I don't want to keep bothering you with nitpicks 😄 Thanks a lot for your patience @DhruvJ22! This is going in 🚀 I will squash all the commits so that we don't have intermediate files that should not have been there.

DhruvJ22:
Thanks a lot for merging and bearing with me as I learn how to write production-quality code.

PR #1571, opened 2022-08-18
Fix GH repo language classification

helgee:
----
:books: Documentation preview :books:: https://poliastro--1571.org.readthedocs.build/en/1571/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.html linguist-detectable=false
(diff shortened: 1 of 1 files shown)


astrojuanlu (approved):
Oh wow, thanks a lot! Didn't occur to me that this could be so easy. Docs: https://github.com/github/linguist/blob/master/docs/overrides.md

PR #1573, opened 2022-08-27
adding libration point calculation using newton raphson method

DhruvJ22:
I like my method of libration point calculation compared to the one in restricted.py that uses brentq:
1. Compatible with cr3bp_char_quant
2. Uses non-dim units, which will be helpful later for cr3bp propagation
3. Uses Newton Raphson instead of brentq as I am not sure if the limit intervals set for brentq can encapsulate all systems, while Newton Raphson is more robust
4. Uses regularized form of the CR3BP EOMs to define a function, whose zeros are the liberation points -> less susceptible numerical errors

a. The function and its test work if I compare their values but L_ND is throwing an error that I am having difficulty getting around. I have tried to use 'u_enabled_units' but that has not resolved the problem. 
b. I would like to add the L_ND and V_ND (added as comments in SystemChars) so that we can use SystemChars' object and not have to define the units in multiple functions.

### DIFF
--- a/src/poliastro/threebody/cr3bp_char_quant.py
+++ b/src/poliastro/threebody/cr3bp_char_quant.py
@@ -36,6 +36,9 @@ class SystemChars:
         self._mu = mu
         self._lstar = lstar
         self._tstar = tstar
+        # self._L_ND = u.def_unit("dist_nd", self.lstar)
+        # self._V_ND = u.def_unit("vel_nd", self.lstar/self.tstar)
+
 
     @classmethod
     def from_primaries(cls, p1, p2):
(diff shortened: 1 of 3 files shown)


DhruvJ22:
Notes to self: need to add T_ND: non-dim unit for time

astrojuanlu (changes requested):
Hi @DhruvJ22, sorry for the huge delay 🙏🏽 

I have one very high level comment: if I understand correctly, `newton_raphson_lib_calc` finds the root of a power series polynomial. However, if I understand correctly, finding the roots of a polynomial does not require a numerical solver, and can be done with a highly efficient linear algebra method:

https://mathworld.wolfram.com/PolynomialRoots.html

In fact, this method is available in NumPy:

https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polynomial.Polynomial.roots.html

as demonstrated here:

```python

In [1]: import numpy as np

In [2]: p = np.polynomial.Polynomial([1, 2, -3])

In [3]: p
Out[3]: Polynomial([ 1.,  2., -3.], domain=[-1,  1], window=[-1,  1])

In [4]: p.roots()
Out[4]: array([-0.33333333,  1.        ])
```

Therefore, could you please

1. Remove the Newton-Raphson implementation and rewrite everything in terms of NumPy polynomials instead, and
2. Uncomment the problematic unit definitions, so that your tests fail and I can see what exactly the problem is?

Issue #1575, opened 2022-09-07
Unit of Orbit.h_vec and Orbit.h_mag is set to km^3/s^2 instead of km^2/s


valdaarhun:
🐞 **Problem**

`Orbit.h_vec` and `Orbit.h_mag` are reported in units of $km^3 / s^2$ instead of $km^2 / s$.

```python
import astropy.units as u
from poliastro.bodies import Earth
from poliastro.twobody import Orbit

r = [6000, 4000, 5000] << u.km
v = [10, 20, 30] << u.km / u.s

orb = Orbit.from_vectors(Earth, r, v)
print("Specific angular momentum = ", orb.h_vec)
print("Magnitude of specific angular momentum = ", orb.h_mag)
```

Output:

```
Specific angular momentum = [  20000. -130000.   80000.] km3 / s2
Magnitude of specific angular momentum = 153948.04318340652 km3 / s2
```

🖥 **Please paste the output of following commands**

* `conda info -a` (only if you have conda)
* `conda list` (only if you have conda)
* `pip freeze`

<details>
  
```bash
# Paste your output here:
-e git+https://github.com/valdaarhun/poliastro.git@a4f02b609d99dfc9ccd6f8e4010271d016b76a83#egg=poliastro
```
</details>

💡 **Possible solutions**

The reason is that in [this](https://github.com/poliastro/poliastro/blob/main/src/poliastro/twobody/orbit/scalar.py#L176) line, `self.v.to(u.km / u.s)` does not strip away the unit.

Because of this `np.cross(r, v)`, inherits `u.km / u.s` and then multiplying it by `u.km**2 / u.s` sets the unit to $km^3/s^2$

📋  **Steps to solve the problem**

This can be fixed by replacing `self.v.to()` with `self.v.to_value()`. This makes the result of `np.cross(r, v)` unitless.

PR #1576, opened 2022-09-07
Fix unit of h_vec in cross product calculation

valdaarhun:
Calculating h_vec using `self.v.to(u.km / u.s)` in `np.cross(r, v)` retains $km/s$ as its unit.

Therefore, multiplying `np.cross(r, v)` with `u.km ** 2 / u.s` sets the unit of h_vec to $km^3/s^2$.

Instead, `self.v.to_value(u.km / u.s)` should be used to keep `np.cross(r, v)` unitless.

Fixes: #1575 

----
:books: Documentation preview :books:: https://poliastro--1576.org.readthedocs.build/en/1576/


astrojuanlu (approved):
Thanks a lot @valdaarhun!

PR #1588, opened 2022-10-17
build: Remove upper bound on requires-python

matthewfeickert:
Remove upper bound on `requires-python` as advocated for in Henry Schreiner's blog post "Should You Use Upper Bound Version Constraints?".
   - c.f. https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special

----
:books: Documentation preview :books:: https://poliastro--1588.org.readthedocs.build/en/1588/

PR #1590, opened 2022-10-25
CI: Add Python 3.11 to testing

matthewfeickert:
* Add Python 3.11 to testing in CI.
* Update PyPI metadata to reflect Python 3.11 support.

----
:books: Documentation preview :books:: https://poliastro--1590.org.readthedocs.build/en/1590/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,13 @@ jobs:
     environment:
       TOXENV: py310
 
+  py311:
+    <<: *test-template
+    docker:
+      - image: python:3.11
+    environment:
+      TOXENV: py311
+
 workflows:
   version: 2
   build_and_test:
@@ -81,5 +88,8 @@ workflows:
           requires:
             - quality
       - py310:
+          requires:
+            - quality
+      - py311:
           requires:
             - quality
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,7 @@ classifiers = [
     "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: Implementation :: CPython",
     "Topic :: Scientific/Engineering",
     "Topic :: Scientific/Engineering :: Physics",
--- a/src/poliastro/maneuver.py
+++ b/src/poliastro/maneuver.py
@@ -26,7 +26,7 @@ class Maneuver:
 
     """
 
-    def __init__(self, *args):
+    def __init__(self, *impulses):
         r"""Constructor.
 
         Parameters
@@ -35,29 +35,31 @@ class Maneuver:
             List of pairs (delta_time, delta_velocity)
 
         """
-        self.impulses = args
-        # HACK: Change API or validation code
-        _dts, _dvs = zip(*args)
+        # Validate units of the impulses provided
+        self.impulses = [
+            self.impulse_has_valid_units(*impulse) for impulse in impulses
+        ]
+        self._dts, self._dvs = zip(*self.impulses)
 
+        # Validate that all delta-V are three-dimensional vectors. Note that the
+        # input delta-V may not even be an iterable. This is why the 'TypeError'
+        # exception is forced
         try:
-            self._dts, self._dvs = self._initialize(
-                [(_dt * u.one).value for _dt in _dts] * (_dts[0] * u.one).unit,
-                [(_dv * u.one).value for _dv in _dvs] * (_dvs[0] * u.one).unit,
-            )
-            if not all(len(dv) == 3 for dv in _dvs):
-                raise TypeError
-        except TypeError:
+            if not all(len(dv) == 3 for dv in self._dvs):
+                raise ValueError
+        except (TypeError, ValueError):
             raise ValueError("Delta-V must be three dimensions vectors")
 
     def __repr__(self):
         return f"Number of impulses: {len(self.impulses)}, Total cost: {self.get_total_cost():.6f}"
 
-    @u.quantity_input(dts=u.s, dvs=u.m / u.s)
-    def _initialize(self, dts, dvs):
-        return dts, dvs
-
     def __getitem__(self, key):
         return self.impulses[key]
+
+    @staticmethod
+    @u.quantity_input(dts=u.s, dvs=u.m / u.s)
+    def impulse_has_valid_units(dts, dvs):
+        return dts, dvs
 
     @classmethod
     def impulse(cls, dv):
--- a/tests/test_maneuver.py
+++ b/tests/test_maneuver.py
@@ -27,12 +27,23 @@ def test_maneuver_constructor_raises_error_if_invalid_delta_v():
 
 
 def test_maneuver_raises_error_if_units_are_wrong():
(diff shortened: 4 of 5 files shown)


matthewfeickert:
Unfortunately `numba` has broken all Python 3.11 build attempts as they have capped `python_requires` to `<3.11`. :cry: 

https://github.com/numba/numba/blob/9d034e91f881f8ff1f2a9051a757421328f02625/setup.py#L22-L47

So until there is a new `numba` release with this fixed there can be _no_ `numba` installs for anyone. This is an example of why I made PR #1588.

astrojuanlu:
@matthewfeickert It actually makes sense that they capped the version, I don't think numba + llvmlite can run on Python 3.11 yet:

https://github.com/numba/numba/issues/8304

Numba + llvmlite do dark magic on the Python bytecode and they need weeks or months to support new Python versions. So, they actively know this version is broken on Python 3.11, and poliastro would too.

I guess in cases like this it's acceptable for them to put an upper cap on the Python version?

matthewfeickert:
> It actually makes sense that they capped the version, I don't think numba + llvmlite can run on Python 3.11 yet:
> 
> [numba/numba#8304](https://github.com/numba/numba/issues/8304)
> 
> Numba + llvmlite do dark magic on the Python bytecode and they need weeks or months to support new Python versions. So, they actively know this version is broken on Python 3.11, and poliastro would too.
> 
> I guess in cases like this it's acceptable for them to put an upper cap on the Python version?

@henryiii had mentioned this as well, but I don't really understand the logic, unless the plan is to never release Python 3.11 wheels for `numba` `v0.56.3` and to only release Python 3.11 wheels for later releases. Then I get it to some extent, but I'm still not sure why the need for the cap unless it is to just save time instead of letting you hit a build failure.

I appreciate that bytecode manipulation isn't easy though, so I'm not upset that they don't have a Python 3.11 release out &mdash; that's hard work in general.

astrojuanlu:
> unless the plan is to never release Python 3.11 wheels for numba v0.56.3 and to only release Python 3.11 wheels for later releases

Yeah, I don't think they publish wheels of old numba versions for old Python versions, precisely because they are not compatible. Just checked with https://pypi.org/project/numba/0.54.1/#files

> unless it is to just save time instead of letting you hit a build failure.

Most likely. That's why we used to cap by the way: to improve the developer experience for less tech-savvy users https://github.com/poliastro/poliastro/issues/1036#issuecomment-731407255

I do understand the problems of this though (and I declare I have read https://iscinumpy.dev/post/bound-version-constraints/). But at least it worked for us in that it reduced the amount of issues that got opened because of this. Bear in mind that most poliastro users are new to Python *and* open source simultaneously.

esc:
> @henryiii had mentioned this as well, but I don't really understand the logic, unless the plan is to never release Python 3.11 wheels for `numba` `v0.56.3` and to only release Python 3.11 wheels for later releases. Then I get it to some extent, but I'm still not sure why the need for the cap unless it is to just save time instead of letting you hit a build failure.

As a Numba developer, I can confirm the first Numba version to support 3.11 will be 0.57.

esc:
> Unfortunately `numba` has broken all Python 3.11 build attempts as they have capped `python_requires` to `<3.11`. 😢
> 
> https://github.com/numba/numba/blob/9d034e91f881f8ff1f2a9051a757421328f02625/setup.py#L22-L47

This statement is not accurate. We no longer pin the max python version in `python_requires` but instead raise an exception at runtime when `setup.py` is run. We did this based on a recommendation from @henryiii since we were bitten by this during previous (3.10 and 3.9) releases. So now, you can try to `pip` install Numba for 3.11 and it will error out, instead of looking for a much older version (several years iirc) that didn't pin via `python_requires` and try to install.

astrojuanlu:
Thanks for chiming in @esc! That's a smart move. I recall there was a thread on Discourse or the pip issue tracker about this "heavy backtracking" triggered by version caps - putting these runtime checks in `setup.py` seems like a decent workaround.

henryiii:
I would consider the fact that the error is working as expected to be a success. `python_requires` was wrong and didn't do the right thing, causing unrelated errors about an old version of llvmlite not compiling (IIRC). Unlike most packages, 0.56.3's source doesn't support 3.11, so you can't ever make a wheel from that source for 3.11. It needs code changes to support 3.11, hence a new version. Numba is very special in this, because of the byte code usage, there is basically a 0% chance that they can support versions they have never seen.

esc:
> I would consider the fact that the error is working as expected to be a success. 

💯 yes @henryiii thank you again ❤️

matthewfeickert:
Nice. Only 1 error isn't bad:

```
=================================== FAILURES ===================================
__________ test_maneuver_constructor_raises_error_if_invalid_delta_v ___________

    def test_maneuver_constructor_raises_error_if_invalid_delta_v():
        dv1 = np.zeros(3) * u.km / u.s
        dv2 = np.ones(2) * u.km / u.s  # Incorrect dv
        with pytest.raises(ValueError) as excinfo:
            with warnings.catch_warnings():
                # Different length numpy arrays generate a deprecation warning.
                warnings.simplefilter(
                    "ignore", category=np.VisibleDeprecationWarning
                )
                Maneuver((0 * u.s, dv1), (2 * u.s, dv2))
>       assert "Delta-V must be three dimensions vectors" in excinfo.exconly()
E       AssertionError: assert 'Delta-V must be three dimensions vectors' in 'ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.'
E        +  where 'ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.' = <bound method ExceptionInfo.exconly of <ExceptionInfo ValueError('setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.') tblen=4>>()
E        +    where <bound method ExceptionInfo.exconly of <ExceptionInfo ValueError('setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.') tblen=4>> = <ExceptionInfo ValueError('setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.') tblen=4>.exconly

tests/test_maneuver.py:26: AssertionError
```

Something to follow up on later.

jorgepiloto:
It looks like the issue comes from NumPy + Astropy quantities. Let me identify the the right version of NumPy in which this fails.

jorgepiloto:
| Numpy version | Test passes |
|---------------|-------------|
| 1.23          | Passes      |
| 1.24          | Fails       |

jorgepiloto:
The issue is caused by this line:

```python
# src/poliastro/maneuver.py @ line 45
[(_dv * u.one).value for _dv in _dvs] * (_dvs[0] * u.one).unit
```

jorgepiloto:
I took advantage of the situation and refactored the initialization of the `Maneuver` class.

jorgepiloto:
Yay! This is finally working.

jorgepiloto (approved):

jorgepiloto:
Thanks everyone for all the hard work in this pull-request and Numba 🚀

